PR #2, opened 2017-07-07
Snapshot API saves and restores time deltas.

scnale:
When running tests, the snapshot API wasn't saving the time adjustments made through the `evm_timeIncrease` JSON-RPC command. I added the code necessary to save and restore the appropriate state but I'm not sure this is the best way to do it. According to my own tests it seems to be working.

### DIFF
--- a/test/time_adjust.js
+++ b/test/time_adjust.js
@@ -64,4 +64,34 @@ describe('Time adjustment', function() {
       })
     })
   })
+
+  it('should revert time adjustments when snapshot is reverted', function(done) {
+    // Adjust time
+    web3.eth.getBlock('latest', function(err, block){
+      if(err) return done(err)
+      var previousBlockTime = block.timestamp
+      var originalTimeAdjustment = provider.manager.state.blockchain.timeAdjustment
+
+      send("evm_snapshot", function(err, result) {
+        // jump forward another 5 hours
+        send("evm_increaseTime", [secondsToJump], function(err, result) {
+          if (err) return done(err);
+
+          var currentTimeAdjustment = provider.manager.state.blockchain.timeAdjustment
+          assert.equal(currentTimeAdjustment, originalTimeAdjustment + secondsToJump)
+
+          // Mine a block so new time is recorded.
+          send("evm_mine", function(err, result) {
+            if (err) return done(err);
+
+            send("evm_revert", function(err, result) {
+              var revertedTimeAdjustment = provider.manager.state.blockchain.timeAdjustment
+              assert.equal(revertedTimeAdjustment, originalTimeAdjustment);
+              done()
+            })
+          })
+        })
+      })
+    })
+  })
 })
--- a/test/time_adjust.js
+++ b/test/time_adjust.js
@@ -1,10 +1,13 @@
-var TestRPC = require("../index.js");
-var assert = require('assert');
+var Ganache = require("../index.js");
+var assert = require('assert-match');
+var matchers = require('assert-match/matchers');
+var gte = matchers.gte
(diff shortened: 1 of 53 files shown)


hiqua:
Is there a reason not to merge these changes?

madvas:
Please merge, much needed

benjamincburns:
@scnale sorry it has taken us so long to circle 'round to this PR! Unfortunately, I don't think it's working. I've pushed a (failing) test to this PR which demonstrates how I think it's meant to work. Can you please either update the test to correct my misunderstanding, or update your solution to make the test pass?

benjamincburns:
I don't know why, but the new test isn't showing in the latest travis build. I promise it's there, and if you run it locally it fails.

scnale:
Ah, I see. Comparing timestamps won't work because the time adjustment is added to the current timestamp. I think the best way to test this would be to:
- take a snapshot
- call `evm_increaseTime` with any value
- restore the snapshot
- call `evm_increaseTime` with value `k` (preferably different from the previous one) and check its return value for equality with `k`.

benjamincburns:
@scnale it's been quite a while - no pressure, but do you intend to fix up the test failures? If not, I'll take it on if I have time, or if I don't I'll just close this PR and flag the issue as `needshelp`.

ouziel-slama:
Please fix/merge, much needed to write real tests. Thanks.

benjamincburns:
@ouziel-slama, per comments above, the changes in this PR don't work as intended. Since I haven't heard back from @scnale in quite a while I'll close this to avoid confusion. If you'd like to fork @scnale's changes and submit a new PR yourself, I strongly encourage it!

PhABC:
@benjamincburns and @scnale the fix works as intended. There is a mistake in the new test case, where revert should revert to snapshots[1]. Line `87` should therefore be `send("evm_revert",[1], function(err, result) { ... });` instead of `send("evm_revert", function(err, result) { ... });`. 

With this change, all the tests related to this PR pass. The original `evm_revert` call does properly revert to the very first `timeAdjustement`, as expected as well. Indeed, calling `evm_increaseTime` at line 46  
increments the `timeAdjustment` by `secondsToJump` seconds, a change that persists throughout the remaining tests. When calling `send("evm_revert", function(err, result) { ... });`, `timeAdjustment` is set to `originalTimeAdjustment - secondsToJump`, as expected. 

### Trace with current test implementation

1. `timeAdjustement` : -50652919 
2. `evm_increaseTime` (line 46)
3. `timeAdjustement` : -50634919
4. `evm_snapshot`: (line 75, snapshot 1, `timeAdjustement` == -50634502)
5. `evm_increaseTime` (line 77)
6. `timeAdjustement` : -50616919
7. `send("evm_revert", function(err, result) {...})` (line 87)
8. `timeAdjustement` : -50616919

The current test assert that `4.` should be equal to `8`, which is false. By specifying snapshot to 1, the timeAdjustment is correctly reverted to the value stored at `4.`

benjamincburns:
> There is a mistake in the new test case

But... but I wrote that test case! :blush:

Thanks so much for pointing this out, @PhABC - I'll have a look!

test/time_adjust.js
```diff
@@ -64,4 +64,34 @@ describe('Time adjustment', function() {
       })
     })
   })
+
+  it('should revert time adjustments when snapshot is reverted', function(done) {
+    // Adjust time
+    web3.eth.getBlock('latest', function(err, block){
+      if(err) return done(err)
+      var previousBlockTime = block.timestamp
```
graemecode: Unused var, fyi

PR #4, opened 2017-07-23
eth_getBlockTransactionCountByNumber add check block undefined

paritygeeker:
eth_getBlockTransactionCountByNumber method should check the block for undefined station, when somebody call `web3.eth.getBlockTransactionCount` from geth or truffle console directly will always get the testrpc crash.
And maybe other method also need some security check, and I will help checkout them if you give me this pleasure, tks.


tcoulter:
Thanks for the heads up! Added similar code to ganache-core that will go out with the next release. Cheers!

PR #9, opened 2017-08-29
1.0.0 calls send with a callback

formula1:
https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-requestmanager/src/index.js#L129


onbjerg:
Why is this not merged? This supports Web3 1.0.0 without breaking backwards compat.

benjamincburns:
Hi @formula1! Thanks for submitting this, but I'm afraid that it conflicts with some changes I have in the mix for websockets. The good news is, once those changes are merged, you'll have this fixed, as all of our tests will be updated to run against web3 1.0

PR #10, opened 2017-08-29
Include hex-encoded operands in the debug logging of opcodes.

nejucomo:
This allows deeper diagnosis of low-level EVM issues when using testrpc by displaying the operands for each opcode in debug mode.

How is `ganache-core` tested, and how can I add test coverage for my changes? Also, does this follow any coding conventions for this project?

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -67,7 +67,10 @@ BlockchainDouble.prototype.initialize = function(accounts, callback) {
 
       if (options.debug == true) {
         self.vm.on('step', function(info){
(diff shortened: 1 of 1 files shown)


benjamincburns:
@nejucomo thanks for this excellent contribution, and sorry it took me so long to get 'round to reviewing it!

I'm not sure that it needs a test to be added for this functionality, per se, but if you'd like to add one, you can see plenty of example tests in the `test` directory. I'd suggest slotting one in wherever it makes sense, or even adding a new file, `test/logging.js`, to catch this one.

PR #12, opened 2017-09-07
Fix broken tests - update findPath reference from merkle-patricia-tree

jakub-wojciechowski:
After installing a fresh version of `ganache-core`,  the tests are failing with:

```
Uncaught TypeError: this._findPath is not a function
```

This is due to referencing an outdated method `_findPath` of `merkle-patricia-tree` library which has been recently made public in version `2.2.0` after [this commit](https://github.com/ethereumjs/merkle-patricia-tree/commit/7dd1fd54cfb41c1a13fd99b10c68fbca39342347).

The dependency to `merkle-patricia-tree` was set to look for anything above `2.1.2` thats why it took an updated version in a fresh built. 

I'm updating both the referenced method and increasing dependency so it won't use old cached version.

### DIFF
--- a/lib/utils/forkedstoragetrie.js
+++ b/lib/utils/forkedstoragetrie.js
@@ -27,7 +27,7 @@ function ForkedStorageTrie(db, root, options) {
 ForkedStorageTrie.prototype.keyExists = function(key, callback) {
   key = utils.toBuffer(key);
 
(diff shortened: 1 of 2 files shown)


benjamincburns:
I hate to say it, but I think this PR no longer applies. Nevertheless thanks for your contribution, @jakub-wojciechowski!

PR #13, opened 2017-09-11
Mine a block exactly at the given timestamp

jakub-wojciechowski:
Following a problem with testing time-checking condition we experienced at [OpenZeppelin](https://github.com/OpenZeppelin/zeppelin-solidity/pull/353),
I'm posting a solution that should make blockchain time checking robust and consistent through tests. 

Currently, setting the blockchain internal time requires first precomputing time period (and rounding) that should be added to it's current clock and then calling two api functions: `evm_increaseTime` and `evm_mine`. This may cause an additional delay during execution which may risk in having a flaky and indeterministic tests or using weaker assertions (like `>=` instead of `==`). 

I'm proposing setting the blockchain time in a single api call `evm_mineAtTimestamp` that contains the expected final `block.timestamp` as a parameter. This should eliminate the errors caused by precomputing time difference with rounding and the delays caused by calling two separate api functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ There’s also special non-standard methods that aren’t included within the or
 * `evm_revert` : Revert the state of the blockchain to a previous snapshot. Takes a single parameter, which is the snapshot id to revert to. If no snapshot id is passed it will revert to the latest snapshot. Returns `true`.
 * `evm_increaseTime` : Jump forward in time. Takes one parameter, which is the amount of time to increase in seconds. Returns the total time adjustment, in seconds.
 * `evm_mine` : Force a block to be mined. Takes no parameters. Mines a block independent of whether or not mining is started or stopped.
+* `evm_mineAtTimestamp` : Force a block to be mined exactly at the given timestamp. Takes one parameter which is the expected timestamp of a new block in seconds.
 
 # Unsupported Methods
 
--- a/README.md
+++ b/README.md
@@ -110,8 +110,7 @@ There’s also special non-standard methods that aren’t included within the or
 * `evm_snapshot` : Snapshot the state of the blockchain at the current block. Takes no parameters. Returns the integer id of the snapshot created.
 * `evm_revert` : Revert the state of the blockchain to a previous snapshot. Takes a single parameter, which is the snapshot id to revert to. If no snapshot id is passed it will revert to the latest snapshot. Returns `true`.
 * `evm_increaseTime` : Jump forward in time. Takes one parameter, which is the amount of time to increase in seconds. Returns the total time adjustment, in seconds.
-* `evm_mine` : Force a block to be mined. Takes no parameters. Mines a block independent of whether or not mining is started or stopped.
-* `evm_mineAtTimestamp` : Force a block to be mined exactly at the given timestamp. Takes one parameter which is the expected timestamp of a new block in seconds.
+* `evm_mine` : Force a block to be mined. Takes one optional parameter, which is the timestamp a block should setup as the mining time. Mines a block independent of whether or not mining is started or stopped.
 
 # Unsupported Methods
 
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -506,9 +506,10 @@ BlockchainDouble.prototype.processBlock = function(block, commit, callback) {
  * pending transactions.
  *
  * @param  {Function} callback Callback when transaction processing is finished.
+ * @param  {number} timestamp at which the block is mined
  * @return [type]              [description]
  */
-BlockchainDouble.prototype.processNextBlock = function(callback) {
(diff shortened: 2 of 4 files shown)


jakub-wojciechowski:
Hi @tcoulter, have you got any thoughts on that? This addition would greatly simplify tests development.

lib/blockchain_double.js
```diff
@@ -546,6 +547,11 @@ BlockchainDouble.prototype.processNextBlock = function(callback) {
   this.createBlock(function(err, block) {
     if (err) return callback(err);
 
+    //Overwrite block timestamp
+    if (timestamp) {
+        block.header.timestamp = to.hex(timestamp);
```
benjamincburns: Is it a requirement that block timestamps be monotonic? If so, we should add a check for this here.

README.md
```diff
@@ -111,6 +111,7 @@ There’s also special non-standard methods that aren’t included within the or
 * `evm_revert` : Revert the state of the blockchain to a previous snapshot. Takes a single parameter, which is the snapshot id to revert to. If no snapshot id is passed it will revert to the latest snapshot. Returns `true`.
 * `evm_increaseTime` : Jump forward in time. Takes one parameter, which is the amount of time to increase in seconds. Returns the total time adjustment, in seconds.
 * `evm_mine` : Force a block to be mined. Takes no parameters. Mines a block independent of whether or not mining is started or stopped.
+* `evm_mineAtTimestamp` : Force a block to be mined exactly at the given timestamp. Takes one parameter which is the expected timestamp of a new block in seconds.
```
benjamincburns: This is confusing to me. Reading this, I'd expect this not to instamine, but rather to wait until `now() >= timestamp` (likely via a `setTimeout`) and mine a block then, with whatever transactions are pending at the time.

Also, I think we are cluttering the `evm_` RPC namespace here. Since this is clearly a testing facility, perhaps we should use this as an opportunity to introduce a `testing_` namespace? I'm on the fence about this, however...

benjamincburns:
Thanks for the change, @jakub-wojciechowski. I've added comments/questions inline with the code.

benjamincburns:
On further thought, why add a whole new RPC call, at all? Why not add an optional `timestamp` argument to the existing `evm_mine` ?

lib/blockchain_double.js
benjamincburns: Even if it _isn't_ a requirement that block timestamps be monotonic, I think we should at least consider logging a warning message or something to that effect when `timestamp` is earlier than the parent block's timestamp.

jakub-wojciechowski: It's a good idea, I've added the warning.

README.md
jakub-wojciechowski: I've updated the description. I hope it's clear now. 
The main reason to have a separate function was to maintain backward compatibility, but I totally agree that I'd be way better to have a more concise api.  So, I followed a pattern you're using in different functions with optional argument repositioning. Let me know if it looks good for you.

jakub-wojciechowski:
@benjamincburns thanks for the awesome feedback! I've tried to follow your suggestions. Let me know if you have any comments.

benjamincburns:
@jakub-wojciechowski sorry it has taken me _ages_ to circle back to this... Per the [travis-ci build above](https://travis-ci.org/trufflesuite/ganache-core/builds/301639582), it seems that your changes cause numerous test failures.

That said, I'm still on the fence about whether or not it's beneficial to include this change. Can you give an example of the sort of logic you're testing which requires an exact equality check on the block timestamp?

My concern is that it's fairly well known that miners can manipulate the block timestamp to some degree (I think ±30 seconds on the major clients). Your contracts should be tolerant to fairly wide variance in block timestamps, especially in `payable` functions, or those which establish state which influences `payable` functions.

benjamincburns:
@jakub-wojciechowski actually, fixing the tests is simple -- The `timestamp` arg of `processNextBlock` should be optional. I'll add the fix to this PR (provided I have the ability to do so).

elie222:
Any updates as to when this will be merged?

benjamincburns:
This should likely get merged, however we need to handle the conflicts.

Ideally this would also trigger a call to `setTime` so that blocks mined after a block with an explicit time set would still have monotonic timestamps.

seesemichaelj:
@benjamincburns it does call `setTime`: https://github.com/trufflesuite/ganache-core/pull/13/files#diff-8f55aa0c345972c969bb69ce03983e2dR563

mikeseese:
@davidmurdoch FYI, looks like cloudflare commented on the wrong PR. Perhaps that comment should have gone on #4272?

mikeseese:
CC @jeffsmale90 ☝️

jeffsmale90:
Thanks @mikeseese !

PR #16, opened 2017-09-26
Support alternate DB providers

### LINKED ISSUE #5: Allow LevelDOWN instance to be provided
We have a test suite that ends up generating loads of files in the [TestRPC](https://github.com/ethereumjs/testrpc) database. When this suite is run on our CI service, during a Docker build, it slows the build down tremendously.

It'd be great if we could provide our own LevelDOWN instance. We could use https://github.com/level/memdown so that TestRPC never has to touch the disk.

I'm filing this issue here first since TestRPC relies on this package.

novemberborn:
Fixes #5 by allowing a LevelDOWN compatible instance to be provided through the 'db' option.

Tests are copied from `test/persistence.js`, which seemed appropriate.

This fork is being used internally so unfortunately I can't allow edits from maintainers on this PR.


benjamincburns:
@novemberborn thanks for the contribution! I had to rework the tests just a bit, but otherwise it's fine as-is.

PR #17, opened 2017-09-28
Continue with Geth API requests immediately after initialization

novemberborn:
Rather than repeatedly waiting for 100ms until the state is initialized,
continue API requests immediately by tracking callbacks and running them
directly.

This fork is being used internally so unfortunately I can't allow edits from maintainers on this PR.

### DIFF
--- a/lib/subproviders/geth_api_double.js
+++ b/lib/subproviders/geth_api_double.js
@@ -19,21 +19,31 @@ function GethApiDouble(options) {
   this.initialized = false;
 
   this.initialization_error = null;
+  this.post_initialization_callbacks = [];
 
   this.state.initialize(this.options, function(err) {
     if (err) {
       self.initialization_error = err;
     }
     self.initialized = true;
+
+    var callbacks = self.post_initialization_callbacks;
+    self.post_initialization_callbacks = [];
+
+    setImmediate(function() {
+      callbacks.forEach(function(callback) {
+        callback();
+      });
+    });
   });
 }
 
 GethApiDouble.prototype.waitForInitialization = function(callback) {
   var self = this;
   if (this.initialized == false) {
-    setTimeout(function() {
+    this.post_initialization_callbacks.push(function() {
       self.waitForInitialization(callback);
-    }, 100);
+    });
   } else {
     callback(this.initialization_error, this.state);
   }
@@ -48,7 +58,7 @@ GethApiDouble.prototype.handleRequest = function(payload, next, end) {
   }
 
   if (this.initialized == false) {
-    setTimeout(this.getDelayedHandler(payload, next, end), 100);
+    this.post_initialization_callbacks.push(this.getDelayedHandler(payload, next, end));
     return;
   }
 
--- a/lib/subproviders/geth_api_double.js
+++ b/lib/subproviders/geth_api_double.js
@@ -30,8 +30,8 @@ function GethApiDouble(options) {
     var callbacks = self.post_initialization_callbacks;
     self.post_initialization_callbacks = [];
 
-    setImmediate(function() {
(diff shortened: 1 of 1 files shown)


lib/subproviders/geth_api_double.js
```diff
@@ -19,21 +19,31 @@ function GethApiDouble(options) {
...
     self.initialized = true;
+
+    var callbacks = self.post_initialization_callbacks;
+    self.post_initialization_callbacks = [];
+
+    setImmediate(function() {
+      callbacks.forEach(function(callback) {
+        callback();
+      });
+    });
```
benjamincburns: I think this would be better as `callbacks.forEach(setImmediate)`

lib/subproviders/geth_api_double.js
```diff
@@ -19,21 +19,31 @@ function GethApiDouble(options) {
...
 }
 
 GethApiDouble.prototype.waitForInitialization = function(callback) {
   var self = this;
   if (this.initialized == false) {
-    setTimeout(function() {
+    this.post_initialization_callbacks.push(function() {
       self.waitForInitialization(callback);
-    }, 100);
+    });
```
benjamincburns: Why `push` an anonymous function w/ a call to `waitForInitialization` here? Why not just `push` the `callback` itself?

novemberborn: I've kept this similar to the previous behavior. Note how in the else condition the error and state are provided to `callback`. Pushing `callback` itself means that has to be replicated in the `initialize` callback above.

lib/subproviders/geth_api_double.js
novemberborn: Ah yes, that would prevent exceptions in one callback from preventing others from being run.

I'm still wrapping the `setImmediate()` call in a function though, to avoid accidentally forwarding the iteration arguments.

benjamincburns: Ah, right - good catch.

lib/subproviders/geth_api_double.js
benjamincburns: I think it'd be good to simplify the control flow a bit by making the functions pushed to `post_initialization_callbacks` have the same signature (`function(err, result)`), as this is a pattern used for callbacks most other places in the code.

If you have the time to do that, I'd appreciate it. Otherwise I'll likely refactor this myself but after I finish up some more pressing changes.

novemberborn: Sure, I just went for the fewest possible changes. 

I filed this PR on behest of a client, but sadly that contract has finished. I won't have time to follow-up on this, so please go ahead and use this as you will 😄

benjamincburns:
@novemberborn do you remember the reason for it? Was it causing some trouble, or were you just tightening things up? If you were just tightening things up, I'll probably close this PR and do something similar at a later date as part of a code quality audit.

I generally regard setTimeout as a code smell, so I like the idea of getting rid of it here, but given that this is a particularly "race conditiony" part of the code, I'm disinclined to change it without a lot more careful testing than I presently have time to give it.

novemberborn:
@benjamincburns I was writing tests (using https://github.com/trufflesuite/ganache-core/pull/16). Each test created its own RPC instance. The slowdown seemed unnecessary.

> I generally regard setTimeout as a code smell, so I like the idea of getting rid of it here, but given that this is a particularly "race conditiony" part of the code, I'm disinclined to change it without a lot more careful testing than I presently have time to give it.

The logic here isn't too bad, just looks like when the code was originally written nobody bothered actually signalling completion, hence the timeout.

benjamincburns:
@novemberborn I accepted the PR with a few minor changes. Thanks so much for your contribution!

PR #19, opened 2017-10-11
upgrade ethereumjs-vm to 2.3.1 (Byzantium)

jwasinger:
This PR upgrades ganache-core to be compatible with the Byzantium release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "ethereumjs-block": "~1.2.2",
     "ethereumjs-tx": "^1.3.0",
     "ethereumjs-util": "~5.1.0",
-    "ethereumjs-vm": "~2.2.0",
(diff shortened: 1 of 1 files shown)


jwasinger:
This is blocked until develop is merged into master.

jwasinger:
Ah this introduces two failing tests.

benjamincburns:
@jwasinger we moved this over to be based on the `develop` branch, as we'll likely merge it there first (`master` is the current release). It looks like the only tests failing are due to the `stateRoot` hash having changed, which given that no other tests failed and that the VM was updated, this is to be expected. I'm happy to accept this PR as-is, and I'll update the tests accordingly.

jwasinger:
Thanks @benjamincburns don't hesitate to reach out to me if anything broken appears as a result of this

PR #20, opened 2017-10-19
Process calls & gas estimates serially

benjamincburns:
Fixes ethereumjs/testrpc#287


benjamincburns:
Will address ethereumjs/testrpc#336 in a separate PR

PR #21, opened 2017-10-20
Fix issues with json responses matching spec

mjkoo:
Currently the json responses from some testrpc methods don't conform to the specification at https://github.com/ethereum/wiki/wiki/JSON-RPC . Using go-ethereum's ethclient module to interact with testrpc fails due to their strict parsing, but as the specification is explicit about things such as hex number formatting etc. their behavior seems to be right and testrpc's wrong.

This PR brings some of the formatting into line with the spec, but probably is not exhaustive.


benjamincburns:
Wow - thanks for this!

It'll take me some time to verify these changes. Also since the names of some things have changed, this is technically a breaking change. That's not a major issue, but we'll need to consider carefully how we bring it into the fold, as it will likely drive a breaking change for TestRPC, and perhaps Truffle as well.

lib/subproviders/geth_api_double.js
```diff
@@ -139,23 +139,24 @@ GethApiDouble.prototype.eth_getBlockByNumber = function(block_number, include_fu
...
       hash: to.hex(block.hash()),
       parentHash: to.hex(block.header.parentHash),
-      nonce: to.hex(block.header.nonce),
+      nonce: to.hexWithLeftPadding(block.header.nonce, 16),
       sha3Uncles: to.hex(block.header.uncleHash),
       logsBloom: to.hex(block.header.bloom),
       transactionsRoot: to.hex(block.header.transactionsTrie),
       stateRoot: to.hex(block.header.stateRoot),
-      receiptRoot: to.hex(block.header.receiptTrie),
+      receiptsRoot: to.hex(block.header.receiptTrie),
```
benjamincburns: Verified: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getblockbyhash

:+1:

lib/subproviders/geth_api_double.js
```diff
@@ -139,23 +139,24 @@ GethApiDouble.prototype.eth_getBlockByNumber = function(block_number, include_fu
...
-      totalDifficulty: to.hex(block.header.difficulty), // TODO: Figure out what to do here.
-      extraData: to.hex(block.header.extraData),
-      size: to.hex(1000), // TODO: Do something better here
-      gasLimit: to.hex(block.header.gasLimit),
-      gasUsed: to.hex(block.header.gasUsed),
-      timestamp: to.hex(block.header.timestamp),
+      difficulty: to.hexWithoutLeadingZeroes(block.header.difficulty),
+      totalDifficulty: to.hexWithoutLeadingZeroes(block.header.difficulty), // TODO: Figure out what to do here.
+      extraData: to.hexWithEvenLength(block.header.extraData),
+      mixHash: to.hex(block.header.mixHash),
```
benjamincburns: @mjkoo I don't see a `mixHash` field here: https://github.com/ethereum/wiki/wiki/JSON-RPC#returns-26

lib/utils/to.js
```diff
@@ -29,8 +29,25 @@ module.exports = {
...
+  hexWithEvenLength: function(val, width) {
+    val = this.hex(val).replace("0x", "");
+    if (val.length % 2 == 1) {
+      val = "0" + val;
+    }
+    return utils.addHexPrefix(val);
+  },
+
+
+  hexWithLeftPadding: function(val, width) {
```
benjamincburns: Should this validate the `width` parameter? or at least treat it as 0 when it's falsey?

benjamincburns:
Just a note to myself for when I come back to this later - padding rules are defined here: https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding

lib/subproviders/geth_api_double.js
mjkoo: Hmm yes you are right, that change was made due to a block deserialization issue in go-ethereum due to this field here https://github.com/ethereum/go-ethereum/blob/master/core/types/block.go#L84 but does not appear in the spec, so this may be a go-ethereum issue.

lib/utils/to.js
mjkoo: Resolved in f3d3c22

benjamincburns: Just coming back to this - there are really only two ways prescribed by [the spec](https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding) to encode hex values: either as a compact representation with no padding for `QUANTITY` types or as a left-padded representation with two digits per byte for fixed-size `DATA` types.

As such, I feel like the correct fix for padding is to make the object's definition read a bit more like the spec, and then centralize the decision as to how best to pad a particular value. I'll go ahead and rework these changes into a new PR @mjkoo, and if you don't mind, can you please review for me when it's ready?

mjkoo: Yeah that makes sense to me, thought about doing that but didn't want to make the patch too intrusive. Sure would be happy to :)

benjamincburns:
Hi @mjkoo, were you finding that the fields you changed weren't being padded correctly?

I've switched `receiptRoot` to `receiptsRoot` in d12ca37. Otherwise, I'm not sure that any other changes are necessary, so I'm going to go ahead and close this. Please let me know if this was in error, and thanks for bringing this to my attention!

mjkoo:
Yes I was, when attempting to interact with testrpc from go-ethereum I was receiving invalid responses. I will test again with the new release soon to verify.

benjamincburns:
@mjkoo I went through every response field I could find and forced strict adherence to the spec, but found that doing so triggered a surprising number of test failures. It will be a little while before I can get back to this one, but please do raise a separate issue for any padding issues you may find. I have the code stashed, so there’s no needn’t for another PR.

mjkoo:
Ack, will do

PR #23, opened 2017-10-26
Add `status` field to receipt, per https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt

benjamincburns:
As of Byzantium, transaction receipts should have a `status` field which have a value of 1 for success, and 0 for failure.

https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt

### DIFF
--- a/test/vm.js
+++ b/test/vm.js
@@ -1,0 +1,98 @@
+var Web3 = require('web3');
+var Transaction = require('ethereumjs-tx');
+var utils = require('ethereumjs-util');
+var assert = require('assert');
+var TestRPC = require("../index.js");
+var solc = require("solc");
+var fs = require("fs");
+var to = require("../lib/utils/to");
+var clone = require("clone");
+
+// Thanks solc. At least this works!
+// This removes solc's overzealous uncaughtException event handler.
+process.removeAllListeners("uncaughtException");
+
+var logger = {
+  log: function(message) {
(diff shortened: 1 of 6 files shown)


test/vm.js
```diff
@@ -0,0 +1,98 @@
...
+
+            assert.notEqual(receipt, null, "Transaction receipt shouldn't be null");
+            assert.equal(receipt.status, 0, "Reverted transactions should have a status of 0.");
+            done();
+          });
+        });
+      }
+
+      if (!instance.address) {
+        web3.eth.getTransactionReceipt(instance.transactionHash, function(err, receipt) {
```
benjamincburns: Probably should always get the tx receipt here and assert that `status` is 1...

benjamincburns: I think we must have a race condition, because I seem to need both code paths for the conditional at line 82 for the test to succeed...

benjamincburns: also I should refactor this to use promises - they're supported in the lts release of node.

benjamincburns:
Going to rebase this one on #27, as the update to solc 0.4.18 shouldn't have been done as part of this PR.

benjamincburns:
Rather than `git push -f` I'll just close this an open it under a new branch.

PR #25, opened 2017-10-29
fix estimateGas bug when basefee > defaultTransactionGasLimit

ewingrj:
The gasLimit for an estimateGas call was set to the
defaultTransactionGasLimit of 90000 gas. This would cause
ethereumjs-vm runCall to throw the error 'base fee exceeds
gas limit'.

By setting the gasLimit to the block gas limit, this allows use to
correctly estimate gas for tx's where the basefee > gasLimit, such as
when deploying a contract.

This is how geth does it. https://github.com/ethereum/wiki/wiki/JSON-RPC#parameters-25

### DIFF
--- a/lib/subproviders/geth_api_double.js
+++ b/lib/subproviders/geth_api_double.js
@@ -277,12 +277,13 @@ GethApiDouble.prototype.eth_call = function(tx_data, block_number, callback) {
 };
 
 GethApiDouble.prototype.eth_estimateGas = function(tx_data, block_number, callback) {
+  tx_data.gas = this.state.blockchain.blockGasLimit;
   this.state.queueTransaction("eth_estimateGas", tx_data, callback);
 };
 
 GethApiDouble.prototype.eth_getStorageAt = function(address, position, block_number, callback) {
   this.state.queueStorage(address, position, block_number, callback);
-}
+};
 
 GethApiDouble.prototype.eth_newBlockFilter = function(callback) {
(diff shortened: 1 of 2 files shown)


benjamincburns:
@perissology thanks for the PR!

Unfortunately, I'm not sure if accepting this change is in the best interest of our goals as a testing tool. While it's true that geth does default to the block limit for `estimateGas` calls, that's not a prescription of the RPC protocol, and other clients may behave differently.

Since the protocol doesn't define a different gas limit for `estimateGas` calls, this seems like a fairly reasonable default. Further, since we're a testing tool, our users will wind up writing more robust Dapps and/or contracts if we choose _less_ permissive defaults, thereby decreasing the likelihood that they'll discover issues like these for the in production rather than during dev/test.

perissology:
@benjamincburns do you know how parity and others do it?

benjamincburns:
@perissology I've recently had my mind changed on this one. See conversation on ethereumjs/testrpc#411.

cc @MicahZoltu - turns out we have a PR waiting in the wings on this one!

Will review and likely merge on Monday.

benjamincburns:
Sorry it took me a little longer than I would've liked to merge this!

Thanks @perissology for your help!

MicahZoltu:
Has this made it to an NPM release of TestRPC yet?  If not, what should I monitor for with regards to a release containing this fix?

benjamincburns:
@MicahZoltu thanks for asking, I think I might've missed this one on the v6.0.3 release notes. It should be in there.

PR #29, opened 2017-10-30
Ensured gasPrice can be set to 0

ellmetha:
It should be possible to set the `gasPrice` to ` 0` if desired but the current implementation doesn't allow it. This commit fixes this behaviour.


benjamincburns:
Easy peasy - thanks for the contribution!

PR #30, opened 2017-11-01
Remove redundant require statements

joojis:


benjamincburns:
Thanks for the cleanup, @joojis!

PR #31, opened 2017-11-12
importRawKey added

jbaylina:
I added support for `personal.importRawKey`.

This is needed when you want to create transactions from a specific account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Both `.provider()` and `.server()` take a single object which allows you to spec
 * `"seed"`: Use arbitrary data to generate the HD wallet mnemonic to be used.
 * `"total_accounts"`: `number` - Number of accounts to generate at startup.
 * `"fork"`: `string` or `object` - When a `string`, same as `--fork` option above. Can also be a Web3 Provider object, optionally used in conjunction with the `fork_block_number` option below.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -101,6 +101,7 @@ The RPC methods currently implemented are:
 * `personal_listAccounts`
 * `personal_lockAccount`
 * `personal_newAccount`
+* `personal_importRawAccount`
```
benjamincburns: I think you meant `personal_importRawKey` here.

lib/subproviders/geth_api_double.js
```diff
@@ -390,6 +390,14 @@ GethApiDouble.prototype.personal_newAccount = function(password, callback) {
   callback(null, account.address);
 };
 
+GethApiDouble.prototype.personal_importRawKey = function(rawKey, password, callback) {
+  var account = this.state.createAccount({ secretKey: rawKey });
```
benjamincburns: This should probably be updated to make use of `options.defaultBalanceEther` (see latest README.md in `develop` branch)

benjamincburns: Actually, scratch that - we likely want this account to be added with a zero balance.

README.md
benjamincburns: No need for you to correct this - I'll do it on merge.

benjamincburns:
Thanks for the contribution, @jbaylina! Merging this one now.

PR #35, opened 2017-11-27
Fix to.hexWithoutLeadingZeroes for input 0

attente:


attente:
Sorry, I totally didn't see https://github.com/trufflesuite/ganache-core/pull/32 until now. Closing.

PR #41, opened 2017-12-25
automated tests for bounty 450

0xNPE:
`mocha -b test/requests.js`

The exceptions are thrown on another thread because of async, so I couldn't use assert.throws to catch them. Comment 1 test out to test the other.

### DIFF
--- a/test/stability.js
+++ b/test/stability.js
@@ -1,6 +1,7 @@
 var Web3 = require('web3');
 var assert = require('assert');
 var TestRPC = require("../index.js");
+var utils = require('ethereumjs-util');
 
 var logger = {
   log: function(message) {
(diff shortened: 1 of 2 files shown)


test/stability.js
```diff
@@ -88,3 +89,52 @@ describe("TestRPC", function(done) {
...
+    web3.eth.getAccounts(function(err, accs) {
+      if (err) return done(err);
+
+      accounts = accs;
+      done();
+    });
+  });
+
+  it("should not cause 'get' of undefined", function(done) {
+    process.prependOnceListener('uncaughtException', function(err) {
```
benjamincburns: These tests demonstrate the problem beautifully, but if the problem is fixed, they'll hang... Ideally this would use `assert.throws`, but I can imagine that not being possible if the error is thrown on a function run by `setImmediate` or similar (my big gripe with the `async` module).

At the very least, perhaps we can call `done()` on a timeout? That feels a bit hokey, but if there really is no way to directly catch the exception, I think that's all that can be done, right?

benjamincburns: Actually, I think that's a fine way to do it (with `setTimeout`, because even in the cases where the "it's fixed" timeout fires before the uncaught exception handler is thrown, mocha will complain to us that `done` was called twice, which in this case indicates a (late) test failure.

benjamincburns: Doh, forget I said that... You can just call `done()` from the `revert` callback on line 120, and the `call` callback on line 135.

benjamincburns: I'll just make this change on merge.

rosario:
The following pull request https://github.com/ethereumjs/merkle-patricia-tree/pull/29 fix the issues.

After removing the skip from the tests, all ganache-core tests pass when merkle-patricia-tree is locally linked.

test/stability.js
```diff
@@ -88,3 +89,52 @@ describe("TestRPC", function(done) {
...
+      done(err);
+    });
+
+    var blockchain = provider.manager.state.blockchain;
+    blockchain.vm.stateManager.checkpoint(); // processCall #1
+    // processNextBlock triggered by interval mining which at some point calls vm.stateManager.commit() and blockchain.putBlock()
+    blockchain.processNextBlock(function(err, tx, results) {
+      blockchain.vm.stateManager.revert(function() { // processCall #1 finishes
+        blockchain.latestBlock(function (err, latestBlock) { 
+          blockchain.stateTrie.root = latestBlock.header.stateRoot; // getCode #1 (or any function with this logic)
```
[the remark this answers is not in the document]
0xNPE: This might not be possible because the function calls cleans up after itself, ie. 
```
// Finally, put the stateRoot back for good
trie.root = currentStateRoot;
```

The scenario requires certain execution to be intertwined

I did a quick test with getAccount and the test passed when it should have failed.

PR #42, opened 2017-12-27
Fixed logging bug

illuzen:
Updating block time triggered caused "only 53 bits of precision" error. Discovered when issuing evm_increaseTime RPC. Apparently block timestamps are already numbers.


illuzen:
I discovered this via stacktrace!

vim /usr/local/lib/node_modules/ganache-cli/build/cli.node.js +90697

benjamincburns:
@illuzen thanks for submitting this. This should already be fixed in current `develop` branch, however. The real problem was that `to.number` wasn't detecting when it was being passed an instance of `Number`. I've corrected this fairly recently.

illuzen:
@benjamincburns should I still be seeing this with 

Ganache CLI v7.0.0-beta.0 (ganache-core: 3.0.0-beta.0)

benjamincburns:
@illuzen I think it was fixed after I cut that release, sorry.

PR #48, opened 2018-01-06
strip leading zeroes for eth_getBalance()

Firescar96:
I got some errors due to formatting of results from ganache preventing it from being used as a drop in replacement for geth. You can reproduce the errors with the following steps.

1.
start ganache
`ganache-cli --deterministic`

2.
and run this go code using the go-ethereum libraries
```go
package main

import (
	"context"
	"log"
	"math/big"

	"github.com/ethereum/go-ethereum/common"
	"github.com/ethereum/go-ethereum/ethclient"
)

func main() {
	const Account = "0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1"

	ethAPI, err := ethclient.Dial("http://localhost:8545")
	if err != nil {
		log.Fatalf("Failed to connect to the Ethereum client: %v", err)
	}

        //result, err := ethAPI.BlockByNumber(context.Background(), big.NewInt(18))
	result, err := ethAPI.BalanceAt(context.Background(), common.HexToAddress(Account), big.NewInt(0))
	if err != nil {
		log.Fatalf("Get Balance failed: %v", err)
	}

	log.Println(result)
}
```
The code here is fix errors for calls to `eth_getBalance` and `eth_getBlockByNumber`

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -208,7 +208,7 @@ StateManager.prototype.gasPrice = function() {
 StateManager.prototype.getBalance = function(address, number, callback) {
   this.blockchain.getBalance(address, number, function(err, balance) {
     if (balance) {
-      balance = to.hex(balance);
+      balance = to.rpcQuantityHexString(balance);
     }
     callback(err, balance);
   });
@@ -217,7 +217,7 @@ StateManager.prototype.getBalance = function(address, number, callback) {
 StateManager.prototype.getTransactionCount = function(address, number, callback) {
   this.blockchain.getNonce(address, number, function(err, nonce) {
     if (nonce) {
-      nonce = to.hexWithoutLeadingZeroes(nonce);
+      nonce = to.rpcQuantityHexString(nonce);
     }
     callback(err, nonce);
   });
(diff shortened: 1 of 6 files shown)


benjamincburns:
Thanks for submitting this, @Firescar96. Unfortunately, I think this perpetuates a somewhat longstanding problem we have in ganache-core, which is that we have serialization logic mixed in with application logic.

I think it'd take a nice step in the right direction if you were to add `rpcDataHexString(item, [byteLength])` and `rpcQuantityHexString(item)` to the `to` module. That way we can be very certain that when serializing output from `Buffer`/`BigNumber`/`BN`/`Number` we conform to the prescribed serialization format.

The `byteLength` field of `rpcDataHexString` should be optional, as for certain fields (e.g. `extraData` this is variable, and should be based on the size of the buffer, with null bytes trimmed.

Firescar96:
What's the difference between `rpcDataHexString(item, [byteLength])` and `rpcQuantityHexString(item)`?

benjamincburns:
@Firescar96, yes, sorry. I should've explained that before. The naming there is to correspond with the `QUANTITY` and `DATA` types defined by the [JSON RPC wiki page](https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding).

benjamincburns:
This change looks great! Thanks for the contribution!

PR #53, opened 2018-01-12
Queue requests

benjamincburns:
Based on #49.

I'm not a huge fan of this approach, but at least it will stop things from crashing...

### DIFF
--- a/lib/provider.js
+++ b/lib/provider.js
@@ -49,6 +49,9 @@ function Provider(options) {
   this.manager = gethApiDouble;
   this.sendAsync = this.send.bind(this);
   this.send = this.send.bind(this);
+  this.close = this.close.bind(this);
+  this._queueRequest = this._queueRequest.bind(this);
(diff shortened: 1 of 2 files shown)


lib/provider.js
```diff
@@ -101,12 +104,50 @@ Provider.prototype.send = function(payload, callback) {
...
+
+  if (args) {
+    self.engine.sendAsync(args.payload, (err, result) => {
+      if (self._requestQueue.length > 0) {
+        setImmediate(self._processRequestQueue)
+      }
+      args.callback(err, result)
+      self._requestInProgress = false;
+    })
+  }
```
gnidan: Looks like you might need to negate `_requestInProgress` if `args` is undefined.

benjamincburns:
Correct. Good catch!

benjamincburns:
Changed base to make this slightly easier to review. Will change back to `develop` once it's mergeable.

PR #57, opened 2018-01-22
Fix for trufflesuite/ganache-cli#354 

yarrumretep:
evm_mine returns true in the result to ensure valid JSON_RPC response.

### DIFF
--- a/test/mining.js
+++ b/test/mining.js
@@ -112,7 +112,7 @@ describe("Mining", function() {
         id: new Date().getTime()
       }, function(err, result) {
         if (err) return reject(err);
-        assert.notDeepEqual(result.result, "0x0", "evm_mine should return result = '0x0'");
+        assert.deepEqual(result.result, "0x0");
         accept(result);
       })
     });
--- a/test/mining.js
+++ b/test/mining.js
@@ -110,9 +110,10 @@ describe("Mining", function() {
         jsonrpc: "2.0",
         method: "evm_mine",
         id: new Date().getTime()
-      }, function(err) {
+      }, function(err, result) {
         if (err) return reject(err);
-        accept();
+        assert.notDeepEqual(result.result, "0x0", "evm_mine should return result = '0x0'");
+        accept(result);
       })
     });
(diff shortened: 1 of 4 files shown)


yarrumretep:
@benjamincburns - ok - updated per your direction.

yarrumretep:
hrm - travis failed, but i think the tests all pass with both of the latest commits?

benjamincburns:
I’ll have a look today. If it just needs a quick fix, I’ll handle it,
provided you ticked the box that lets project members push changes. Thanks
for your contribution!

yarrumretep:
I ticked "allow edits from maintainers" - assume that's what you are referring to.  

It's really a minor contribution - but happy to have my feet wet - hopefully I can contribute more meaningfully in the future!

Rock on!

test/mining.js
```diff
@@ -111,9 +111,10 @@ describe("Mining", function() {
         jsonrpc: "2.0",
         method: "evm_mine",
         id: new Date().getTime()
-      }, function(err) {
+      }, function(err, result) {
         if (err) return reject(err);
-        accept();
+        if(result.result !== '0x0') return reject(new Error("evm_mine should return result = '0x0':" + JSON.stringify(result)))
```
benjamincburns: This works fine, but it's more idiomatic to write `assert.notDeepEqual(result.result, '0.0')`

benjamincburns: No need to resubmit - I'll change it when I merge

benjamincburns:
Looks like the test failure was an unrelated test timing out, which seems to happen on the node lte/boron release from time to time. I'm rerunning that build and provided it doesn't fail a second time around, I'll merge.

benjamincburns:
rebased changes on current develop and pushed back to upstream repo prior to merging

benjamincburns:
Testing snob protip: when one corrects a contributor's nonidiomatic assertion logic, one looks rather foolish when the "correction" breaks the contributor's test. :joy:

Fixed ;-)

PR #61, opened 2018-01-30
Allow newer versions of Solidity 0.4.x

celeduc:
solc should not be restricted to only to version 0.4.18 because [semantic versioning](http://solidity.readthedocs.io/en/develop/layout-of-source-files.html#version-pragma) specifies that only non-breaking changes will be included in the 0.4.x series. Locking it to 0.4.18 means that users of Truffle cannot use Solidity 0.4.19 with its bug fixes, nor 0.4.20 when it comes along, etc.

Did some copyediting on the README to trigger builds, as Travis was being difficult.

See also https://github.com/trufflesuite/truffle/pull/784


celeduc:
For whatever reason this is [failing](https://travis-ci.org/trufflesuite/ganache-core/builds/335379659) on the lts/boron image for trufflesuite, but [passing](https://travis-ci.org/celeduc/ganache-core/builds/335384415) on my identical Travis config. I am stumped.

PR #62, opened 2018-01-31
fixes trufflesuite/ganache-cli#471 tx rejection error reporting

benjamincburns:


lib/provider.js
```diff
@@ -9,6 +9,7 @@ var GethApiDouble = require('./subproviders/geth_api_double');
 var BlockTracker = require('./block_tracker');
 
 var RuntimeError = require("./utils/runtimeerror");
+var TXRejectedError = require("./utils/txrejectederror");
```
benjamincburns: unused import

lib/provider.js
```diff
@@ -168,47 +176,39 @@ var transactionMethods = [
...
 
   if (!err) return newResponse;
 
-  // make sure we always return the transaction hash on failed transactions so
-  // the caller can get their tx receipt
-  if (self._isTransactionRequest(request) && err.hashes) {
-    newResponse.result = err.hashes[0];
+  // TODO: for next major release: move reporting of tx hash on error to error
+  // field to prevent poorly-written clients which assume that the existence of
+  // the "result" field implies no errors from breaking.
```
benjamincburns: I've witnessed this with web3 1.0 - when both `error` and `result` were set on the response sometimes it'd interpret the error correctly and sometimes it wouldn't.

lib/provider.js
```diff
@@ -168,47 +176,39 @@ var transactionMethods = [
...
+      // the caller can get their tx receipt. This breaks JSONRPC 2.0, but it's how
+      // we've always done it.
+      newResponse.result = err.hashes[0]
+
+      // If we executed code as part of this request, check whether we're supposed
+      // to cause web3 and friends to throw by including the error field in our
+      // response. If we don't report an error, the caller must fetch inspect the
+      // `status` field of the receipt to detect an error, and check the ganache
+      // logging output to see the error message. This is how all major clients
+      // work today, so this is the default option.
```
benjamincburns: Need to update this comment, as it's not the default anymore

lib/utils/txrejectederror.js
```diff
@@ -0,0 +1,15 @@
+var inherits = require("util").inherits;
+
+// raised when the transaction is rejected prior to running it in the EVM.
+function TXRejectedError(message) {
```
benjamincburns: Creating this type wasn't absolutely necessary in the end, but I thought it helped to express the intent behind the errors being returned, so I left it in.

lib/utils/txrejectederror.js
```diff
@@ -0,0 +1,15 @@
...
+
+  // Why not just Error.apply(this, [message)? See
```
benjamincburns: Typo.

test/hex.js
```diff
@@ -30,15 +30,24 @@ describe("to.rpcQuantityHexString", function() {
   });
 });
 
-function noLeadingZeros(result) {
+function noLeadingZeros(method, result, path) {
```
benjamincburns: This test needs improvement, as it sometimes fails when fixed-width fields naturally have their most significant nibble zeroed out (e.g. hashes, etc). Further, if taken as documentation/specification, it's just plain wrong. It's not that we care about leading zeros, it's that we care about whether the field is fixed width or not.

further improvement is out of scope for this PR, however. I just added these changes because the test failed on me on one of the runs, and I wanted it to print out a better message next time.

test/requests.js
```diff
@@ -1066,7 +1066,7 @@ var tests = function(web3) {
     it("should return more than 0 accounts", function(done) {
       web3.eth.personal.getAccounts(function(err, result) {
         if (err) return done(err);
-        assert.equal(result.length, 3);
+        assert.equal(result.length, 13);
```
benjamincburns: Bit of a yak shave, this one...

In order to lock `accounts[1]` in the `transaction_rejection` tests, I had to fix the bug which prevents locking of the initially generated accounts. That is, that they aren't registered as personal accounts from the start. This in turn forced this change.

benjamincburns:
This also fixes ganache-cli#405

PR #67, opened 2018-02-14
Move ganache-cli init logic into ganache-core, makes debugging way easier

benjamincburns:


benjamincburns:
@davidmurdoch is doing some work which invalidates the need for this

PR #69, opened 2018-02-15
Allow defaultBlock parameter for calls

cgewecke:
Modifies `processCall` to accept a `defaultBlock` parameter and run against the specified block. Borrows logic from `processTrace` as suggested by @benjamincburns in a discussion last week. 

Relevant open issue is [ganache-cli 385](https://github.com/trufflesuite/ganache-cli/issues/385). The feature request there is to support `defaultBlock` where ever the RPC spec permits. At the moment this PR is just one step in that direction - maybe it should be expanded? 

Happy to make whatever revisions might be helpful :)

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -420,22 +420,24 @@ BlockchainDouble.prototype.sortByPriceAndNonce = function() {
 
 BlockchainDouble.prototype.processCall = function(tx, blockNumber, callback) {
   var self = this;
-  var startingStateRoot = self.stateTrie.root;
+  var startingStateRoot;
 
   var cleanUpAndReturn = function (err, result, changeRoot) {
-    // For defaultBlock, undo state root changes
-    if (changeRoot){
-      self.stateTrie.root = startingStateRoot;
-    }
-
     self.vm.stateManager.revert(function (e) {
+      // For defaultBlock, undo state root changes
+      if (changeRoot){
+        self.stateTrie.root = startingStateRoot;
+      }
       callback(err || e, result);
     });
   };
 
   var runCall = function(tx, changeRoot, err, block){
+    if (err) return callback(err);
+
     // For defaultBlock, use that block's root
     if (changeRoot){
+      startingStateRoot = self.stateTrie.root;
       self.stateTrie.root = block.header.stateRoot;
     }
 
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -77,11 +77,11 @@ BlockchainDouble.prototype.initialize = function(accounts, callback) {
             try {
               number = to.number(number)
             } catch (e) {
-              // Do nothing; must be a block hash. 
+              // Do nothing; must be a block hash.
             }
 
             self.getBlock(number, done)
-          } 
+          }
         },
         enableHomestead: true,
         activatePrecompiles: true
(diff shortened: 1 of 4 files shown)


lib/blockchain_double.js
```diff
@@ -418,29 +418,35 @@ BlockchainDouble.prototype.sortByPriceAndNonce = function() {
...
-  self.latestBlock(function (err, latestBlock) {
-    if (err) {
-      return callback(err);
+    self.vm.stateManager.revert(function (e) {
+      callback(err || e, result);
+    });
+  };
+
+  var runCall = function(tx, changeRoot, err, block){
+    // For defaultBlock, use that block's root
```
cgewecke: Need to handle this err.

lib/blockchain_double.js
```diff
@@ -418,29 +418,35 @@ BlockchainDouble.prototype.sortByPriceAndNonce = function() {
   self.pending_transactions = sorted_transactions;
 };
 
-BlockchainDouble.prototype.processCall = function(tx, callback) {
+BlockchainDouble.prototype.processCall = function(tx, blockNumber, callback) {
   var self = this;
+  var startingStateRoot = self.stateTrie.root;
```
benjamincburns: I wouldn't address this now, but the fact that we have this root updating nonsense slung all over the place is a bit of of a code smell, suggesting that we're really not encapsulating this concern very well.

I have a branch that creates a new instance of the trie and EVM on demand with the correct root, but the way it's structured really isn't much better. If you feel like taking this on, let me know. Even if you don't, it'd be good to hash out a plan of attack with someone else on zoom if you're willing.

lib/blockchain_double.js
```diff
@@ -418,29 +418,35 @@ BlockchainDouble.prototype.sortByPriceAndNonce = function() {
...
 
+  var cleanUpAndReturn = function (err, result, changeRoot) {
+    // For defaultBlock, undo state root changes
+    if (changeRoot){
+      self.stateTrie.root = startingStateRoot;
+    }
```
benjamincburns: No need for the changeRoot conditional here. There's no direct side-effect to the assignment. I think root is a property, but under the hood it's just updating a field to point at whatever buffer object you give it. IIRC, it doesn't even lock (hence trufflesuite/ganache-cli#453

lib/blockchain_double.js
benjamincburns: Generally I feel like if we're defining functions in functions like this, we've got a cohesion problem (trying to do too much in one place).

Also we should probably be using `pify` or `util.promisify` to promisify our callback hell, as well.

Not sure whether to suggest this be changed, though - I'm like inches away from a big refactoring here, anyway. Might be easier to just solve it then.

test/requests.js
```diff
@@ -703,6 +705,38 @@ var tests = function(web3) {
...
 
+    // NB: relies on the previous test setting value to 25 and the contract deployment setting
+    // original value to 5. `contractCreationBlockNumber` is set in the first test of this
+    // describe block.
+    it("should be able to read data via a call at a specified block (eth_call)", function(done){
+      var starting_block_number = null;
+      var call_data = contract.call_data;
+      call_data.to = contractAddress;
+
+      // TODO: Removing this callback hell would be nice.
```
benjamincburns: Callback hell can be somewhat removed now - web3 1.0 returns PromiEvents!

Also I wonder if we're not leaning too heavily on `requests.js` for our testing... I almost feel like this whole file should be broken out by method name.

Again, nothing that you need to change here, just idle musing.

benjamincburns: Actually, ideally we'd test setting a specific block number, as well as `initial`, `latest`, and `pending`... If you're up for it, perhaps you could create `test/call.js` if it doesn't already exist, and do that there?

cgewecke: Definitely - that's a good idea.

lib/blockchain_double.js
cgewecke: Yes!

test/requests.js
cgewecke: @benjamincburns Close to pushing in the revisions - would you like to move everything `eth_call` (and maybe even `estimateGas`?) into `test/call.js`? The file currently exists but only contains tests for an edge case where running a call could exceed the gasLimit.

lib/blockchain_double.js
cgewecke: In the end this seemed to be necessary and needs to happen after the `vm.revert`. No idea why. Expanded the tests to include a case where we call `latest` immediately after calling a specific block and verify that the response is correct.

test/requests.js
cgewecke: @benjamincburns Actually - restructuring the `requests.js` is likely a good idea but WDYT about doing it in a separate PR? Have a stack of PRs headed your way this week and maybe that one would be a good way to cap them off.

test/requests.js
```diff
@@ -703,6 +705,78 @@ var tests = function(web3) {
...
+
+        startingBlockNumber = result;
+        return web3.eth.call(call_data)
+
+      }).then(function(result){
+
+        assert.equal(to.number(result), 25, "value retrieved from latest block should be 25");
+        return web3.eth.call(call_data, contractCreationBlockNumber)
+
+      }).then(function(result){
```
cgewecke: Is this too spacey?

benjamincburns: doesn't bother me any – generally I'm pro whitespace

PR #72, opened 2018-02-20
Return storage changes in structLogs on debug_traceTransaction, fixes #46

### LINKED ISSUE #46: Return storage in debug_traceTransaction structLogs

benjamincburns:


test/debug.js
```diff
@@ -99,7 +99,9 @@ describe("Debug", function() {
 
         assert.equal(lastop.op, "STOP");
         assert.equal(lastop.gasCost, 1);
-        assert.equal(lastop.pc, 131);
+        assert.equal(lastop.pc, 141);
+        assert.equal(lastop.storage['0x0000000000000000000000000000000000000000000000000000000000000000'], '0x000000000000000000000000000000000000000000000000000000000000001a')
+        assert.equal(lastop.storage['0x0000000000000000000000000000000000000000000000000000000000000001'], '0x000000000000000000000000000000000000000000000000000000000000001f')
```
gnidan: I believe the trace reports values without leading `0x` for `memory` and `stack` – I can't tell from this PR if that behavior is preserved here or not (test suggests maybe no?)

lib/blockchain_double.js
```diff
@@ -787,6 +796,53 @@ BlockchainDouble.prototype.processTransactionTrace = function(hash, params, call
...
+
+  var argsNum = event.opcode.in;
+  var args = event.stack.slice(-argsNum)
+    .map((arg) => to.hex(arg));
+
+  if (storageStack.currentDepth > event.depth) {
+    storageStack.pop()
+  } if (storageStack.currentDepth < event.depth) {
+    storageStack.stack.push({})
+  }
```
gnidan: Just a note even though it's probably not worth being actionable right now: I'd be apprehensive about introducing another stateful variable passed around as reference (or maybe I am just shilling Redux too hard)

PR #75, opened 2018-03-01
implement eth_estimateGas as binary search

### LINKED ISSUE #26: estimateGas returns invalid value when tx has a gasRefund
When calling `eth_estimateGas`, the returned value is incorrect for the following cases:

1. transaction set a non-0 state variable to 0 thus resulting in a gas refund. The returned value is the transaction receipt gasUsed value, but in reality you need the tx receipt gasUsed + gasRefund to execute the tx, as gas is refunded at the end of the tx.

2. If there is a strange gas check in the contract function. ex:

```
contract FunkyGasPattern {
	string public field;

	function SetField(string value) {
	    // This check will screw gas estimation! Good, good!
   	    if (msg.gas < 100000) {
		    throw;
	    }
	    field = value;
        }
}
```

 

## Expected Behavior

estimateGas provides the correct gas to successfully execute a tx.
## Current Behavior

estimateGas does not always provide the correct gas to successfully execute a tx.
## Possible Solution

A simple fix for case 1 is in `statemanager.processGasEstimate` return the following:
``` return callback(null, to.hex(results.gasUsed.iadd(results.vm.gasRefund))) ```

A more thorough solution would be to implement a binary search for estimate gas as geth & metamask do. Possibly just updating the `web3-provider-engine` version and using their estimateGas method would work, assuming it doesn't break anything else

https://github.com/ethereum/go-ethereum/pull/3587
https://github.com/MetaMask/provider-engine/pull/119/files

I'd be happy to implement either, but thought I'd post for feedback to see what would be preferred.

1.
2.
3.
4.

## Your Environment

* Version used:
* Server type and version:
* Operating System and version:
* Link to your project:

perissology:
fixes #26 

This fix implements a binary search for the gas estimate. The main disadvantage is that `eth_estimateGas` calls are slower as the tx needs to be processed 2 - ~12 times to find a `gasLimit` that will successfully run the tx.

I've also added an option to set the accuracy threshold. The higher the threshold, the less accurate (will always be > then required) the gas estimate and the faster the response.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Both `.provider()` and `.server()` take a single object which allows you to spec
 * `"ws"`: Enable a websocket server. This is `true` by default.
 * `"vmErrorsOnRPCResponse"`: Whether to report runtime errors from EVM code as RPC errors. This is `true` by default to replicate the error reporting behavior of previous versions of ganache.
 * `"hdPath"`: The hierarchical deterministic path to use when generating accounts. Default: "m/44'/60'/0'/0/"
+* `"estimateGasThreshold"`: Minimum threshold for estimateGas response. The bigger the number, the faster but less accurate (response is always >= actual gas required for the tx) the estimateGas response is. Default is 50000
 
 # IMPLEMENTED METHODS
 
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -21,6 +21,8 @@ var random = require('./utils/random');
 var txhelper = require('./utils/txhelper');
 var TXRejectedError = require('./utils/txrejectederror');
 
+var DEFAULT_ESTIMATE_GAS_THRESHOLD = 50000;
+
 StateManager = function(options, provider) {
   var self = this;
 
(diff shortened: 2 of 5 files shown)


benjamincburns:
Hi @perissology,

First, apologies for letting this sit for so long! I think this a great change, however I'm not sure whether I should accept it. Rather than repeating myself, please see the discussion and concerns that I've raised over on ethereum/interfaces#8.

benjamincburns:
Calling this one blocked for now until the ambiguity in the RPC interface is resolved.

In the mean time please be aware that there are only a few operations for which a gas stipend is required (i.e. `CALL` and friends). Because of this, I believe the most efficient way to implement this is to add a `STEP` handler to the EVM, tracking the stipend-requiring opcodes being executed.

perissology:
@benjamincburns considering the lack of movement on #80, and that geth & parity clients use a binary search is it possible to make a decision and merge something that will fix this issue? It's a major annoyance w/ testrpc & I'm sure countless hours have been wasted b/c of this bug.

I do think a step handler would be a better solution, and I'd be happy to implement it if it would be merged

benjamincburns:
@perissology thanks for this contribution, however we found a more efficient way to handle this than to run a binary search. Expect that our estimates will be fixed sometime in the next couple of releases.

PR #76, opened 2018-03-03
Handle method eth_signTypedData

jstoxrocky:
Adding a function to `geth_api_double.js` and `statemanager.js` to handle the method `eth_signTypedData`. This is in reference to ethereum/EIPs#712. Although this EIP is not yet a standard, it has been experimentally implemented in Metamask 3.11.0+ and is also being touted by much of the community as the best way forward for signing data.

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -1,5 +1,6 @@
 var Account = require('ethereumjs-account');
 var Block = require('ethereumjs-block');
+var Abi = require('ethereumjs-abi');
 var VM = require('ethereumjs-vm');
 var RuntimeError = require('./utils/runtimeerror');
 var Trie = require('merkle-patricia-tree');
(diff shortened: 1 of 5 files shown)


benjamincburns:
Hi @jstoxrocky, thanks for this!

Have geth and/or parity implemented this yet? Ideally we'd only add this to ganache-cli once it's accepted to the standard and those clients add it, as otherwise it's too easy for devs to code against it, then find that their dapp is actually nonportable once it comes time to deploy elsewhere.

jstoxrocky:
@benjamincburns Discussion is still ongoing on the EIP. You're right, probably best to hold off until that is solidified. An experimental version has already been added to MetaMask (not sure how they are implementing it on whatever nodes they run) but nothing on Geth or Parity yet.

benjamincburns:
In that case, let's call this one blocked until the EIP is decided.

benjamincburns:
Looks like the EIP is merged but in draft status now. This one is still blocked pending some sign of implementation from geth/parity.

Relevant PR from web3 project. ethereum/web3.js#1073

benjamincburns:
Hi @jstoxrocky - I'm happy to report that we now support `eth_signTypedData` - we went with a different PR than yours because it was a bit more concise (made use of an external library so we'd have less code to maintain). Apologies for going a different route, but hopefully you'll be pleased to hear that support for this call should be released tomorrow morning!

PR #77, opened 2018-03-06
Added eth_ecrecover function

agnescameron:
ecrecover recovers the ethereum address used to sign a message with ecsign (which is already implemented in repository), and is useful for verification. ecrecover has already been used in the repo as part of a test for ecsign in tests/requests.js (both are called from the ethereumjs-util library), and so is implemented using the same style.

thanks!


agnescameron:
hi @benjamincburns, would you be able to give this a look? thanks.

PR #79, opened 2018-03-08
Ignore EIP 170

seesemichaelj:
I'm creating a real-time Solidity debugger runtime for an [Augur bounty](https://github.com/augurProject/augur-bounties#-bounty-2-portable-solidity-debugger).

To do this sanely, compiler optimization needs to be turned off for the developer to be follow a logical stepping process; otherwise it's possible for the compiler to optimize things away and cause unexpected behavior from a debugging perspective.

Unfortunately, turning off optimization can mean very large compiled contracts. [EIP 170](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md) prevents contracts with more than `0x6000` bytes (or 24KB) from being created. The rationale for EIP 170 (as I interpreted it) was really to address scaling from a global perspective, however I can see it acceptable to allow users to ignore EIP 170, specifically for debugging purposes.

This PR adds the option for ignoring EIP 170. **The option will not do anything until the associated `ethereumjs-vm` PR (https://github.com/ethereumjs/ethereumjs-vm/pull/282) has been merged.**

At first I figured `ganache-core` would always want the EIP 170 ignored, but after learning more about `Ganache` it seems that applications will use `ganache-core` with a "production blockchain" mindset as well; therefore I am using the existing debug flag to indicate whether or not this flag should be ignored.

I'm open to the idea to adding an additional external option for ignoring EIP 170 if that would make more sense.


benjamincburns:
Thanks for submitting this, @seesemichaelj!

Given that we're oftentimes the first client new developers use when learning smart contract development, and that most users learn by trial-and-error, I think we'll need to think carefully about this change.

Off the bat, I have three primary concerns:

1. We likely don't want this to be triggered automatically by the `debug` flag, as we generally want flags to be named such that it's abundantly clear what behavior they're changing. I think for ganache-core we'd likely want something like a boolean `checkContractSize`, or better yet an integer `maxContractSize` (I've commented over on ethereumjs/ethereumjs-vm#282 about this).
2. I'll need to think a bit on how to make it abundantly clear to the user that setting this option causes the EVM to behave differently than it would in production. We don't want new users changing this when their contracts get too big, thinking that they'll be able to tweak a similar setting once they reach public networks.
3. Finally, before we can accept the change, can you please also update the options section in `README.md` and add the appropriate tests? Might want to wait on this one until we decide if any other action is necessary, however.

seesemichaelj:
1. For posterity's sake, please see https://github.com/ethereumjs/ethereumjs-vm/pull/282 for a discussion on how we reached to the conclusion of using a `Boolean` instead of a `Number`.
2. I agree this is a tough one. Even if the user first understands the purpose of the flag and why it's used, I can see users forgetting it's enabled and scratch their head while they're getting OOG errors on contract initialization in a production environment.
3. I'll see what I can come up with in the `README.md` (sorry I didn't originally add it) to address concern 2

tcoulter:
Hey there. I made a comment on the ethereumjs-vm ticket: https://github.com/ethereumjs/ethereumjs-vm/pull/282#issuecomment-371883470

Would much rather prefer an API for driving the data layer rather than flags for bypassing protocol restrictions. This would be a long term goal, however. For the short term, if we accept this change, we should likely choose the same flag name as ethereumjs-vm and ensure this is off by default.

seesemichaelj:
FYI I will go ahead and make all of the appropriate changes for review to this once https://github.com/ethereumjs/ethereumjs-vm/pull/282 is accepted. I believe we're going to go ahead and stick with the constructor option for now. I'm interested to discuss more about what you're envisioning for the API @tcoulter in another issue; I'm not quite sure I'm on the same page as you haha

seesemichaelj:
@benjamincburns definitely, still a WIP (been busy with debugger stuff). Will let you know when it's ready for your review. Sorry for the lengthy delay!

seesemichaelj:
@benjamincburns this is ready for your review now

seesemichaelj:
Scratch that. Something got messed up in the rebase causing all this extra stuff to make this confusing to review. I'm going to try to make this easier. I'm going to go ahead and close this PR in favor for another PR as it added a bunch of participants that are probably going to get notified about something they are not impacted by.

PR #81, opened 2018-03-08
[added temporal snapshot/revert, added evm_decreaseTime, added tests for both]

chris-shyft:
ftw <3

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -1043,9 +1043,9 @@ BlockchainDouble.prototype.increaseTime = function(seconds) {
 };
 
 BlockchainDouble.prototype.decreaseTime = function(seconds) {
-    if (seconds > 0) seconds = 0;
-    this.timeAdjustment -= seconds;
-    return this.timeAdjustment;
+  if (seconds < 0) seconds = 0;
+  this.timeAdjustment -= seconds;
+  return this.timeAdjustment;
 };
 
 BlockchainDouble.prototype.setTime = function(date) {
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -1042,6 +1042,12 @@ BlockchainDouble.prototype.increaseTime = function(seconds) {
   return this.timeAdjustment;
 };
 
+BlockchainDouble.prototype.decreaseTime = function(seconds) {
+    if (seconds > 0) seconds = 0;
+    this.timeAdjustment -= seconds;
+    return this.timeAdjustment;
+};
+
 BlockchainDouble.prototype.setTime = function(date) {
   var now = new Date().getTime() / 1000 | 0;
   var start = date.getTime() / 1000 | 0;
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -717,7 +717,9 @@ StateManager.prototype.snapshot = function(callback) {
     if (err) return callback(err);
 
     self.snapshots.push({
-      blockNumber: blockNumber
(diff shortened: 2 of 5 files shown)


lib/blockchain_double.js
```diff
@@ -1042,6 +1042,12 @@ BlockchainDouble.prototype.increaseTime = function(seconds) {
   return this.timeAdjustment;
 };
 
+BlockchainDouble.prototype.decreaseTime = function(seconds) {
```
naure: Any reason not to let `increaseTime` accept negative values? That'd be one fewer public method.

chris-shyft: mainly because "increaseTime" as read by a developer would indicate time is going up, and a negative value on the variable is easy to mistake on a glance, while reading right to left of a function is pretty solid.

increaseTime(forwardValue)
increaseTime(-forwardValue)

vs

increaseTime(forwardValue)
decreaseTime(backwardsValue)

chris-shyft: plus reading pure function definitions in the readme (which I didn't update in this PR) would be immediately visible, while you'd have to create creating an extra note surrounding a negative value being acceptable in the alternative.

benjamincburns:
Hi @chris-shyft!

Thanks for this! Just so I don't repeat myself, can you please review my comments over on #95 and let me know what you think?

Also FWIW, given that this PR seems to more directly address trufflesuite/ganache-cli#390, I'm slightly more inclined to accept it than #95. Perhaps you and @PhABC might want to get in touch and coordinate changes?

benjamincburns:
> I wanted to keep these known values rolling within the tests instead of making notes that while X value was changed, it has been reverted in the next it()

Just to make sure you're aware, Truffle automatically runs `evm_snapshot` and `evm_revert` between tests. Given that #2 is now merged, I think you'll likely find you don't need `evm_decreaseTime`.

I agree however that without truffle doing the snapshot/revert adding an `evm_decreaseTime` is a bit more readable, and therefore nicer.

chris-shyft:
yep I'm aware. it wasn't between test files (contract("xyx") {}) it was between tests within the same contract("xyz") {}.

PhABC:
@benjamincburns If #2 was merged and `revert` is ran between every tests, why is the timestamp still not reverted between unit tests?

I'm specifically referring to unit tests within the same test, such as ; 

```javascript
it('should REVERT when expiration is passed', async function () {
  ...

  //Increasing time
  await web3.currentProvider.send({
    jsonrpc: "2.0", 
    method: "evm_increaseTime", 
    params: [timeToExpire], 
    id: 0
  });
  ...
 }); 

it('should PASS when expiration is NOT passed', async function () {
  ...
 }); 
```

benjamincburns:
@PhABC I believe I was mistaken before - truffle calls `evm_snapshot` and `evm_revert` between test files, not individual tests. If we were to support `evm_decreaseTime` without reverting state, we'd also need to ensure that we were following the requirements around monotonicity of timestamps imposed by the protocol, meaning that most of the time an `evm_decreaseTime` followed by an `evm_mine` would still result in an error.

The correct way to handle this case is still to use  `evm_snapshot` and `evm_revert`, as this causes the blockchain to actually behave as though it's gone back in time. You can call these methods yourself manually in your tests.

In the mean time we've added support for mining a block at an exact timestamp, and we've added the ability to set the time to a specific timestamp. between these new features and snapshot/revert, hopefully your needs are covered.

PR #87, opened 2018-03-15
Fix for deployment transactions without a to address.

ricmoo:
Fix for https://github.com/trufflesuite/ganache/issues/425.

Transactions without a `to` should have `null`, not `0x0000...0000`.

I believe this may be an issue for other values in this function, but this is the only one I have confirmed (and tested against. It is causing issues in [ethers.js](https://github.com/ethers-io/ethers.js), since it is a valid address, the `creates` is not being populated properly.

### DIFF
--- a/lib/utils/txhelper.js
+++ b/lib/utils/txhelper.js
@@ -12,6 +12,8 @@ module.exports = {
         break;
       }
     }
+
+
     var resultJSON = {
       hash: to.hex(tx.hash()),
       nonce: to.hex(tx.nonce),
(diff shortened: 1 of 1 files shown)


benjamincburns:
@ricmoo thanks for this change! Before I accept, can you please add a test which would have failed before the change was made, but passes now that it is applied?

Aside - `pending` tests are tests which we've marked as ignored. They represent incomplete behavior, or behavior which we've decided to accept as temporarily broken, usually due to the fix being blocked by some other change that's already in progress.

benjamincburns:
Original submitter hasn't responded in several months - we'll keep the original issue around and work it into our backlog.

ricmoo:
Thanks. I have not had time to create the required test cases.

Issue #51, opened 2018-01-10
zero-length unformatted data fields should return 0x, not 0x0


aakilfernandes:
Not sure if this is an issue or not, but `eth_getCode` returns `0x0` for non-contract accounts. My assumption is that it should return `0x`.

benjamincburns:
Hi @aakilfernandes! Funny story. I just started writing up a nice polite response telling you how wrong your assumption is, and then I went to grab a reference link from the [Ethereum RPC wiki](https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding) and discovered what an incorrect, smug bastard I am! :joy:

You're absolutely correct - unformatted data values which are zero length should be returned as `0x.`

Does this break anything for you?

benjamincburns:
Was hoping to knock this out quickly, but I hit some snags. Will need to come back to it later. PRs welcome.

jordanschalm:
A possible side-effect of this issue is sometimes non-zero fields are prefixed with a zero, which Geth no longer accepts as valid. For example `0x1234` would be returned by Ganache as `0x01234`.

I have seen this occurring with `eth_getBlockByNumber`. Ganache returns the following transaction within the JSON response:
```
{  
    "hash":"0x3f8d3b83b6b14a020d8bb0a088027fc0ae5004feeb063e11841103ce6a331153",
    "nonce":"0x0",
    "blockHash":"0xcf57c1d749ce9c28cb6424c99e50f67641ebe78cc27f43a9751de64344f66df8",
    "blockNumber":"0x01",
    "transactionIndex":"0x00",
    "from":"0x627306090abab3a6e1400e9345bc60c78a8bef57",
    "to":"0xf17f52151ebef6c7334fad080c5704d77216b732",
    "value":"0x5af3107a4000",
    "gas":"0x5208",
    "gasPrice":"0x04a817c800",
    "input":"0x0"
}
```

`gasPrice` (non-zero) and `nonce`/`input` (zero) are both prefixed by a zero. This causes Geth's standard `ethclient` to fail while unmarshalling the response with the following error:
```
json: cannot unmarshal hex number with leading zero digits into Go struct field rpcBlock.transactions of type *hexutil.Big
```

crzndx:
i also realized, that empty arrays return inconsistent results if empty, e.g. [0x0, 0x, 0x, ...]

benjamincburns:
[...]

PR #92, opened 2018-03-29
Align block_helper.toJSON with hashed values

AusIV:
I have a project where we're trying to use Ganache to test our services that
are written in Go. The RPC client that ships with Geth ignores the
hash provided by RPC and recalculates it instead. Due to some discrepancies
between the value used to calculate the hash and the values served over
RPC, the Go client is unable to compute the same hash as Ganache.

Two of the issues I ran into are in Ganache core. One was that the RPC
value for mixHash did not match the value used to calculate the checksum.
For reasons that are not clear even after reviewing the commit history,
rather than including the mixHash used to calculate the hash, block_helper
was including a string of repeating "10"s. I have changed this to reflect
the mixHash stored in the block header.

The other issue I ran into was extraData. If block.header.extraData was
an empty buffer, it would get encoded as "0x00", which decodes to a buffer
containing a single null byte. When a client uses that value to compute
the hash, they obviously get different results. This modifies
to.rpcDataHexString to pass empty buffers through as "0x" instead of
padding it with an extra 0 byte.

My mission of being able to compute the same hash with Ganache and  the Geth
RPC client has one more stop at ethereumjs-block, which defaults to having
an empty nonce instead of an appropriately sized buffer of null bytes.

### DIFF
--- a/lib/utils/block_helper.js
+++ b/lib/utils/block_helper.js
@@ -7,7 +7,7 @@ module.exports = {
       number: to.rpcQuantityHexString(block.header.number),
       hash: to.hex(block.hash()),
       parentHash: to.hex(block.header.parentHash), //common.hash
-      mixHash: "0x" + (new Array(32).fill("10").join("")), // TODO: Figure out what to do here.
+      mixHash: to.hex(block.header.mixHash),
       nonce: to.rpcDataHexString(to.hex(block.header.nonce), 16),
       sha3Uncles: to.hex(block.header.uncleHash),
       logsBloom: to.hex(block.header.bloom),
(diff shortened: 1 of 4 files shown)


AusIV:
The failing checks don't relate to the changes in this PR. At the last commit to develop, web3-providers-ws was at version 1.0.0-beta.30. By 1.0.0-beta.33 web3-provider-ws they changed the invocation of the listener function from cb(err, response) to cb(response), which is causing these tests to fail.

benjamincburns:
Thanks for this one, @AusIV! I think this fixes trufflesuite/ganache-cli#387 and #51.

benjamincburns:
Also kudos for adding/updating the test suite - this was a very easy change to accept!

nickysemenza:
I can confirm this works 👍, my errors were while trying to talk to ganache-cli from go-ethereum: `cannot unmarshal hex number with leading zero digits into Go struct field Log.logIndex of type hexutil.Uint`

Any chance you can re-push the docker image so that it has these changes in it? (i've pushed my own fork of the image so it's not a huge deal)

PR #93, opened 2018-03-30
Added noMethodLog option

satorumpen:
I added a option do not output RPC method name.


benjamincburns:
Hi @satorumpen. Thanks for this contribution. Can you please explain why this change is needed?

I think rather than having special logging flags in our already overgrown options object, it's likely better to switch to a proper logging library with logging keys and log levels. I've used `winston` for this in the past.

satorumpen:
@benjamincburns Thanks for your reply.

I use it via ganache-cli, but I want to suppress the display of Polling method like `eth_getFilterChanges`. In this case, I think that the way to replace the logger with the logger option as you say is too excessive.

Please let me know your opinion.

PR #94, opened 2018-03-31
Fix tests for web3-provider-ws@1.0.0-beta.33

AusIV:
In web3-provider-ws v1.0.0-beta.33 they changed the callback behavior
from cb(err, response) to cb(response), but it seems some instances
make the callback without passing the error while other instances make
the callback with an error and response. This change deals with both
possible inputs to the callback function.

An alternative might be to pin web3-provider-ws to 1.0.0-beta.30.


AusIV:
This addresses the build failures in #92 & #93.

PR #95, opened 2018-04-02
Adding prototype.evm_decreaseTime

PhABC:
It's currently hard to perform tests that are timestamp dependent since the timestamp changes between tests are shared (E.g. see #2, [Ganache-cli-issue-390](https://github.com/trufflesuite/ganache-cli/issues/390) and [here](https://ethereum.stackexchange.com/questions/30452/is-there-the-opposite-function-for-evm-increasetime)) . Using `evm_increaseTime` leads to non-reversible timestamp changes that persist through all the tests. 

Adding `evm_decreaseTime` allow developers to revert the timestamp change after each test, using a logic such as 
```javascript
beforeEach(async function () {
  await web3.currentProvider.send({
      jsonrpc: "2.0", 
      method: "evm_increaseTime", 
      params: [timeToIncrease], 
      id: 0
  });
});

afterEach(async function () {
  await web3.currentProvider.send({
      jsonrpc: "2.0", 
      method: "evm_decreaseTime", 
      params: [timeToIncrease], 
      id: 0
  });
});
```

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -1042,6 +1042,12 @@ BlockchainDouble.prototype.increaseTime = function(seconds) {
   return this.timeAdjustment;
 };
 
+BlockchainDouble.prototype.decreaseTime = function(seconds) {
+  if (seconds < 0) seconds = 0;
+  this.timeAdjustment -= seconds;
+  return this.timeAdjustment;
+};
+
 BlockchainDouble.prototype.setTime = function(date) {
   var now = new Date().getTime() / 1000 | 0;
   var start = date.getTime() / 1000 | 0;
--- a/lib/subproviders/geth_api_double.js
+++ b/lib/subproviders/geth_api_double.js
@@ -473,6 +473,11 @@ GethApiDouble.prototype.evm_increaseTime = function(seconds, callback) {
   callback(null, this.state.blockchain.increaseTime(seconds));
 };
 
+GethApiDouble.prototype.evm_decreaseTime = function(seconds, callback) {
+  callback(null, this.state.blockchain.decreaseTime(seconds));
+};
+
+
 GethApiDouble.prototype.evm_mine = function(callback) {
   this.state.processBlocks(1, function(err) {
     callback(err, '0x0');


benjamincburns:
@roderik could you elaborate, please? By any chance is your code open? A link to a repo or a gist would help my feeble mind tremendously! 😄

PhABC:
@benjamincburns Thank you for the response!

 I believe [trufflesuite/ganache-cli#390](https://github.com/trufflesuite/ganache-cli/issues/390) is fixed in [this Pull Request](https://github.com/trufflesuite/ganache-core/pull/2) (see my comments on that pull request).

I'm in favor of either method.

roderik:
Sure,

look at this repo: https://github.com/DataBrokerDAO/dtx-crowdsale-contracts
I worked around it in this case though.

I like my source files small, so it is more easy to get an overview of what is happening. Which means I would make lots of test files with specific things I wanted to test. 

But multiple tests run at a random order. So if the first test that runs is one that jumps in time, all the other tests would run at that time in the future.

So If i wanted to test "cannot contribute before the sale" and "can contribute during the sale", it would be total guess work if they would work, because this depends if the "during saler" test runs first because it jumps ahead in time.

While the entire idea of the snapshots is that you start "clean" (but still faster than redeploy every time) at the beginning of each file.

benjamincburns:
> While the entire idea of the snapshots is that you start "clean" (but still faster than redeploy every time) at the beginning of each file.

@roderik as an aside, I've suggested internally that a cleaner approach to this might be if our tests were to use the forking feature of `ganache-core`. That would also enable tests to be run in parallel against multiple chains which were forked from an initial chain post-setup.

I'll raise it again during our next internal planning meeting.

roderik:
No, seems fine, do not have anything to test it on right now :)

benjamincburns:
@roderik I think you were the only one still advocating for this issue now that #2 is merged. Can you please reach out to me at ben@trufflesuite.com if this is still important to you?

Otherwise in the mean time I'm closing this to clean up our backlog some.

wjmelements:
I would find this useful for gas-profiling contracts where a timestamp is passed as a parameter, because the zeroeness of the bytes of the timestamp affect the gas consumption. I can then profile as if it was way in the past without reconfiguring ganache for just this test.

Instead of `evm_decreasetime`, I think a better interface would be passing a negative value to `evm_increasetime`.

PR #99, opened 2018-04-10
deps - bump web3-provider-engine

kumavis:
allows for looser id formatting for subs + filters


benjamincburns:
Thanks, @kumavis!

PR #100, opened 2018-04-11
fix (most) global leaks

Xcrux:
There were some variable leaks to the global scope, I fixed most of them. one of them was in another repo and I have a pending PR there (https://github.com/ethereumjs/merkle-patricia-tree/pull/42)


benjamincburns:
Thanks so much for this, @Xcrux!

PR #101, opened 2018-04-12
Rename blocktime to blockTime

kosecki123:
This is missing part of https://github.com/trufflesuite/ganache-cli/pull/483 which renames `blocktime` to `blockTime` and pass that property in options while ganache-core still expects property named `blocktime`

This PR renames `blocktime` to `blockTime` in ganache-core.

Fixes  https://github.com/trufflesuite/ganache-cli/issues/506

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -60,8 +60,8 @@ StateManager = function(options, provider) {
   this.gasPriceVal = to.hex(options.gasPrice);
 
   this.is_mining = true;
-  this.blocktime = options.blocktime;
-  this.is_mining_on_interval = !!options.blocktime;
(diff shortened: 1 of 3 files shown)


benjamincburns:
Thanks for your diligence on this, @kosecki123! Apologies it took me a while to review, as I've been away for the past couple weeks dealing with an international move.

I notice there are two tests failing in the boron build. I'm fairly sure that this is an existing bug exposed by your changes here.  As a result I'll go ahead an accept this PR as-is and track down the cause of those test failures separately.

Thanks again!

EralpB:
Can you please make this into a release 🙏

PR #102, opened 2018-04-17
Mine n blocks, skip blocks

filiplazovic:
I made two methods:

- `evm_mineBlocks:` - Mines multiple blocks. Accepts one arguments: number of blocks to mine.

- `evm_skipBlocks:` Will create one block with fake block number. Takes one argument: number of blocks you want to add on top of current block height. Useful when testing big block-number based constraints (eg. it would take ~1h to mine 500k blocks). It is not guaranteed that everything will this function, so you should save a snapshot of the blockchain before calling this function, then revert afterwards. Should be mainly used for checking block-number based constraints.

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -180,6 +180,26 @@ BlockchainDouble.prototype.getEffectiveBlockNumber = function(number, callback) 
   }
 };
 
+BlockchainDouble.prototype.getRealBlockNumber = function(number, callback) {
+  if (typeof number != "string") {
+    number = to.hex(number);
+  }
+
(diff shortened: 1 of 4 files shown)


test/forking.js
```diff
@@ -441,13 +441,9 @@ describe("Forking", function() {
...
-    var callcount = 0
     event.on('data', function(log) {
-      callcount++
-      if (callcount == 2) {
-        event.removeAllListeners()
-        done()
-      }
+      event.removeAllListeners()
+      done()
     });
```
filiplazovic: This part is left to be fixed. Only 1 event was firing

PR #73, opened 2018-02-23
Fix JSON encoding of Log & Receipt

protolambda:
Fix JSON encoding of Log & Receipt; respect data/quantity format types (no leading zeroes for quantities)

As you may already know, the RPC format defines `DATA` and `QUANTITY` hex formats:
- `DATA` has leading zeroes, padding length to a multiple of 2
- `QUANTITY` may not have any leading zeroes

The log/receipt JSON encodings in the Ganache RPC turned out to be invalid sometimes, i.e. leading zeroes in quantities. Go-ethereum failed to decode the `eth_getLogs` RPC responses, as it strictly enforces the formatting as defined in the spec.

You can find a clear spec of the `DATA`/`QUANTITY` properties in the ethereum wiki:

Receipt: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt
Logs: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getfilterchanges

I changed the `toJSON` methods of `Log` and `Receipt` to call the `to.rpcQuantityHexString` and `to.rpcDataHexString`, to force proper formatting (these methods where already made available for some other RPC formatting).
Note that the previous implementation, `to.hex`, didn't pad with zeroes, but also didn't enforce the spec. (E.g. call `hex("0x0123")` for a quantity and you get `"0x0123"`).

PR #103, opened 2018-04-18
Update blockchain_double.js to properly hex encode Log without leading zeroes 

nickysemenza:
When trying to talk to `ganache-cli` overRPC from go-ethereum: `cannot unmarshal hex number with leading zero digits into Go struct field Log.logIndex of type hexutil.Uint`

This was based off of changes I observed in #92

### DIFF
--- a/lib/utils/txhelper.js
+++ b/lib/utils/txhelper.js
@@ -13,17 +13,17 @@ module.exports = {
       }
     }
     var resultJSON = {
-      hash: to.hex(tx.hash()),
-      nonce: to.hex(tx.nonce),
-      blockHash: to.hex(block.hash()),
-      blockNumber: to.hex(block.header.number),
-      transactionIndex: to.hex(transactionIndex),
-      from: to.hex(tx.from),
-      to: to.hex(tx.to),
-      value: to.hex(tx.value),
-      gas: to.hex(tx.gasLimit),
-      gasPrice: to.hex(tx.gasPrice),
-      input: to.hex(tx.data),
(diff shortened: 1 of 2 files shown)


nickysemenza:
nevermind, looks like #73 accomplished this with a different approach?

nickysemenza:
looks like issue still persists even with #73 for some endpoints?

lib/utils/txhelper.js
```diff
@@ -13,17 +13,17 @@ module.exports = {
...
-      blockNumber: to.hex(block.header.number),
-      transactionIndex: to.hex(transactionIndex),
-      from: to.hex(tx.from),
-      to: to.hex(tx.to),
-      value: to.hex(tx.value),
-      gas: to.hex(tx.gasLimit),
-      gasPrice: to.hex(tx.gasPrice),
-      input: to.hex(tx.data),
+      hash: to.rpcDataHexString(tx.hash()),
+      nonce: to.rpcDataHexString(tx.nonce),
```
benjamincburns: Nonce should be a quantity string per https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionbyhash

benjamincburns:
I think this needs to be rebased on current develop. Also if you could please add tests for the fields you've changed that would be ideal.

seesemichaelj:
I'm going to move this to our backlog to do what @benjamincburns asked a month ago as we can complete the requested changes. I think this may be related to #166

benjamincburns:
I've just gone ahead and rebased this on current `develop`

Issue #109, opened 2018-04-29
undefined "blockchain" parameter on server.listen() callback


mpolci:
When used as a general http server:
```
var Ganache = require("ganache-core");
var server = Ganache.server();
callback = function(err, blockchain) {
  // blockchain is always undefined
}
server.listen(port, callback);
```
The parameter _blockchain_ of the function _callback()_ is always undefined.

benjamincburns:
@mpolci thanks once again for your PR! This will be fixed in the next release!

PR #110, opened 2018-04-29
Fix missing server.listen() callback parameters

mpolci:
Fix issue #109


benjamincburns:
Thanks for fixing this, @mpolci! I must admit culpability for this bug, as I think it's something I broke during my first week on the job.

I'll go ahead and add a test for this and merge.

benjamincburns:
Rebased on current develop and added test. Build is broken on Node v10 but not due to this change (we're troubleshooting that separately).

PR #112, opened 2018-05-08
upgrade ethereum vm to fix crucial bug

Xuefeng-Zhu:


modong:
To provide a bit context here, BYTE opcode was broken in js EVM. This is preventing easy testing for a lot of stuff. I second this PR to upgrade to newer version of JS EVM.

PR #113, opened 2018-05-17
removed if statements for v, r, s

noot:
previously, the if statement around v, r, s was never satisfied, so v, r, s were never returned in the transaction
geth and parity returned these values, but ganache doesn't
I removed the if statement and now it returns them
needed v, r, s for testing merkle proofs.

### DIFF
--- a/lib/utils/txhelper.js
+++ b/lib/utils/txhelper.js
@@ -12,7 +12,7 @@ module.exports = {
         break;
       }
     }
-    var resultJSON = {
+    return {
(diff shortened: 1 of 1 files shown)


benjamincburns:
Hi @ebinks - can you please add a test which shows that this works for your use case?

benjamincburns:
@noot is this still a problem? I haven't heard issues from other users about this and I haven't heard back from you about the test. Please let me know if it is and we'll reopen.

PR #115, opened 2018-05-18
Fix #114: Return error message if vmErrorsOnRPCResponse is disabled

### LINKED ISSUE #114: Revert should not throw VM expection on call if vmErrorsOnRPCResponse is disabled
## Expected Behavior
If `vmErrorsOnRPCResponse` is set to false I would expect that no error is thrown. Also if `revert` was used with the message the encoded error message should be returned.

## Current Behavior
If there is an error in the contract it throws and the `revert` message is not returned

## Possible Solution
RuntimeError should not be triggered if the option is enabled:

https://github.com/trufflesuite/ganache-core/blob/develop/lib/blockchain_double.js#L481

## More info
trufflesuite/ganache-cli#521

rmeissner:


benjamincburns:
Hi @rmeissner - can you please add tests to this showing the intended behavior with and without the flag set?

benjamincburns:
Sorry for my previous comment - there are already tests here! This change looks good to me.

benjamincburns:
Thanks so much for your contribution, @rmeissner!

benjamincburns:
@rmeissner I just published `ganache-core@2.1.1` and `ganache-cli@6.1.2`, both of which contain this fix.

rmeissner:
Very cool, thank you

Issue #117, opened 2018-05-25
Ganache shouldn't assume that every transaction with data is a contract call


bordalix:
I think I found a bug, where Ganache assumes that every transaction with data must be a contract call, which forbids to send transactions with data to non-contract accounts.

Imagine I'm sending a transaction from address1 to address2 (both non-contract accounts) with some data on it.

The function [_transactionIsContractCall](https://github.com/trufflesuite/ganache-core/blob/c7a2f69cad6bd8d73a67c300e8eb9816b44d694c/lib/statemanager.js#L849) will return true since the transaction has some data, defining it as a contract call (which is wrong).

On [line 832](https://github.com/trufflesuite/ganache-core/blob/c7a2f69cad6bd8d73a67c300e8eb9816b44d694c/lib/statemanager.js#L832) of the same file, since the **_transactionIsContractCall** function returned true, Ganache will now expects for the transaction to have some code, which it don't (since I'm not calling any function, I'm just sending some Ether to address2), so a new **TXRejectedError** will be raised, rejecting the transaction.

Shouldn't transactions to non-contract accounts be able to send some data?

Here is what the [Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) states on point 4.2:
> 
> There are two types of transactions: those which result in message calls and those which result in the creation of new accounts with associated code (known informally as ‘contract creation’). Both types specify a number of common fields:
> ...
> Additionally, a contract creation transaction contains:
> **init**: An unlimited size byte array specifying the EVM-code for the account initialisation procedure, forma formally Ti.
> ...
> In contrast, a message call transaction contains:
> **data**: An unlimited size byte array specifying the input data of the message call, formally Td.

Is my understanding that the Ethereum protocol doesn't distinguish between message calls to External Owned Accounts (personal accounts) from message calls to Contract Accounts, so we should be able to send data in any message call. And if my understanding is correct, then this behaviour from Ganache should be considered as an incompatibility bug.

guipozo:
I'm trying to send a transaction with some data to an account (not a contract) and i'm receiving:

`Attempting to run transaction which calls a contract function, but recipient address 0x6e2e3accbe4b1320107118147a7cb6f5b689ea2f is not a contract address`.

Ethereum "blocks/disable" transactions (Acc2Acc) with data? It's normal?

davidmurdoch:
[...]

PR #119, opened 2018-05-31
There is no need to check if the recipient of a transaction is a contract or not

bernardpeh:
Everytime ganache sees a "data" field, it assumes that it is trying to call a contract function and that the "to" field must be a contract. This should not be the case.

It should be perfectly fine to do this:

```
web3.eth.sendTransaction({from:web3.eth.accounts[0],to:web3.eth.accounts[1],value:web3.toWei(0.02,“ether”),data:web3.toHex(“hellow world”)})
```

This above snippet below is working in testnet but not in ganache.

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -810,28 +810,9 @@ StateManager.prototype.createFakeTransactionWithCorrectNonce = function(rawTx, f
       }
     }
 
-    // If we're calling a contract, check to make sure the address specified is a contract address
-    if (_transactionIsContractCall(rawTx)) {
-      self.getCode(to.hex(rawTx.to), 'latest', function(err, code) {
-        if (err) {
-          callback(err);
-        } else if (code === '0x0') {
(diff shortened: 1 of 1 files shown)


benjamincburns:
@bernardpeh Thanks for this. 

I added this check to be helpful in the case where the caller is attempting to call a contract but the contract address is incorrect, but you're correct that in doing so I inadvertently prevented users from sending transactions with arbitrary data.

IIRC, it's possible to check whether the `data` field is formatted a function call. In the case when it is formatted as such, I'd like to keep this check in place. Could you update your PR to handle this?

Also it seems that you didn't run tests before submitting, as this change caused a test to fail. Please be sure to check your changes with `npm test` and update tests as necessary.

benjamincburns:
Going to keep #117 around but since I haven't heard back from @bernardpeh on this one and since it causes numerous test failures I'm going to close in hopes that a better submission comes along.

@bernardpeh if you want to update this to make tests pass please feel free to submit a new PR.

PR #121, opened 2018-06-13
gasPrice should be encoded as rpc quantity hex

erickearns:
go-ethereum code errors as follows when deserializing `eth_gasPrice` response:

`failed to suggest gas price: json: cannot unmarshal hex number with leading zero digits into Go value of type *hexutil.Big`


benjamincburns:
Thanks for the contribution, @erickearns!

PR #124, opened 2018-06-19
Add test for checking value after contract deployed after forking

seesemichaelj:
This was brought up by @tcoulter mentioning that sending value to a contract during creation on a chain that resulted from a fork would not persist. This test now checks for the presence of this issue. A separate PR will be made for the resolution of this test.

### DIFF
--- a/test/forking_deploy_after_fork.js
+++ b/test/forking_deploy_after_fork.js
@@ -5,8 +5,6 @@ var assert = require('assert');
 var Ganache = require("../index.js");
 var fs = require("fs");
 var solc = require("solc");
-var to = require("../lib/utils/to.js");
-var async = require("async");
 
 // Thanks solc. At least this works!
 // This removes solc's overzealous uncaughtException event handler.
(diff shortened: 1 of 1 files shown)


seesemichaelj:
The test added in this PR failed (as it should) here:
- https://travis-ci.org/trufflesuite/ganache-core/jobs/394347307#L2599
- https://travis-ci.org/trufflesuite/ganache-core/jobs/394347308#L2551

You can see here that the test shows that the current codebase will fail the test. #125 addresses this issue.

test/forking_deploy_after_fork.js
```diff
@@ -0,0 +1,177 @@
...
+          assert.notEqual(code, "0x0");
+
+          done();
+        })
+      });
+    });
+  });
+
+  it("should send 1 ether to the created contract, checked on the forked chain", function(done) {
+    mainWeb3.eth.getBalance(contractAddress, function(err, balance) {
```
benjamincburns: we can async/await now! `web3.eth.getBalance` returns a `PromiEvent`

seesemichaelj:
I started to put this in `forking.js` but the file is already getting a little out of hand. `forking.js` uses `forkedWeb3` instead of `mainWeb3` to deploy the contract. It's just a different set of steps and I felt it should be separated as such. I think `forking.js` should probably be split up as well.

PR #125, opened 2018-06-19
Flush VM Cache Before Pushing Contract Code

seesemichaelj:
In our monkey-patching/duck-punching function for putContractCode, we pull the account from the state trie to put the code. However, putContractCode is called before the cache will be flushed, causing any balance in the account to be lost.

This PR flushes the VM cache before getting the account details.

This addresses #124 and should be merged in after #124.


seesemichaelj:
The tests can be seen passing here:
- https://travis-ci.org/trufflesuite/ganache-core/jobs/394351758#L2183
- https://travis-ci.org/trufflesuite/ganache-core/jobs/394351759#L2180

PR #80, opened 2018-03-08
Consume gasRefund From Potential ethereumjs-vm Addition

seesemichaelj:
This fixes case 1 in #26, but will require the merging of PR https://github.com/ethereumjs/ethereumjs-vm/pull/284 before fixed.

This PR adds the gasRefund to the estimated gas call. This can be merged prior to https://github.com/ethereumjs/ethereumjs-vm/pull/284 being merged, but may be fruitless to do so until https://github.com/ethereumjs/ethereumjs-vm/pull/284 is merged to maintain cohesion and ensuring the problem does end up fixing #26.

More discussion can be found on #26 and https://github.com/ethereumjs/ethereumjs-vm/issues/283

PR #141, opened 2018-08-09
Consume gasRefund From Potential ethereumjs-vm Addition

davidmurdoch:
This is the same as PR #80 with added tests. Fixes issue #26: `estimateGas returns invalid value when tx has a gasRefund`.


seesemichaelj:
@davidmurdoch FYI, sometimes you can push directly to the PR branch if they checked the box to "allow reviewers to modify it". no problem though!

davidmurdoch:
@seesemichaelj I used Github's "open this in GitHub Desktop" link on the PR and it automatically created a new branch instead of just opening your branch and so I just rolled with it from there. `¯\_(ツ)_/¯`

I'll be sure to do it The Right Way next time.

lib/statemanager.js
```diff
@@ -594,7 +594,7 @@ StateManager.prototype.processGasEstimate = function (from, rawTx, blockNumber,
       }
       var result = '0x0'
       if (!results.error) {
-        result = to.hex(results.gasUsed)
+        result = results.gasRefund ? to.hex(results.gasUsed.add(results.gasRefund)) : to.hex(results.gasUsed);
```
benjamincburns: We expect the `gasRefund` field to always be there, right? If so, this squelches an error and makes things look like they're working if the `gasRefund` field suddenly disappears.

benjamincburns:
It _might_ be worth adding a test to check whether our gas estimation accounts for explicit gas stipends properly (e.g. something like `mycontract.depth.gas(100000)(y-1)`)

Relevant docs [here](http://solidity.readthedocs.io/en/latest/control-structures.html#external-function-calls).

lib/statemanager.js
davidmurdoch: fixed in latest commits

PR #142, opened 2018-08-11
Add User-Agent token to allowed headers

dinostabOMG:
Related: https://github.com/ethereum/web3.js/issues/1803

Also related: https://github.com/trufflesuite/ganache/issues/359

Although the latter is a broader issue, I fixed my immediate problem of not being able to connect from Firefox. I will look into the rest in a separate branch as requested by @davidmurdoch

### DIFF
--- a/lib/httpServer.js
+++ b/lib/httpServer.js
@@ -62,7 +62,7 @@ module.exports = function (provider, logger) {
           break;
         default:
           response.writeHead(400, {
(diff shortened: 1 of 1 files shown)


benjamincburns:
Thanks for this one, @dinostabOMG! I've opened #142 to prevent issues like this one from cropping up in the future.

davidmurdoch:
The correct issue for tracking the CORS fix is actually in #143

PR #149, opened 2018-08-14
Add eth_signTypedData

rmeissner:
Adds support for `eth_signTypedData` according to [EIP-712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md)

This is required for trufflesuite/ganache-cli#515


rmeissner:
How should conflicts in npm-shrinkwrap be resolved?

benjamincburns:
Hi @rmeissner! Thanks for the contribution.

I think shrinkwrap conflicts should be handled by making sure that existing versions don't change.

Otherwise I'm a little hesitant about this change, primarily because EIP-712 is still listed as draft status, and I can't find where either geth or Parity support it yet. I see that MetaMask has added support, however - so maybe that's good enough since most web apps will be interacting with the client via them or some other wallet middleware...

PR #157, opened 2018-08-24
Fix race conditions causing problems identified in #417

kn:
This PR fixes the problem described in [here](https://github.com/trufflesuite/ganache-cli/issues/453).

**Problem:**
The root cause of the problem is that we use `blockchain.vm.stateManager.[checkpoint|commit|revert]()` assuming there is no other functions performing transactions (meaning db transaction instead  of blockchain transaction :)) on `stateTrie`. Doing this without a lock can cause race conditions. For example, the following events can cause race conditions:
1. `checkpoint()` is called by `processCall()`
1. `checkpoint()` is called by `processNextBlock()`
1. `revert()` is called by `processCall()` <- this reverts the checkpoint created by `processNextBlock()`
1. `commit()` is called by `processNextBlock()` <- this commits the checkpoint created by `processCall()`

**Solution:**
This PR solves this problem by introducing a semaphore lock on the stateTrie transactions to prevent functions unintentionally committing or reverting a checkpoint created by other functions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ TODO
 .tern-port
 .vscode
 yarn.lock
+*.swp
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -54,6 +54,18 @@ BlockchainDouble.prototype._applyDefaultOptions = function(options) {
   return _.merge(options, defaultOptions, Object.assign({}, options));
 }
 
+BlockchainDouble.prototype._stateCheckpoint = function() {
+  this.vm.stateManager.checkpoint();
+}
+
+BlockchainDouble.prototype._stateCommit = function(callback) {
+  this.vm.stateManager.commit(callback);
+}
+
+BlockchainDouble.prototype._stateRevert = function(callback) {
+  this.vm.stateManager.revert(callback);
+}
+
 BlockchainDouble.prototype.initialize = function(accounts, callback) {
   var self = this;
 
@@ -437,7 +449,7 @@ BlockchainDouble.prototype.processCall = function(tx, blockNumber, callback) {
   var startingStateRoot;
 
   var cleanUpAndReturn = function (err, result, changeRoot) {
-    self.vm.stateManager.revert(function (e) {
+    self._stateRevert(function (e) {
       // For defaultBlock, undo state root changes
       if (changeRoot){
         self.stateTrie.root = startingStateRoot;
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@ TODO
 .tern-port
 .vscode
 yarn.lock
+*.swp
```
kn: FYI: I added this because I use vim and it leaves these temporary files.

benjamincburns:
@kn I'm going to close this PR because the problem needs to be solved by having a robust underlying data structure (the merkle-patricia-tree) rather than by carefully sidestepping problems in that structure.

I _strongly_ encourage you to keep trying, though!

kn:
Thanks for the feedback! Happy to look into a better solution.

One question on the past attempt to solve this issue. From the comment you made in the issue page, it sounds like someone attempted to solve this by updating ‘checkpoint()’ to take callback and make sure there won’t be race conditions -during- a call to the ‘checkpoint()’ by using a semaphore lock.

I think this doesn’t solve the issue entirely since the example problem I described above still can happen i.e. we have no control over which checkpoint to commit or revert when there are multiple async functions creating checkpoints for isolated contexts.

If this is true, we probably need some kind of database transaction support instead of linear checkpoint feature.

Does this statement align with your understanding of the issue?

kn:
Here are the changes that demonstrate the idea above:
https://github.com/ethereumjs/merkle-patricia-tree/compare/master...kn:i417
https://github.com/ethereumjs/ethereumjs-vm/compare/master...kn:i417
https://github.com/trufflesuite/ganache-core/compare/develop...kn:i417_2

Run openzeppelin-solidity tests with `asyncRequestProcessing` option set to `true` a few times and they always pass also, except for this one test failing consistently called `Contract: Bount against broken contract can claim reward` which fails even without the changes and `asyncRequestProcessing` option set to `false`.

benjamincburns:
@kn - I'll try to run the `zeppelin-solidity` tests again this afternoon. I didn't look too carefully when I ran last time -- it's possible I saw the same test fail and mistook it for the bug still hanging around. Will reopen if I reproduce your results.

benjamincburns:
And thanks for sticking with this!

ceresstation:
Hey @kn just wanted to check in to see how things are going on this front. Seconding @benjamincburns, appreciate you sticking with this!

PR #158, opened 2018-08-24
[FIX] Refine isContract call to support Present design

### LINKED ISSUE #117: Ganache shouldn't assume that every transaction with data is a contract call

adibas03:
Fixes #117 
@bernardpeh created #119 which completely removes the contract call check, which was closed as tests were not passing.
This handles the possible scenarios under which it is a contract call and only checks for for code presence if it should already exist.

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -883,15 +883,20 @@ StateManager.prototype.createFakeTransactionWithCorrectNonce = function(rawTx, f
 
     // If we're calling a contract, check to make sure the address specified is a contract address
     if (_transactionIsContractCall(rawTx)) {
-      self.getCode(to.hex(rawTx.to), 'latest', function(err, code) {
-        if (err) {
-          callback(err);
-        } else if (code === '0x0') {
-          callback(new TXRejectedError(`Attempting to run transaction which calls a contract function, but recipient address ${to.hex(rawTx.to)} is not a contract address`))
-        } else {
-          callback(null, tx)
-        }
-      });
+      // Contract just being deployed, code yet to exist at adress
+      if (!rawTx.to || rawTx.to === '0x0' || rawTx.to === '0x' || rawTx.to === '0') {
+        callback(null, tx)
+      } else {
+        self.getCode(to.hex(rawTx.to), 'latest', function(err, code) {
+          if (err) {
+            callback(err);
+          } else if (code === '0x0') {
+            callback(new TXRejectedError(`Attempting to run transaction which calls a contract function, but recipient address ${to.hex(rawTx.to)} is not a contract address`))
+          } else {
+            callback(null, tx)
+          }
+        });
+      }
     } else {
       callback(null, tx)
     }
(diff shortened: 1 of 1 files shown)


adibas03:
@benjamincburns I don't know if this would be a preferred solution or just completely removing as in #119   would be preferred.

benjamincburns:
Thanks for the contribution, @adibas03! I think the right way to fix this is to remove the check, as in #119. My desire in adding the check in the first place was to give people feedback during development when they sent would-be contract call transactions to simple (non-contract) addresses. It seems that this check has caused more pain than it has alleviated, however.

lib/statemanager.js
```diff
@@ -883,15 +883,20 @@ StateManager.prototype.createFakeTransactionWithCorrectNonce = function(rawTx, f
...
-        if (err) {
-          callback(err);
-        } else if (code === '0x0') {
-          callback(new TXRejectedError(`Attempting to run transaction which calls a contract function, but recipient address ${to.hex(rawTx.to)} is not a contract address`))
-        } else {
-          callback(null, tx)
-        }
-      });
+      // Contract just being deployed, code yet to exist at adress
+      if (!rawTx.to || rawTx.to === '0x0' || rawTx.to === '0x' || rawTx.to === '0') {
```
benjamincburns: FWIW, I think contract deployment only happens when the `to` field is completely missing. It's completely valid to send a transaction with data to address zero.

benjamincburns: That said, `to` is a DATA field (per the [RPC wiki](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sendtransaction)), so the zero-length value of `0x` should probably also be considered a contract deployment.

benjamincburns:
@adibas03 I'm going to close this one for now but I strongly encourage you to submit a new PR that picks up where #119 left off.

PR #164, opened 2018-08-31
Add tests to ensure testrpc doesn't fails to clear mapping storage slots

davidmurdoch:
Closes trufflesuite/ganache-cli#294

### DIFF
--- a/test/EstimateGas.sol
+++ b/test/EstimateGas.sol
@@ -32,8 +32,6 @@ contract EstimateGas {
     mapping (uint => uint) public uints;
     // Sets the uints[1] slot to a value;
     function store(uint _uint) public { uints[1] = _uint;}
-    // Fails under testRPC provider on non empty store
-    // "callback contain no result Error: Error: base fee exceeds gas limit"
     function clear() public { delete uints[1]; }
 
     mapping(bytes32 => uint) index;
--- a/test/EstimateGas.sol
+++ b/test/EstimateGas.sol
@@ -27,6 +27,15 @@ contract EstimateGas {
         triggerRselfdestructRefund();
     }
 
+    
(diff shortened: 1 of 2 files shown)


test/gas.js
```diff
@@ -114,6 +114,26 @@ describe("Gas", function() {
...
+      assert.equal(uintsa, "0", "initial value is not correct");
+
+      const receipta = await estimateGasInstance.methods.store(1).send(options);
+      assert.equal(receipta.status, true, "storing value did not work");
+
+      const uintsb = await estimateGasInstance.methods.uints(1).call();
+      assert.equal(uintsb, "1", "set value is incorrect");
+
+      const receiptb = await estimateGasInstance.methods.clear().send(options);
+      assert.equal(receiptb.status, true, "storing 0 did not work");
```
seesemichaelj: This assert message is technically not correcting; `deleting` and `storing` are not the same thing, although `deleting` will resolve in a `0` value when calling it again.

I would just saying `clearing value did not work`

test/EstimateGas.sol
```diff
@@ -27,6 +27,15 @@ contract EstimateGas {
         triggerRselfdestructRefund();
     }
 
+    
+    // https://github.com/trufflesuite/ganache-cli/issues/294
+    mapping (uint => uint) public uints;
+    // Sets the uints[1] slot to a value;
+    function store(uint _uint) public { uints[1] = _uint;}
+    // Fails under testRPC provider on non empty store
+    // "callback contain no result Error: Error: base fee exceeds gas limit"
```
seesemichaelj: Looks like these were just copied straight from https://github.com/trufflesuite/ganache-cli/issues/294, but I'm not sure if these two comments are applicable in our test, because it doesn't fail. I'd remove these for posterity's sake.

PR #168, opened 2018-09-13
Add webpack as a `devDependency` so it works w/out requiring `npm -g install`

davidmurdoch:
Also, allow npm to do its dependency reordering and shrinkwrap updating things it likes to do to make life more exciting.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5,197 +5,193 @@
   "requires": true,
   "dependencies": {
     "@webassemblyjs/ast": {
-      "version": "1.5.13",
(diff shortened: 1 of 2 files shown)


seesemichaelj:
There isn't much change in the package.json file. I'm cool with merging this, but I'm unsure what effective changes are being made

davidmurdoch:
I actually misread the package.json due to the reordering of the packages (thanks, npm) and thought it was adding the `webpack` dependency. When running a build it was erroring out for me so I ran `npm install webpack` -- when npm reordered the deps I just misread it and thought I was now adding webpack.

No need to merge.

PR #170, opened 2018-09-19
httpServer: return 400 if attempting to eth_subscribe on httpServer

tramp87:
fixes https://github.com/trufflesuite/ganache-core/issues/169

### DIFF
--- a/lib/httpServer.js
+++ b/lib/httpServer.js
@@ -31,15 +31,18 @@ module.exports = function (provider, logger) {
           break;
         case "POST":
           //console.log("Request coming in:", body);
+          
+          var send400 = function() {
+            headers[ "Content-Type" ] = "text/plain";
+            response.writeHead(400, headers);
+            response.end("400 Bad Request");
+          };
 
           var payload;
           try {
             payload = JSON.parse(body);
           } catch (e) {
-            headers[ "Content-Type" ] = "text/plain";
-            response.writeHead(400, headers);
-            response.end("400 Bad Request");
-            return;
+            return send400();
           }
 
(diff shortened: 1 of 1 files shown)


tramp87:
test failures seem to be unrelated timeout errors

lib/httpServer.js
```diff
@@ -31,15 +31,18 @@ module.exports = function (provider, logger) {
           break;
         case "POST":
           //console.log("Request coming in:", body);
+          
+          var send400 = function() {
```
benjamincburns: This functionality was sort of a last resort measure for unhandled errors. Sadly it breaks JSON-RPC 2.0, as all Requests MUST return a Response, which doesn't happen in this case.

lib/httpServer.js
```diff
@@ -48,9 +51,15 @@ module.exports = function (provider, logger) {
             for (var i = 0; i < payload.length; i++) {
               var item = payload[ i ];
               logger.log(item.method);
+              if (item.method === 'eth_subscribe') {
+                return send400();
```
benjamincburns: Because of my other comment, it would be best if this returned a proper JSON-RPC error, perhaps with a message like "Method not allowed over HTTP transport."

tramp87:
Thanks for the review.

Just wanted to let you know that I don't think I'll have the capacity to work on this soon (I might do so, but can't promise any ETA). I'm not sure how I'd go about making an automated test for this code path either, as I'm not extremely familiar with ganache codebase. Feel free to take over the PR or the issue if you consider it to be urgent enough.

seesemichaelj:
Closing this in favor of #252

PR #171, opened 2018-09-19
Additional fix to #51 (eth_getCode on a non-contract should return 0x)

### LINKED ISSUE #51: zero-length unformatted data fields should return 0x, not 0x0

HowardBraham:
Fixes #51 

I don't think this issue was completely fixed before.  In particular, the original comment was:

> Not sure if this is an issue or not, but `eth_getCode` returns `0x0` for non-contract accounts. My assumption is that it should return `0x`.

eth_getCode was still returning `0x0`.

I deleted 3 lines from `to.js`, wrote 2 new test cases, and changed some old test cases from expecting `0x0` to expecting `0x`.

Let me know if anything is wrong or missing.

### DIFF
--- a/lib/utils/to.js
+++ b/lib/utils/to.js
@@ -4,9 +4,7 @@ module.exports = {
   // Note: Do not use to.hex() when you really mean utils.addHexPrefix().
   hex: function(val) {
     if (typeof val == "string") {
-      if (val === "0x") {
-        return "0x0";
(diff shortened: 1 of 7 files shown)


benjamincburns:
Thanks for contributing, @HowardBraham!

PR #183, opened 2018-09-26
Vrs signature parameters

nicholasjpaterno:
This adds v,r and s signature parameters to the eth_getTransaction and eth_getTransactionReceipt methods.  It is not part of the official web3 [API](https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethgettransaction), however both Parity and Geth provide these parameters allowing the user to verify transactions and blocks.


benjamincburns:
Looks great, thanks for this one, Nick!

PR #191, opened 2018-10-12
Gas usage cumalative gas usage

nicholasjpaterno:
When multiple transactions occur in a block, each transaction's .gasUsed should reflect the gas used in that transaction alone and the cumulativeGas should equal the totalGasUsed at the time of the transaction instead of total gasUsed in the block. This PR should fix the gasUsed and cumulativeGasUsed values to accurately reflect their respective values.

Resolves https://github.com/trufflesuite/ganache-cli/issues/445.

### DIFF
--- a/test/gas.js
+++ b/test/gas.js
@@ -5,6 +5,7 @@ var fs = require("fs");
 var path = require("path");
 var solc = require("solc");
 var to = require("../lib/utils/to.js");
+const pify = require("pify");
 const RSCLEAR_REFUND = 15000;
 const RSELFDESTRUCT_REFUND = 24000;
 // Thanks solc. At least this works!
@@ -30,8 +31,9 @@ describe("Gas", function() {
     var EstimateGasContract;
     var estimateGasContractData;
     var estimateGasContractAbi;
+
     before("compile source", function() {
-      this.timeout(10000);
+      this.timeout(30000);
 
       var source = fs.readFileSync(path.join(__dirname, "EstimateGas.sol"), "utf8");
       var result = solc.compile({sources: {"EstimateGas.sol": source}}, 1);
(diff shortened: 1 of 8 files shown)


test/gas.js
```diff
@@ -272,5 +272,92 @@ describe("Gas", function() {
...
       return testGasExpenseIsCorrect(gasPrice, false, new Web3(Ganache.provider({ mnemonic, gasPrice })))
     })
-  })
+    it('should calculate cumalativeGas and gasUsed correctly when multiple transactions are in a block', function (done){
+      let tempWeb3 = new Web3(Ganache.provider({
+        blockTime: .5, // seconds
+        mnemonic: "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat"
+      }));
+
+      const accounts = [ // configured with mnemonic
```
davidmurdoch: I think these accounts are already available in the `accounts` variable defined previously in the `describe` scope above. Any reason you don't want to just use the ones from that array?

test/gas.js
```diff
@@ -272,5 +272,92 @@ describe("Gas", function() {
       let gasPrice = '0x2000'
       return testGasExpenseIsCorrect(gasPrice, false, new Web3(Ganache.provider({ mnemonic, gasPrice })))
     })
-  })
+    it('should calculate cumalativeGas and gasUsed correctly when multiple transactions are in a block', function (done){
+      let tempWeb3 = new Web3(Ganache.provider({
+        blockTime: .5, // seconds
+        mnemonic: "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat"
```
davidmurdoch: This appears to be the same mnemonic as the one defined in the `mnemonic` variable above. Don't see why we can't just use that variable here, too.

test/gas.js
```diff
@@ -272,5 +272,92 @@ describe("Gas", function() {
...
+      }));
+
+      const accounts = [ // configured with mnemonic
+        '0x627306090abab3a6e1400e9345bc60c78a8bef57',
+        '0xf17f52151ebef6c7334fad080c5704d77216b732',
+      ];
+      const secretKeys = [ // configured with mnemonic
```
davidmurdoch: I don't see `secretKeys` used anywhere. Perhaps this was just left over from debugging?

PR #194, opened 2018-10-16
Set 'to' to null when deploy contract

sc0Vu:
`to` should be null when it is a transaction for contraction creation.

Resolves [trufflesuite/ganache-cli/issues/579](https://github.com/trufflesuite/ganache-cli/issues/579)


davidmurdoch:
Thanks for this PR, @sc0Vu!

I was already work this (and other `0x`-related bugs) so I'm closing in favor of the fix in https://github.com/trufflesuite/ganache-core/pull/197

PR #197, opened 2018-10-17
Audit the use of `0x`, `0x0`, and quantity and fix

davidmurdoch:
Fixes https://github.com/trufflesuite/ganache-cli/issues/579
Fixes https://github.com/trufflesuite/ganache-cli/issues/582
Fixes https://github.com/trufflesuite/ganache/issues/907
Fixes https://github.com/trufflesuite/ganache-core/issues/59
Fixes https://github.com/trufflesuite/ganache/issues/425


lib/utils/to.js
```diff
@@ -71,10 +77,11 @@ module.exports = {
       return val;
     }
     if (typeof val == "string") {
-      if (val.indexOf('0x') != 0) {
-        return parseInt(val)
+      if (val.indexOf('0x') !== 0) {
+        return parseInt(val, 10)
```
nicholasjpaterno: Is it possible that val exceed's Number.MAX_SAFE_INTEGER here?

davidmurdoch: Yes. May we RIP if and when this happens. I'll hopefully handle these edge cases in the upcoming refactor.

PR #209, opened 2018-10-29
fixes forked debugging: refactor vm & state trie so they are short-lived objects

benjamincburns:
I think this may work around trufflesuite/ganache-cli#453 as well, but I haven't tested yet.

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -26,9 +26,6 @@ function StateManager(options, provider) {
     this.blockchain = new BlockchainDouble(options);
   }
 
-  this.vm = this.blockchain.vm;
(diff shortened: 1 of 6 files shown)


lib/statemanager.js
```diff
@@ -21,14 +21,11 @@ function StateManager(options, provider) {
   this.options = options = this._applyDefaultOptions(options || {});
 
   if (options.fork) {
-    this.blockchain = new ForkedBlockchain(options);
+    this.blockchain = new ForkedBlockchain(options, this);
```
benjamincburns: this is stale/unnecessary and creates a circular dependency... will fix

PR #212, opened 2018-10-31
add to and from fields to transaction receipts and tests

eshaben:
getTransactionReceipt was not previously returning a `to` and `from` address field. This PR adds those fields to the transaction receipt.

### DIFF
--- a/test/requests.js
+++ b/test/requests.js
@@ -494,6 +494,32 @@ const tests = function(web3) {
     });
   });
 
+  // This test relies on the one above it running
+  describe("eth_getTransactionReceipt", function() {
+    it("should return to and from address fields in the receipt", function(done) {
+      var provider = web3.currentProvider;
+
+      provider.send(
+        {
+          jsonrpc: "2.0",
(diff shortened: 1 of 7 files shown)


test/requests.js
```diff
@@ -494,6 +494,32 @@ const tests = function(web3) {
...
+  // This test relies on the one above it running
+  describe("eth_getTransactionReceipt", function() {
+    it("should return to and from address fields in the receipt", function(done) {
+      var provider = web3.currentProvider;
+
+      provider.send(
+        {
+          jsonrpc: "2.0",
+          method: "eth_getTransactionReceipt",
+          params: [ "0xe0be65aa192f931857f9e42d3a8d772522fd0196d72981566bd1c8367ae12eb8" ],
```
davidmurdoch: Instead of copying the hash from a previous test let's just run our own transaction within this test and get the receipt for it here.

eshaben: Ok, I made those changes 👌plus a couple others that Nick had pointed out.

PR #214, opened 2018-11-02
add support and tests for ejsvm emitFreeLogs option - Used for debug

jpgonzalezra:
Hey guys, some tools for debugging and creating coverage reports of contracts relay on adding events before compiling and testing, this approach was proven to work on the solidity-coverage project https://github.com/sc-forks/solidity-coverage

By doing this, they elevate the cost of all the calls, and they can't cover views and pure functions, because the EVM will throw when those functions try to emit an event.

The "emitFreeLogs" option allows instantiating an EVM instance without this limitation, enabling any contract to emit an unlimited quantity of events without modifying the block gas limit, and also to emit events during STATICCALLS.

In ethereumjs-vm was merged this parameter. see (https://github.com/ethereumjs/ethereumjs-vm/pull/378).

Related:
sc-forks/solidity-coverage#234
OpenZeppelin/openzeppelin-solidity#876 (comment)


nventuro:
Is there anything blocking this? `ethereumjs-vm` has supported `emitFreeLogs` since v2.5.0 (v3.0.0 was released a couple days ago), and we need `ganache-core` to expose this option so that the cli can enable it.

davidmurdoch:
Closing due to significant conflicts due this PR's age. Sorry about this! I've opened an issue for the feature and referenced this PR:

Issue #148, opened 2018-08-13
Exceeds block gas limit error to call view functions


benjamincburns:
@xxorg3 commented on [Wed Jul 25 2018](https://github.com/trufflesuite/ganache/issues/853)

With solidity view function, you can specific a large number of gas to use. Since it is not a state changing transaction, all the gas cost is not charged for, and the client should be able to allow the call function to run with a gas limit larger than the block limit. 

![image](https://user-images.githubusercontent.com/7900175/43245547-0a96cb8a-90f2-11e8-9549-cbcfbdb5f20a.png)

benjamincburns:
"Needs validation" in that we need someone from our team to verify that geth/parity also support view calls in excess of the block gas limit.

Further, we probably don't want to allow infinite gas on view calls, or if we do we likely want to implement some timeout functionality as otherwise a view which never terminates will take out the whole node.

Uxio0:
I can add to this issue that I just checked and at least Parity has no problems using more gas for calls than block gasLimit

rstormsf:
I can confirm it's the issue with ganache
Parity and geth clients have no problem with calling function with large gas limit even though the real gas limit is 8M. I can easily call geth with 200M gas if it's `eth_call` method used

TripleSpeeder:
FYI, geth has the following option to limit gas usage for calling functions:
> --rpc.gascap value                  Sets a cap on gas that can be used in eth_call/estimateGas (default: 0)

(checked with geth 1.9.3-stable)

davidmurdoch:
Closing as this is now fixed in https://github.com/trufflesuite/ganache-core/pull/449. 

@TripleSpeeder, The new option, `callGasLimit`, wasn't added to the readme, i'll do that now.

PR #215, opened 2018-11-02
Block gas limit when running a view or pure function #148

ccowell:
Added blockgaslimit.js  that test excessive block gas limits when running an expensive `view` and `pure` function found in BlockGasLimit.sol.


ccowell:
Please see [PR #216 ](url)

PR #216, opened 2018-11-02
Sender gas limit vs block gas limit. Also Travis update.

ccowell:
Testing block gas limit with very expensive `view` and `pure` functions.
Added LTS/* to build.

### DIFF
--- a/test/blockgaslimit.js
+++ b/test/blockgaslimit.js
@@ -4,6 +4,8 @@ const Ganache = require("../index");
 const path = require("path");
 const compileAndDeploy = require("./helpers/contracts").compileAndDeploy;
 
+const TARGET = process.env.TARGET || "ganache";
+const PARITY_UNLOCKED_ADDR = process.env.PARITY_UNLOCKED_ADDR || "0x00a329c0648769A73afAc7F9381E08FB43dBEA72";
 const mnemonic = "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
 
 function setUp(options = { mnemonic }, contractName = "BlockGasLimit") {
@@ -17,9 +19,7 @@ function setUp(options = { mnemonic }, contractName = "BlockGasLimit") {
   };
 
   before("setup web3", async function() {
-    const TARGET = "ganache";
-
-    // Ganache provider
+    // Enable Ganache provider
     if (TARGET === "ganache") {
       // eslint-disable-next-line new-cap
       context.provider = new Ganache.provider(context.options);
@@ -34,19 +34,14 @@ function setUp(options = { mnemonic }, contractName = "BlockGasLimit") {
       context.web3 = new Web3(context.provider);
     }
 
-    // Parity setup
+    // Enable Parity provider
     if (TARGET === "parity") {
-      // const address = "0x268918bf9b500219d2823216fdae375fa5e4c3c6";
-      // const password = "c87509a1c067bbde78beb793e6fa76530b6382a4c0241e5e4a9ec0a0f44dc0df";
-
       context.options.blockTime = 2000;
       context.provider = new Web3(new Web3.providers.HttpProvider("http://localhost:8545"));
       context.web3 = new Web3(context.provider);
 
-      const address = "0x00a329c0648769A73afAc7F9381E08FB43dBEA72";
-      await context.web3.eth.personal.unlockAccount(address, "", null);
-      // const accounts = await context.web3.eth.personal.getAccounts();
-      // console.log(accounts[0]);
+      const response = await context.web3.eth.personal.unlockAccount(PARITY_UNLOCKED_ADDR, "", null);
+      assert.strictEqual(response, true);
     }
   });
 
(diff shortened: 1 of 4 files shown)


test/blockgaslimit.js
```diff
@@ -0,0 +1,73 @@
...
+
+    const isVeryHighGasLimit = clientGasLimit > block.gasLimit && clientGasLimit < Number.MAX_SAFE_INTEGER;
+    assert.strictEqual(isVeryHighGasLimit, true);
+
+    try {
+      // Attempt to run an expensive view function
+      await context.instance.methods.expensiveOperation(iterations).call({ gas: clientGasLimit });
+      assert.fail("Expecting a block gas limit error when executing a expensive 'view' function");
+    } catch (error) {
+      assert.strictEqual(error.message, "Exceeds block gas limit");
```
nicholasjpaterno: It looks like [Uxio0](https://github.com/trufflesuite/ganache-core/issues/148#issuecomment-435048323) was able to confirm that parity allows for view calls exceeding the the block gas limit.  Were you able to [verify](https://github.com/trufflesuite/ganache-core/issues/148#issuecomment-416342014) if geth does as well?

test/blockgaslimit.js
```diff
@@ -0,0 +1,73 @@
+const Web3 = require("web3");
+const assert = require("assert");
+const Ganache = require("../index");
+const path = require("path");
+const compileAndDeploy = require("./helpers/contracts").compileAndDeploy;
+
+const mnemonic = "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
+
+function setUp(options = { mnemonic }, contractName = "BlockGasLimit") {
+  let context = {
```
nicholasjpaterno: We can declare this as const.  Same with lines 41-43.

test/blockgaslimit.js
ccowell: Geth does allows a sender to specify a gasLimit > block.gasLimit. But have not confirmed Parity locally yet.

test/blockgaslimit.js
ccowell: Update is on the way.

test/blockgaslimit.js
nicholasjpaterno: If that's the case ( and assuming parity does as well ) then we definitely need to discuss supporting this feature, as well as the implications, with the team - pinging @davidmurdoch 
@seesemichaelj.  As @benjamincburns alluded, this potentially offers an avenue for a slowloris like 'attack' against an individual node.  Maybe we can explore what 'safe-guard' (if any) geth/parity have in place.

test/blockgaslimit.js
```diff
@@ -0,0 +1,119 @@
...
+    accounts: [],
+    contractArtifact: {},
+    instance: null
+  };
+
+  before("setup web3", async function() {
+    // Enable Ganache provider
+    if (TARGET === "ganache") {
+      // eslint-disable-next-line new-cap
+      context.provider = new Ganache.provider(context.options);
```
davidmurdoch: just use `Ganache.provider(context.options)` without the `new` keyword here and remove the eslint comment.

test/blockgaslimit.js
```diff
@@ -0,0 +1,119 @@
...
+      context.options.blockTime = 2000;
+      context.web3 = new Web3(context.provider);
+    }
+
+    // Enable Geth provider
+    if (TARGET === "geth") {
```
davidmurdoch: There is a lot of code duplication between the multiple test targets here so I think these should all be refactored so there isn't so much duplication between them.

Also, we'll soon be migrating nearly all of our tests to support multiple targets (like you've done here) and this multi target configuration you've made would be very useful to have available in a reusable `./helpers/` file. The tricky part will be that we'll need some sort of automated mechanism for spooling up geth and parity locally for testing.

Could you pull your `before("setup web3, fn)" functionality out into it's own file that would could reuse?

test/blockgaslimit.js
```diff
@@ -0,0 +1,119 @@
...
+      context.web3 = new Web3(context.provider);
+    }
+
+    // Enable Parity provider
+    if (TARGET === "parity") {
+      context.options.blockTime = 2000;
+      context.provider = new Web3(new Web3.providers.HttpProvider("http://localhost:8545"));
+      context.web3 = new Web3(context.provider);
+
+      const response = await context.web3.eth.personal.unlockAccount(PARITY_UNLOCKED_ADDR, "", null);
```
davidmurdoch: Is there a way to programmatically get the parity address?

test/blockgaslimit.js
```diff
@@ -0,0 +1,119 @@
...
+      assert.fail("Expecting a block gas limit error when executing a expensive 'view' function");
+    } catch (error) {
+      assert.strictEqual(error.message, "Exceeds block gas limit");
+    }
+  });
+
+  it("when calling a 'pure' function should generate a block gas limit error", async function() {
+    const block = await context.web3.eth.getBlock("latest");
+
+    const isVeryHighGas = clientGasLimit > block.gasLimit && clientGasLimit < Number.MAX_SAFE_INTEGER;
```
davidmurdoch: Why do you need to check `clientGasLimit < Number.MAX_SAFe_INTEGER` here?

test/blockgaslimit.js
davidmurdoch: It sounds to me like this call _shouldn't_ throw? Assuming we want to align with geth and parity, is this test backwards?

ccowell: The test is backwards in order to expose the different responses from the tool. Will start working on a fix to align Ganache with Geth and Parity.

test/blockgaslimit.js
ccowell: I can start placing more focus on a refactoring effort. Also decided to use code duplication to help validate my local Geth and Parity environments. Figured it would be a good setup for other until we all share a consistent geth/parity environment.

PR #218, opened 2018-11-07
Issue 678: Testing solidity strings with getters/setters

ccowell:
Unable to confirm reported bug.

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -243,26 +243,14 @@ BlockchainDouble.prototype.putBlock = function(block, logs, receipts, callback) 
     var requests = [
       self.database.blocks.push.bind(self.database.blocks, block),
       self.database.blockLogs.push.bind(self.database.blockLogs, logs),
-      self.database.blockHashes.set.bind(
-        self.database.blockHashes,
-        to.hex(block.hash()),
-        length
-      )
+      self.database.blockHashes.set.bind(self.database.blockHashes, to.hex(block.hash()), length)
     ];
 
     block.transactions.forEach(function(tx, index) {
       var txHash = to.txHash(tx);
       requests.push(
-        self.database.transactions.set.bind(
-          self.database.transactions,
-          txHash,
-          tx
-        ),
-        self.database.transactionReceipts.set.bind(
-          self.database.transactionReceipts,
-          txHash,
-          receipts[index]
-        )
+        self.database.transactions.set.bind(self.database.transactions, txHash, tx),
+        self.database.transactionReceipts.set.bind(self.database.transactionReceipts, txHash, receipts[index])
       );
     });
 
@@ -305,28 +293,15 @@ BlockchainDouble.prototype.popBlock = function(callback) {
         var txHash = to.txHash(tx);
 
         requests.push(
-          self.database.transactions.del.bind(
-            self.database.transactions,
-            txHash
-          ),
-          self.database.transactionReceipts.del.bind(
-            self.database.transactionReceipts,
-            txHash
-          )
+          self.database.transactions.del.bind(self.database.transactions, txHash),
+          self.database.transactionReceipts.del.bind(self.database.transactionReceipts, txHash)
         );
       });
 
       requests.push(
-        self.database.blockLogs.pop.bind(
-          self.database.blockLogs
-        ),
-        self.database.blockHashes.del.bind(
-          self.database.blockHashes,
-          blockHash
-        ),
-        self.database.blocks.pop.bind(
-          self.database.blocks
-        ) // Do this one last in case anything relies on it.
+        self.database.blockLogs.pop.bind(self.database.blockLogs),
+        self.database.blockHashes.del.bind(self.database.blockHashes, blockHash),
+        self.database.blocks.pop.bind(self.database.blocks) // Do this one last in case anything relies on it.
       );
 
       async.series(requests, function(err) {
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,10647 @@
+{
+  "name": "ganache-core",
+  "version": "2.2.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.0.0"
+      }
+    },
+    "@babel/highlight": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0.tgz",
+      "integrity": "sha512-UFMC4ZeFC48Tpvj7C8UgLvtkaUuovQX+5xNWrsIoMG8o2z+XFKjKaN9iVmS84dPwVN00W4wPmqvYoZF3EGAsfw==",
+      "dev": true,
+      "requires": {
+        "chalk": "^2.0.0",
+        "esutils": "^2.0.2",
+        "js-tokens": "^4.0.0"
+      },
+      "dependencies": {
+        "ansi-styles": {
+          "version": "3.2.1",
+          "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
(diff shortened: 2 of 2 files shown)


benjamincburns:
Looks like a lot of commits for relatively small net change - I'd recommend squashing the commits and doing `git push -f` to keep the change log easier to read

test/DynamicStringLength.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.4.2;
+
+contract DynamicStringLength {
+  string public testString;
+  uint public testId = 246;
```
benjamincburns: looks like this & the event aren't used

test/solidity_dynamic_strings.js
```diff
@@ -0,0 +1,54 @@
+const Web3 = require("web3");
+const assert = require("assert");
+const Ganache = require("../index");
+const path = require("path");
+const compileAndDeploy = require("./helpers/contracts").compileAndDeploy;
+
+const mnemonic = "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
```
benjamincburns: The mnemonic defines the list of accounts that ganache will use - but you're using web3.eth.getAccounts below (which is good) so there's no need to hard code a mnemonic.

It's better if tests use randomized accounts - we want tests to look like best practice code examples.

test/solidity_dynamic_strings.js
```diff
@@ -0,0 +1,54 @@
...
+      contractName,
+      context.web3
+    );
+    context.instance = context.contractArtifact.instance;
+  });
+  return context;
+}
+
+describe("Solidity variable string length", function() {
+  const context = setUp();
```
benjamincburns: readability: since this is the only place the `setUp` function is called, I'm not sure what value it's providing. The usual flow for mocha tests is as shown in the snippet below, and this breaks that flow

```
describe(...
  describe( ...
    before(...
    it(...
    after(...
  }
}
```

test/solidity_dynamic_strings.js
```diff
@@ -0,0 +1,54 @@
...
+    provider: null,
+    web3: null,
+    accounts: [],
+    contractArtifact: {},
+    instance: null
+  };
+
+  before("setup web3", async function() {
+    context.provider = Ganache.provider(context.options);
+    context.options.blockTime = 2000;
```
benjamincburns: I wouldn't set a block time - will just make the test slow

benjamincburns: FWIW, only ever set block time if you're explicitly testing interval mining

test/solidity_dynamic_strings.js
```diff
@@ -0,0 +1,54 @@
...
+    context.web3 = new Web3(context.provider);
+  });
+
+  before("compile source", async function() {
+    this.timeout(10000);
+    context.contractArtifact = await compileAndDeploy(
+      path.join(__dirname, ".", `${contractName}.sol`),
```
benjamincburns: no need to do `__dirname, ".", contractFileName` -- the dot is implicit - it's like saying `/./contractName.sol` which is the same as `/contractName.sol`

test/solidity_dynamic_strings.js
```diff
@@ -0,0 +1,54 @@
...
+      contractName,
+      context.web3
+    );
+    context.instance = context.contractArtifact.instance;
```
benjamincburns: unless you're using `contractArtifact`, I'd drop `contractArtifact` altogether and do

```
const { instance } = await compileAndDeploy( ...
```

^ sweet ES6 destructuringing syntax will pull the `instance` field out of the object returned by `compileAndDeploy` and assign it to a const variable of the same name.

test/solidity_dynamic_strings.js
```diff
@@ -0,0 +1,54 @@
...
+  });
+  return context;
+}
+
+describe("Solidity variable string length", function() {
+  const context = setUp();
+  it("replacing a long string with a short string", async function() {
+    let response;
```
benjamincburns: usually I only defined vars when they're first assigned unless for some unavoidable reason the variable needs to be accessed in a higher-level scope than the one in which it was assigned

ccowell: Great suggestions for overall refactoring efforts.

test/solidity_dynamic_strings.js
ccowell: Great suggestions for overall refactoring efforts.

test/solidity_dynamic_strings.js
ccowell: Great suggestions for overall refactoring efforts.

test/solidity_dynamic_strings.js
ccowell: Great suggestions for overall refactoring efforts.

test/solidity_dynamic_strings.js
ccowell: Great suggestions for overall refactoring efforts.

test/DynamicStringLength.sol
ccowell: Good catch.

test/helpers/compile_deploy.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {Array|String} contractFileNames List of imported contracts
+ * @param {Object} web3 Web3 interface
+ * @returns {Object} context
+ */
+async function compileAndDeploy(contractPath, mainContractName, contractFileNames = [], web3) {
+  const selectedContracts = contractFileNames.length === 0 ? [mainContractName] : contractFileNames;
+  const contractSources = selectedContracts.map((contractName) => {
+    return { [`${contractName}.sol`]: readFileSync(`${contractPath}${contractName}.sol`, "utf8") };
+  });
+  const sources = Object.assign(...contractSources);
```
nicholasjpaterno: I'm not sure we need this line here since this is the last usage of `contractSources`.  If we do need a copy (in case `compile` mutates `sources` on line 19) I would change this to explicitly use an empty object as the source parameter `Object.assign({}, contractSources)` or just the spread operator like `const sources = {...contractSources}` and use the original `contractSources` in the return on line 40.

test/requests.js
```diff
@@ -505,22 +505,6 @@ const tests = function(web3) {
     });
   });
 
-  describe("eth_getTransactionReceipt", function() {
```
nicholasjpaterno: I think maybe this was removed by accident?  This looks like it was part of @eshaben's [commit](https://github.com/trufflesuite/ganache-core/commit/4d3c606852ce7445a35e9767d770386ae5b8f6a8) for PR [212](https://github.com/trufflesuite/ganache-core/pull/212)

test/requests.js
```diff
@@ -661,11 +645,6 @@ const tests = function(web3) {
       assert(receipt.contractAddress, "should have deployed a contract");
     });
 
-    it("should return null for the to field due to contract creation (eth_getTransactionReceipt)", async function() {
```
nicholasjpaterno: Same as above ^ this may have been removed by accident from PR [212](https://github.com/trufflesuite/ganache-core/pull/212)?

test/solidity_dynamic_strings.js
```diff
@@ -0,0 +1,51 @@
...
+
+      // Only confirm the long string
+      let status = await instance.methods.confirmSetting(text).call();
+      assert.strictEqual(status, true);
+
+      // Retrieve and confirm the long string
+      let response = await instance.methods.testString().call();
+      assert.strictEqual(response, text);
+
+      // Set and confirm (within the EVM) a long string setting to a public variable
```
nicholasjpaterno: short* string

test/solidity_dynamic_strings.js
```diff
@@ -0,0 +1,51 @@
+const assert = require("assert");
+const pretest = require("./helpers/pretest_setup");
```
nicholasjpaterno: We can destructure `setup` here.

test/helpers/compile_deploy.js
ccowell: Agreed. Wanted to keep the code readable.

test/requests.js
ccowell: This may be some squashing magic gone wrong

test/requests.js
ccowell: This may be some squashing magic gone wrong

test/requests.js
eshaben: This was definitely from me

test/solidity_dynamic_strings.js
ccowell: Are you suggesting to rename the variable? `testString` is being used to store a long string followed by a short string.

test/solidity_dynamic_strings.js
ccowell: I was opting for readability on this one. Fortunately this is the only export in pretest_setup for now.

test/solidity_dynamic_strings.js
ccowell: Simple fix 👍

test/solidity_dynamic_strings.js
ccowell: I'll just rename `setup` to some thing more descriptive. 👍

PR #224, opened 2018-11-14
Issue 678: Storing very long and short strings into a variable

ccowell:
- Starting to store contracts into `test/contracts` directory
- Added a new compileAndDeploy method (preparing to deprecate previous implementation)
- Extracted pretest setups into a `test/helper` module

### DIFF
--- a/test/contracts/DynamicStringLength.sol
+++ b/test/contracts/DynamicStringLength.sol
@@ -5,12 +5,13 @@ contract DynamicStringLength {
   string public testString;
   DynamicStringLengthCheck validate = new DynamicStringLengthCheck();
 
-  function set(string _s) public {
+  function setAndConfirm(string _s) public {
     testString = _s;
-    validate.stringSetting(_s, testString);
+    require(confirmStringInternally(_s));
+    validate.stringSettingExternally(_s, testString);
   }
 
-  function confirmSetting(string _s) public view returns (bool) {
+  function confirmStringInternally(string _s) public view returns (bool) {
     return keccak256(abi.encodePacked(_s)) == keccak256(abi.encodePacked(testString));
   }
 }
--- a/test/contracts/DynamicStringLength.sol
+++ b/test/contracts/DynamicStringLength.sol
@@ -1,0 +1,16 @@
+pragma solidity ^0.4.2;
+import "./DynamicStringLengthCheck.sol";
+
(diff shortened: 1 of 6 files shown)


test/contracts/DynamicStringLength.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity ^0.4.2;
+import "./DynamicStringLengthCheck.sol";
+
+contract DynamicStringLength {
+  string public testString;
+  DynamicStringLengthCheck validate = new DynamicStringLengthCheck();
```
benjamincburns: I'm not sure I follow why this extra contract is necessary?

ccowell: The extra contract is there to validate the test string during contract-to-contract interaction.

test/helpers/compile_deploy.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Compile and deploy the selected contract(s)
+ * @param {String} contractPath  Path to contracts directory
+ * @param {String} mainContractName  Name of the main contract (without .sol extension)
+ * @param {Array|String} contractFileNames List of imported contracts
+ * @param {Object} web3 Web3 interface
+ * @returns {Object} context
+ */
+async function compileAndDeploy(contractPath, mainContractName, contractFileNames = [], web3) {
```
benjamincburns: doesn't this duplicate `test/helpers/contracts.js` ?

test/contracts/DynamicStringLengthCheck.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity ^0.4.2;
+
+contract DynamicStringLengthCheck {
+  function stringSetting(string oldText, string newText) pure public {
+    require(keccak256(abi.encodePacked(oldText)) == keccak256(abi.encodePacked(newText)));
```
benjamincburns: I was thinking that ideally this would be in the same contract above. I'm not sure why this would need to be in a separate contract?

test/helpers/compile_deploy.js
ccowell: It adds multi-contract support and also export accounts. Also some refactoring.

ccowell:
Will reopen when `helper` files are finalized.

PR #227, opened 2018-11-18
Pending transactions

eshaben:
eth_getTransactionByHash used to return null for pending transactions, but according to JSON RPC spec, if the transaction is pending only these fields should return null: `blockHash`, `blockNumber`, and `transactionIndex`. This PR fixes this bug so now we return the transaction object with the above mentioned fields as null for pending transactions. And for eth_getTransactionByReceipt on pending transactions, we return null.


eshaben:
Made a fundamental mistake and the problem still exists. Will create a new PR when I am ready!

lib/subproviders/geth_api_double.js
```diff
@@ -212,7 +212,13 @@ GethApiDouble.prototype.eth_getTransactionReceipt = function(hash, callback) {
     var result = null;
 
     if (receipt) {
-      result = receipt.toJSON();
+      // if the block is undefined the transaction is still pending
+      if (typeof receipt.block === "undefined") {
```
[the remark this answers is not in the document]
eshaben: much simpler! thanks!

lib/subproviders/geth_api_double.js
```diff
@@ -227,7 +233,20 @@ GethApiDouble.prototype.eth_getTransactionByHash = function(hash, callback) {
     var result = null;
 
     if (receipt) {
-      result = txhelper.toJSON(receipt.tx, receipt.block);
+      // if the block is undefined the transaction is still pending
+      if (typeof receipt.block === "undefined") {
```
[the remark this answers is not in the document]
eshaben: 👍

lib/blockchain_double.js
```diff
@@ -1152,6 +1127,16 @@ BlockchainDouble.prototype.getTransaction = function(hash, callback) {
 
 BlockchainDouble.prototype.getTransactionReceipt = function(hash, callback) {
   hash = to.hex(hash);
+  const pendingTxs = this.pending_transactions;
+
+  if (pendingTxs.length > 0) {
```
davidmurdoch: No need for the `length > 0` check here, we can let the `for` loop logic handle the 0 length case.

lib/blockchain_double.js
```diff
@@ -305,28 +293,15 @@ BlockchainDouble.prototype.popBlock = function(callback) {
...
         requests.push(
-          self.database.transactions.del.bind(
-            self.database.transactions,
-            txHash
-          ),
-          self.database.transactionReceipts.del.bind(
-            self.database.transactionReceipts,
-            txHash
-          )
+          self.database.transactions.del.bind(self.database.transactions, txHash),
```
davidmurdoch: I think these will need to be updated to reference `self.data.transactions` to make this merge-conflict free.

lib/subproviders/geth_api_double.js
```diff
@@ -212,7 +212,9 @@ GethApiDouble.prototype.eth_getTransactionReceipt = function(hash, callback) {
     var result = null;
 
     if (receipt) {
-      result = receipt.toJSON();
+      if (receipt.block) {
```
davidmurdoch: Combine this `receipt.block` truthiness check with the `if (receipt) {` check above it, i.e., `if (receipt && receipt.block) {`.

lib/subproviders/geth_api_double.js
```diff
@@ -227,9 +229,19 @@ GethApiDouble.prototype.eth_getTransactionByHash = function(hash, callback) {
     var result = null;
 
     if (receipt) {
+      // if there is no block then the transaction is still pending
+      if (!receipt.block) {
+        // assemble the block object with null values for pending transactions
+        receipt.block = {
```
davidmurdoch: Would just using `new Block()` here work instead of mocking a `Block` object?

eshaben: I don't think that using `new Block()` here would work because it generates a `hash` for you and I can't seem to overwrite it by passing in null. I am able to overwrite `header.number` and set that to `null` but not able to with `hash` and that needs to be null when it's pending.

eshaben:
@davidmurdoch I have made all of the changes except the `new Block()` suggestion which I left a comment about!

test/requests.js
```diff
@@ -1200,6 +1200,73 @@ const tests = function(web3) {
...
+        method: "eth_getTransactionByHash",
+        params: [result.result]
+      });
+
+      assert.strictEqual(null, transactionObject.result.blockHash);
+      assert.strictEqual(null, transactionObject.result.blockNumber);
+      assert.strictEqual(null, transactionObject.result.transactionIndex);
+    });
+
+    it("should return values for blockHash, blockNumber, and transactionIndex once mined", async function() {
```
davidmurdoch: Looks like this test needs to be updated so it does mine the block, then does `eth_getTransactionByHash`, then checks that the values are present. Actually, you can go ahead and combine what this test here intended to test with the previous test; just add a call to `evm_mine` after the `asserts` then get call `eth_getTransactionByHash` again and check that it has the values you expect.

PR #231, opened 2018-11-21
Incorrect tx receipt trie root

nicholasjpaterno:
Neither the `transactionsTrie` nor the `receiptsTrie` were being calculated and the resulting `transactionsRoot` and `receiptsRoot` were always identical (EMPTY_TRIE_ROOT) when there were 1 or more transactions in a block.

This PR suggests a fix by:
Creating 2 empty Tries and populating them with tx index (within the block) as the keys and transactions/receipts for their respective values.  The resulting Trie's roots are then copied to `block.header.transactionsTrie` and `block.header.receiptTrie` respectively.


lib/blockchain_double.js
```diff
@@ -583,6 +583,21 @@ BlockchainDouble.prototype.processBlock = function(vm, block, commit, callback)
 
       block.header.gasUsed = utils.toBuffer(to.hex(totalBlockGasUsage));
 
+      const txTrie = new Trie();
+      const rcptTrie = new Trie();
+      const promises = [];
+      const putInTrie = function(trie, key, val) {
+        return new Promise((resolve, reject) => {
+          try {
+            trie.put(key, val, (...args) => {
```
davidmurdoch: `...args` seems unnecessary here. I think we should be handling an `error` param in this callback (with just `reject(error)`).

davidmurdoch: actually... let's just use node's `promisify` on `trie.put` instead of wrapping it with `new Promise`. I don't think the `try`/`catch` is necessary as the `key` and `val` passed in will always be `Buffers` in this code, at least 2 bajillion tests will fail if someone comes in here and tries changes things.

test/block_tags.js
```diff
@@ -35,128 +34,77 @@ var contract = {
...
-        if (err) {
-          return done(err);
-        }
-        initial = result;
-        initial.nonce = to.number(initial.nonce);
-        done();
-      }
-    );
+  before("Get initial balance and nonce", async function() {
+    const [balance, nonce] = [
```
davidmurdoch: love that you got rid of `async.parallel` here!

test/block_tags.js
```diff
@@ -35,128 +34,77 @@ var contract = {
...
+      ...(await Promise.all([
```
davidmurdoch: I don't think we need to destructure the array here. Once awaited the result is already an array of resolved values (not pending Promises).

For example:

```
const [a, b] = await Promise.all([
  new Promise((resolve)=>{resolve("A")}),
  new Promise((resolve)=>{resolve("B")})
]);
console.log(a, b);
```
outputs `"A"`, `"B"`

test/block_tags.js
nicholasjpaterno: Same :joy:

test/block_tags.js
nicholasjpaterno: Agreed!  :100:

lib/blockchain_double.js
nicholasjpaterno: Agree 1000%

nicholasjpaterno:
> I'd prefer not to see test refactor changes in these commits (I know we've done these drive-by refactorings in the past) since chris is currently refactoring everything to async/await right now. There may end up being merge conflicts with #313.

Agreed, just for background this test was updated before any refactoring started.  I did make sure to notify and remind chris of the changes to hopefully avoid any conflicts.

PR #241, opened 2018-12-03
Retrieve an uncle id by index

ccowell:
### PR Summary
**geth_api_double.js**
- Added RPC methods for get eth_getUncleBy*AndIndex

**block_helper.js**
- Set uncle array

**forkedblockchain.js**
- Added uncle stats to a new forked block

**compileAndDeploy.js, pretest_setup.js**
- Adding these files to each test as a refactoring effort until merged into `develop`

**contract.js**
- Added some comments (preparing to migrate to compileAndDeploy.js)

**rpc.js**
- Starting to modularizing the rpc methods as part of the refactoring effort.


lib/utils/forkedblockchain.js
```diff
@@ -216,6 +218,7 @@ ForkedBlockchain.prototype.getFallbackBlock = function(numberOrHash, cb) {
     block.header.gasUsed = utils.toBuffer(json.gasUsed);
     block.header.timestamp = utils.toBuffer(json.timestamp);
     block.header.extraData = utils.toBuffer(json.extraData);
+    block.uncleHeaders = json.uncles;
```
davidmurdoch: I wonder if we should guard against `json.uncles` not being an array, or not being an array with data formatted as we'd expect?

test/helpers/rpc.js
```diff
@@ -0,0 +1,25 @@
...
+  return send({
+    id: "1",
+    jsonrpc: "2.0",
+    method,
+    params
+  });
+};
+
+module.exports = {
+  rpcSend
```
davidmurdoch: This helper could be useful in the future within the forking methods ganache-core in the future. Maybe we should move it to `./lib/utils` now? Thoughts on doing this now vs later?

Also, rename `rpcSend` to just `send` in both spots in this file.

test/helpers/rpc.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Array} params Argument for method call
+ * @param {Object} web3 Web3 interface
+ */
+
+const rpcSend = (method = "", params = [], web3) => {
+  const send = pify(web3._provider.send.bind(web3._provider));
+  // an "id" is required here because the web3 websocket provider (v1.0.0-beta.35) throws if it is
+  // missing (it's probably just a bug on their end)
+  return send({
+    id: "1",
```
davidmurdoch: An auto-incrementing id here would probably be better, I think.

The [JSON-PRC spec](https://www.jsonrpc.org/specification) intends for the `id` to be used to match requests and responses. Not sure how useful this is for us when using an http provider, but it would be useful in websocket communication.

test/eth_getUncles.js
```diff
@@ -0,0 +1,113 @@
+const assert = require("assert");
+const Web3 = require("web3");
+const { provider } = require("../index");
```
davidmurdoch: Use:

```
var Ganache = require(process.env.TEST_BUILD
  ? "../build/ganache.core." + process.env.TEST_BUILD + ".js"
  : "../index.js");
```

here and all other tests that import ganache-core so we can also run our tests against the webpacked version of the library.

test/eth_getUncles.js
```diff
@@ -0,0 +1,113 @@
...
+      assert.strictEqual(result, validDoubleUncleIds.position0);
+
+      method = "eth_getUncleByBlockHashAndIndex";
+      params = [testBlock.hash, 0];
+      ({ result } = await rpcSend(method, params, forkedWeb3));
+      assert.strictEqual(result, validDoubleUncleIds.position0);
+
+      // Validate single uncle id using Ganache rpc methods
+      method = "eth_getUncleByBlockNumberAndIndex";
+      params = [twoUncleBlockNumber, 1];
```
davidmurdoch: If you use `"0x1"` here instead of 1 does everything still work? Can you add a test proving this? Same with the block number -- make sure it works with hex encoding and provide a test.

test/eth_getUncles.js
```diff
@@ -0,0 +1,113 @@
...
+const Web3 = require("web3");
+const { provider } = require("../index");
+const { rpcSend } = require("./helpers/rpc");
+
+describe("Uncle support", function() {
+  describe("checking uncle id through a forked blockchain", function() {
+    // Set up forking from mainnet
+    const forkedWeb3 = new Web3(
+      provider({
+        fork: "https://mainnet.infura.io"
```
davidmurdoch: Use `https://mainnet.infura.io/v3/f6194ba837b640bba3db3d29cc658b07` instead, as project ids are now required by infura.

lib/subproviders/geth_api_double.js
```diff
@@ -568,30 +568,83 @@ GethApiDouble.prototype.eth_getUncleCountByBlockNumber = function(blockNumber, c
...
-  callback(null, {});
+GethApiDouble.prototype.eth_getUncleByBlockNumberAndIndex = function(blockNumber, index, callback) {
+  // Check for valid uncle index
+  if (index !== 0 && index !== 1) {
+    const error = new Error("Uncle array index must be 0 or 1");
+    callback(error);
+    return;
+  }
+
+  this.state.getBlock(blockNumber, function(err, block) {
```
nicholasjpaterno: Can flatten these ifs? Im not sure if that would be more or less readable..  Something like:
```
  this.state.getBlock(hash, function(err, block) {
    let error = err;
    if (err && err.message && err.message.indexOf("Key not found in database") >= 0) {
      error = new Error("Unknown block hash");
    } else if (index >= block.uncleHeaders.length) {
      error = new Error("Uncle array index is out of range");
    } else {
      callback(null, uncleHeaders[index]);
      return;
    }
    callback(error);
  });
```

lib/utils/forkedblockchain.js
ccowell: That particular validation logic may not be reachable. Not sure if I can think of a scenario where uncles are malformed and we don't have more fundamental issues.

test/helpers/rpc.js
ccowell: Thinking ... we should hold off on updating `./lib/utils` for new tests.

Can definitely swap `rpcSend` to `send`, didn't want it to be confused with existing send().

test/helpers/rpc.js
ccowell: I'm not sure either, but can add auto-incrementing but not sure how to test this yet.

lib/subproviders/geth_api_double.js
```diff
@@ -568,30 +568,83 @@ GethApiDouble.prototype.eth_getUncleCountByBlockNumber = function(blockNumber, c
...
- * @param {Object} result - A block object,
+ * @throws - Block hash not found
+ * @throws {OutOfRange} - Uncle array index must be 0 or 1
+ * @throws {OutOfRange} - Uncle array index is out of range
+ * @returns {number} result - Integer count of uncles in this block.
  */
 GethApiDouble.prototype.eth_getUncleByBlockHashAndIndex = function(hash, index, callback) {
-  callback(null, {});
+  // Check for valid uncle index
+  if (index !== 0 && index !== 1) {
```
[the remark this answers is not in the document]
ccowell: This still looks good to me. We want to fail early if an invalid uncle index is detected. Anything that is not a `0` or `1` should not fetch a block.

Assumption: Max uncle count is 2 for all ethereum blocks.

lib/geth_api_double.js
```diff
@@ -0,0 +1,836 @@
+var utils = require("ethereumjs-util");
```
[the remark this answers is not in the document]
ccowell: You caught me! lol.

lib/subproviders/geth_api_double.js
ccowell: That can be updated.

lib/subproviders/geth_api_double.js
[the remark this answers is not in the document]
nicholasjpaterno: @ccowell good call - I must have been half asleep when I reviewed this!  `!(index===0 || index===1)`

lib/subproviders/geth_api_double.js
ccowell: On second thought, this may make things more unreadable considering that `callback` accept any parameter(s).

lib/subproviders/geth_api_double.js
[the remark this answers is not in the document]
ccowell: Reversing the logic may be more readable.

lib/subproviders/geth_api_double.js
nicholasjpaterno: My thought process here is eliminating the number of `callback()` calls.  Such that it either succeeds or it doesn't and reduces the amount of nested blocks which (maybe?) makes it easier to follow.  I wrote this quickly and didn't test this or anything and I realize now I may have missed something and in addition maybe this can be simplified even further.  I'll test quickly in the morning with a fresh pair of eyes.

lib/subproviders/geth_api_double.js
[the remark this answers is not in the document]
nicholasjpaterno: I think the `&&` are potentially more perfomant, it will short circuit if the first conditional evals to false.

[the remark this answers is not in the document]
ccowell: I agree with all your points. lol.

lib/subproviders/geth_api_double.js
nicholasjpaterno: So what I originally suggested was slightly flawed - if err exists but the rest of the conditional fails we might not return an error as expected.  Either way - just some ideas..
```
 this.state.getBlock(hash, function(err, block) {
    let error;
    if (err) {
        error = err.message && err.message.indexOf("Key not found in database") >= 0
            ? new Error("Unknown block hash")
            : err;
    } else if (index >= block.uncleHeaders.length) {
      error = new Error("Uncle array index is out of range");
    } else {
      callback(null, uncleHeaders[index]);
    }
    callback(error);
  });
```
or with *1* callback call but likely even less readable:
```
 this.state.getBlock(hash, function(err, block) {
    let params;
    if (err) {
        params = err.message && err.message.indexOf("Key not found in database") >= 0
            ? [new Error("Unknown block hash")]
            : [err];
    } else if (index >= block.uncleHeaders.length) {
      params = [new Error("Uncle array index is out of range")];
    } else {
      params = [null, uncleHeaders[index]];
    }
    callback(...params);
  });
```

ccowell: This may break a linting rule but may be worth the override.

nicholasjpaterno: If we have to override the linter then the changes aren't worth it in my opinion.  Let's just aim for reducing the number of callback()'s and let's opt for the early return on invalid index's rather than wrapping the block in a conditional.

ccowell:
Will reopen when a `helper` files are finalized.

PR #243, opened 2018-12-04
Refactor Transactions throughout core

davidmurdoch:
Fixes issue where transaction hashes weren't correctly calculated.


lib/utils/transaction.js
```diff
@@ -0,0 +1,328 @@
...
+    this.type = type;
+
+    fixProps(this, data);
+    initData(this, data);
+  }
+
+  static get types() {
+    // values must be squares of 2
+    return {
+      none: 0,
```
nicholasjpaterno: Do we want `none` to be 1  ie: `2^0` instead of 0?

Also I think the comment should read 'powers' of 2.  Ie: `2^n`

lib/utils/transaction.js
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * In a signed transaction a nonce of Buffer([]) is the same as Buffer([0]),
+   * but in a fake transaction Buffer([]) is null and Buffer([0]) is 0.
+   *
+   * @param {Buffer} expectedNonce The value of the from account's next nonce.
+   */
+  validateNonce(expectedNonce) {
+    let nonce;
+    if (this.isSigned()) {
+      nonce = this.nonce.length === 0 ? BUFFER_ZERO : this.nonce;
```
nicholasjpaterno: Thoughts on a small change for readability?  Something like:
```
validateNonce(expectedNonce) {
    let nonce;
    if (this.isSigned() && this.nonce.length === 0) {
      nonce = BUFFER_ZERO;
    } else {
      nonce = this.nonce;
    }
    return nonce.equals(expectedNonce);
  }
```

lib/utils/transaction.js
```diff
@@ -0,0 +1,328 @@
...
+   * we don't recalc the chainId when we set the v to soemthing else.
+   *
+   * Note: If the transaction is a fake transaction this hash method gets
+   * overridden in the constructor.
+   *
+   * @param {Boolean} [includeSignature=true] whether or not to inculde the signature
+   * @return {Buffer}
+   */
+  hash(includeSignature) {
+    if (includeSignature === undefined) {
```
nicholasjpaterno: Thoughts on eliminating this conditional and using a default parameter?

lib/blockchain_double.js
```diff
@@ -478,9 +451,10 @@ BlockchainDouble.prototype.sortByPriceAndNonce = function() {
   var sortedByNonce = {};
 
   self.pending_transactions.forEach((tx) => {
-    const from = to.hex(tx.from);
-    if (sortedByNonce[from]) {
-      sortedByNonce[from].push(tx);
+    const from = tx.from.toString("hex");
```
nicholasjpaterno: Thoughts on generating hex values on tx instantiation and held as member variables instead of recalculating `value.toString("hex")` each time?

PR #268, opened 2019-01-01
bug fix for transaction receipts with multiple transactions in one block

davidbancor:
When calling web3.eth.getTransactionReceipt() for a transaction which was mined with at least one other transaction in the same block, the logs returned include the logs of every other transaction in that block. As you can imagine reproducing the bug isn't trivial, but if you'd like i'll provide a detailed way to reproduce this issue. The fix I added in receiptserializer.js filters the logs in the block the transaction was included in to only return logs that match the original transaction hash.

### DIFF
--- a/lib/database/receiptserializer.js
+++ b/lib/database/receiptserializer.js
@@ -32,12 +32,16 @@ ReceiptSerializer.prototype.decode = function(json, done) {
             return done(err);
           }
 
+          const logs = result.logs.reduce((acc, curr) => {
+            return curr.transactionHash === json.transactionHash ? acc.concat(curr) : acc;
+          }, []);
+
           done(
             null,
             new Receipt(
               tx,
               result.block,
-              result.logs,
+              logs,
               json.gasUsed,
               json.cumulativeGasUsed,
               json.contractAddress,
--- a/lib/database/receiptserializer.js
+++ b/lib/database/receiptserializer.js
@@ -18,7 +18,7 @@ ReceiptSerializer.prototype.decode = function(json, done) {
   // If you can figure out how to test this please feel free to add a test!
   var txHash = json.transactionHash.toLowerCase();
 
-  this.database.transactions.get(json.transactionHash, function(err, tx) {
+  this.database.transactions.get(txHash, function(err, tx) {
     if (err) {
       return done(err);
     }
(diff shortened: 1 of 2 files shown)


seesemichaelj:
@balexander4 verified this issue: https://github.com/trufflesuite/ganache-core/issues/273#issuecomment-456473186

davidmurdoch:
Thanks, @davidbancor and @seesemichaelj! We'll need to write some tests for this before merging, but the fix itself LGTM!

davidbancor:
Could it perhaps be an issue that the fix isn't normalizing the transaction hash to be lowercase, and then when calling web3.eth.getTransactionReceipt you can get different results for passing in a non lowercase transaction hash?

davidmurdoch:
@davidbancor 	It could be. We'll need to add a test for that.

davidmurdoch:
@davidbancor Opened an issue for handling mixed-case or uppercase transaction hashes, as Ganache doesn't currently support it (at least not for getting transaction receipts): https://github.com/trufflesuite/ganache-core/issues/308

I don't think it is currently possible to generate a log or transaction that records an upper-mixed-case transaction hash in the context of the filter logging code this PR addresses, however I do think adding the `toLowerCase` check could prevent possible issues down the line so I'll go ahead and make the change. Thanks again!

lib/database/receiptserializer.js
```diff
@@ -11,8 +11,14 @@ ReceiptSerializer.prototype.encode = function(receipt, done) {
...
 ReceiptSerializer.prototype.decode = function(json, done) {
   var self = this;
-
-  this.database.transactions.get(json.transactionHash, function(err, tx) {
+  // Make sure we can handle mixed/upper-case transaction hashes
+  // it doesn't seem possible to record a transaction hash that isn't
+  // already lower case, as that's the way ganache generates them, however
+  // I don't think it will hurt anything to normalize here anyway.
+  // If you can figure out how to test this please feel free to add a test!
+  var txHash = json.transactionHash.toLowerCase();
```
nicholasjpaterno: Is it worth the effort to preserve the case?  We talked briefly about potentially unintended consequences when forking.

davidmurdoch: I changes the `get` part back to the way it was before.

PR #269, opened 2019-01-04
Overload eventemitter on method to handle data listeners

### LINKED ISSUE #267: MaxListenersExceededWarning encountered when running with Web3
For more detailed history: https://github.com/ethereum/web3.js/issues/1648

When running Ganache provider (for my tests) and then subsequently interacting them with Web3.js, I get warnings like the following:

```
(node:38062) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 data listeners added. Use emitter.setMaxListeners() to increase limit
```

According to this specific [comment](https://github.com/ethereum/web3.js/issues/1648#issuecomment-431047839), the issue seems to be due to the Ganache Provider.

Interestingly, @cgewecke suppressed these warnings in Truffle in this PR: https://github.com/trufflesuite/truffle/pull/1088

Not exactly sure if this belongs here, but I believe it is worth the investigation.

seesemichaelj:
This essentially bypasses the memoryleak warning for data event listeners by increasing the max listeners as needed.

This fixes #267, but I'm not sure if this approach is even a sane one. A better implementation is fixing the issue at the core in web3.js (see #267 for more details).

### DIFF
--- a/lib/provider.js
+++ b/lib/provider.js
@@ -62,6 +62,7 @@ const defaultOptions = {
 };
 
 Provider.prototype = Object.create(EventEmitter.prototype);
+Provider.prototype.oldOn = Provider.prototype.on;
 Provider.prototype.constructor = Provider;
 
 Provider.prototype._applyDefaultOptions = function(options) {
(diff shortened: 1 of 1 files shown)


davidmurdoch:
Eh. Just going to drop web3 soon.

PR #274, opened 2019-01-11
Add Constantinople support

davidmurdoch:


lib/blockchain_double.js
```diff
@@ -503,30 +514,31 @@ BlockchainDouble.prototype.processCall = function(tx, blockNumber, callback) {
...
-        skipBalance: true,
-        skipNonce: true
-      };
-
-      self.vm.runTx(runArgs, function(vmerr, result) {
-        // This is a check that has been in there for awhile. I'm unsure if it's required, but it can't hurt.
-        if (vmerr && vmerr instanceof Error === false) {
-          vmerr = new Error("VM error: " + vmerr);
-        }
+      self.vm.stateManager._cache.clear();
```
seesemichaelj: hmmmm why is this necessary?

lib/utils/forkedblockchain.js
```diff
@@ -319,40 +318,6 @@ ForkedBlockchain.prototype.getCode = function(address, number, callback) {
   });
 };
 
-ForkedBlockchain.prototype.putCode = function(address, value, callback) {
```
seesemichaelj: why was this removed?

lib/blockchain_double.js
davidmurdoch: REASONS, SEESE. REASONS!

When we update the stateRoot to the stateRoot of a different block the data in the cache isn't updated with it. If, for instance, an account is in the cache from a previous call to access the account from the `latest` block, but we are at this time in the code operating in the context of an older block where the account may have different values, `getAccount` will return the wrong values, despite the `stateRoot` being updated to the correct block.

I do think we need to rework our implementation of the `vm` and its `stateManager` in the near future to avoid these kinds of shenanigans.

lib/blockchain_double.js
```diff
@@ -327,7 +337,9 @@ BlockchainDouble.prototype.putAccount = function(account, address, callback) {
       return callback(err);
     }
 
-    self.vm.stateManager.cache.flush(callback);
+    // stateManager.cache.flush is no longer public, but `getStateRoot`
+    // does the same thing now.
+    self.vm.stateManager.getStateRoot(callback);
```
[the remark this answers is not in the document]
davidmurdoch: I've filed an issue requesting a public method here: https://github.com/ethereumjs/ethereumjs-vm/issues/418

But I agree with you that using `_cache.flush` is better and will update to use it.

lib/blockchain_double.js
```diff
@@ -162,7 +167,12 @@ BlockchainDouble.prototype.initialize = function(accounts, callback) {
 };
 
 BlockchainDouble.prototype.createStateTrie = function(db, root) {
-  return new Trie(this.data.trie_db, root);
+  const trie = new Trie(db, root);
+  // activate precompiles:
```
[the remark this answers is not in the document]
davidmurdoch: They'll release when their ready. I don't mind shipping with this patch as it is.

PR #278, opened 2019-01-17
Make sure call uses an ephemeral vm and that evm_revert clears the cache

davidmurdoch:


lib/blockchain_double.js
```diff
@@ -399,7 +411,7 @@ BlockchainDouble.prototype.getQueuedNonce = function(address, callback) {
   if (nonce != null) {
     return callback(null, Buffer.from([nonce + 1]));
   }
-  this.vm.stateManager.getAccount(addressBuffer, function(err, val) {
+  this.stateTrie.get(addressBuffer, function(err, val) {
```
davidmurdoch: I had recently changed this line to `this.vm.stateManager.getAccount` but that is actually wrong as we _shouldn't_ be reading from the cache here, as the cache could be dirty (I don't think i can be dirty now, but in the future when we enable `asyncRequestProcessing` I think it could be). We should actually always be going to the trie itself when reading out data like this.

lib/blockchain_double.js
```diff
@@ -466,22 +478,10 @@ BlockchainDouble.prototype.sortByPriceAndNonce = function() {
...
-      if (startingStateRoot) {
-        stateManager.setStateRoot(startingStateRoot, () => {
-          callback(err || e, result);
-        });
-      } else {
-        callback(err || e, result);
-      }
-    });
-  };
+  var vm = this.createVMFromStateTrie(this.stateTrie.copy());
```
davidmurdoch: `processCall` now uses an ephemeral copy of the vm based on the current state trie (it ignores the cache since things in the cache aren't guaranteed to be written to the trie, especially in an `asyncRequestProcessing` context).

lib/blockchain_double.js
```diff
@@ -534,13 +527,8 @@ BlockchainDouble.prototype.processCall = function(tx, blockNumber, callback) {
       if (err) {
         return callback(err);
       }
-      stateManager.getStateRoot((err, startingStateRoot) => {
```
davidmurdoch: we don't need to get the current state root (which we were only using so we could put it back again later) because we are just going to throw this whole VM away now.

lib/utils/forkedblockchain.js
```diff
@@ -41,6 +41,12 @@ function ForkedBlockchain(options) {
 
   BlockchainDouble.call(this, options);
 
+  this.createVMFromStateTrie = function() {
+    var vm = BlockchainDouble.prototype.createVMFromStateTrie.apply(this, arguments);
+    this.patchVM(vm);
```
davidmurdoch: Forking requires that the vm be "special" otherwise things get all broken on us. When forking calls `processCall` over in `BlockChainDouble` (or traces a transaction, which we actually don't test for) it needs to have some special monkey patching set up deep in the VM's stateManager in order to continue working.

area:
Now this is merged, it is still necessary to queue `eth_call` requests, or can they be [serviced immediately rather than queued](https://github.com/area/ganache-core/commit/b2e11beac7447a43152e28298e15ba6f66e18f10)? Our dapp makes a lot of `eth_call` requests to read state from chain, and servicing them one-at-a-time slows performance to a crawl (whereas if using parity / geth etc, they're serviced in parallel). This has worked in my limited testing so far without issue, but I'm wary of what's lurking beneath the surface here in terms of subtleties...

davidmurdoch:
There is an ongoing refactor/rewrite into typescript (see the `ts` branch for a WIP -- but don't try using it -- it doesn't actually work yet) that does/will run `eth_call` (and others) in parallel. If I can find some time to work on this rewrite a bit more we should see major performance improvements in ganache-core, cli, and UI, as well as much better feature and behavior parity with Geth and Parity clients Soon™.

PR #281, opened 2019-01-19
Add personal_sign & personal_ecRecover

ellis2323:
Hi,

I have added these two standard methods:

 * personal_sign
 * personal_ecRecover

and i have added a unit test which check the personal_sign and personal_ecRecover works well together.

As the EIP-712 is not yet widely adopted, many use these two methods.

Best regards,


lib/statemanager.js
```diff
@@ -449,6 +449,20 @@ StateManager.prototype.sign = function(address, dataToSign) {
...
+
+  if (!account) {
+    throw new Error("cannot sign data; no private key");
+  }
+
+  return sigUtil.personalSign(account.secretKey, { data: dataToSign });
+};
+
+StateManager.prototype.personal_ecRecover = function(dataThatWasSigned, signature) {
+  return sigUtil.recoverPersonalSignature({ data: dataThatWasSigned, sig: signature });
```
princesinha19: I think it's not good to use the abbreviation. I would rename **sig** to **signature**.

lib/subproviders/geth_api_double.js
```diff
@@ -511,6 +511,32 @@ GethApiDouble.prototype.personal_sendTransaction = function(txData, password, ca
...
+  return callback(error, result);
+};
+
+GethApiDouble.prototype.personal_ecRecover = function(dataThatWasSigned, signature, callback) {
+  var result;
+  var error;
+
+  try {
+    result = this.state.personal_ecRecover(dataThatWasSigned, signature);
+  } catch (e) {
```
princesinha19: same here. I would replace **e** with **error**.

lib/subproviders/geth_api_double.js
```diff
@@ -511,6 +511,32 @@ GethApiDouble.prototype.personal_sendTransaction = function(txData, password, ca
   });
 };
 
+GethApiDouble.prototype.personal_sign = function(dataToSign, address, password, callback) {
+  var result;
+  var error;
+
+  try {
+    result = this.state.personal_sign(address, dataToSign);
+  } catch (e) {
```
princesinha19: same here. I can remove abbreviation.

lib/statemanager.js
ellis2323: ok

davidmurdoch:
Closing this PR since this is based off of the old v6 architecture. We're tracking the implementation in v7 in https://github.com/trufflesuite/ganache/issues/995

PR #282, opened 2019-01-20
Access-Control-Allow-Credentials and Access-Control-Allow-Origin

dgendill:
I was working through [this tutorial](https://medium.com/coinmonks/interacting-with-ethereum-smart-contracts-through-web3-js-e0efad17977) and serving my app on a local server. When I tried to call my contract I got the Access-Control-Allow-Origin error in the console. It looks like that issue was solved with [this recent PR](https://github.com/trufflesuite/ganache-core/pull/226).

However, I think web3.js sets [Request.credentials](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) to 'include', which means we also need Access-Control-Allow-Credentials set true and replace the wildcard in Access-Control-Allow-Origin to a specific domain (Request.include can't be used with a wildcard](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials). This was the error I was seeing...

image

So I think setting Access-Control-Allow-Origin to request.headers.origin and setting Access-Control-Allow-Credentials might be helpful.

### DIFF
--- a/lib/httpServer.js
+++ b/lib/httpServer.js
@@ -15,21 +15,62 @@ module.exports = function(provider, logger) {
         // At this point, we have the headers, method, url and body, and can now
         // do whatever we need to in order to respond to this request.
 
-        var headers = {
-          "Access-Control-Allow-Origin": request.headers.origin,
-          "Access-Control-Allow-Credentials": "true"
-        };
+        var headers = {};
+
+        // https://fetch.spec.whatwg.org/#http-requests
+        let isCORSRequest = request.headers.hasOwnProperty("origin");
+        let isCORSPreflight = isCORSRequest &&
+                              method === "OPTIONS" &&
+                              request.headers.hasOwnProperty("access-control-request-headers") &&
+                              request.headers.hasOwnProperty("access-control-request-method");
+
+        if (isCORSRequest) {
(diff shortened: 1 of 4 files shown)


davidmurdoch:
Thanks for the PR, @dgendill! Looks like there is a [Related issue at web3](https://github.com/ethereum/web3.js/issues/1802), too.

I'll need to add some tests for this and look into if we can conditionally set the `Access-Control-Allow-Credentials` header only when the client is asking for it (same with changing to `*` vs `request.headers.origin`). Feel free to add tests and conditional logic as we've got some other high-priority issues we currently working on.

Additionally, it is possible that `request.headers.origin` is undefined, so this case will need to be properly handled as well.

Thanks again!

davidmurdoch:
Really great work @dgendill! We'll take a closer look at this and get it merged in soon!

davidmurdoch:
Merged in https://github.com/trufflesuite/ganache-core/commit/9633943c9944c5004bde90e30674834b963ecefd! Thanks again @dgendill for this excellent PR!

dgendill:
Thanks, David!

davidmurdoch:
@dgendill we at Truffle were all seriously impressed by your thoroughness in this PR. Thanks again!

PR #286, opened 2019-01-25
Refactored gas tests

ccowell:
- Moved all gas tests into a subfolder
- Extracted gas specific helper functions into a lib subfolder


ccowell:
Missing too much history. Will create a new PR with better history preservation.

PR #289, opened 2019-01-28
Refactored gas tests

ccowell:
Added missing documentation to bootstrap module
Refactored and grouped gas related tests
Fixed reference errors to old helper files
Fixed a couple of module export definitions


test/call.js
```diff
@@ -19,7 +19,7 @@ describe("eth_call", function() {
   var estimateGasContractAbi;
   var EstimateGasContract;
   var estimateGasInstance;
-  var source = fs.readFileSync(path.join(__dirname, "EstimateGas.sol"), "utf8");
+  var source = fs.readFileSync(path.join(__dirname, "/contracts/gas/EstimateGas.sol"), "utf8");
```
benjamincburns: shouldn't have slashes in the `path.join` call, as `path.join` is there to figure out the path separators for you

should be: `path.join(__dirname, "contracts", "gas", "EstimateGas.sol")`

test/call/undefined.js
```diff
@@ -1,11 +1,11 @@
 const assert = require("assert");
-const bootstrap = require("../helpers/bootstrap");
+const bootstrap = require("../helpers/contract/bootstrap");
 
 describe("Undefined", () => {
   describe("Calls", () => {
     const mainContract = "Call";
     const contractFilenames = [];
-    const contractPath = "../contracts/call/";
+    const contractPath = "../../contracts/call/";
```
benjamincburns: separation of concerns: if we're going to use a directory structure pattern for storing contracts everywhere along with `bootstrap` & co, should tests need to be concerned with path resolution issues, or should it be the job of bootstrap, etc to map out the storage/path pattern for the test?

test/gas/custom_gas_price.js
```diff
@@ -0,0 +1,22 @@
+const assert = require("assert");
+const getWeb3 = require("../helpers/web3/getWeb3");
+
+describe("Custom Gas Price", () => {
+  it("should return gas price of 15 when specified as a decimal", async() => {
```
benjamincburns: `it("should return gas price of 15 when specified as a decimal")` -- this doesn't make any sense to me - whenever I specify gas price as a decimal it should come back as 15 regardless of what the value is?

test/gas/custom_gas_price.js
```diff
@@ -0,0 +1,22 @@
...
+      gasPrice: 15
+    };
+    const { web3 } = await getWeb3(options);
+    const result = await web3.eth.getGasPrice();
+    assert.strictEqual(parseInt(result), 15);
+  });
+
+  it("should return gas price of 15 when specified as hex", async() => {
+    const options = {
+      gasPrice: 0xf
```
benjamincburns: I would have expected this to be a hex string rather than a hex literal. If you're specifying a hex literal this test is logically equivalent to the previous one

test/gas/gasLimit.js
```diff
@@ -0,0 +1,16 @@
+const assert = require("assert");
+const to = require("../../lib/utils/to.js");
+const preloadWeb3 = require("../helpers/web3/preloadWeb3");
+
+describe("options:gasLimit", () => {
+  const mnemonic = "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
```
benjamincburns: Probably not necessary for this review, but we should move away from mnemonics being specified in tests.

benjamincburns: That is, unless a test requires a very specific mnemonic (instead of a mnemonic that doesn't change from run to run) the mnemonic in question should be defined in one of the test helpers - `getWeb3` seems like a good choice from what I've seen of it so far

benjamincburns:
So many of these tests are validating that an option was properly handled/observed. Given that this is mostly following a pattern of "set option to nondefault value then validate that the option was set", I wonder if this couldn't/shouldn't be handled by a test helper. That way we could have a single "it observes gas options" test which just runs through a list of options

test/gas/lib/customContractDeploy.js
```diff
@@ -0,0 +1,6 @@
+const deployContract = (abi, accounts, bytecode, web3) => {
```
benjamincburns: where is this used? why not export it from the base `compileAndDeploy` helper?

test/gas/lib/isGasExpenseCorrect.js
```diff
@@ -0,0 +1,42 @@
+const assert = require("assert");
+const to = require("../../../lib/utils/to.js");
+
+const isGasExpenseCorrect = async(expectedGasPrice, setGasPriceOnTransaction = false, web3, accounts) => {
+  const transferAmount = web3.utils.toBN(web3.utils.toWei("5", "finney"));
```
benjamincburns: why hard code the transfer amount?

test/gas/lib/isGasExpenseCorrect.js
```diff
@@ -0,0 +1,42 @@
...
+    from: accounts[0],
+    to: accounts[1],
+    value: transferAmount
+  };
+
+  if (setGasPriceOnTransaction) {
+    params.gasPrice = expectedGasPriceBN;
+  }
+
+  const receipt = await web3.eth.sendTransaction(params);
```
benjamincburns: The name of the function (the "is" prefix, specifically) makes me think it's "pure" - meaning it has no side effects, but we're changing state on this line

I think we're asking too much of this function, however. I'd rather see it get broken down into smaller chunks instead of having its name changed

benjamincburns: Also re: encapsulation - based on the name I'd expect this function to contain the logic that validates a given transaction's gas usage is correct, but we depend on the caller to do that.

That said, I don't think we should be rewriting gas calculation logic into our test suite. I take this as yet more indication that we're asking too much of this function

test/gas/lib/isGasExpenseCorrect.js
benjamincburns: Also it seems a bit weird that we get our BN dependency via web3 - we should just `const BN = require("bn.js")` and `const transferAmount = new BN(...);`

test/gas/lib/isGasExpenseCorrect.js
```diff
@@ -0,0 +1,42 @@
...
+  const actualGasPrice = gasExpense.div(gasUsed);
+
+  assert(
+    expectedGasPriceBN.eq(actualGasPrice),
+    `Gas price used by EVM (${to.hex(actualGasPrice)}) was different from` +
+      ` expected gas price (${to.hex(expectedGasPriceBN)})`
+  );
+};
+
+module.exports = isGasExpenseCorrect;
```
benjamincburns: where is this module used? ctrl-f is failing me here

benjamincburns: overall it feels like this module is a test rather than a helper

test/gas/lib/transactionEstimate.js
```diff
@@ -0,0 +1,14 @@
+const assert = require("assert");
+
+const testTransactionEstimate = async(contractFn, args, options, instance) => {
```
benjamincburns: Where is it used?

test/helpers/contract/bootstrap.js
```diff
@@ -1,12 +1,18 @@
...
-
 const { join } = require("path");
 const { compileAndDeploy } = require("./compileAndDeploy");
 
-const setUp = (mainContractName = "", subContractNames = [], options = {}, contractPath = "../contracts/") => {
+/**
+ * @param {string} mainContractName Main contract filename (withouth file extension)
+ * @param {string|Array} subContractNames Array of supporting contract filenames (without file extension)
+ * @param {Object} options Provider options
+ * @param {string} contractPath Relative path to contract
```
benjamincburns: Relative to what? This module, or the caller? Will absolute work?

test/helpers/contract/bootstrap.js
```diff
@@ -1,12 +1,18 @@
 const Ganache = require(process.env.TEST_BUILD
   ? "../build/ganache.core." + process.env.TEST_BUILD + ".js"
-  : "../../index.js");
+  : "../../../index.js");
```
benjamincburns: The above require should be wrapped up in a helper somewhere so we don't need to remember to copypasta it in every test file

test/gas/lib/customContractDeploy.js
benjamincburns: looks like this duplicates the `deploy` function that's already in `compileAndDeploy`, actually

test/helpers/web3/getWeb3.js
```diff
@@ -0,0 +1,21 @@
...
+  ? "../build/ganache.core." + process.env.TEST_BUILD + ".js"
+  : "../../../index.js");
+const Web3 = require("web3");
+
+const getWeb3 = async(options = {}) => {
+  const provider = Ganache.provider(options);
+  const web3 = new Web3(provider);
+  const accounts = await web3.eth.getAccounts();
+
+  return Object.assign(
```
benjamincburns: Why `Object.assign` here?

test/helpers/web3/getWeb3.js
```diff
@@ -0,0 +1,21 @@
+const Ganache = require(process.env.TEST_BUILD
+  ? "../build/ganache.core." + process.env.TEST_BUILD + ".js"
+  : "../../../index.js");
+const Web3 = require("web3");
+
+const getWeb3 = async(options = {}) => {
```
benjamincburns: This doesn't return a web3 instance - it returns an object which contains a web3 instance. perhaps it should be `initializeTestProvider` or something like that?

Also bear in mind that we'll need the ability to initialize tests which operate using both remote and local providers.

test/gas/gasLimit.js
```diff
@@ -0,0 +1,16 @@
...
+  const options = { mnemonic };
+  const services = preloadWeb3(options);
```
benjamincburns: why use this and not `getWeb3` - what's the difference?

test/helpers/web3/preloadWeb3.js
```diff
@@ -0,0 +1,24 @@
+const Ganache = require(process.env.TEST_BUILD
+  ? "../build/ganache.core." + process.env.TEST_BUILD + ".js"
+  : "../../../index.js");
+const Web3 = require("web3");
+
+const preloadWeb3 = (options = {}) => {
```
benjamincburns: As I said elsewhere, I don't understand the functional difference between this module and `getWeb3` - seems like a pretty direct duplication of code

test/transaction_rejection.js
```diff
@@ -25,7 +25,7 @@ describe("Transaction rejection", function() {
   var estimateGasContractAbi;
   var EstimateGasContract;
   var estimateGasContractAddress;
-  var source = fs.readFileSync(path.join(__dirname, "EstimateGas.sol"), "utf8");
+  var source = fs.readFileSync(path.join(__dirname, "/contracts/gas/EstimateGas.sol"), "utf8");
```
benjamincburns: same comment as above re: `path.join` & path separators

test/gas/gasLimit.js
ccowell: Another options would be to store mnemonics in a .env implementation.

test/gas/gas.js
```diff
@@ -0,0 +1,563 @@
...
+    promiEvent.on("receipt", (receipt) => {
+      deploymentReceipt = receipt;
+    });
+  });
+
+  describe("Refunds", () => {
+    it(
+      "accounts for Rsclear Refund in gasEstimate when a dirty storage slot is reset and it's original " +
+        " value is 0",
+      async() => {
```
benjamincburns: we shouldn't be using arrow functions to encapsulate tests

https://remarkablemark.org/blog/2016/08/20/arrow-functions-in-mocha/

test/gas/lib/toBytes.js
```diff
@@ -0,0 +1,11 @@
+const to = require("../../../lib/utils/to.js");
+
+const toBytes = (message) => {
```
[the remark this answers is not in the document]
benjamincburns: Now that I see where this is used - I think until it's clear that another test needs this, it should be kept as an unexported function alongside the two tests that use it (and it should be called something like `toBytesHexString`)

test/gas/lib/customContractDeploy.js
benjamincburns: Found where it's used - but this still looks like unnecessary duplication

test/gas/gas.js
```diff
@@ -0,0 +1,563 @@
+const assert = require("assert");
```
benjamincburns: Why are some tests split into their own files, while the bulk of tests remain in this one large file?

test/gas/gas.js
```diff
@@ -0,0 +1,563 @@
...
+        await instance.methods.initialSettingOfX().send(options);
+
+        // update storage and then reset it to 0
+        const method = instance.methods.triggerRsclearRefund();
+
+        const gasEstimate = await method.estimateGas(options);
+
+        const { gasUsed, cumulativeGasUsed } = await method.send({ from, gas: gasEstimate });
+
+        if (provider.options.hardfork === "byzantium") {
```
benjamincburns: do we test both fork types?

test/gas/gasLimit.js
ccowell: preloadWeb3 sets up a provider within a mocha `before` hook

test/gas/gas.js
ccowell: The test counts were preserved for each of the test files.

test/call.js
ccowell: Can change this but will be fixed in a future PR.

test/call/undefined.js
ccowell: Bootstrap defaults to `../../contracts` but can be updated to concatenate `call`, per this example.

test/gas/gas.js
davidmurdoch: @ben, as far as arrow functions in tests, the _only_ reason not to is if you really _want_ to use `this` inside of the test to reference the test instance itself. There are no instances where we actually need to do that within an `it`. Within `before` we absolutely _can't_ use arrow functions.

I like the arrow functions here, personally. :-)

test/gas/custom_gas_price.js
ccowell: You are correct and both tests in this file backs the assumptions. Whether you specify `15` or `0xf` for `gasPrice`, `getGasPrice()` successfully returns the decimal value of `15`.

test/gas/custom_gas_price.js
ccowell: `gasPrice` is specified as a hex on line 16. I can update the description to be clearer.

test/gas/gas.js
```diff
@@ -0,0 +1,563 @@
...
+  const options = { mnemonic };
+  const services = bootstrap(mainContract, contractFilenames, options, contractPath);
+
+  let deploymentReceipt;
+
+  before("Adding funds to the contract", async function() {
+    this.timeout(10000);
+    const { accounts, bytecode, contract } = services;
+
+    const promiEvent = contract.deploy({ data: bytecode }).send({
```
benjamincburns: why do we call `contract.deploy` here? shouldn't we improve `bootstrap`, `compileAndDeploy`, and whatever else to make it possible to transfer eth on the deploy?

test/gas/gas.js
```diff
@@ -0,0 +1,563 @@
...
+
+        // prime storage by making sure it is set to 0
+        await instance.methods.reset().send(options);
+
+        // update storage and then reset it back to 0
+        const method = instance.methods.triggerRsclearRefund();
+
+        const gasEstimate = await method.estimateGas(options);
+
+        const { gasUsed, cumulativeGasUsed } = await method.send({ from, gas: gasEstimate });
```
benjamincburns: No need to change this now, but for future refactoring let's avoid switching to destructuring syntax here (we used to keep a reference to the entire receipt before).

On the one hand destructuring is nice because there's less to type later on down the line, but on the other it squashes  everything into the local namespace/scope. I think keeping things namespaced (e.g. `receipt.gasUsed` instead of `gasUsed`) makes it easier to understand the source of the data in question, and makes it less likely to have naming collisions as logic is added over time.

benjamincburns: That said, I also wouldn't change things that are already in a destructured style to go back to the object reference style. There isn't enough benefit either way to make it worth your time

benjamincburns:
As I read through `gas.js` I'm seeing a bunch of different refactoring steps. E.g. "switched to destructured syntax," "switched variable definitions from `var` or `let` to `const` where applicable"

It would be easier to review and to understand the rationale behind some changes if the commits reflected this. That way we can review commit-by-commit on the merits of specific changes rather than wholesale like this. Doing this should also make this work more cherry-pickable

test/gas/gas.js
```diff
@@ -0,0 +1,563 @@
...
+        assert.strictEqual(gasUsed, transactionCostMinusRefund);
+      } else if (provider.options.hardfork === "constantinople") {
+        // since storage was initially primed to 0 and we call triggerAllRefunds(), which then
+        // resets storage back to 0, 19800 gas is added to the refund counter per Constantinople EIP 1283
+        transactionCostMinusRefund =
+          gasEstimate - RSELFDESTRUCT_REFUND - RSCLEAR_REFUND_FOR_RESETTING_DIRTY_SLOT_TO_ZERO;
+        assert.strictEqual(gasUsed, transactionCostMinusRefund);
+      }
+
+      const [tx1, tx2, tx3] = await Promise.all(hashes.map((hash) => web3.eth.getTransactionReceipt(hash)));
```
benjamincburns: Two things here:

This used to be an array called `receipts` - why change to `tx`? Often we use `tx` to mean "parameters for a transaction we're about to run" - sometimes it also means "the hash of the transaction we just ran" or "the result of `eth_getTransactionBy*` - let's avoid overloading the term further.

This is definitely a case where I would've preferred not to destructure. Any time you have a bunch of variables named the same thing with but differentiated by a numeric suffix, you should have an array. Keeping them as an array makes it easier to do bulk FP-style operations.

test/gas/gas.js
```diff
@@ -0,0 +1,563 @@
...
+process.removeAllListeners("uncaughtException");
+
+describe("Gas", () => {
+  const mainContract = "EstimateGas";
+  const contractFilenames = [];
+  const contractPath = "../../contracts/gas/";
+
+  const mnemonic = "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
+  const options = { mnemonic };
+  const services = bootstrap(mainContract, contractFilenames, options, contractPath);
```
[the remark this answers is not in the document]
benjamincburns: The name `services` means something entirely different to me than what this object is. Being a plural noun, I'd expect it to be an Array. As what I think of when I read `service` here, I either think it's a reference to the client API of some external network-accessible service, or I think it's a reference to our architecture's service layer. If you're not familiar with that, see [this quora post](https://www.quora.com/In-software-architecture-what-is-a-service-layer/answer/Scott-Nash).

Either way, I think `testContext` or similar is a better name.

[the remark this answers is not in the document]
benjamincburns: Also FWIW, TypeScript (with `noImplicitAny` enabled) kind of slaps you over the head by forcing you to think about and define the semantics of structures like this. Part of the reason why a lot of people (myself included) think duck typed languages are bad for big projects is that it's too easy to define structures like this without considering their semantics. This makes it tough to reason about how people will use them in the future, which in turn makes it tough to write maintainable code.

test/gas/lib/isGasExpenseCorrect.js
ccowell: Good alternative. Never saw this as a big issue consider web3 already exists.

test/gas/custom_gas_price.js
davidmurdoch: @ccowell @ben is just saying that `0xf === 16` whereas `"0xf" !== 16` (strictly). One version tests if ganache can understand a _`Number`_ and the other a hex _`string`_

test/gas/lib/isGasExpenseCorrect.js
ccowell: The hardcoded transfer amount seems good enough for validating gas used. Not seeing how using a different transfer amount will affect gas price validation.

test/gas/lib/isGasExpenseCorrect.js
ccowell: Can change the name of the function and break up the function.

test/gas/lib/isGasExpenseCorrect.js
ccowell: This used to be an embedded function used by 3 tests.

test/gas/lib/toBytes.js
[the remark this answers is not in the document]
ccowell: This function is also referenced in call.js.

test/gas/lib/transactionEstimate.js
ccowell: It is used in gas.js.

test/helpers/contract/bootstrap.js
ccowell: It's relative to the module. It's defaults to the contract directory where all contracts files will eventually live.

test/helpers/contract/bootstrap.js
ccowell: I anticipate the refactoring will reduce this declaration to a few files. This can also go into an `.env` when the refactoring get to that stage.

test/helpers/web3/getWeb3.js
ccowell: Can change to {}.

test/helpers/web3/getWeb3.js
ccowell: Good change request.

test/helpers/web3/preloadWeb3.js
ccowell: preloadWeb3 sets up a provider within a mocha `before` hook

test/transaction_rejection.js
ccowell: Can change for now but will be removed in a future PR

test/gas/gas.js
ccowell: Currently only testing `constantinople`.

test/gas/gas.js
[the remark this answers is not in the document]
ccowell: > I think any test setup like this (especially setup which performs I/O or has other types of side effects) should be in a `before` block

As far I can tell, `bootstrap` is a `before` block.

> The name `services` means something entirely different to me than what this object is. Being a plural noun, I'd expect it to be an Array. As what I think of when I read `service` here, I either think it's a reference to the client API of some external network-accessible service, or I think it's a reference to our architecture's service layer. If you're not familiar with that, see [this quora post](https://www.quora.com/In-software-architecture-what-is-a-service-layer/answer/Scott-Nash).
> 
> Either way, I think `testContext` or similar is a better name.

Can change back to `context`.

test/gas/custom_gas_price.js
benjamincburns: What @davidmurdoch said, but my original comment was that today both tests appear to check if ganache can understand a `Number` and no tests check if ganache can understand a hex-encoded `string`

test/gas/gasLimit.js
benjamincburns: The issue I'm raising here isn't that mnemonics should be kept secret, but rather that there's no need for tests which need an unchanging mnemonic to define their own. It's about encapsulation and code duplication rather than security. These are test phrases published in the open - we should expect that every one of them is being scraped.

benjamincburns: further, adding a requirement for a .env file that isn't checked into source control would increase the burden of running tests, meaning less people would be likely to run tests before submitting PRs, and fewer still would write tests as part of their contributions

test/gas/lib/isGasExpenseCorrect.js
benjamincburns: BN is a separate library from web3, so it's weird that web3 provides a toBN function at all. Using a transitive dependency as a direct dependency is a code smell.

> Not seeing how using a variable transfer amount will affect gas price validation
This is moot because I don't think this helper should be changing state, but it's more to do with cohesion/improper encapsulation. In this case we're taking a test parameter that should be encoded explicitly in the test itself and making it implicit by hiding it within a test helper.

test/gas/custom_gas_price.js
ccowell: The initial round of refactoring effort focused on cleaning up code and not creating new tests. That test can be added for completeness.

test/gas/gasLimit.js
ccowell: So far it makes sense to move the mnemonics into the `getWeb3`

test/gas/gas.js
```diff
@@ -0,0 +1,563 @@
...
+
+  let deploymentReceipt;
+
+  before("Adding funds to the contract", async function() {
```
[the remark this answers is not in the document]
ccowell: Thanks for the reference.

test/gas/gas.js
ccowell: ok, will keep this in mind.

test/gas/gas.js
ccowell: Can change this back to a receipts array.

test/call.js
ccowell: Changed.

test/call/undefined.js
ccowell: Changed.

test/gas/custom_gas_price.js
ccowell: Added new tests.

test/gas/custom_gas_price.js
ccowell: Updated.

test/gas/lib/customContractDeploy.js
ccowell: Updated compileAndDeploy to support options.

ccowell: Updated `deploy` to support options.

test/gas/lib/isGasExpenseCorrect.js
ccowell: Ended up using `number-to-bn` over bn.js for also handling stringed values.

test/helpers/web3/getWeb3.js
ccowell: Changes were made to `initializeTestProvider.js`.

test/helpers/web3/getWeb3.js
ccowell: Module and filename changed to `initializeTestProvider`.

test/gas/gas.js
ccowell: Added byzantium support to this test file.

test/gas/lib/customContractDeploy.js
ccowell: This file was removed.

test/gas/lib/isGasExpenseCorrect.js
ccowell: This file was renamed.

test/gas/lib/isGasExpenseCorrect.js
ccowell: This file was renamed.

test/gas/gas.js
ccowell: We have `receipts` again.

test/helpers/contract/bootstrap.js
ccowell: Once the refactoring is complete. This requirement will only appear in a few files.

test/gas/gas.js
ccowell: Reverted back to `receipts`.

test/gas/gas.js
ccowell: `deploy` now exports the `receipt`.

test/gas/gasLimit.js
ccowell: @davidmurdoch Perhaps using a seed to generate mnemonics/accounts may be more secure and traceable.

benjamincburns: @ccowell what's the security concern here?

test/helpers/contract/bootstrap.js
```diff
@@ -0,0 +1,48 @@
...
+    const web3 = new Web3(provider);
+
+    const { abi, accounts, bytecode, contract, instance, receipt, sources } = await compileAndDeploy(
+      mainContractName,
+      subContractNames,
+      join(__dirname, "..", "..", "contracts", `${contractSubdirectory}/`),
+      web3
+    );
+
+    Object.assign(context, {
```
benjamincburns: Why Object.assign here?

test/helpers/web3/preloadTestProvider.js
```diff
@@ -0,0 +1,19 @@
+const initializeTestProvider = require("./initializeTestProvider");
+
+const preloadTestProvider = (options = {}) => {
+  let context = {};
+  before("Setting up web3", async function() {
+    this.timeout(10000);
+    const { accounts, provider, web3 } = await initializeTestProvider(options);
+
+    Object.assign(context, {
```
benjamincburns: Why object.assign here?

test/gas/gasLimit.js
benjamincburns: > preloadWeb3 sets up a provider within a mocha before hook.

If that's the only distinction, I think it's better to nix this function altogether. Better to make the `before` blocks explicit in the actual test file.

test/gas/gasLimit.js
ccowell: No huge security concerns other than having mnemonics out there. More about best practices.

test/gas/gasLimit.js
ccowell: Ok, can remove this.

test/helpers/contract/bootstrap.js
```diff
@@ -1,24 +1,31 @@
 const Ganache = require(process.env.TEST_BUILD
   ? "../build/ganache.core." + process.env.TEST_BUILD + ".js"
```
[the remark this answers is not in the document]
ccowell: This should be fixed now.

test/call.js
```diff
@@ -54,14 +54,14 @@ describe("eth_call", function() {
...
-      .add(toBytes("Tim"), toBytes("A great guy"), 5)
+      .add(toBytesHexString("Tim"), toBytesHexString("A great guy"), 5)
       .call({ from: accounts[0] })
       .then((result) => {
         assert.strictEqual(result, true);
       });
   });
 
-  function toBytes(s) {
+  function toBytesHexString(s) {
```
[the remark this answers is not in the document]
ccowell: call.js is scheduled for a future PR.

test/gas/lib/toBytesHexString.js
```diff
@@ -0,0 +1,11 @@
+const to = require("../../../lib/utils/to.js");
+
+const toBytesHexString = (message) => {
```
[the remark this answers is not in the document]
ccowell: rpcDataHexString does some extra padding and break. Would require a significant workaround.

test/helpers/contract/compileAndDeploy.js
```diff
@@ -64,9 +71,11 @@ async function deploy(abi, bytecode, web3) {
  * @param {Object} web3 Web3 interface
```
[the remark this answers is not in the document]
ccowell: Updated.

test/helpers/utils/sleep.js
```diff
@@ -0,0 +1,7 @@
+const sleep = async(milliseconds) => {
```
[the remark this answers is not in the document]
ccowell: Separated them out because they are unrelated functionality.

test/helpers/contract/bootstrap.js
```diff
@@ -0,0 +1,48 @@
...
+ * @returns {Object} abi, accounts, bytecode, contract, instance, provider, sources, web3
+ */
+const bootstrap = (mainContractName = "", subContractNames = [], options = {}, contractSubdirectory = "") => {
+  const context = {};
+
+  before("Setting up web3 and contract", async function() {
+    this.timeout(10000);
+
+    // Setup contract path
+    const provider = Ganache.provider(options);
```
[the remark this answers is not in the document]
ccowell: Able to make this work by adding existing account support to `deploy` method.

test/gas/lib/toBytesHexString.js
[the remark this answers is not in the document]
davidmurdoch: What about moving this over to utils since call.js will use it eventually as well?

test/call.js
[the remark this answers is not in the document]
davidmurdoch: Since you've already refactored this file in this PR go ahead and swap out this duplicate function with the one used by the gas tests.

test/gas/lib/toBytesHexString.js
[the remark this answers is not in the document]
davidmurdoch: What extra padding and what do you mean by "break"?

test/helpers/contract/bootstrap.js
ccowell: Used to update a variable outside the `before` hook.

ccowell: Can remove the `before` enclosure.

test/helpers/web3/preloadTestProvider.js
ccowell: File was removed in favor of keep `before` context to the test files.

test/helpers/contract/bootstrap.js
ccowell: This was removed. Planning to keep all mocha hooks localized to test files.

test/gas/gas.js
```diff
@@ -0,0 +1,567 @@
...
+    const contractSubdirectory = "gas";
+    const mnemonic = "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
+    const options = {
+      mnemonic,
+      hardfork
+    };
+    const context = bootstrap(mainContract, contractFilenames, options, contractSubdirectory);
+
+    describe("Refunds", function() {
+      it(
```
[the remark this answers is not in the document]
ccowell: https://github.com/trufflesuite/ganache-core/issues/296

test/gas/gasLimit.js
davidmurdoch: @benjamincburns 	I think the concern is that people may use our tests as examples of how to use ganache-cli/core, and we're committing our mnemonics. Obviously everyone _should_ already know that you shouldn't be doing this with mainnet mnemonics, but these sorts of security blunders still happen.

With that said, a seed is basically still a mnemonic (it's actually just used to generate one, internally), but at least wallets don't provide users with a `seed` they can copy and then paste into their code.

So... to reiterate: the concern is that someone may say to themselves "Ganache commits mnemonics to their repo, I have a mnemonic, I have a repo, I should commit it to my repo!". By using a seed in our tests instead of a mnemonic we may prevent users from making this mistake.

Alternatively, we could just leave big old `"/* KEEP YOUR MAINNET MNEMONICS SECURE AND PRIVATE OR YOU WILL LOOSE ALL YOUR FUNDS */"` comments wherever we do use a mnemonic, but a `seed` still seems easier.

ccowell: I'm leaning towards seeding atm.

ccowell: Replaced mnemonic with a randomly generated seed for refactored tests only.

test/call/undefined.js
```diff
@@ -1,19 +1,25 @@
...
 
 describe("Undefined", () => {
   describe("Calls", () => {
-    const mainContract = "Call";
-    const contractFilenames = [];
-    const contractPath = "../contracts/call/";
-    const options = {
-      vmErrorsOnRPCResponse: false
-    };
+    let context = {};
```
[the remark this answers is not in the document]
ccowell: Yup. Forgot to update this when reincorporating the `before` hook.

test/gas/gasLimit.js
```diff
@@ -0,0 +1,22 @@
...
+const initializeTestProvider = require("../helpers/web3/initializeTestProvider");
+
+describe("Gas", function() {
+  describe("options:gasLimit", function() {
+    let context = {};
+    before("Setting up web3", async function() {
+      this.timeout(10000);
+      const mnemonic = "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
+      const options = { mnemonic };
+      Object.assign(context, await initializeTestProvider(options));
```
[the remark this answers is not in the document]
ccowell: Updated to use `let context`.

test/gas/gas.js
```diff
@@ -0,0 +1,567 @@
...
+          const gasEstimate = await method.estimateGas(options);
+
+          const receipt = await method.send({ from, gas: gasEstimate });
+
+          if (provider.options.hardfork === "byzantium") {
+            assert.strictEqual(receipt.gasUsed, gasEstimate - RSCLEAR_REFUND);
+          } else if (provider.options.hardfork === "constantinople") {
+            // since storage was initially primed to 0 and we call triggerRsclearRefund(), which then
+            // resets storage back to 0, 19800 gas is added to the refund counter per Constantinople EIP 1283
+            assert.strictEqual(receipt.gasUsed, gasEstimate - RSCLEAR_REFUND_FOR_RESETTING_DIRTY_SLOT_TO_ZERO);
```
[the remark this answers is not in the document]
ccowell: So noted.

ccowell:
> So many of these tests are validating that an option was properly handled/observed. Given that this is mostly following a pattern of "set option to nondefault value then validate that the option was set", I wonder if this couldn't/shouldn't be handled by a test helper. That way we could have a single "it observes gas options" test which just runs through a list of options

Extracting the options to a helper file would be ideal and a great start to increasing coverage stats. We can target this during the next refactoring pass through.

PR #309, opened 2019-02-19
Test refactor: account.js, debug.js

ccowell:
Updated with async/awaits
- account/account.js
- debug/debug.js

Note:
- ~~helpers/utils/rpc.js (added a `generateSend` for compatibility to legacy tests).~~
- now exporting `generateSend` from initializeTestProvider as a custom `send` function


test/helpers/utils/rpc.js
```diff
@@ -5,7 +5,7 @@ const pify = require("pify");
  * @param {Object} provider Ganache provider
  * @returns {Promise} Response object
  */
-const send = (provider) => (method = "", ...params) => {
+const generateSend = (provider) => (method = "", ...params) => {
```
davidmurdoch: I understand the need for backwards compatibility while we upgrade our tests, but `generateSend` and `send` are nearly identical. Can you DRY this up a bit?

test/debug/debug.js
```diff
@@ -0,0 +1,79 @@
...
+    assert.strictEqual(
+      storage["0000000000000000000000000000000000000000000000000000000000000000"],
+      "000000000000000000000000000000000000000000000000000000000000001a"
+    );
+    assert.strictEqual(
+      storage["0000000000000000000000000000000000000000000000000000000000000001"],
+      "000000000000000000000000000000000000000000000000000000000000001f"
+    );
+
+    // await instance.methods.value().call({ from: accounts[0], gas });
```
davidmurdoch: What's this line? Is it supposed to test `assert.strictEqual(value, expectedValueBeforeTrace);`?

test/helpers/utils/rpc.js
ccowell: Sure, let me check with @nicholasjpaterno on some of his finer points.

test/debug/debug.js
ccowell: Correct.

test/helpers/utils/rpc.js
ccowell: It turns out we can export a `send` function when we `initializeTestProvider`. Keeping `generateSend`.

test/accounts.js
```diff
@@ -1,263 +1,195 @@
...
-  : "../index.js");
-var assert = require("assert");
+const assert = require("assert");
+const initializeTestProvider = require("./helpers/web3/initializeTestProvider");
 
-describe("Accounts", function() {
-  var expectedAddress = "0x604a95C9165Bc95aE016a5299dd7d400dDDBEa9A";
-  var mnemonic = "into trim cross then helmet popular suit hammer cart shrug oval student";
+describe("Accounts", async() => {
+  const expectedAddress = "0x604a95c9165bc95ae016a5299dd7d400dddbea9a";
```
davidmurdoch: I wonder if removing the checksumming here has unintended consequences.

test/accounts.js
```diff
@@ -1,263 +1,195 @@
...
-        done();
+  it("should lock all accounts when specified", async() => {
+    const options = {
+      mnemonic,
+      secure: true
+    };
+    const { accounts, web3 } = await initializeTestProvider(options);
+
+    accounts.forEach(async(account) => {
+      try {
```
davidmurdoch: Maybe we can do something like https://nodejs.org/api/assert.html#assert_assert_rejects_asyncfn_error_message instead of the try/catch (anti-)pattern?

test/accounts.js
```diff
@@ -1,263 +1,195 @@
...
-        return new Promise(function(resolve, reject) {
-          web3.eth.getBalance(accounts[0], function(err, balance) {
-            if (err) {
-              return reject(err);
-            }
+    assert.strictEqual(accounts.length, 2, "The number of accounts created should be 2");
+  });
 
-            var balanceInEther = web3.utils.fromWei(balance, "Ether");
+  it("should create a 7 accounts when ", async() => {
```
davidmurdoch: Can you update this so it says `"should create the correct number of accounts as specified by total_accounts"`  or something like it?

test/debug.js
```diff
@@ -1,143 +1,80 @@
...
+    const contractRef = {
+      contractFiles: ["DebugContract"],
+      contractSubdirectory: "debug"
+    };
 
-    return DebugContract.deploy({ data: code })
-      .send({ from: accounts[0], gas: 3141592 })
-      .then((instance) => {
-        debugContract = instance;
+    const ganacheProviderOptions = {};
```
davidmurdoch: Bootstrap `options` defaults to `{}` so let's just leave this out.

test/helpers/utils/rpc.js
```diff
@@ -1,17 +1,25 @@
-const pify = require("pify");
+const { promisify } = require("util");
```
davidmurdoch: <3

test/helpers/utils/rpc.js
```diff
@@ -1,17 +1,25 @@
...
+ * @returns {Function} Send method
  */
-const send = (provider) => (method = "", ...params) => {
-  return pify(provider.send.bind(provider))({
-    id: `${new Date().getTime()}`,
-    jsonrpc: "2.0",
-    method,
-    params: [...params]
-  });
+const send = function(provider) {
```
davidmurdoch: can you rename this to `generateSend`?

test/accounts.js
ccowell: Any objections to restoring `expectedAddress` back to case-sensitive form and removing the use of `toLowerCase()` for address comparisons?

test/accounts.js
ccowell: That can be done.

test/accounts.js
```diff
@@ -1,263 +1,195 @@
...
-      Ganache.provider({
-        mnemonic: mnemonic,
-        secure: true,
-        unlocked_accounts: [0, secondAddress]
-      })
-    );
+    if (typeof balanceInEther === "string") {
+      balanceInEther = parseFloat(balanceInEther);
+    } else {
+      balanceInEther.toNumber();
```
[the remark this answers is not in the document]
ccowell: Using a `gasPrice` of 0 looks to be the better option. Not noticing any side effects. Good idea.

test/accounts.js
```diff
@@ -21,17 +21,15 @@ describe("Accounts", async() => {
     const { accounts, web3 } = await initializeTestProvider(options);
 
     accounts.forEach(async(account) => {
-      try {
-        await web3.eth.sendTransaction({
+      assert.rejects(
```
davidmurdoch: You need to await calls to `assert.rejects`.

ccowell: Turns out `assert.rejects` was added in v10.0 and breaks when tested against `carbon`. The `try/catch` pattern is the cleanest approach that passes Travis CI across the board.

test/accounts.js
ccowell: `assert.rejects` was added to v10 and breaks when ran with `carbon` in Travis CI. `try/catch` appears to be the cleanest approach. I'm open to other suggestions.

test/accounts.js
davidmurdoch: Node docs show that it was added in V8.13.0: https://nodejs.org/docs/latest-v8.x/api/assert.html#assert_assert_rejects_block_error_message

davidmurdoch: The second arg must be one of `<RegExp> | <Function> | <Object>`. You could replace the `error.message` string comparison check with something like `/signer account should be locked/` (this regexp is the same as calling `err.message.toLowerCase().indexOf("signer account should be locked") !== -1`)

test/accounts.js
ccowell: It turns out that adding an `async` wrapper to the executing function did the trick.
```
      await assert.rejects(
        async() => {
          await web3.eth.sendTransaction({
            from: account,
            to: badAddress,
            value: web3.utils.toWei("1", "ether"),
            gasLimit: 90000
          });
        },
        /signer account is locked/,
        "should not be able to unlock the count"
      );
```

test/accounts.js
```diff
@@ -1,263 +1,178 @@
...
-  ? "../build/ganache.core." + process.env.TEST_BUILD + ".js"
-  : "../index.js");
-var assert = require("assert");
+const assert = require("assert");
+const initializeTestProvider = require("./helpers/web3/initializeTestProvider");
 
-describe("Accounts", function() {
-  var expectedAddress = "0x604a95C9165Bc95aE016a5299dd7d400dDDBEa9A";
-  var mnemonic = "into trim cross then helmet popular suit hammer cart shrug oval student";
+describe("Accounts", async() => {
```
davidmurdoch: I can't remember why, but I think you said we should be converting to arrow functions on `describe` callbacks. Does that apply here?

test/accounts.js
```diff
@@ -1,263 +1,178 @@
...
-    );
+  it("should lock all accounts when specified", async() => {
+    const options = {
+      mnemonic,
+      secure: true
+    };
+
+    const { accounts, web3 } = await initializeTestProvider(options);
+
+    accounts.forEach(async(account) => {
```
davidmurdoch: This doesn't quite work the way you'd expect.

`forEach` will not await on the return `Promise`. You can use `await Promise.all(accounts.map(` instead.

test/accounts.js
```diff
@@ -1,263 +1,178 @@
...
+      await assert.rejects(
+        async() => {
```
davidmurdoch: I don't think you should be `await`ing the inner `web3.eth.sendTransaction` Promise and you can remove the `async` keyword here.

test/accounts.js
```diff
@@ -1,263 +1,178 @@
...
+  it("should lock all accounts when specified", async() => {
+    const options = {
+      mnemonic,
+      secure: true
+    };
+
+    const { accounts, web3 } = await initializeTestProvider(options);
+
+    accounts.forEach(async(account) => {
+      await assert.rejects(
```
davidmurdoch: If you do the `await Promise.all(accounts.map(` thing above you can remove this `await` and just return the `assert.reject` `Promise` itself.

test/accounts.js
ccowell: We should definitely avoid arrow functions with hooks such as `before`, `after`, etc...  because it unbinds `this` (aka Mocha context) and prevents configuring timeouts for a block.

test/accounts.js
```diff
@@ -1,263 +1,178 @@
...
+      mnemonic,
+      secure: true,
+      unlocked_accounts: [expectedAddress]
+    };
+
+    const { accounts, web3 } = await initializeTestProvider(options);
+
+    accounts.forEach(async(account) => {
+      try {
+        await web3.eth.sendTransaction({
```
davidmurdoch: this doesn't work because of the `forEach` `await` behavior described above.

test/accounts.js
```diff
@@ -1,263 +1,213 @@
...
+    const options = {
+      mnemonic,
+      secure: true
+    };
+
+    const { accounts, web3 } = await initializeTestProvider(options);
+
+    await Promise.all(
+      accounts.map((account) => {
+        assert.rejects(
```
davidmurdoch: You need to return the `assert.rejects` `Promise` in order to `await` on it in the `Promise.all`

ccowell: It appears that the tests are functional in the latest commit.

PR #310, opened 2019-02-19
Improve websocket coverage

nicholasjpaterno:


test/subscriptions.js
```diff
@@ -7,6 +7,7 @@ const assert = require("assert");
 const PORT = 8545;
 const HOST = "127.0.0.1";
 const HTTPADDRESS = `http://${HOST}:${PORT}`;
+var Web3WsProvider = require("web3-providers-ws");
```
davidmurdoch: Since `web3-providers-ws` isn't in our package.json it shouldn't be used here (I know we already have tests that do this - shame on them!). Either included it in package.json or use `Web3.providers.WebsocketProvider` instead (which is what I prefer)

test/subscriptions.js
```diff
@@ -39,6 +40,54 @@ const testHttp = function(web3) {
   });
 };
 
+const testWebSocket = function(web3) {
+  let web3send;
+
+  before("get personal accounts", async function() {
+    // accounts = await web3.eth.getAccounts();
```
davidmurdoch: can we just delete this before section?

test/subscriptions.js
```diff
@@ -39,6 +40,54 @@ const testHttp = function(web3) {
...
+
+describe("WebSockets Server:", function() {
+  const Web3 = require("web3");
+  const web3 = new Web3();
+  let server;
+
+  before("Initialize Ganache server", async function() {
+    server = Ganache.server({
+      seed: "1337"
+      // so that the runtime errors on call test passes
```
davidmurdoch: unneeded comment?

PR #311, opened 2019-02-19
Test refactor: hex.js

ccowell:
Updated hex.js

Note:
- Moved noLeadingZeros into `security/lib/`


test/security/hex.js
```diff
@@ -1,148 +1,78 @@
...
-        id: 2
-      };
+    let method = "eth_getTransactionCount";
+    let params = [accounts[0], "pending"];
 
-      provider.sendAsync(request, function() {
-        // Ignore eth_sendTransaction result, it returns the transaction hash.
-        // A transaction hash is a 'DATA' type, which can have leading zeroes
-        // to pad it to an even string length (4 bit per char, so whole bytes).
+    let result = send(method, params, web3);
```
davidmurdoch: `send` needs to be `await`ed, right?

davidmurdoch: @ccowell 	☝️

ccowell: Good catch but we skip this test. Should we keep it?  @davidmurdoch

davidmurdoch: Yes, we should keep it.

PR #316, opened 2019-02-19
Test refactor: publicExports.js

ccowell:
Had to restore keepAliveTimeout to its original location in order. Repairing an unintended updated from a different branch.


test/publicExports.js
```diff
@@ -1,10 +1,10 @@
-var Ganache = require(process.env.TEST_BUILD
+const assert = require("assert");
```
davidmurdoch: Why'd you rename this file?

test/keepAliveTimeout.js
```diff
@@ -1,24 +1,22 @@
 const assert = require("assert");
```
davidmurdoch: Why is the file a part of this PR?

test/publicExports.js
ccowell: Thought we agreed on camelCase.

test/keepAliveTimeout.js
ccowell: Looks like a `git mv` on an incorrect branch while plowing through a bunch updates. This can be restored.

test/publicExports.js
ccowell: I opened an issue for all occurrences in test files. https://github.com/trufflesuite/ganache-core/issues/353

PR #324, opened 2019-02-20
Test refactor: unlimitedContractSize.js

ccowell:


test/unlimited_contract_size.js
```diff
@@ -1,92 +1,70 @@
...
-        })
-        .catch(function(_) {
-          done();
-        });
+    it("should fail deployment", async function() {
+      this.timeout(10000);
+      const { web3 } = context;
+      const { abi, bytecode } = contract;
+      const errorMessage = "succeeded deployment when it should have failed";
+      try {
```
davidmurdoch: Update to `await assert.rejects` pattern (unless we really can't in v8 LTS).

test/unlimited_contract_size.js
```diff
@@ -1,92 +1,70 @@
...
-        .then(function(instance) {
-          done();
-        })
-        .catch(function(_) {
-          done(new Error("failed deployment when it should have succeeded"));
-        });
+    it("should succeed deployment", async function() {
+      const { web3 } = context;
+      const { abi, bytecode } = contract;
+      await deploy(abi, bytecode, web3, { gas: 20000000 });
```
davidmurdoch: This should use `await assert.doesNotReject(deploy(abi, bytecode, web3, { gas: 20000000 }), "failed deployment when it should have succeeded");` (https://nodejs.org/docs/latest-v8.x/api/assert.html#assert_assert_doesnotreject_block_error_message) if possible.

ccowell: Can attempt pattern with v8.

PR #325, opened 2019-02-21
Test refactor: vm.js

ccowell:


test/vm.js
```diff
@@ -47,28 +47,30 @@ describe("revert opcode", function() {
...
       });
+
+      // TODO: ugly workaround - not sure why this is necessary.
+      if (!instance._requestManager.provider) {
+        instance._requestManager.setProvider(web3.eth._provider);
+      }
+      receipt = await instance.methods.alwaysReverts(5).send({ from: testContext.accounts[0] });
+    } catch (error) {
+      assert.strictEqual(error.results[error.hashes[0]].error, "revert", "Expected error result not returned.");
+      return web3.eth.getTransactionReceipt(error.hashes[0]);
```
davidmurdoch: Shouldn't return here.

test/vm.js
```diff
@@ -47,28 +47,30 @@ describe("revert opcode", function() {
...
+      const instance = await RevertContract.deploy({ data: revertCode }).send({
+        from: testContext.accounts[0],
+        gas: 3141592
       });
+
+      // TODO: ugly workaround - not sure why this is necessary.
+      if (!instance._requestManager.provider) {
+        instance._requestManager.setProvider(web3.eth._provider);
+      }
+      receipt = await instance.methods.alwaysReverts(5).send({ from: testContext.accounts[0] });
```
davidmurdoch: Update to `await assert.rejects` pattern.

ccowell: It turns out that we need to grab the transaction hash from the error object in order to grab the receipt.

test/vm.js
ccowell: Fixed.

PR #327, opened 2019-02-21
Test refactor: transactionRejection.js

ccowell:


test/transactionRejection.js
```diff
@@ -138,7 +133,7 @@ describe("Transaction rejection", function() {
     };
 
     // don't send with web3 because it'll inject its own checks
-    provider.send(request, (_, response) => {
+    provider.send(request, async(_, response) => {
```
davidmurdoch: promisify this.

test/transactionRejection.js
```diff
@@ -1,59 +1,54 @@
...
+  before("get accounts", async function() {
+    accounts = await web3.eth.getAccounts();
   });
 
   before("lock account 1", function() {
     return web3.eth.personal.lockAccount(accounts[1]);
   });
 
-  before("compile source", function() {
+  before("compile source", async function() {
```
davidmurdoch: Can this use your compileAndDeploy?

davidmurdoch (changes requested):
Updates needed.

test/transactionRejection.js
ccowell: Note: Created an issue for this https://github.com/trufflesuite/ganache-core/issues/351

test/transactionRejection.js
```diff
@@ -1,62 +1,34 @@
...
-        }
-        estimateGasContractAddress = to.hex(instance.options.address);
-      });
+  before("lock account 1", function() {
+    const { accounts, web3 } = context;
+    web3.eth.personal.lockAccount(accounts[1]);
   });
 
   it("should reject transaction if nonce is incorrect", function(done) {
+    this.timeout(10000);
```
davidmurdoch: Why add these `timeout`s here? I think if these tests take 10 seconds they're probably all broken.

ccowell: They are gone now. :)

test/transactionRejection.js
```diff
@@ -153,27 +135,17 @@ describe("Transaction rejection", function() {
...
-                )
-              );
-            }
-          })
-          .catch((error) => {
-            return done(error);
-          });
+        const result = await web3.eth.getTransactionReceipt(response.result);
+        if (to.number(result.status) === 0) {
+          assert.fail("TX rejections should return error, but returned receipt with zero status instead");
```
davidmurdoch: Since this test hasn't been `async`/`await` ed, but still executes asynchronously it needs to call `done` even in the case of an error. The test will still fail with these `assert.fail` calls, but they'll just take the entire timeout length to do so. And the error will be from the timeout, and we'll _also_ see an `UnhandledPromiseRejectionWarning` due to the `assert.fail` throwing an exception.

All that to say: either promisify this and fully convert to async/await (no mocha `done` should be used) or put the `done(new Error(...` statements back.

This also no longer handles the exceptional case where `getTransactionReceipt` throws (`.catch((error)`), like the test used to on lines 172-174.

ccowell:
Ok, will add that back to my list.

PR #328, opened 2019-02-21
Test refactor: call.js

ccowell:


test/call.js
```diff
@@ -1,59 +1,31 @@
...
-  it("should use the block gas limit if no gas limit is specified", function() {
+  it("should use the block gas limit if no gas limit is specified", async function() {
+    const { accounts, instance } = context;
+
+    const name = "0x54696d"; // Byte code for "Tim"
+    const description = "0x4120677265617420677579"; // Byte code for "A great guy"
+    const value = 5;
+    const status = await instance.methods.add(name, description, value).call({ from: accounts[0] });
+
     // this call uses more than the default transaction gas limit and will
```
davidmurdoch: This comment should be moved up to the line before the `.call` call

test/call.js
```diff
@@ -1,59 +1,31 @@
...
     this.timeout(10000);
-    var result = solc.compile({ sources: { "EstimateGas.sol": source } }, 1);
+    const contractRef = {
+      contractFiles: ["EstimateGas"],
+      contractSubdirectory: "gas"
+    };
 
-    estimateGasContractData = "0x" + result.contracts["EstimateGas.sol:EstimateGas"].bytecode;
-    estimateGasContractAbi = JSON.parse(result.contracts["EstimateGas.sol:EstimateGas"].interface);
+    const ganacheProviderOptions = {};
```
davidmurdoch: We don't need this because `{}` is the default already, right?

test/call.js
```diff
@@ -1,59 +1,30 @@
...
+    const { accounts, instance } = context;
+
+    const name = "0x54696d"; // Byte code for "Tim"
+    const description = "0x4120677265617420677579"; // Byte code for "A great guy"
+    const value = 5;
+
     // this call uses more than the default transaction gas limit and will
+    const status = await instance.methods.add(name, description, value).call({ from: accounts[0] });
+
     // therefore fail if the block gas limit isn't used for calls
```
davidmurdoch: Can you move the rest of this comment up, too?

PR #329, opened 2019-02-22
Fix contract storage caching

davidmurdoch:
Fixes https://github.com/trufflesuite/ganache-cli/issues/624


lib/blockchain_double.js
```diff
@@ -123,7 +123,7 @@ BlockchainDouble.prototype.initialize = function(accounts, callback) {
   });
 };
 
-BlockchainDouble.prototype.createVMFromStateTrie = function(state) {
+BlockchainDouble.prototype.createVMFromStateTrie = function(state, activatePrecompiles) {
```
nicholasjpaterno: Thoughts on using a default argument?  It would simplify line 149 too.
`function(state,activatePrecompiles = false)`

PR #331, opened 2019-02-23
Tries

davidmurdoch:


lib/blockchain_double.js
```diff
@@ -635,11 +635,12 @@ BlockchainDouble.prototype.processBlock = function(vm, block, commit, callback)
...
           result.createdAddress,
           receipt.status,
-          to.hex(result.bloom.bitvector)
+          to.hex(receipt.bitvector)
         );
         receipts.push(rcpt);
 
-        const rcptBuffer = Buffer.from(JSON.stringify(rcpt.toJSON()));
+        const rawReceipt = [receipt.status, receipt.gasUsed, receipt.bitvector, receipt.logs];
+        const rcptBuffer = utils.rlp.encode(rawReceipt);
```
eshaben: Is this just a preferred way to convert to buffer?

davidmurdoch: The previous code was actually wrong and this corrects it. The variable name is wrong and I'll fix it. :-D

PR #348, opened 2019-03-04
Proposal for additions to documentation

nicholasjpaterno:
Adds new headings over each section.
Adds navigation
Adds some types to the options area.
Adds links to corresponding rpc methods
Moves `miner_stop` and `miner_start` to custom area
Increases documentation of custom RPC methods with curl examples


README.md
```diff
@@ -89,62 +94,94 @@ The RPC methods currently implemented are:
...
-* `personal_sendTransaction`
-* `shh_version`
-* `rpc_modules`
-* `web3_clientVersion`
-* `web3_sha3`
-
-There’s also special non-standard methods that aren’t included within the original RPC specification:
-
-* `evm_snapshot` : Snapshot the state of the blockchain at the current block. Takes no parameters. Returns the integer id of the snapshot created. A snapshot can only be used once. After a successful `evm_revert`, the same snapshot id cannot be used again. Consider creating a new snapshot after each `evm_revert` if you need to revert to the same point multiple times.
+* <a href="https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_accounts" target="_blank">eth_accounts</a>
```
davidmurdoch: Any reason not to use markdown links for these instead of HTML anchors?

README.md
```diff
@@ -89,62 +94,94 @@ The RPC methods currently implemented are:
...
+          http://localhost:8545
+  ```
+* `evm_mine` : Force a block to be mined. Takes one optional parameter, which is the timestamp a block should setup as the mining time. Mines a block **independent** of whether mining is *started* or *stopped*.
+  ```bash
+  # Ex: new Date("2009-01-03T18:15:05+00:00").getTime()
+  curl -H "Content-Type: application/json" -X POST --data \
+          '{"id":1337,"jsonrpc":"2.0","method":"evm_mine","params":[1231006505000]}' \
+          http://localhost:8545
+  ```
+* `miner_start` : Resumes mining within ganache.
```
davidmurdoch: `miner_start` and `miner_stop` are actually specced in the [Management APIs wiki](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#list-of-management-apis), along with `debug_traceTransaction`. Maybe these can have their own section?

README.md
```diff
@@ -4,7 +4,10 @@
 [![Coverage Status](https://coveralls.io/repos/github/trufflesuite/ganache-core/badge.svg?branch=develop)](https://coveralls.io/github/trufflesuite/ganache-core?branch=develop)
 # Ganache Core
 
-This is the core code that powers the Ganache application and the the Ganache command line tool.
+[Usage](#usage) | [Options](#options) | [Implemented Methods](#implemented-methods) | [Custom Methods](#custom-methods) | [Unsupported Methods](#unsupported-methods) | [Testing](#testing)
+--- | --- | --- | --- | --- | ---
+
+This is the core code that powers the Ganache application and the Ganache command line tool.
 
 # INSTALLATION
```
davidmurdoch: While we're making things purdy, what do you think about making these subheadings 2nd level headings (by using `## Installation` instead of `# Installation`)? That and switching to Title Case instead of UPPER CASE?

README.md
```diff
@@ -4,7 +4,10 @@
 [![Coverage Status](https://coveralls.io/repos/github/trufflesuite/ganache-core/badge.svg?branch=develop)](https://coveralls.io/github/trufflesuite/ganache-core?branch=develop)
 # Ganache Core
 
-This is the core code that powers the Ganache application and the the Ganache command line tool.
+[Usage](#usage) | [Options](#options) | [Implemented Methods](#implemented-methods) | [Custom Methods](#custom-methods) | [Unsupported Methods](#unsupported-methods) | [Testing](#testing)
```
davidmurdoch: This is so pretty. I wonder if it will render OK in Docker hub? (not that it looks good in Docker right now)

README.md
nicholasjpaterno: I was hoping to leverage the target property so links would open in a new tab (and not take the user away from our docs).  It didnt seem to work unfortunately, we should definitely switch it back to the markdown markup.

README.md
nicholasjpaterno: I actually found the Management API docs monday or teusday afternoon,  but I've been waiting for the weekend to update!  Looking to circle back and dig into those docs more thoroughly as well!

README.md
nicholasjpaterno: Sounds good to me!

README.md
nicholasjpaterno: Hmm, can we test it on a beta branch (when the time comes)?  Or get a preview somehow

README.md
nicholasjpaterno: Defintely, I think they should get their own section.  I was also thinking I should put the curl results in the examples for each of these

README.md
davidmurdoch: Tested it:

image

It works! 👍

README.md
```diff
@@ -4,11 +4,11 @@
 [![Coverage Status](https://coveralls.io/repos/github/trufflesuite/ganache-core/badge.svg?branch=develop)](https://coveralls.io/github/trufflesuite/ganache-core?branch=develop)
 # Ganache Core
 
-This is the core code that powers the Ganache application and the the Ganache command line tool.
-
 [Usage](#usage) | [Options](#options) | [Implemented Methods](#implemented-methods) | [Custom Methods](#custom-methods) | [Unsupported Methods](#unsupported-methods) | [Testing](#testing)
 --- | --- | --- | --- | --- | ---
 
+This is the core code that powers the Ganache application and the the Ganache command line tool.
```
davidmurdoch: Remove the the the :-D

README.md
```diff
@@ -6,7 +6,10 @@
 
 This is the core code that powers the Ganache application and the the Ganache command line tool.
```
davidmurdoch: Mean to comment about the the the here. :-D

README.md
```diff
@@ -53,111 +56,152 @@ const server = ganache.server();
...
 * `"allowUnlimitedContractSize"`: `boolean` - Allows unlimited contract sizes while debugging. By setting this to `true`, the check within the EVM for contract size limit of 24KB (see [EIP-170](https://git.io/vxZkK)) is bypassed. Setting this to `true` **will** cause `ganache-core` to behave differently than production environments. (default: `false`; **ONLY** set to `true` during debugging).
-* `"gasPrice"`: Sets the default gas price for transactions if not otherwise specified. Must be specified as a hex string in wei. Defaults to `"0x77359400"`, or 2 gwei.
-* `"gasLimit"`: Sets the block gas limit. Must be specified as a hex string. Defaults to `"0x6691b7"`.
-* `"keepAliveTimeout"`: If using `.server()` - Sets the HTTP server's `keepAliveTimeout` in milliseconds. See the [NodeJS HTTP docs](https://nodejs.org/api/http.html#http_server_keepalivetimeout) for details. `5000` by default.
+* `"gasPrice"`: `String::hex` Sets the default gas price for transactions if not otherwise specified. Must be specified as a `hex` encoded string in `wei`. Defaults to `"0x77359400"` (2 gwei).
```
davidmurdoch: If you want to format "wei", do it for "gwei" here, too. I don't have a preference either way.

README.md
```diff
@@ -53,111 +56,152 @@ const server = ganache.server();
...
+* [net_peerCount](https://github.com/ethereum/wiki/wiki/JSON-RPC#net_peerCount)
+* [net_listening](https://github.com/ethereum/wiki/wiki/JSON-RPC#net_listening)
+* [eth_uninstallFilter](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_uninstallFilter)
+* [eth_syncing](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_syncing)
+* [web3_clientVersion](https://github.com/ethereum/wiki/wiki/JSON-RPC#web3_clientVersion)
+* [web3_sha3](https://github.com/ethereum/wiki/wiki/JSON-RPC#web3_sha3)
+* `bzz_hive`
+* `bzz_info`
+
+#### Implemented Management API Methods
```
davidmurdoch: Drop the word `Implemented`, pretty please.

README.md
nicholasjpaterno: I don't see the the issue? :laughing:

PR #349, opened 2019-03-04
Test refactor: enable_constantinople_hardfork.js

ccowell:


test/enable_constantinople_hardfork.js
```diff
@@ -1,57 +1,34 @@
...
+      };
 
-      let dummyContractInstance = await promiEvent;
+      context = await bootstrap(contractRef, ganacheProviderOptions);
+    });
+
+    it("should fail execution", async function() {
+      const { instance } = context;
 
       try {
```
davidmurdoch: Update to `await assert.rejects` pattern

PR #376, opened 2019-03-12
Fix forkedblockchain.js, correctly handle buffer encoded hash values

nicholasjpaterno:
In forkedblockchain.js, some cases (ex: debug_traceTransaction) getBlock is passed a buffer as a value of a hash, this crashes ganache as we expect the hash to be a hex encoded string.


davidmurdoch:
Closing in favor of https://github.com/trufflesuite/ganache-core/pull/396

I tried to fix too much in this PR and will be cherry picking parts from it

PR #377, opened 2019-03-12
Test refactor: bad_input.js

ccowell:


test/helpers/web3/initializeTestServer.js
```diff
@@ -0,0 +1,28 @@
+const Ganache = require(process.env.TEST_BUILD
+  ? "../../../build/ganache.core." + process.env.TEST_BUILD + ".js"
+  : "../../../index.js");
+const Web3 = require("web3");
+
+const intiailizeTestServer = (tests, port = 12345) => {
```
davidmurdoch: I don't think this file should be here. Also, `initialize` is spelled wrong. :-D

test/vm.js
```diff
@@ -1,74 +1,39 @@
...
-
-var web3 = new Web3();
-web3.setProvider(
-  Ganache.provider({
-    /* blockTime: 100, */
-    logger: logger,
-    seed: "1337"
-  })
-);
+const assert = require("assert");
```
davidmurdoch: Is this file supposed to be here?

ccowell: No. I'll restore from `develop` in the next commit.

test/helpers/web3/initializeTestServer.js
ccowell: Where do you think it should go?

Issue #339, opened 2019-02-26
Add `eth_chainId` RPC endpoint


davidmurdoch:
https://github.com/ethereumproject/go-ethereum/wiki/JSON-RPC#eth_chainid

fireblockdev:
After discussion with nividia from web3js, this method is needed to use ganache with web3js.

fireblockdev:
I did the same but he didn't fix the problem for me

projectoblio:
Thank you @tcichowicz for not being a psycho capitalist

projectoblio:
@davidmurdoch Is this issue closed now with the push today? Tcichowicz code should be merged right

lucashenning:
Any update on this? The issue in the web3.js repo has been closed (https://github.com/ethereum/web3.js/issues/2610) but I'm still getting the same error in web3 `1.0.0-beta.55` and Ganache `2.0.2-beta.0 (2.0.2-beta.0.648)`.

```
{"message":"Method eth_chainId not supported.","code":-32000,"data":{"stack":"Error: Method eth_chainId not supported.\n    at GethApiDouble.handleRequest (/Applications/Ganache.app/Contents/Resources/app.asar/node_modules/ganache-core/lib/subproviders/geth_api_double.js:67:16)\n    at next (/Applications/Ganache.app/Contents/Resources/app.asar/node_modules/ganache-core/node_modules/web3-provider-engine/index.js:116:18)
```

imbenwolf:
@davidmurdoch please review #419 so we can finally use latest web3 and ganache versions together

davidmurdoch:
I've hesitated adding this as the value it returns should be configurable. I'll go ahead and merge this into develop now regardless and follow up with a new Issue to add a chain_id configuration option to core, cli, and UI.

Thanks for keeping on top of this @imbenwolf and @lucashenning!

Aside, it may be worth it to know that web3@beta.37 is the version of web3 that will be released as stable. PR here: https://github.com/ethereum/web3.js/pull/2905

davidmurdoch:
Followup issue: https://github.com/trufflesuite/ganache-core/issues/441

Thanks again!

saurik:
Is this API supposed to be live on https://public-node.rsk.co/?

```
$ curl -s -X POST --data '{"jsonrpc":"2.0","id":"","method":"eth_chainId","params":[]}' -H 'content-type: application/json' https://public-node.rsk.co/
[...]

PR #395, opened 2019-03-25
fix attempt for issue #339

iccicci:
Hi all,

probably there are thousands of better ways to fix this issue, but at least this will allow to use web3.js version 1.0.0-beta.47 or higher.

Hope this helps,
iCC


iccicci:
I'm sorry for the confusion, I'm clearly so tired

PR #398, opened 2019-03-26
Fix forked transaction trace storage so it returns the correct data and doesn't modify the root trie

davidmurdoch:
This should fix trufflesuite/ganache-cli/issues/439 and #362.


lib/utils/forkedblockchain.js
```diff
@@ -60,21 +60,15 @@ ForkedBlockchain.prototype.initialize = function(accounts, callback) {
...
-    BlockchainDouble.prototype.initialize.call(self, accounts, function(err) {
-      if (err) {
-        return callback(err);
-      }
-
-      self.patchVM(self.vm);
-
-      callback();
-    });
+    BlockchainDouble.prototype.initialize.call(self, accounts, callback);
```
davidmurdoch: We don't need to call `patchVm` since it already gets called within `createVMFromStateTrie`.

lib/utils/forkedblockchain.js
```diff
@@ -107,7 +101,7 @@ ForkedBlockchain.prototype.createGenesisBlock = function(callback) {
 
     // Update the relevant block numbers
     self.forkBlockNumber = self.options.fork_block_number = blockNumber;
-    self.stateTrie.forkBlockNumber = blockNumber;
+    self.stateTrie.forkBlockNumber = self.stateTrie.options.forkBlockNumber = blockNumber;
```
davidmurdoch: Reason for change: The `stateTrie` is now `copy`ed instead of created via `new ForkedStorageTrie` (this aligns with how the merkle patricia trie does it internally). When copying we need to also copy the options object, so here I am making sure the forkBlockNumber is set on the object.

lib/utils/forkedblockchain.js
```diff
@@ -122,30 +116,20 @@ ForkedBlockchain.prototype.createGenesisBlock = function(callback) {
...
-};
-
-ForkedBlockchain.prototype.lookupStorageTrie = function(address, callback) {
-  address = to.hex(address);
-
-  if (this.storageTrieCache[address] != null) {
-    return callback(null, this.storageTrieCache[address]);
-  }
+ForkedBlockchain.prototype.getLookupStorageTrie = function(stateTrie, lookupAccount) {
+  lookupAccount = lookupAccount || this.getLookupAccount(stateTrie);
```
davidmurdoch: Reason for change: `getLookupStorageTrie` is usually called when the ephemeralVm is created; when that happens, we already have a `lookupAccount` fn generated, so we used that. There are other ways to call getLookupStorageTrie and those don't have (easy) access to a valid `lookupAccount` fn. So sometimes we have to create it.

lib/utils/forkedblockchain.js
```diff
@@ -122,30 +116,20 @@ ForkedBlockchain.prototype.createGenesisBlock = function(callback) {
...
+  return (address, callback) => {
+    const storageTrie = stateTrie.copy();
```
davidmurdoch: Reason for change: DRY and in-line with the way the MPT lib does it.

lib/utils/forkedblockchain.js
```diff
@@ -122,30 +116,20 @@ ForkedBlockchain.prototype.createGenesisBlock = function(callback) {
...
+    storageTrie.address = address;
```
davidmurdoch: Reason for change: ForkedStorageTrie needs the `address`.

lib/utils/forkedblockchain.js
```diff
@@ -161,7 +145,8 @@ ForkedBlockchain.prototype.isFallbackBlock = function(value, callback) {
 };
 
 ForkedBlockchain.prototype.isBlockHash = function(value) {
-  return typeof value === "string" && value.indexOf("0x") === 0 && value.length > 42;
+  const isHash = typeof value === "string" && value.indexOf("0x") === 0 && value.length > 42;
+  return isHash || (Buffer.isBuffer(value) && value.byteLength > 20);
```
davidmurdoch: Reason for change: Sometimes the value is a Buffer and we obviously should check the buffer to see if it big enough to be a blockhash.

lib/utils/forkedblockchain.js
```diff
@@ -171,6 +156,10 @@ ForkedBlockchain.prototype.isFallbackBlockHash = function(value, callback) {
     return callback(null, false);
   }
 
+  if (Buffer.isBuffer(value)) {
```
davidmurdoch: Reason for change: `blockHashes.get`, below, only allows hex strings.

lib/utils/forkedblockchain.js
```diff
@@ -234,52 +216,46 @@ ForkedBlockchain.prototype.getFallbackBlock = function(numberOrHash, cb) {
 };
 
 ForkedBlockchain.prototype.getBlock = function(number, callback) {
-  var self = this;
-
-  this.isFallbackBlockHash(number, function(err, isFallbackBlockHash) {
+  let checkFn;
```
davidmurdoch: Reason for change: simplifying some of the code, which also eliminates from duplicate checks.

lib/utils/forkedblockchain.js
```diff
@@ -234,52 +216,46 @@ ForkedBlockchain.prototype.getFallbackBlock = function(numberOrHash, cb) {
...
+        getBlock(number, callback);
       });
-    });
+    }
   });
 };
 
 ForkedBlockchain.prototype.getStorage = function(address, key, number, callback) {
-  this.lookupStorageTrie(address, function(err, trie) {
+  this.getLookupStorageTrie(this.stateTrie)(address, (err, trie) => {
```
davidmurdoch: Reason for change: Make sure we use the new and improved storage trie instead of the old and busted one (which has been removed)

lib/utils/forkedblockchain.js
```diff
@@ -122,30 +116,20 @@ ForkedBlockchain.prototype.createGenesisBlock = function(callback) {
...
-    fork: this.fork,
-    forkBlockNumber: this.forkBlockNumber
-  });
-
-  this.storageTrieCache[address] = trie;
-
-  return trie;
-};
-
-ForkedBlockchain.prototype.lookupStorageTrie = function(address, callback) {
```
davidmurdoch: Reason for change: we have a better one now. And the cache is mostly unnecessary as it complicates things (like when the stateTrie root is from a  previous block but the cache contains data from a newer block) and the cache is already properly handled in the VM, which most of the calls to this fn come from.

lib/utils/forkedblockchain.js
```diff
@@ -327,6 +303,28 @@ ForkedBlockchain.prototype.getCode = function(address, number, callback) {
   });
 };
 
+ForkedBlockchain.prototype.getLookupAccount = function(trie) {
```
davidmurdoch: Reason for change: we needed to be able to lookup an account based on a trie, not a block number.

lib/utils/forkedblockchain.js
```diff
@@ -640,33 +638,4 @@ ForkedBlockchain.prototype.getBlockLogs = function(number, callback) {
   });
 };
 
-ForkedBlockchain.prototype._checkpointTrie = function() {
```
davidmurdoch: Reason for change: These don't get used.

lib/utils/forkedstoragetrie.js
```diff
@@ -70,12 +58,14 @@ ForkedStorageTrie.prototype.get = function(key, blockNumber, callback) {
           callback(null, account.serialize());
         });
       } else {
+        if (to.number(blockNumber) > to.number(self.forkBlockNumber)) {
```
davidmurdoch: Reason for change: Whenever we go to the main chain we shouldn't use a block number after we forked because that data is not valid for us.

lib/utils/forkedstoragetrie.js
```diff
@@ -85,16 +75,30 @@ ForkedStorageTrie.prototype.get = function(key, blockNumber, callback) {
   });
 };
 
-// I don't want checkpoints to be removed by commits.
-// Note: For some reason, naming this function checkpoint()
-// -- overriding the default function -- prevents it from
-// being called.
-ForkedStorageTrie.prototype.customCheckpoint = function() {
```
davidmurdoch: These were never used.

lib/utils/forkedstoragetrie.js
```diff
@@ -1,41 +1,26 @@
...
+const MerklePatriciaTree = require("merkle-patricia-tree");
+const BaseTrie = require("merkle-patricia-tree/baseTrie");
+const checkpointInterface = require("merkle-patricia-tree/checkpoint-interface");
 var utils = require("ethereumjs-util");
 var inherits = require("util").inherits;
 var Web3 = require("web3");
 var to = require("./to.js");
 
-inherits(ForkedStorageTrie, MerklePatriciaTree);
+inherits(ForkedStorageBaseTrie, BaseTrie);
```
davidmurdoch: Reason for change:

Most of these changes are just code reorg. We now have a "Base" trie and our normal trie. This is just to align with the MPT implementation. I wanted to use the `copy` mechanism, but couldn't do so without this small refactor... because of reasons (things crashed).

PR #406, opened 2019-04-09
replace Web3() by Web3(provider)

fireblockdev:
The Web3() syntax is not valid anymore for the web3js project owner. To upgrade web3js to a recent version (> beta 37), it is an operation needed. However, this is not sufficient.

Other changes:
 * Web3.utils.toWei to Web3Utils.toWei (with the package web3-utils)
 * Web3.utils.BN to Web3Utils.BN
 * remove web3-providers-ws and replace by Web3.providers.WebsocketProvider

But after these changes, many unit tests doesn't work anymore

### DIFF
--- a/test/bad_input.js
+++ b/test/bad_input.js
@@ -187,15 +187,14 @@ var tests = function(web3) {
 };
 
 describe("Provider:", function() {
-  var web3 = new Web3();
-  web3.setProvider(Ganache.provider({}));
+  let web3 = new Web3(Ganache.provider({}));
   tests(web3);
 });
 
 describe("Server:", function(done) {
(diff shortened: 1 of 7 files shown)


ellis2323:
With the new version web3js 1.0.52, it seems that we need to provide these options:
```
new Web3(provider, null, { transactionConfirmationBlocks: 1, transactionBlockTimeout: 5 });
```

davidmurdoch:
Thanks for the PR, @ellis2323. Since the next stable version of web3 will be 1.0.0-beta.37, and in the near future we'll be removing nearly all web3-dependent tests from our codebase, I'm going to go ahead and close this.

PR #424, opened 2019-05-07
improvement: bump ethereumjs-wallet

andremedeiros:
This pull-request upgrades `ethereumjs-wallet`, which has upgraded the underlying dependency on scrypt.js to 0.3.0, making scrypt an optional dependency and offering a pure JS version as a fallback.

The reasoning behind this is that scrypt is problematic to install in some systems, particularly those that don't have node-gyp setup and we have seen some weird issues when installing with elevated privileges (i.e. `sudo npm install -g scrypt`)

This change would also require following up with bumping all dependencies on `ganache-core` to use the latest version too, which I can do as well.

cc. @iurimatias @michaelsbradleyjr

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -468,7 +468,6 @@
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
       "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
       "dev": true,
-      "optional": true,
       "requires": {
         "mime-types": "~2.1.18",
         "negotiator": "0.6.1"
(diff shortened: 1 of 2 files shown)


andremedeiros:
Huh, not sure if the CI failure is related to my change.

davidmurdoch:
This looks great! Thanks! I'll bump ganache and ganache-cli when the next ganache-core version is released. truffle will do the same shortly after that.

Thanks again!

davidmurdoch:
The CI failure is due to the transitive dependency `keccakjs` requiring node-sha3@1.x which doesn't build on Node 12. We are currently working to resolve this issue in the keccakjs repo.

Issue #425, opened 2019-05-20
Import fails on MacOS with Node 12


markspanbroek:
I'm unable to use ganache-core on MacOS with Node 12. I seem to have hit a combination of #297 (optional depencency on scrypt is not so optional) and https://github.com/barrysteyn/node-scrypt/issues/192 (scrypt doesn't and will never work on MacOS and Node 12).

## Expected Behavior
When I import ganache-core, I expect no errors.

## Current Behavior
I see the following error:
```
Error: Cannot find module 'scrypt'
Require stack:
- /Users/mark/Development/ganache-core/node_modules/web3-eth-accounts/node_modules/scrypt.js/node.js
- /Users/mark/Development/ganache-core/node_modules/web3-eth-accounts/src/index.js
- /Users/mark/Development/ganache-core/node_modules/web3-eth/src/index.js
- /Users/mark/Development/ganache-core/build/ganache.core.node.js
- /Users/mark/Development/ganache-core/index.js
etc..
```

## Possible Solution
Updating ethereumjs-wallet and web3 to the latest versions should remove the dependency on scrypt.

## Steps to Reproduce (for bugs)
Luckily the problem also manifests itself when running the unit tests of ganache-core, so the steps to reproduce are simple:

1. Checkout ganache-core on a Mac with Node 12
2. `yarn`
3. `yarn build`
4. `yarn test`

## Context
I use ganache-core as a web3 provider in Truffle for testing.

## Your Environment
* Version used: latest develop (cba166610008262266a60908d7ab4eb20fd856be)
* Environment name and version: Node 12.1.0
* Operating System and version: MacOS Mojave
* Link to your project: https://github.com/sustainablesource/sustainablesource

davidmurdoch:
Thanks for this! These abandoned native transitive deps have been quite the headache for many projects here at Truffle!

adrianmcli:
@markspanbroek Are you sure that upgrading `ethereumjs-wallet` to the latest version will remove the dependency on Scrypt? Because I see that they still list Scrypt.js as a dependency: https://github.com/ethereumjs/ethereumjs-wallet/blob/master/package.json#L46

markspanbroek:
@adrianmcli that's it exactly. Updating to scrypt.js 0.3.0 marks the native scrypt dependency as optional. With Node 12.1.0 the native scrypt doesn't compile, but as long as it's optional that doesn't matter.

adrianmcli:
[...]

PR #426, opened 2019-05-20
Fix webpack replacement of 'scrypt' with 'js-scrypt'.

### LINKED ISSUE #297: Ignore optional installs fails to find scrypt.js
Adding ganache-core as a dependency and installing all dependencies with:
```
yarn --frozen-lockfile --ignore-optional
```
Results in Ganache failing to load as it cannot find the scrypt dependency.
```
ganache-core-optional λ node
> const ganache = require("ganache-core");
{ Error: Cannot find module 'scrypt.js'
    at call (/Users/jacob/tmp/zz/ganache-core-optional/node_modules/ganache-core/build/webpack:/Ganache/webpack/bootstrap:19:1)
    at Object.__webpack_require__ (/Users/jacob/tmp/zz/ganache-core-optional/node_modules/ganache-core/build/webpack:/Ganache/node_modules/ethereumjs-wallet/index.js:8:16)
    at call (/Users/jacob/tmp/zz/ganache-core-optional/node_modules/ganache-core/build/webpack:/Ganache/webpack/bootstrap:19:1)
    at Object.require (/Users/jacob/tmp/zz/ganache-core-optional/node_modules/ganache-core/build/webpack:/Ganache/external "scrypt.js":1:1)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Module.require (internal/modules/cjs/loader.js:636:17)
    at Function.Module._load (internal/modules/cjs/loader.js:506:25)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15) code: 'MODULE_NOT_FOUND' }
```

 

## Expected Behavior
Ganache-core doesn't fail on load if optional dependencies are skipped.

## Current Behavior
Ganache-core fails on load if optional dependencies are skipped.

1. Add ganache-core as a dependency
2. Install packages ignoring optionals: `yarn --frozen-lockfile --ignore-optional`
3. Load ganache-core
4. Cannot find module `scrypt.js`

## Context

We are publishing a lean and mean Docker image and wanting to remove sizeable dependencies which are not required.

## Your Environment

* Version used: 2.2.0
* Server type and version:
* Operating System and version: Linux
* Link to your project:

markspanbroek:
Fixes #297.
Provides a workaround for #425.

The webpack code that replaced 'scrypt' with its javascript-only equivalent 'js-scrypt' didn't work because of two reasons that are corrected in this pull request:

1. js-scrypt was not installed
2. the dependency path has changed to `web3#web3-eth#web3-eth-accounts#scrypt.js#scrypt`

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "ethereumjs-wallet": "0.6.3",
     "ethers": "4.0.26",
     "husky": "^1.1.3",
+    "js-scrypt": "0.2.0",
     "lint-staged": "^8.0.4",
     "memdown": "^1.3.1",
     "mocha": "5.2.0",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -468,6 +468,7 @@
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
       "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
       "dev": true,
+      "optional": true,
       "requires": {
         "mime-types": "~2.1.18",
         "negotiator": "0.6.1"
@@ -552,7 +553,8 @@
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/any-promise/-/any-promise-1.3.0.tgz",
       "integrity": "sha1-q8av7tzqUugJzcA3au0845Y10X8=",
-      "dev": true
+      "dev": true,
+      "optional": true
     },
     "anymatch": {
       "version": "2.0.0",
@@ -612,7 +614,8 @@
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/array-flatten/-/array-flatten-1.1.1.tgz",
       "integrity": "sha1-ml9pkFGx5wczKPKgCJaLZOopVdI=",
-      "dev": true
+      "dev": true,
+      "optional": true
     },
     "array-union": {
       "version": "1.0.2",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
     "ethereumjs-util": "5.2.0",
     "ethereumjs-vm": "2.6.0",
     "heap": "0.2.6",
+    "js-scrypt": "^0.2.0",
```
davidmurdoch: This needs to be moved to `devDependencies` and pinned to `0.2.0`. The shrinkwrap file will have to be updated once moved as well.

davidmurdoch:
Thanks for this! Now I wonder how our tests have been passing as we test the built library before releasing.

package.json
markspanbroek: Agreed, it should be in devDependencies. Well spotted! I'll push an updated package.json in a minute.

Just out of curiosity, why would you like to have it pinned to 0.2.0? Shouldn't the npm-shrinkwrap.json take care of that?

markspanbroek:
I just pushed some changes; `js-scrypt` is in `devDependencies` now, and pinned to `0.2.0`.

It looks as though the [Travis CI builds for the develop branch](https://travis-ci.org/trufflesuite/ganache-core/branches) have not been successful since Node 12 came out.

package.json
davidmurdoch: Thanks.

For `yarn` users and so we can better control when we update transitive dependencies. If a developer working on ganache-core runs `npm install` it will automatically _update_ the shrinkwrap file, which can make things difficult to maintain and test consistently if dependencies are always updating (which still happens for transitive dependencies, unforatunely). Running `npm ci` avoids this, but isn't what it is meant for a takes longer to run (as it wipes out the node_modules dir first).

So, we try to keep core dependencies (dependencies used by core, and not just by our test suite) pinned.

markspanbroek: Thanks, that makes sense.

PR #432, opened 2019-06-11
Change transaction signature value encoding to QUANTITY

fjl:
Even though the Ethereum JSON-RPC spec says that transaction signature
values R and S are DATA, all clients encode them as QUANTITY. The
signature values are defined as numbers in the yellow paper, suggesting
QUANTITY as the correct encoding. This change ensures ganache is compatible
with Geth, Parity and any future spec change.

I found this issue through https://github.com/ethereum/go-ethereum/issues/19461 on
the go-ethereum tracker. See https://gist.github.com/fjl/69454fe6b56d04b3c4932d8673dedc63
for a comprehensive reproducer script.

### DIFF
--- a/lib/utils/transaction.js
+++ b/lib/utils/transaction.js
@@ -300,8 +300,8 @@ module.exports = class Transaction extends EthereumJsTransaction {
       gasPrice: to.rpcQuantityHexString(this.gasPrice),
       input: to.rpcDataHexString(this.data),
       v: to.nullableRpcQuantityHexString(this.v),
-      r: to.nullableRpcDataHexString(this.r),
-      s: to.nullableRpcDataHexString(this.s)
+      r: to.nullableRpcQuantityHexString(this.r),
+      s: to.nullableRpcQuantityHexString(this.s)
     };
 
     return resultJSON;


fjl:
There is no downside to merging this change, I think. Web3.js and others will likely accept any encoding for these values.

davidmurdoch:
Thanks for doing the leg work on this!

I understand this issue is a bit frustrating, but to be clear, this is NOT a bug in ganache-core, as we _are_ compliant with the spec, so I'd really prefer that you didn't label it as such in github issues and comments going forward as it may paint us in a bad light in the eyes of developers that don't understand the context of the issue. :-)

I had brought this up in the all core devs call a few weeks ago and it ended up turning into a conversation about maintaining and further developing the JSON-RPC specification rather than just changing this one return type.

That said, I realize I'm being a bit stubborn about changing this, but I feel that if I don't stand my ground, at least a little bit, the spec will likely go unchanged and become even further incomplete. I'd love to be able to take on spec maintenance myself but I don't currently have the bandwidth to do so.

fjl:
I agree the spec should change, and I will change it. Just want to bring it up in the call first. 
You're right about ganache being correct and everyone else being wrong. It's not a bug in ganache, that was just my first reaction to the go-ethereum issue when I found out about this problem :). Sorry :).

davidmurdoch:
@fjl 	Are you aware of any progress or updates on this issue on either the geth or spec side?

davidmurdoch:
@fjl 	Any updates on getting the spec updated? I know on the last core devs call it was agreed that it should be changed. I'd do it my self but I don't have edit privs for https://github.com/ethereum/wiki/wiki/JSON-RPC#returns-28 :-)

fjl:
I don't have edit permissions either. Will acquire them and update the spec today.

fjl:
Changed! https://github.com/ethereum/wiki/wiki/JSON-RPC/_compare/c6bc8bf4657f8bf0bb839e2b4a69e518dc559c4d...f4e624855ae05371b3fb78e02f5052679063b0b2

davidmurdoch:
Closing in favor of https://github.com/trufflesuite/ganache-core/pull/461. Thanks @fjl!

PR #442, opened 2019-07-03
Fix #111 `block.timestamp` `block.number` varies within block

### LINKED ISSUE #111: `block.timestamp` value varies within block
@vzts reported this behavior in [truffle 921](https://github.com/trufflesuite/truffle/issues/921) - there's more detail there. In summary it's unclear what the expected behavior is and @benjamincburns suggested we follow the other major clients, favoring static timestamps if there's no agreement.

Have just tested intra-block time values on `geth dev` and Parity POA over 4 calls at 1 second intervals: both have static timestamps. 

Will send a PR fixing this at [processCall](https://github.com/trufflesuite/ganache-core/blob/develop/lib/blockchain_double.js#L456) - just opening here in the meantime in case this change would impact any other timing related issues or there are implementation details to consider.

davidqhr:
Try to fix #921 #111

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -494,6 +494,13 @@ BlockchainDouble.prototype.processCall = function(tx, blockNumber, callback) {
       if (err) {
         return callback(err);
       }
+
+      // use correct block number and block timestamp to process the tx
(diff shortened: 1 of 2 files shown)


davidmurdoch:
Thanks for this! This PR was nearly there, but not quite. Closing in favor of #443.

davidqhr:
NP！

PR #443, opened 2019-07-03
fix: ensure timestamp, block number, and parentHash are correct

### LINKED ISSUE #111: `block.timestamp` value varies within block

davidmurdoch:
Closes #111, supersedes PR #442


lib/blockchain_double.js
```diff
@@ -354,12 +354,17 @@ BlockchainDouble.prototype.clearPendingTransactions = function() {
...
  */
-BlockchainDouble.prototype.createBlock = function(parent, callback) {
+BlockchainDouble.prototype.createBlock = function(parent, emulateParent, callback) {
   var self = this;
 
   if (typeof parent === "function") {
     callback = parent;
     parent = null;
+    emulateParent = false;
+  } else if (typeof emulateParent === "function") {
```
nicholasjpaterno: Looks like maybe we can reuse code here if we combine the conditions.

`typeof parent === "function" || typeof emulateParent === "function"`

davidmurdoch: Good catch. This was actually a bug in the logic. Fixing.

PR #444, opened 2019-07-03
Update solc to v5

### LINKED ISSUE #439: Need a `CREATE2` test for gas exactimation

nicholasjpaterno:
Update all contracts to compiler ^0.5.0
Update all tests to reflect new contracts and new solc input/output
Upgrade bootstrap, compile, deploy for new solc  AND
	add the ability to pass compiler options
Add new helper function to compile individual files
Add CREATE2 test fixes #439


test/gas/gas.js
```diff
@@ -249,7 +288,11 @@ describe("Gas", function() {
             const options = { from, gas: 5000000 };
 
             // prime storage by making sure it is set to 0
-            await instance.methods.reset().send(options);
+            try {
```
davidmurdoch: why do we need to `try`/`catch` here?

test/helpers/contract/compileAndDeploy.js
```diff
@@ -49,6 +62,7 @@ async function deploy(abi, bytecode, web3, options = {}, existingAccounts = [])
 
   if (existingAccounts.length) {
     block = await web3.eth.getBlock("latest");
+    accounts = existingAccounts;
```
davidmurdoch: nice catch!

test/gas/gas.js
nicholasjpaterno: Sorry! We don't, that code is vestigial from when I was testing.  Removing now :+1:

PR #452, opened 2019-07-15
fix: prevent crash from badly-encoded revert string

haltman-at:
This PR prevents Ganache from crashing if a transaction reverts with a badly-encoded revert string.  It's possible for `ethereumjs-abi`'s decoder to throw an error in certain cases, and as we're not prepared to catch it, it'll crash Ganache.  So, I've just added a try block around the decoding; and if anything goes wrong, well, we just leave the reason empty, same as if one wasn't provided or the error signature `08c379a0` is missing.

For examples of contracts that can cause this, you can see my `returndata-test` project in our `solidity-test-cases` repository.  Specifically, the file `ErrorError.sol`; try running the functions `wayOverlong` or `pointBeforeBeginning` and see what happens.  (Well, Ganache will crash, is what will happen.)

Why does `ethereumjs-abi` throw an error here?  Well, ABI encoding a 1-tuple containing a string (which is how revert strings are actually encoded, beyond the signature -- as a 1-tuple containing a string, rather than a string by itself, because ABI encoded data *always* has a tuple as the outermost layer) requires first following a pointer to find the length of the string, then reading the length of the string, then finally reading the string itself.  The `ethereumjs-abi` decoder attempts to convert both the pointer and the length to a `number` so that it can actually, well, use them; however, if the result is too large to fit in a `number`, this will throw an exception.  So, we need to be prepared to catch that exception; hence this PR.


haltman-at:
Note: I have no idea why Travis is failing here.  I'm going to quit restarting it for now and leave that sort of thing to you...

davidmurdoch:
Thanks, @haltman-at! I'll look into porting the relevant part of your test into ganache's tests soon so we don't regress in the future.

haltman-at:
OK, could somebody else add the tests?  I don't know how Ganache's tests are set up.  I mentioned two tests you can use in the original PR writeup though, so you can just integrate those into whatever your current test setup is.

haltman-at:
I put this PR up back when the build was broken.  Now that it's no longer broken, I thought I'd merge in `develop` and remove that red X...

PR #463, opened 2019-08-06
Updated Receipt's `logIndex`to be  index position in the _block_,. #596

sushilbal:
Updated the log index to be in sync with the number of logs in the block rather than determined by the number of transactions


nicholasjpaterno:
@sushilbal nice work!  The only thing we would need to merge this is a test case!  This is actually a good chance for us to formally test `eth_getFilterChanges` directly for the first time in our test suite! 
```javascript
  describe("eth_getFilterChanges", function() {
    it("should contain the correct logIndex", function() {
      
    });
  });
```
Something like this in `test/requests.js` with just one passing test would be ideal!

davidmurdoch:
Closing as this has been fixed in the `ts` branch.

PR #466, opened 2019-08-16
chore: update `ethereumjs-vm` to v4.1.0

davidmurdoch:
Once https://github.com/ethereumjs/ethereumjs-vm/pull/578 is pulled in and released this PR will need to updated to point at the new version before being merged in.


davidmurdoch:
This commit has a perf regression in ethereumjs-vm@4.1.0 but will be released as a special `istanbul` tag to npm until the regression is fixed.

PR #496, opened 2019-10-30
Allow constructor arguments for bootstrap

nicholasjpaterno:


davidmurdoch:
what does this change affect?

nicholasjpaterno:
Previously, you couldnt deploy a contract with args to its constructor.  This fixes that.  I can add a test to demo?

PR #517, opened 2019-12-10
Feat: update ganache-core to support the Istanbul hardfork

nicholasjpaterno:


bogdan:
Is it possible to release this version for the `web3 1.2.4` and enhanced node 12 support?

PR #529, opened 2020-01-14
Update README.md

nicholasjpaterno:


README.md
```diff
@@ -190,7 +190,7 @@ Special non-standard methods that aren’t included within the original RPC spec
   ```json
   { "id": 1337, "jsonrpc": "2.0", "result": "060" }
   ```
-* `evm_mine` : Force a block to be mined. Takes one optional parameter, which is the timestamp a block should setup as the mining time. Mines a block independent of whether or not mining is started or stopped.
+* `evm_mine` : Force a block to be mined (independent of mining status: started | stopped). Takes one **optional** parameter, which is the timestamp a block should setup as the mining time. NOTE: block timestamps should be truncated to `seconds` and **not** `milliseconds` precision.
```
davidmurdoch: Maybe change to something like
```NOTE: the timestamp parameter should be specified in seconds since the epoch. In JavaScript you would calculate it like this: `Math.floor(Date.now() / 1000);`.
```

PR #530, opened 2020-01-14
Fixes `TypeError: Cannot read property 'pop' of undefined`

davidmurdoch:
This fixes an issue caused by writing AND reading the same key multiple times simultaneously. Sometimes the read operation would end up reading the value as 0 bytes due to the way writes are performed in node. To fix this, we implemented a queue so only a single read or write operation can occur at a time for each key; basically an in-memory lock.

Additionally, during testing we found that it was possible for a write operation to fail due to program termination. This failure would sometimes cause the key to _exist_ but contain 0 bytes (which is always invalid). To fix this we write to a temp file, and only if it works do we move this temp file to its correct key location. This prevents early termination from writing partial/empty values.

Of course, all this will eventually be for nothing as we are migrating the db to actual an leveldb implementation that doesn't use a separate file for every key Soon(TM).

### DIFF
--- a/lib/database/filedown.js
+++ b/lib/database/filedown.js
@@ -3,6 +3,8 @@ var AbstractLevelDOWN = require("abstract-leveldown").AbstractLevelDOWN;
 var async = require("async");
 var fs = require("fs");
 var path = require("path");
+var tmp = require("tmp");
+tmp.setGracefulCleanup();
 
 util.inherits(FileDown, AbstractLevelDOWN);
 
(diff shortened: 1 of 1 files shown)


lib/database/filedown.js
```diff
@@ -3,6 +3,8 @@ var AbstractLevelDOWN = require("abstract-leveldown").AbstractLevelDOWN;
 var async = require("async");
 var fs = require("fs");
 var path = require("path");
+var tmp = require("tmp");
+tmp.setGracefulCleanup();
```
nicholasjpaterno: :+1:

lib/database/filedown.js
```diff
@@ -15,17 +17,78 @@ FileDown.prototype._open = function(options, callback) {
...
+    let cache = accessQueue.cache[lKey];
+    if (cache) {
+      cache.push(callback);
+    } else {
+      accessQueue.cache[lKey] = [];
+      callback();
+    }
+  },
+  cache: {}
+};
```
nicholasjpaterno: :+1: on the abstraction.  Much cleaner

PR #550, opened 2020-02-12
Update ethereumjs-util to v6.2.0

nicholasjpaterno:
Addresses https://github.com/trufflesuite/ganache-cli/issues/717

### DIFF
--- a/lib/utils/transaction.js
+++ b/lib/utils/transaction.js
@@ -109,7 +109,8 @@ function initData(tx, data) {
       const keys = Object.keys(data);
       tx._fields.forEach(function(field) {
         if (keys.indexOf(field) !== -1) {
-          self[field] = data[field];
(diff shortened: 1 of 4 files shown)


lib/utils/transaction.js
```diff
@@ -109,7 +109,8 @@ function initData(tx, data) {
       const keys = Object.keys(data);
       tx._fields.forEach(function(field) {
         if (keys.indexOf(field) !== -1) {
-          self[field] = data[field];
+          const val = data[field];
+          self[field] = val === "" ? "0x" : val;
```
davidmurdoch: I think if `val` is a string that isn't prefixed with 0x we should probably just prefix it here? Just to be safe?

nicholasjpaterno: :+1:

PR #578, opened 2020-05-28
fix: set correct `gasCost` values in `debug_traceTransaction`

### LINKED ISSUE #277: Ganache produces shifted or even negative gas costs in debug_traceTransaction
## Expected Behavior
I expect the trace to contain all the gas used by the transaction assigned to opcodes.

## Current Behavior
In reality, when I sum all the costs found in a trace I get much less than the gas found on a transaction receipt.

screen shot 2019-01-17 at 5 48 31 pm
(The image is actually from geth and is here just for visual purposes)

## Possible Solution
Investigate and either fix or explain why is it like that.

1. Create `truffle` project from a `metacoin` box.
2. Run JS test that calls `sendCoin`
3. Look at ganache logs to see the gas usage.
4. Request the transaction trace and sum up gas costs of opcodes.

You don't need that exact box. Just any smart contract transaction. This is probably the simplest one.
You could probably also use remix.

## Context
I'm building a profiler based on traces and I can only assign half of the gas to opcodes. I'd like to know where is the other half.

BTW, Geth has the same problem, so there is probably some fundamental misunderstanding on my side about how gas/traces work.

## Your Environment

* Version used: Latest and latest beta.
* Server type and version: 
* Operating System and version: Mac latest
* Link to your project: https://0x.org

iamdefinitelyahuman:
### What I did
Fix the `gasCost` values within `debug_traceTransaction` by shifting the values one step to the left.

Fixes #277

### How I did it
Currently, the value given for the first step is indicative of the cost of broadcasting the transaction (21k + calldata). The remaining values are all correct when applied to the previous step in the trace. 

I solve assigning `gasUsedThisStep` to the previous object in `structLogs`, instead of the object currently being built.

The final instruction in any transaction is `STOP`, `REVERT`, or `RETURN`. These are all priced at 0, so there is no issue with the last step being incorrectly priced.


WyseNynja:
I think I'm seeing problems caused by this (or at least related to this) in my tests that are comparing GST2 and 1inch's new CHI.

haltman-at:
> The final instruction in any transaction is `STOP`, `REVERT`, or `RETURN`. These are all priced at 0, so there is no issue with the last step being incorrectly priced.

I'm going to nitpick and point out that this is not true.  There's one more common way for a transaction to end, which is on an `INVALID`.  But really a transaction can end on any instruction due to running out of gas.  And there are plenty of other ways to get an exceptional halt as well, even if they're less common.

That said, I don't know how much the pricing of the last step really matters, even it might be nice to get right.  This PR sounds like an improvement one way or another. :)  I especially don't know how much it matters for exceptional halts (including `INVALID` and out of gas), since in those cases, as I understand it, the final instruction is never actually executed and paid for.  So like this may be good enough for the purposes!  But I had to point out that that statement wasn't quite true. :)

Actually there's another way it's not true: If I understand correctly, `RETURN` and `REVERT` can indeed use gas if the area of memory they return extends beyond the current bounds of memory.  Not very likely, obviously, but worth noting as technically possible!  (Again, attempting to account for this may simply be not worth the trouble, this may be fine as-is, but thought I'd point it out.)

iamdefinitelyahuman:
@haltman-at very interesting..  I hadn't thought about memory expansion during a `REVERT` or `RETURN`, that's good to know!  Thank you for explaining.

If there is a more accurate way to solve this problem I'm all for it. Currently we're applying the same shifting technique downstream, it'd be nice to have it solved where it's happening.

haltman-at:
Yes, like I said, I'm nitpicking; it's entirely possible what you're already doing is good enough for the purposes.  But I had to point it out. :)

PR #581, opened 2020-06-10
perf: cache requests to the underlying fork provider

fubhy:
We managed to improve the performance of the ganache fork we use for testing by a factor of 10 to 20 by caching identical requests to the underlying fork (our slowest test used to take ~250 seconds and frequently crash because of exceeding rate limits on our archive node / timeouts - It's now at 13 seconds consistently).

To achieve this, I simply patched the web3 provider used by the fork to intercept all calls to `send` and run them through a LRU cache.


fubhy:
I fixed that one failing test (serialization / deserialization issue in lru-cache apparently).

I wanted to add a test to verify that the caching actually works but noticed that a) you are using mocha (which doesn't have native method spies) for your test suite and b) there's currently no method spy dependency installed (spy.js / sinon). I didn't want to add yet another dependency as part of this PR so how'd you want to implement such a test without spying on the `.send` method?

davidmurdoch:
This looks really great! I'm working on some subtle fixes to the code, as well as writing some tests.

davidmurdoch:
Closing. Will be merging via #585.

PR #584, opened 2020-06-17
perf: cache requests to the underlying fork provider

davidmurdoch:
Based off #581

### DIFF
--- a/lib/utils/forkedblockchain.js
+++ b/lib/utils/forkedblockchain.js
@@ -69,19 +69,43 @@ function ForkedBlockchain(options) {
     // simply cache every non-error result because all requests to the
     // fork should be deterministic.
     this.fork.send = (payload, callback) => {
-      const key = JSON.stringify(Object.assign({}, payload, { id: null }));
-
-      if (cache.has(key)) {
-        callback(null, Object.assign({}, JSON.parse(cache.get(key)), { id: payload.id }));
+      let payloads;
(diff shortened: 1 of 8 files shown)


lib/utils/forkedblockchain.js
```diff
@@ -56,7 +57,57 @@ function ForkedBlockchain(options) {
   } else {
     this.fork = options.fork;
   }
+
   this.forkBlockNumber = options.fork_block_number;
+  this.forkCacheSize = parseInt(options.fork_cache_size == null ? 10000 : options.fork_cache_size);
```
nicholasjpaterno, line 62: Should we default to `10000`? I wonder if this should be opt in, though I don't suspect this will interfere with any current tests.

test/forking.js
```diff
@@ -193,6 +193,34 @@ describe("Forking", function() {
...
+    };
+    const provider = mainWeb3.currentProvider;
+    provider.send(tx, (_, result) => {
+      const result1 = Object.assign({}, result, { id: null });
+      assert.strictEqual(parseInt(result1.result), 7, "return value is incorrect");
+      tx.id = 2;
+      provider.send(tx, (_, result) => {
+        const result2 = Object.assign({}, result, { id: null });
+        assert.deepStrictEqual(result2, result1);
+        assert.strictEqual(callCount, 1, "cache didn't work");
```
nicholasjpaterno, line 217: :+1:

Issue #562, opened 2020-04-14
Husky throws an error if `ganache-core` is a dependency in a project


ylv-io:
## Expected Behavior
Expect Husky not to throw an error on Git commands.

## Current Behavior
Git `commit` would cause an error.

```bash
not found: husky-run
Can't find Husky, skipping pre-commit hook
You can reinstall it using 'npm install husky --save-dev' or delete this hook
On branch master
```

## Steps to Reproduce (for bugs)
1. git init
2. npm init
3. npm i -D ganache-core
4. git commit -m 'test'

## Context
You have to delete Git hooks to continue the development.

## Your Environment
* Version used: 2.10.2
* Operating System and version: MacOS Catalina 10.15.3 (19D76)

alcuadrado:
I'm seeing the exact same problem. The issue is that you are including husky in your shrinkwrap file.

frangio:
Are there any plans to fix this?

davidmurdoch:
Yes! I'll try to get it done next week.

alcuadrado:
I can confirm that `2.11.0-beta.0` fixes this

alcuadrado:
Hey @davidmurdoch, this happened once again, and `ganache-core@2.12.1` pulls its dev dependencies, including husky, once again.

davidmurdoch:
Thanks, @alcuadrado! It should already be fixed in the latest alpha release that I'm smoke testing now.

alcuadrado:
I just tried the alpha release, and it doesn't include husky, nor compiles any native dependency 🎉

When can we expect this to be released as non-alpha nor beta?

davidmurdoch:
Thanks for giving it a try @alcuadrado! What version of node did you install with?

And for release to beta, maybe today, and if that goes well, stable should be released by end of week, maybe even sooner.

alcuadrado:
Awesome news! 

I tried it with node v10.22.1 and npm v6.14.6

davidmurdoch:
Fixed (again) in https://github.com/trufflesuite/ganache-core/releases/tag/v2.13.0-beta.0 and https://github.com/trufflesuite/ganache-cli/releases/tag/v6.11.1-beta.1

PR #593, opened 2020-06-23
fix: update prepublishOnly script to remove husky from shrinkwrap

alcuadrado:
Hey @davidmurdoch,

I took a look at #562, and fixed it.

This PR modifies the `prepublishOnly` script so that `devDependencies` aren't included in the shrinkwrap file, and hence, not installed be packages depending on `ganache-core`.

The problem was twofold:

1. `npm prune` takes a `--production` flag instead of `--only=prod`. It's very unfortunate that `npm` just silently accepts unused params. You can learn more about this here: https://docs.npmjs.com/cli-commands/prune.html

2. `npm prune` only uninstall dependencies, but doesn't update the shrinkwrap file, so it has to be done manually after running `prune`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "build": "webpack-cli --config ./webpack/node/core.webpack.config.js",
     "build-web": "webpack-cli --config ./webpack/web-experimental/core.webpack.config.js",
     "format": "prettier --write \"{lib,perf,test}/**/*.js\" && eslint --fix --ignore-path .gitignore .",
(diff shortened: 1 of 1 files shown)


alcuadrado:
PS: This doesn't affect the `postpublish` script, so your publishing workflow doesn't need to be changed.

alcuadrado:
Travis is failing for some unrelated reason. This only happens in node 8. Is it a known issue?

davidmurdoch:
Thanks @alcuadrado! I was actually just starting to look into this issue when I got the notification of this PR!

The Node 8 failure is to be expected. I'll land the PR fixing that before merging this one in (and we expect to drop Node 8 support in the next month or two 🥳!)

My hunch as to what changed is that the behavior of `--only=prod` changed at some point, unless this has been a problem the whole time we've been using this publish script.

Thanks again!

alcuadrado:
Thanks for replying and merging this so quickly, @davidmurdoch !

PR #607, opened 2020-08-14
Add support for eth_signTypedData_v4

dmihal:
Add's support for the `eth_signTypedData_v4` RPC method. Also supports signing typed data where the payload is a string (which is how many nodes expect it).

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -505,6 +505,10 @@ StateManager.prototype.sign = function(address, dataToSign) {
 };
 
 StateManager.prototype.signTypedData = function(address, typedDataToSign) {
+  if (typeof typedDataToSign === 'string') {
+    typedDataToSign = JSON.parse(typedDataToSign);
+  }
+
   var account = this.accounts[to.hex(address).toLowerCase()];
   if (!account) {
     throw new Error("cannot sign data; no private key");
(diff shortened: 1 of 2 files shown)


davidmurdoch:
@dmihal do you know of nodes that have implemented `eth_signTypedData` and/or `eth_signTypedData_v4`?

dmihal:
@davidmurdoch Looks like Geth has eth_signTypedData, but not eth_signTypedData_v4

However, metamask includes eth_signTypedData_v4

gnidan:
Hey @dmihal sorry to drop the ball on this. Do you know if the situation around this method has changed at all? Does geth support it yet (e.g.)?

Regardless, for this PR, it might make sense to target `develop` now, which represents the upcoming Ganache v7.0.0 release. We _could_ get this into the existing major release, but we _are_ seeking to move forward.

Also it might be worth noting that it's been a long time since August of last year; this v4 method might already exist but without the `_v4` suffix. Sounds like we'll need to do some investigation.

Thanks!

dogutaskiran:
Any updates on getting this into the build?

davidmurdoch:
Released in https://github.com/trufflesuite/ganache/releases/tag/ganache%407.0.0-alpha.2

Issue #571, opened 2020-04-28
Forking mainnet results in erroneous revert reason


mds1:
This issue has been noticed by the Compound team before when forking the mainnet, and can be reproduced using the tests in [this commit](https://github.com/ScopeLift/endaoment/tree/13bd9eb75e28f64705207da1c12b178fae565d66).

### Steps to Reproduce
1. Clone the above repository at the specified commit
2. Run `npm install`
3. Compile contracts and run tests with `npx oz compile && npx mocha --exit --timeout 0 test/endaoment-test.js`
4. The tests will fail with the following console output

```
$ npx oz compile && npx mocha --exit --timeout 0 test/endaoment-test.js
Nothing to compile, all contracts are up to date.


  Endaoment
    ✓ should see the deployed Endaoment contract
    1) should allow a membership proposal & vote


  1 passing (39s)
  1 failing

  1) Endaoment
       should allow a membership proposal & vote:
     Error: Returned error: VM Exception while processing transaction: revert Endaoment::processProposal - token transfer to guild bank failed -- Reason given: re-entered.
     at PromiEvent (node_modules/@truffle/contract/lib/promievent.js:9:30)
      at TruffleContract.processProposal (node_modules/@truffle/contract/lib/execute.js:169:26)
      at Context.it (test/endaoment-test.js:61:29)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

### Explanation

In the file `endaoment-test.js` we are failing in the test called "should allow a membership proposal & vote" at the line `await this.instance.processProposal(0, {from: summoner});`

This line calls the function `processProposal()` in `Endaoment.sol`. The failure occurs in the following section of that function:

```javascript
// transfer tokens to guild bank
require(
    guildBank.deposit(proposal.tokenTribute),
    "Endaoment::processProposal - token transfer to guild bank failed"
);
```

This line is calling the `deposit()` function in `GuildBank.sol`. This function is shown below for convenience:

```javascript
function deposit(uint256 amount) public onlyOwner returns (bool) {
    bool transferSuccess = approvedToken.transferFrom(msg.sender, address(this), amount);
    if (!transferSuccess) {
        return false;
    }
    
    uint256 mintCode = cToken.mint(amount);
    if (0 == mintCode) {
        return false;
    }
    emit Deposit(amount);
    return true;
}
```

[...]

PR #610, opened 2020-08-15
test: add simple tests that fork Infura networks

seesemichaelj:
This PR adds a handful of simple tests to test forking against non-ganache networks, specifically Infura. I have configured Travis to inject the `INFURA_KEY` environment variable as a secret. Unfortunately, this adds ~10 seconds to tests but I think it's a valuable addition.

The tests will be skipped if the environment variable isn't defined (by default for dev environments)

I can add a `dotenv` requirement to consume a non-committed `.env` file if you want? For now, I'm manually injecting the environment variable when I run `INFURA_KEY=... npm run test`


lib/forking/forked_blockchain.js
```diff
@@ -800,4 +800,11 @@ ForkedBlockchain.prototype.getBlockLogs = function(number, callback) {
   });
 };
 
+ForkedBlockchain.prototype.close = function(callback) {
+  if (this.fork.disconnect) {
+    this.fork.disconnect();
+  }
+  BlockchainDouble.prototype.close.call(this, callback);
+};
```
seesemichaelj, line 808: This function was necessary to get the websockets test to stop. without this mocha won't exit because of the open websocket connection

test/forking/infura/simple.js
```diff
@@ -0,0 +1,156 @@
...
+  log: function(msg) {
+    /* console.log(msg) */
+  }
+};
+
+describe("Simple Infura", function() {
+  let INFURA_KEY;
+
+  before(function() {
+    if (shouldSkip()) {
```
nicholasjpaterno, line 18: Minor nitpick, seems like a premature abstraction to drop this in a new file when its basically a one liner and only used once.

test/forking/infura/simple.js
```diff
@@ -0,0 +1,156 @@
...
+      this.skip();
+    } else {
+      INFURA_KEY = process.env.INFURA_KEY;
+    }
+  });
+
+  it("forks mainnet https", async function() {
+    // https://etherscan.io/block/10661638
+    const blockHeight = 10661638;
```
nicholasjpaterno, line 27: Does the infura key have archival access? Will this blockHeight need to be adjusted to stay viable?

seesemichaelj: It does have archival access; there's no need to keep it updated

test/forking/infura/simple.js
seesemichaelj: I have future plans to use this abstraction later. It also can help others as we add more tests to the infura folder. It will keep this organized so that when someone 6 months from now wants to make a new test needs to figure out how to skip the tests, they don't need to remember the specific environment key

I'm happy to remove it if y'all really want to, but I believe it to be good organization skills

nicholasjpaterno: I generally subscribe to the rule of three. https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

nicholasjpaterno: Im sure its a fine abstraction, its just too few examples to know whether its ultimately the right one.

seesemichaelj: I just know that ganache-core's tests are already a mess and organization is important to prevent that from happening further, but definitely your call. I will remove the abstraction

nicholasjpaterno: No need to remove it, or at least I would just defer to @davidmurdoch on whether thats actually necessary. As I said, it simply seems premature from my perspective as your intention to use it later on was an unknown to me. I personally wait to abstract until there is actual code duplication as too many abstractions is actually counterproductive to organization. Too many one liner utility functions can become unwieldy.

davidmurdoch:
@seesemichaelj What are your thoughts on mocking these infura network requests?

seesemichaelj:
@davidmurdoch I feel like mocking them provides no additional value. The purpose of these tests are to do the actual non-mocked requests. None of the current tests do any real-world forking, and I believe it's important we do those tests. The costs are very minimal in CI and don't affect devs during development testing. I rather they are skipped than mocked because the mocked responses are easily covered in other ganache-forking-ganache tests.

davidmurdoch:
Are you ultimately trying to test Infura's failure scenarios? Or is this just for regression testing?

What specific test issues are better solved using an external source rather than mocking it locally? My concern is that archive-state Infura-based tests makes our testing closed-source.

seesemichaelj:
We have a use case where users use Infura as a forking endpoint (i.e. Teams and basically any other tutorial or workshop out there that talks about Ganache forking); this PR tests that explicit use case. It seems a bit silly and unproductive for us to solely rely on whether or not forking works in tests based on testing with our own provider.

Later down the road, I may add performance based tests for TDD to increase the performance of forking off of Infura.

I could make these tests non-archive-state, but I don't believe non-archive-state really tests performance and it's likely to be added in a later PR. I think it's totally okay for us to have "you must use an archival node to make these tests work or they'll be skipped"; if a user forks our repo and wants to test they will be skipped as if they didn't exist. If it ever becomes obsolete, they can be removed. However, I would consider these additional tests as value-added to the `ganache-core` testing suite for the time being.

I could also move all these tests to the `marshmallow` repo (an internal Truffle Suite project in case someone else stumbles on this comment), but I honestly believe they belong in `ganache-core`

seesemichaelj:
Further, it's possible that there will be some issues that are quicker to use mainnet states for reproduction than it is to spend the time to figure out the root problem. For example, see the test [Compound made](https://github.com/compound-finance/ganache-core/pull/5/files#diff-fc1a61905586f586c0eaab20e6999348) to reproduce the issue in https://github.com/trufflesuite/ganache-core/issues/571

davidmurdoch:
I can't take the time to write out a full/thoughtful response at the moment, but in case you are going to work the #571 make sure you check the history of this delete method: https://github.com/trufflesuite/ganache-core/pull/482#issue-317104527 . I don't think the solution is to replace delete with `put(0)`, this will likely just move the goalpost.

seesemichaelj:
> I don't think the solution is to replace delete with put(0), this will likely just move the goalpost.

Good, I'm also skeptical

davidmurdoch:
Okay, so I'm :+1: with running these tests as part of CI, but I'll want the test directories reorganized so we have a clear separation between smoke tests[*](#user-content-one), like these, and our existing unit and integration tests.

I currently (manually :sweat_smile:) run a bunch of other smoke tests before publishing, I can make these Infura tests part of that process. My goal has always been to fully automate these pre-publish smoke tests before deployment, it just hasn't happened yet.

Perhaps a decent solution is to divide tests within the `test` directory into two parts: `smoke` and `local` (or some other name if you can think of something better). A new npm script should be added to run these smoke tests, and this npm script should be configured to be run by CI (Travis, currently), as well as on prepublish. Any secrets that are used by these tests should be added to Travis (which you've already done, thanks!), and to [GitHub Secrets](https://github.com/organizations/trufflesuite/settings/secrets) as we'll be switching Soon:tm:.

--- 

In case my lack of rebuttal isn't satisfactory: if it isn't clear what a test is testing, the test itself becomes a maintenance liability. For every problem/bug/feature we should have a test to exercise it specifically. Committing tests without an understanding of the root of the problem may save some time now, but can cause headaches later. I'm not definitively _for_ automated smoke tests, but for every issue brought to light by a smoke test, the solution should have an accompanying unit/integration test that targets the issue.

---

Thanks for diving into these issues!

---

<sub><a name="one">*</a> a better term for these tests might be "system tests", but if they aren't testing a specific feature of the system, that might not be too accurate a term either... but I digress.</sub>

seesemichaelj:
> In case my lack of rebuttal isn't satisfactory: if it isn't clear what a test is testing, the test itself becomes a maintenance liability. For every problem/bug/feature we should have a test to exercise it specifically.

I argue this tests a feature: being able to use the Infura service as a forking candidate

> but I'll want the test directories reorganized so we have a clear separation between smoke tests, like these, and our existing unit and integration tests...Perhaps a decent solution is to divide tests within the test directory into two parts: `smoke` and `local`

Can you define exactly what you want me to do? Do you just want me to move `test/*` into `test/local/*` and move `test/forking/infura` to `test/smoke/infura`? Are there other current tests that are supposed to be moved? Ganache's testing is already poorly organized, and I'm not sure if I'm the right person to start determining what's an actual test and what's a smoke test.

This was supposed to be a simple value-added PR, and we've spent more time talking about the semantics of testing and organization than actually solving issues. If you don't want this PR, I'm totally fine with closing this PR and keeping these tests in my local branch to help me with my personal development usage of ganache. I just thought this was going to 1) help me solve forking bugs, 2) help future devs with forking bugs, and 3) provide a dead-simple set of tests that tests a fundamental feature of forking

davidmurdoch:
I do appreciate the value-add of these specific tests, if that wasn't clear. I only want to develop a definitive separation of the types of tests developers and contributors must run. I try to avoid to temptation to pile on to the mess just because this is the way things are already.

> Can you define exactly what you want me to do?

I was hoping you could align with me on these organization goals and further develop the testing methodology. If that's not something you are interested in, moving the directories as you've suggested is fine.

> Are there other current tests that are supposed to be moved

There are no other smoke/system tests to move.

> we've spent more time talking about the semantics of testing and organization than actually solving issue

I don't want you to become apathetic to ganache's development processes, or lack thereof; I highly appreciate your feedback and strongly value your expertise. If you do want to discuss further perhaps we should sync over Zoom (git me up on Slack if you'd like!).

seesemichaelj:
@davidmurdoch this is ready for review

seesemichaelj:
Alright @davidmurdoch this one is ready again

PR #612, opened 2020-08-18
fix(forking): invalidate `_deleted` in `ForkedStorageTrie` if a subsequent `put` happens

### LINKED ISSUE #571: Forking mainnet results in erroneous revert reason

seesemichaelj:
This PR resolves #571 by wrapping the `Trie.put` method in `ForkedStorageTrie`, checking if `_deleted.get(rpcKey) === 1`, and setting it to `0` if it is. This PR adds a simple test that gets to the root of the issue: `_deleted` was being set for a storage slot 0 when a bool was changed from `true` to `false`, but we never "undelete" it if it's set back to `true`. This caused an erroneous error when contracts looked at that boolean again and got `0` because `ForkedStorageTrie` determined it was set to `0` (interpreted as `false`) [here](https://github.com/trufflesuite/ganache-core/blob/34504feaa59d7e59fcc360c0332df10f57131b51/lib/forking/forked_storage_trie.js#L55-L59)

To add a reason for supporting merging #610, those Infura tests gave me a simple, ready to use way to reproduce the error and understand the root cause. From there, I determined I could create a non-Infura reproduction of the error.

### DIFF
--- a/lib/forking/forked_storage_trie.js
+++ b/lib/forking/forked_storage_trie.js
@@ -97,6 +97,23 @@ ForkedStorageBaseTrie.prototype.keyExists = function(key, callback) {
   });
 };
 
+const originalPut = ForkedStorageBaseTrie.prototype.put;
+ForkedStorageBaseTrie.prototype.put = function(key, value, callback) {
+  let rpcKey = to.rpcDataHexString(key);
+  if (this.address) {
+    rpcKey = `${to.rpcDataHexString(this.address)};${rpcKey}`;
+  }
+  this._deleted.get(rpcKey, (_, result) => {
+    if (result === 1) {
+      this._deleted.put(rpcKey, 0, () => {
+        originalPut.call(this, key, value, callback);
+      });
+    } else {
+      originalPut.call(this, key, value, callback);
+    }
+  });
+};
+
 ForkedStorageBaseTrie.prototype.keyIsDeleted = function(key, callback) {
   let rpcKey = to.rpcDataHexString(key);
   if (this.address) {
--- a/test/contracts/forking/StorageDelete.sol
+++ b/test/contracts/forking/StorageDelete.sol
@@ -1,0 +1,18 @@
+pragma solidity ^0.6.0;
+
+contract StorageDelete {
+  bool entered;
+
+  constructor() public {
+    entered = true;
+  }
+
+  function test() public nonReentrant {}
+
+  modifier nonReentrant() {
+    require(entered, "re-entered");
+    entered = false;
+    _;
+    entered = true;
+  }
+}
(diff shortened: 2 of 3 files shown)


davidlucid:
@seesemichaelj thank you so much!

chanhosuh:
This is great!  Any idea when this will get merged into ganache-cli?

seesemichaelj:
@chanhosuh You can try it out now with [v6.10.1-beta.2](https://github.com/trufflesuite/ganache-cli/releases/tag/v6.10.1-beta.2)!

PR #613, opened 2020-08-20
fix(forking): leverage the requested block number when getting forked storage

### LINKED ISSUE #573: eth_getStorageAt with block number parameter doesn't return correct result when using fork feature
JSON RPC method [eth_getStorageAt](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getStorageAt) always returns the latest state when using fork feature regardless of the third parameter

## Expected Behavior
When changing the storage value in contract c at key k from v1 to v2 in block b, `eth_getStorageAt(c, k, b-1)` should return v1.

## Current Behavior
`eth_getStorageAt(c, k, b-1)` returns v2 for any b.
For storage values that didn't change after fork, `eth_getStorageAt` works as expected

1. `ganache-cli -f mainnet_node --networkId 1`
2. `truffle migrate`
3. `eth_getStorageAt("[migration_contract]", "0x1", "0x0")`

## Your Environment
* Version used: Ganache CLI v6.9.1 (ganache-core: 2.10.2)
* Operating System and version: macOS Catalina 10.15.4

seesemichaelj:
This PR resolves #573 leveraging the direction [proposed](https://github.com/trufflesuite/ganache-core/issues/573#issuecomment-676670387) by @nebojsa94. If we're getting storage prior to the fork, we don't even bother with ForkedStorageTrie and use `eth_getStorageAt` in `ForkedBlockchain.getStorage`. If we do end up hitting the forked storage trie, then we make sure we get the requested block state root prior to looking up storage

I believe there is a possible cleaner architecture, but with other refactoring efforts going on plus the time to refactor in the old codebase, I feel like this is an acceptable solution.

### DIFF
--- a/lib/forking/forked_blockchain.js
+++ b/lib/forking/forked_blockchain.js
@@ -403,11 +403,32 @@ ForkedBlockchain.prototype.getStorage = function(address, key, number, callback)
     if (err) {
       return callback(err);
     }
-    this.getEffectiveBlockNumber(number, (err, number) => {
+    this.getEffectiveBlockNumber(number, (err, blockNumber) => {
       if (err) {
         return callback(err);
       }
-      trie.get(utils.setLengthLeft(utils.toBuffer(key), 32), number, callback);
+
+      if (blockNumber > this.forkBlockNumber) {
+        // only hit the ForkedStorageTrieBase if we're not looking
+        // for something that's on the forked chain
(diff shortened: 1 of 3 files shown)


nebojsa94:
@seesemichaelj Thank you for implementing these changes, looks good!

Issue #588, opened 2020-04-22
Allow `--unlock` functionality when ganache is already running


iamdefinitelyahuman:
* I asked about this on [gitter](https://gitter.im/ConsenSys/truffle): [x]
* Link to my question on gitter: https://gitter.im/ConsenSys/truffle?at=5ea068f0f39e5c2fbb970216

## Summary
It is possible to unlock accounts with the `--unlock` flag when launching ganache. It would be useful to be able to do this while ganache is already running.

## Context
When combined with`--fork`, `--unlock` is very powerful in that it lets the me simulate mainnet transactions from accounts I don't control. However, I like to play around in the console and so often don't know which accounts I want to unlock ahead of time. Currently each new unlock means exiting, restarting, and repeating a (sometimes lengthy) series of transactions.

## Possible Solution
I have three ideas for possible implementations:

1. Add a cli flag `--unlock-all` that arbitrarily allows transactions to be broadcast from any address, without having to explicitly unlock first. In my mind this is the best approach.
2. Add a new custom RPC endpoint that accepts an address as a single parameter, and unlocks that address.
3. Modify the behavior of [`personal_unlockAccount`](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal_unlockAccount) so that any account may be unlocked with a passphrase of `null`. Optionally add a cli flag that enables this behavior.

iamdefinitelyahuman:
An RPC method would work fine for me..  I see in my initial writeup I suggested a flag as the best approach, but today I feel the custom RPC method is actually better as it doesn't require me to do anything special prior to launching.

Thanks for your work on this tool btw!

ItsNickBarry:
It would be useful to be able to simply unlock all accounts, preventing the need for RPC calls.  I can't think of a test case that would rely on an account's being locked.

ItsNickBarry:
I now see that that's already been suggested.  In that case, I agree that `--unlock-all` would be a useful addition.

iamdefinitelyahuman:
Implemented in #622

surajsjain:
Is there any way I can use this with the latest version (or a different branch) of Ganache CI so that I can unlock all the accounts?

PR #622, opened 2020-09-01
add evm_unlockUnknownAccount

rymnc:
This feature is much-needed for on the fly transaction creation, I hope you accept this PR!
I can write more tests if required

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "ganache-core",
-  "version": "2.11.3-beta.0",
+  "version": "2.11.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ganache-core",
-  "version": "2.11.3-beta.0",
+  "version": "2.11.3",
   "main": "./index.js",
   "types": "./typings/index.d.ts",
   "engines": {


davidmurdoch:
Thanks for the PR. We'll need to discuss implementation details (such as the RPC endpoint) before mergeing.

Related issue: #588

aaryamannchallani:
> Thanks for the PR. We'll need to discuss implementation details (such as the RPC endpoint) before mergeing.
> 
> Related issue: #588

Thank You!

aaryamannchallani:
@davidmurdoch would you be open to using the
`web3.extend` module?
It could help if you are planning of extending the json rpc api
[Documentation](https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#extend)
It has i/o formatters as well

davidmurdoch:
I don't think this is something that could be done from within ganache-core, as we don't use web3 (except for a couple of RPC calls when in forking mode). Maybe trufflesuite/truffle could extend their injected web3 with ganache's missing RPC methods, if they don't already.

aaryamannchallani:
yes, that's true. Just had a little doubt though. does truffle use ganache-core?

davidmurdoch:
Yes, it does for the default development chain.

aaryamannchallani:
@davidmurdoch any update on integrating this endpoint into the suite?

davidmurdoch:
@aaryamannchallani We're discussing this internally. For now, let's change the namespace to `evm_unlockAccount`, as we use that for our non-standard rpc endpoints.

There are some security considerations to be aware of here. As is, this will allow any user to unlock any account, even an intentionally locked account (accounts that are in `accounts` but not in `unlocked_accounts`, which can happen at start up or via some of the `personal_` RPC calls).

So we need to a think of a way to secure this RPC endpoint by default, as ganache is often used in collaborative environments.

aaryamannchallani:
got it, will make that change
as for the locked accounts, I'll have to read the codebase a bit more

lib/subproviders/geth_api_double.js
```diff
@@ -574,6 +574,18 @@ GethApiDouble.prototype.debug_traceTransaction = function(txHash, params, callba
...
+// Unlocks any account after server start
+GethApiDouble.prototype.debug_unlockAccount = function(address, callback) {
+  // Checks if given address is already unlocked
+  if (this.state.unlocked_accounts[address.toLowerCase()] !== true) {
+    // Unlocks it
+    this.state.unlocked_accounts[address.toLowerCase()] = true;
+    return callback(null, true);
+  }
+  // If Account exists, this is the callback
+  callback(null, "Account Already Unlocked");
```
davidmurdoch, line 586: Let's make this should return `false` instead of `"Account Already Unlocked"`.

lib/subproviders/geth_api_double.js
```diff
@@ -574,6 +574,18 @@ GethApiDouble.prototype.debug_traceTransaction = function(txHash, params, callba
   this.state.queueTransactionTrace(txHash, params, callback);
 };
 
+// Unlocks any account after server start
+GethApiDouble.prototype.debug_unlockAccount = function(address, callback) {
```
davidmurdoch, line 578: We chatted about this internally and decided on this method should be named `evm_unlockUnknownAccount`.

lib/subproviders/geth_api_double.js
```diff
@@ -574,6 +574,18 @@ GethApiDouble.prototype.debug_traceTransaction = function(txHash, params, callba
...
+  // Checks if given address is already unlocked
+  if (this.state.unlocked_accounts[address.toLowerCase()] !== true) {
+    // Unlocks it
+    this.state.unlocked_accounts[address.toLowerCase()] = true;
```
davidmurdoch, line 582: Before we can merged this the behavior needs to be changed so that it _does not_ unlock accounts found in `this.state.personal_accounts`, as these can already be unlocked (if previously locked at start up or via `personal_lockAccount`) via `personal_unlockAccount`.

davidmurdoch: In the case where the account cannot be unlocked, return an error in the callback:

```
  callback(new Error("cannot unlock known/personal account"));
```

davidmurdoch (changes requested):
We chatted about this internally and decided on this method should be named `evm_unlockUnknownAccount`.

Before we can merge this the behavior needs to be changed so that it _does not_ unlock accounts found in `this.state.personal_accounts`, as these can already be unlocked (if previously locked at start up or via `personal_lockAccount`) via `personal_unlockAccount`.

lib/subproviders/geth_api_double.js
aaryamannchallani: Resolved in this [commit](https://github.com/trufflesuite/ganache-core/pull/622/commits/010e21e05210423bcce925c8fdd7167745a8e2d0)

aaryamannchallani:
@davidmurdoch let me know if this is what you require!

davidmurdoch:
Can you enable "allow edits from maintainers" on this PR so I can commits some additional work directly to this PR?

aaryamannchallani:
It is enabled!

aaryamannchallani:
ah thank you!
will it be in the next version 🙂

davidmurdoch:
I didn't mean to close this just yet! I just pushed the wrong local branch :-). I'm working on fixing it back up now.

aaryamannchallani:
ah okay cool

davidmurdoch:
@aaryamannchallani Merged!

aaryamannchallani:
Awesome! let me know if I can help anywhere else

surajsjain:
How do I use this now for unlocking all the accounts from a forked chain?

davidmurdoch:
You'll still need to unlock them individually before using each one. We don't yet have a way of unlocking _all_ accounts.

```javascript
// unlock an address
await provider.send("evm_addAccount", [address, passphrase] );
await provider.send("personal_unlockAccount", [address, passphrase] );
```

MicaiahReid:
To add some more detail for any future readers, see #3603

PR #627, opened 2020-09-09
fix(forking): fix snapshots for forking

### LINKED ISSUE #611: Transaction status is 1, but last opcode is REVERT
I have a project that uses brownie and ganache-cli --fork against a geth node. (Geth is running according to [these steps](https://gist.github.com/WyseNynja/89179917d209d10e6ea27c5f2f8f88f1)).

A few of my tests check the return_value of my transactions. Most succeed, but one has no return value despite not reverting.

* I asked about this on [gitter](https://gitter.im/ConsenSys/truffle): [X]
* Link to my question on gitter: https://gitter.im/ConsenSys/truffle?at=5ed03f164412600ccd7497f0

## Expected Behavior

A transaction with a status of 1 should have a return value.

## Current Behavior

- eth_getTransactionReceipt.status is 1
- debug_traceTransaction final instruction is REVERT

## Steps to Reproduce

1. git clone https://github.com/SatoshiAndKin/argobytes-contracts-brownie (at ecf07c4)
2. [Follow the readme](https://github.com/SatoshiAndKin/argobytes-contracts-brownie#install)
3. `./scripts/test.sh tests/test_uniswap_v1_arbitrage.py -I`

```
./scripts/test.sh tests/test_uniswap_v1_arbitrage.py -I
+ [ -z /home/ski/code/argobytes-contracts-brownie/venv ]
+ brownie test --network mainnet-fork tests/test_uniswap_v1_arbitrage.py -I
Brownie v1.8.9 - Python development framework for Ethereum

===================================================================================== test session starts =====================================================================================
platform linux -- Python 3.8.2, pytest-5.4.2, py-1.8.1, pluggy-0.13.1
rootdir: /home/ski/code/argobytes-contracts-brownie
plugins: eth-brownie-1.8.9, forked-1.1.3, web3-5.10.0, hypothesis-5.15.0, xdist-1.32.0
collecting ... Attached to local RPC client listening at '127.0.0.1:8575'...
collected 1 item                                                                                                                                                                              

tests/test_uniswap_v1_arbitrage.py F
address_zero = '0x0000000000000000000000000000000000000000', argobytes_atomic_trade = <ArgobytesAtomicTrade Contract '0xF024fE5EF0b1Bf88caf9B1222E49d1A5CC312a71'>
dai_erc20 = <Dai Contract '0x6B175474E89094C44Da98b954EedeAC495271d0F'>, argobytes_owned_vault = <ArgobytesOwnedVault Contract '0xb66F46436fa703237A9Fd26f4DA63919D3cc943D'>
example_action = <ExampleAction Contract '0xDb43EFA90874d427e38Ff1e5F08572c0a2619930'>, chi = <ChiToken Contract '0x0000000000004946c0e9F43F4Dee607b0eF1fA1c'>
uniswap_v1_factory = <Vyper_contract Contract '0xc0a47dFe034B400B47bDaD5FecDa2621de6c4d95'>, uniswap_v1_action = <UniswapV1Action Contract '0x006ba56B025c63aE46c510a904422945F2389d6F'>
usdc_erc20 = <FiatTokenProxy Contract '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48'>

    def test_uniswap_arbitrage(address_zero, argobytes_atomic_trade, dai_erc20, argobytes_owned_vault, example_action, chi, uniswap_v1_factory, uniswap_v1_action, usdc_erc20):
        assert argobytes_owned_vault.balance() == 0
        assert example_action.balance() == 0
    
        value = 1e18
    
        # send some ETH into the vault
        accounts[0].transfer(argobytes_owned_vault, value)
        # send some ETH into the sweep contract to simulate arbitrage profits
        accounts[0].transfer(example_action, value)
    
        # mint some gas token
        # TODO: how much should we make?
        argobytes_owned_vault.mintGasToken(chi, 26, {"from": accounts[0]})
    
        # make sure balances match what we expect
        assert argobytes_owned_vault.balance() == value
        assert example_action.balance() == value
    
        usdc_exchange = uniswap_v1_action.getExchange(uniswap_v1_factory, usdc_erc20)
        dai_exchange = uniswap_v1_action.getExchange(uniswap_v1_factory, dai_erc20)
    
        # sweep a bunch of times to use up gas
        encoded_actions = argobytes_atomic_trade.encodeActions(
            [
                example_action,
                uniswap_v1_action,
                uniswap_v1_action,
                uniswap_v1_action,
            ],
            [
                # add some faked profits
                example_action.sweep.encode_input(uniswap_v1_action, address_zero),
    
                # trade ETH to USDC
                # uniswap_v1_action.tradeEtherToToken(address to, address exchange, address dest_token, uint dest_min_tokens, uint trade_gas)
                uniswap_v1_action.tradeEtherToToken.encode_input(uniswap_v1_action, usdc_exchange, usdc_erc20, 1, 0),
    
                # trade USDC to DAI
                # uniswap_v1_action.tradeTokenToToken(address to, address exchange, address src_token, address dest_token, uint dest_min_tokens, uint trade_gas)
                uniswap_v1_action.tradeTokenToToken.encode_input(
                    uniswap_v1_action, usdc_exchange, usdc_erc20, dai_erc20, 1, 0),

                # trade DAI to ETH
                # uniswap_v1_action.tradeTokenToEther(address to, address exchange, address src_token, uint dest_min_tokens, uint trade_gas)
                uniswap_v1_action.tradeTokenToEther.encode_input(address_zero, dai_exchange, dai_erc20, 1, 0),
            ],
        )
    
        arbitrage_tx = argobytes_owned_vault.atomicArbitrage(
            chi, argobytes_atomic_trade, address_zero, [address_zero], value, encoded_actions, {'from': accounts[1]})
    
        assert argobytes_owned_vault.balance() > value
    
        # make sure the transaction succeeded
        # there should be a revert above if status == 0, but something is wrong
        assert arbitrage_tx.status == 1
>       assert arbitrage_tx.return_value is not None
E       AssertionError: assert None is not None
E        +  where None = <Transaction '0x903f37e1596487075620d2c83dea892287b553c0147d97fb8bf4642a2083e177'>.return_value

tests/test_uniswap_v1_arbitrage.py:67: AssertionError

Interactive mode enabled. Use quit() to continue running tests.
>>> arbitrage_tx.trace[-1]['op']                                                                                                                                                               
'REVERT'
>>> arbitrage_tx._trace[-1]['op']                                                                                                                                                              
'REVERT'
>>>  
```

## Context

Due to previous issues with reverting traces pointing to incorrect lines, I've disabled solc's runs, so that shouldn't be the problem.

I suspected that gastoken2/chi might be related to the issue, but removing that code didn't change anything.

I'm sure the transaction isn't actually reverting because my test balance increases after the transaction is completed. If the transaction had reverted, the balance would not increase.

I don't think Brownie is the problem here. Brownie just gets the trace from ganache.

## Your Environment
* Version used: ganache-cli 6.9.1 and 6.10.0-beta.1
* Version of Truffle/Remix/Other tools used: eth-brownie-1.8.9
* NodeJS Version: [ ] 6.x, [ ] 7.x (unsupported), [ ] 8.x, [ ] 9.x, [x] v12.16.3
* Operating System and version (include distro if Linux): Ubuntu 20.04
* Link to your project or repro gist: https://github.com/SatoshiAndKin/argobytes-contracts-brownie
* Commit hash to use with above link for reproduction: ecf07c4
* I intend to submit a pull request to fix this issue: [ ]

### LINKED ISSUE #623: Ensure `ForkedBlockchain.prototype.getAccount` works
This issue pulls out #3 from https://github.com/trufflesuite/ganache-core/issues/483#issuecomment-680185528:

Yes, I meant `ForkedBlockchain.prototype.getAccount`. There might be a separate bug in this fn, I haven't checked if there are tests for this scenario yet:
- contract account A exists at block 9 on the original chain. it has a non-zero balance
- we fork at block 10
- we `selfdestruct()` the contract on our fork at block 11, thus transferring its value to another address.
- `getAccount(account A, block 12)` may return the account from the original chain. A call to `eth_getBalance` may return the non-zero account value.

### LINKED ISSUE #624: Ensure `ForkedBlockchainBaseTrie.prototype.del` works with snapshots
I have a feeling this may be a duplicate of #611, but I'm pulling #4 out of https://github.com/trufflesuite/ganache-core/issues/483#issuecomment-680185528

> I wonder if we can handle evm_snapshot + .del + evm_revert. 👀

mikeseese:
This issue fixes #624 and fixes #611 with the following approach:
- Track what keys we've touched (put/delete) rather than what keys we've deleted (not sure if this was necessary, but it is a cleaner approach)
    - Track when we first touched these keys by block number and vice versa (block number -> which keys)
- Handle `popBlock` in `ForkedBlockchain` to forget that we touched respective keys in that block

A couple of side effects:
- `ForkedStorageBaseTrie` no longer overloads `get` with a block number (we use `this.blockchain.blocks.last()` to get the current block now)
- This also fixes #623. I can separate this into another PR if you want

### DIFF
--- a/lib/forking/forked_storage_trie.js
+++ b/lib/forking/forked_storage_trie.js
@@ -85,6 +85,7 @@ ForkedStorageBaseTrie.prototype.touch = function(key, callback) {
   if (this.address) {
     rpcKey = `${to.rpcDataHexString(this.address)};${rpcKey}`;
   }
+  rpcKey = rpcKey.toLowerCase();
 
   this._touched.get(rpcKey, (err, result) => {
     if (err && err.type !== "NotFoundError") {
@@ -126,6 +127,7 @@ ForkedStorageBaseTrie.prototype.getTouchedAt = function(key, callback) {
   if (this.address) {
     rpcKey = `${to.rpcDataHexString(this.address)};${rpcKey}`;
   }
+  rpcKey = rpcKey.toLowerCase();
 
   this._touched.get(rpcKey, function(err, result) {
     if (err && err.type !== "NotFoundError") {
--- a/lib/forking/forked_storage_trie.js
+++ b/lib/forking/forked_storage_trie.js
@@ -95,6 +95,7 @@ ForkedStorageBaseTrie.prototype.touch = function(key, callback) {
         const number = lastBlock === null ? self.forkBlockNumber : to.number(lastBlock.header.number);
         if (this.persist) {
           this._touched.put(rpcKey, number + 1);
+          this.blockchain._touchedKeys.push(rpcKey);
         }
         callback();
       });
--- a/lib/forking/forked_storage_trie.js
+++ b/lib/forking/forked_storage_trie.js
@@ -2,7 +2,6 @@ const Sublevel = require("level-sublevel");
 const MerklePatriciaTree = require("merkle-patricia-tree");
 const BaseTrie = require("merkle-patricia-tree/baseTrie");
 const checkpointInterface = require("merkle-patricia-tree/checkpoint-interface");
-const Account = require("ethereumjs-account").default;
 var utils = require("ethereumjs-util");
 var inherits = require("util").inherits;
 var Web3 = require("web3");
@@ -12,7 +11,7 @@ inherits(ForkedStorageBaseTrie, BaseTrie);
 
 function ForkedStorageBaseTrie(db, root, options) {
   BaseTrie.call(this, db, root);
-  this._deleted = Sublevel(this.db).sublevel("deleted");
+  this._touched = Sublevel(this.db).sublevel("touched");
 
   this.options = options;
   this.address = options.address;
@@ -20,101 +19,40 @@ function ForkedStorageBaseTrie(db, root, options) {
   this.blockchain = options.blockchain;
   this.fork = options.fork;
   this.web3 = new Web3(this.fork);
+  this.persist = typeof options.persist === "undefined" ? true : options.persist;
 }
 
 // Note: This overrides a standard method whereas the other methods do not.
-ForkedStorageBaseTrie.prototype.get = function(key, blockNumber, callback) {
+ForkedStorageBaseTrie.prototype.get = function(key, callback) {
   var self = this;
-  let blockNumberProvided = true;
-
-  // Allow an optional blockNumber
-  if (typeof blockNumber === "function") {
-    callback = blockNumber;
-    blockNumber = this.forkBlockNumber;
-    blockNumberProvided = false;
-  }
 
   key = utils.toBuffer(key);
 
-  // If the account/key doesn't exist in our state trie, get it off the wire.
-  this.keyExists(key, function(err, exists) {
+  this.getTouchedAt(key, function(err, touchedAt) {
     if (err) {
       return callback(err);
     }
 
-    if (exists) {
-      // I'm checking to see if a blockNumber is provided because the below
-      // logic breaks for things like nonce lookup, in which we should just
-      // use the root trie as is. I'm guessing there's a cleaner architecture
-      // that doesn't require such checks
-      if (blockNumberProvided) {
-        // this logic is heavily influenced by BlockchainDouble.prototype.getStorage
-        // but some adjustments were necessary due to the ForkedStorageTrieBase context
-        self.blockchain.getBlock(blockNumber, function(err, block) {
+    if (typeof touchedAt !== "undefined") {
+      MerklePatriciaTree.prototype.get.call(self, key, function(err, r) {
+        callback(err, r);
+      });
+    } else {
+      // If this is the main trie, get the whole account.
+      if (self.address == null) {
+        self.blockchain.fetchAccountFromFallback(key, self.forkBlockNumber, function(err, account) {
           if (err) {
             return callback(err);
           }
 
-          // Manipulate the state root in place to maintain checkpoints
-          const currentStateRoot = self.root;
-          self.root = block.header.stateRoot;
-
-          MerklePatriciaTree.prototype.get.call(self, utils.toBuffer(self.address), function(err, data) {
-            if (err != null) {
-              // Put the stateRoot back if there's an error
-              self.root = currentStateRoot;
-              return callback(err);
-            }
-
-            const account = new Account(data);
-
-            self.root = account.stateRoot;
-            MerklePatriciaTree.prototype.get.call(self, key, function(err, value) {
-              // Finally, put the stateRoot back for good
-              self.root = currentStateRoot;
-
-              if (err != null) {
-                return callback(err, value);
-              }
-
-              callback(null, value);
-            });
-          });
+          callback(null, account.serialize());
         });
       } else {
-        MerklePatriciaTree.prototype.get.call(self, key, function(err, r) {
-          callback(err, r);
-        });
-      }
-    } else {
-      self.keyIsDeleted(key, (err, deleted) => {
-        if (err) {
-          return callback(err);
-        }
-
-        if (deleted) {
-          // it was deleted. return nothing.
-          callback(null, Buffer.allocUnsafe(0));
-          return;
-        }
-
-        // If this is the main trie, get the whole account.
-        if (self.address == null) {
-          self.blockchain.fetchAccountFromFallback(key, blockNumber, function(err, account) {
-            if (err) {
-              return callback(err);
-            }
-
-            callback(null, account.serialize());
-          });
-        } else {
-          if (to.number(blockNumber) > to.number(self.forkBlockNumber)) {
-            blockNumber = self.forkBlockNumber;
-          }
-          self.web3.eth.getStorageAt(to.rpcDataHexString(self.address), to.rpcDataHexString(key), blockNumber, function(
-            err,
-            value
-          ) {
+        self.web3.eth.getStorageAt(
+          to.rpcDataHexString(self.address),
+          to.rpcDataHexString(key),
+          self.forkBlockNumber,
+          function(err, value) {
             if (err) {
               return callback(err);
             }
@@ -122,9 +60,9 @@ ForkedStorageBaseTrie.prototype.get = function(key, blockNumber, callback) {
             value = utils.rlp.encode(value);
 
             callback(null, value);
-          });
-        }
-      });
+          }
+        );
+      }
     }
   });
 };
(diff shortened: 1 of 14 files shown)


davidmurdoch:
Ganache currently accepts non-standard input as a "feature". For instance, you can pass in a Buffer in place of a hex address in many RPC methods and ganache will happily go about using it. The `toLowerCase()` introduced here changes that behavior.

This also causes other interesting changes, for example, passing something like `null`, or other non-string types, to `eth_getBalance` _currently_ returns "0x", but with these changes it returns an error.

Also note that in any case where `toLowerCase()` is used on a hex string that will be converted to a Buffer, the casing doesn't matter, as `Buffer.from("AA", "hex").equals(Buffer.from("aa", "hex")) === true` because node's `Buffer` treats hex as case-insensitive.

I'm leaning towards reverting the broad `toLowerCase()` changes here, and only changing what is necessary.

seesemichaelj:
The whole reason to doing the `toLowerCase` was I was under the impression `Buffer.from("AA", "hex").equals(Buffer.from("aa", "hex"))`  was `false`. This is because I didn't add `hex` as the radix, so it's pretty cool that Buffer handles that properly

I can do what's only necessary then which is at the function level rather than at the interface level and only do it for the forked storage/code/account functions

davidmurdoch:
> I can do what's only necessary then which is at the function level rather than at the interface level and only do it for the forked storage/code/account functions

Sounds good!

lib/forking/forked_storage_trie.js
```diff
@@ -12,119 +11,58 @@ inherits(ForkedStorageBaseTrie, BaseTrie);
...
-            blockNumber = self.forkBlockNumber;
-          }
-          self.web3.eth.getStorageAt(to.rpcDataHexString(self.address), to.rpcDataHexString(key), blockNumber, function(
-            err,
-            value
-          ) {
+        self.web3.eth.getStorageAt(
+          to.rpcDataHexString(self.address),
+          to.rpcDataHexString(key),
+          self.forkBlockNumber,
```
davidmurdoch, line 54: So I ran some hacked together tests to test a hunch I had about this `blockNumber` change. Here is the test:

```javascript
  it.only("gets values at specified blocks on the original change", async function() {
    const { send, accounts: [from], abi, web3: originalWeb3, instance: {_address: contractAddress}, provider: originalProvider } = forkedContext;
    const originalContract = new originalWeb3.eth.Contract(abi, contractAddress);

    const txParams = {from};
    
    await originalContract.methods.test().send(txParams);
    const initialValue = await originalContract.methods.value().call();
    await send("evm_mine", null);
    const {result: preForkBlockNumber} = await send("eth_blockNumber");

    // Fork the "original" chain _now_
    const { web3 } = await intializeTestProvider({
      fork: originalProvider,
      logger,
      seed: "forked provider"
    });

    const instance = new web3.eth.Contract(abi, contractAddress);

    // get the value as it was before we forked
    instance.defaultBlock = preForkBlockNumber;
    // there is a bug possibly unrelated to this PR that prevents this from
    // returning the wrong value (it crashes with a `pop of undefined` instead).
    const finalValue = await instance.methods.value().call();
    assert.equals(finalValue, initialValue);
  });
```

There is a problem with this test though... another bug seems to be preventing the final `await instance.methods.value().call();` from returning a value, due to a `'pop' of undefined` error.

The immediate issue I was investigating was the use of `self.forkBlockNumber` instead of `blockNumber`, but perhaps the issue is further up in the stack and we shouldn't be visiting this function anyway.

All that said, this change _might_ cause a regression, _but_ it also might not currently (i.e., `develop`) affect user code due to the other bug getting in the way. So, shipping it with a `// TODO` and a new Issue might not be the worst thing in the world. :sweat_smile:

davidmurdoch (changes requested):
I like the approach where the storage trie isn't directly block-number aware. Hopefully any issues I've highlighted won't necessitate bringing the `blockNumber` back here.

I had some questions and possible changes, but other than these, things are looking good!

lib/forking/forked_storage_trie.js
```diff
@@ -137,41 +75,59 @@ ForkedStorageBaseTrie.prototype.keyExists = function(key, callback) {
...
-      this._deleted.put(rpcKey, 0, () => {
-        originalPut.call(this, key, value, callback);
+
+  this._touched.get(rpcKey, (_, result) => {
+    if (typeof result === "undefined") {
+      // key doesnt exist
+      this.blockchain.data.blocks.last((err, lastBlock) => {
+        if (err) {
+          console.log(new Error("shouldn't happen"));
+          callback();
```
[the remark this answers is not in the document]
seesemichaelj: lack of due diligence to revert. i think i was temporarily curious on whether or not it could happen

lib/forking/forked_storage_trie.js
seesemichaelj: my fix seems to coincidentally remove the `'pop' of undefined` error 🤷

seesemichaelj:
@davidmurdoch jk, i still need to fix the `toLowerCase` thing

seesemichaelj:
alright @davidmurdoch, I think this is ready to review now

seesemichaelj:
@davidmurdoch since this one hasn't been finished merging, i pushed a small commit (and accompanying test) that fixes an issue where I also need to check if the key exists on the trie. I wouldn't quite say it was a regression for the case I was solving, but it could be a regression for some past-block lookups

seesemichaelj:
I'll check into why these 2 forking tests are now failing, but this should still be good to review

seesemichaelj:
alright @davidmurdoch, tests pass locally for me with the latest commit. i didnt need to rearchitect anything; I just changed `ForkedStorageBaseTrie.delete` to do a `put(key, 0)`. This should be ready to review now

WyseNynja:
Thank you!

seesemichaelj:
@WyseNynja I'll ping this thread when a release goes out with these fixes (probably a beta release)

seesemichaelj:
This was released in https://github.com/trufflesuite/ganache-core/releases/tag/v2.12.2-beta.0

mrice32:
@davidmurdoch is it possible that this PR re-broke https://github.com/trufflesuite/ganache-core/issues/571  and caused https://github.com/trufflesuite/ganache-core/issues/797.

mikeseese:
FWIW forking (especially in it's current state) is a beast, so I wouldn't be surprised if this PR caused more issues.

However, I don't think it rebroke #571 as I believe I implemented specific test scenarios to pinpoint the causes of that issue (and others with similar symptoms). It's likely a different issue with the same symptom. It's very possible that @0xTimepunk's issue is even another

PR #633, opened 2020-09-22
test: throw if test contracts fail compilation

seesemichaelj:
This a simple improvement that will throw if the test contracts fail compilation. Otherwise, the error gets squashed and you'll get some other "<ContractName> of undefined" error at line 44 when you try to read `result.contracts`.


davidmurdoch (approved):
A++ PR. Micaiah would be proud!

davidmurdoch:
Looks like CI is failing. Will merge if we can get some green 🙂

seesemichaelj:
alright tests are passing @davidmurdoch

PR #651, opened 2020-10-18
Add `removed` field to Log JSON

### LINKED ISSUE #336: TX Receipts and `getFilterChanges` responses should return a `removed` property for each log.
See: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getfilterchanges

iostat:
Fixes #336

This is a fairly trivial change, especially considering Ganache will never reorg and thus the `removed` property is moot (which reasonably explains its omission in the first place).

However, for those of us who care whether or not a log was reorged as part of their application logic, and aren't using JavaScript (i.e., care about well-formed JSON-RPC responses and can't use the truthiness of `log.removed === undefined`), this lack of `removed` renders Ganache unusable as a Web3 provider.

### DIFF
--- a/lib/utils/log.js
+++ b/lib/utils/log.js
@@ -31,7 +31,8 @@ Log.prototype.toJSON = function() {
     address: to.rpcDataHexString(this.address),
     data: to.rpcDataHexString(this.data),
     topics: this.topics,
-    type: "mined"
+    type: "mined",
+    removed: (this.removed || false)
   };
 };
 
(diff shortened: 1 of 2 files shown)


davidmurdoch:
Looks good. I'd like to see a simple test to check the value of the field before meeting though.

Aside: this field has also been added in a separate branch where we are rewriting ganache-core: https://github.com/trufflesuite/ganache-core/blob/4a081acfad686e2f94c49f6087df6bb9722d2d1d/src/chains/ethereum/src/things/blocklogs.ts#L137. We do plan on stimulating reorgs in the future.

iostat:
Hope that was a reasonable place to put the test. I didn't want to add an entire transaction flow just to test the presence of a field, and the headline of test suggested it's appropriate anyway.

Issue #650, opened 2020-10-16
The latest version can't be installed with node v12.19.0 and npm  v6.14.8


alcuadrado:
Hey @davidmurdoch,

I've seen multiple reports of npm failing to install ganache-core lately. Initially I couldn't reproduce it until I switched to node 12.

This is a pretty weird error, as running `npm init -y && npm i ganache-core` doesn't fail. But if you have a project with a `package.json` and `package-lock.json` that already have `ganache-core`, you get this error:

```
pato@pmbp /t/npmproject-3> npm i
npm ERR! code ENOLOCAL
npm ERR! Could not install from "node_modules/ganache-core/node_modules/web3-provider-engine/node_modules/eth-sig-util/ethereumjs-abi@git+https:/github.com/ethereumjs/ethereumjs-abi.git" as it does not contain a package.json file.
```

This may be a bug in npm/node, but you could avoid it by upgrading `web3-provider-engine`, as it doesn't use a git-based dependency anymore.

We haven't touched the `ethereumjs/ethereumjs-abi` repo in a long time, and it does contain a `package.json`. So I'm inclined to think that it's npm's fault.

davidmurdoch:
Hm, thanks for opening this. I'll look into it!

I couldn't figure out `web3-provider-engine`'s breaking changes regarding the Subscription provider, otherwise I would just update it.

davidmurdoch:
Do you have more information to reproduce? I've been unable to do so.

justinjmoses:
It's currently inside our synthetix bundle now (`v2.30.2-ovm`)

Try cloning and installing this: https://github.com/Synthetixio/synthetix-cli - you should see it pop up immediately.

I wonder if the usage of `npm-shrinkwrap.json` instead of a lockfile in ganache is a problem?

alcuadrado:
Something I noticed from the latest release is that the shrink wrap file has been generated with npm 7, maybe those are incompatible/trigger a bug in npm 6.

drptbl:
@davidmurdoch @alcuadrado I was able to fix this by removing:
`ethereumjs-abi@` prefix from:
`ethereumjs-abi@git+https://github.com/ethereumjs/ethereumjs-abi.git`
in `package-lock.json` (in your case it would be `npm-shrinkwrap.json`)
resulting in:
`git+https://github.com/ethereumjs/ethereumjs-abi.git`
everything seems to work now.

davidmurdoch:
[...]

PR #653, opened 2020-10-22
fix: ethereumjs-abi ENOLOCAL error

drptbl:
changes:
- fixed `ethereumjs-abi` ENOLOCAL error (#650)
- reverted workaround (#652)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -29581,7 +29581,7 @@
         },
         "ethereumjs-abi": {
           "version": "git+ssh://git@github.com/ethereumjs/ethereumjs-abi.git#1ce6a1d64235fabe2aaf827fd606def55693508f",
-          "from": "ethereumjs-abi@git+https://github.com/ethereumjs/ethereumjs-abi.git",
+          "from": "git+https://github.com/ethereumjs/ethereumjs-abi.git",
           "requires": {
             "bn.js": "^4.11.8",
             "ethereumjs-util": "^6.0.0"
--- a/package.json
+++ b/package.json
@@ -26,8 +26,7 @@
     "coverage": "nyc report --reporter=text-lcov | coveralls"
   },
   "bundleDependencies": [
-    "keccak",
-    "web3-provider-engine"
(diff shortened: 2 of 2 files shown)


drptbl:
Not sure why tests on CI has failed, seems like not related to my changes at all (probably flaky?).

davidmurdoch:
Thanks @drptbl! Because of the way npm treats shrinkwraps*, this PR won't actually fix the issue once the package is published.

So, what i'm going to do for today is create a release with your fix by manually editing the shrinkwrap before it published to npm. I hope to upstream a fix this week so ganache no longer relies on this git package.

Thanks for this PR, btw... it really is saving me some extra headache! I close this issue once I'm finished with this manual release.

* if we publish the `npm-shrinkwrap.json` as is, when a user runs `npm install ganache-core` npm will install all of ganache's `devDependencies`, despite them being marked as `"dev": true` in the npm-shrinkwrap.json. So what we do on pre-publish is `npm prune --production` and then run `npm shrinkwrap` to recreate the shrinkwrap with only production dependencies. This restores the buggy `ethereumjs-abi@git+https://github.com/ethereumjs/ethereumjs-abi.git` URL.

_note: I'm using npm v7 to maintain the shrinkwrap file due to another npm v6 bug that creates entries in the `npm-shrinkwrap.json` that omits package some urls... urls that npm will attempt to use on `npm install ganache-core`, which causes npm to just crash._

davidmurdoch:
Update: this npm v7 issue is making it tricky to publish: https://github.com/npm/cli/issues/2045

davidmurdoch:
I've published a beta and it looks like the fix works!

https://www.npmjs.com/package/ganache-core/v/2.13.1-beta.1

@drptbl, @alcuadrado do you mind installing this version out and letting me know if you have any issues?

drptbl:
@davidmurdoch Indeed issues is fixed. Just tested. Thanks ;).

davidmurdoch:
Fixed in 2.13.1 https://www.npmjs.com/package/ganache-core/v/2.13.1

PR #661, opened 2020-11-10
eth-sig-util@3.0.0

rekmarks:
`eth-sig-util@2.5.3` (currently used by `ganache-core`) accidentally shipped breaking changes due to messing up its exports. Ganache is not actually impacted due to its use of the package, but `2.5.3` is deprecated, and all future non-breaking changes will be shipped under the `3.0.0` major version.


rekmarks:
I have no idea why the shrinkwrap diff is so large. 🤔

davidmurdoch:
shrinkwrap diff is large because I used npm v7 to work around some other npm bug (https://github.com/trufflesuite/ganache-core/commit/4b3f588a7ded29ee07e27139222e9cb2a4c162c1)

davidmurdoch:
Oh, and thanks for this! I've got a couple more PRs I want to get in the next release, so this should go out with those soon!

davidmurdoch:
fixed by https://github.com/trufflesuite/ganache-core/pull/711

Issue #700, opened 2021-01-05
Separate flavor packages into separate options packages


seesemichaelj:
This issue rearchitects flavor implementations to be mini-monorepos to have "options" packages.

These options packages make it so that you can include the options packages without the actual core packages. This will help with future potential hot-loading of core packages

i.e.

```
/src
  /chains
    /ethereum
      /ethereum
        package.json
      /options
        package.json
    /filecoin
      /filecoin
        package.json
      /options
        package.json
```

mikeseese:
Resolved in #703 and #709

PR #704, opened 2021-01-11
Ensure proper usage of yargs + npm `create` script without using `--`

seesemichaelj:
This PR is part of #700

There was a reason for me adding the `--` requirement and your changes (reflected in #703) just steamrolled over those necessary changes without discussion. I am providing you 2 different options that I strongly encourage you picking and merging one. I also would appreciate PR reviews rather than just coding over what you want/don't want; that way we can have a discussion when you have a disagreement so I can explain why I made the decisions I did.

**This PR is option 1 (without using `--`, but using positional args)**

--

If you don't want to use the `--` that NPM requires you to use to pass `--flag` arguments (what yargs calls "options"), then you need to change the `create` script to use strictly positional arguments.

This is because without the `--`, NPM parses your `--location` and `--folder` as NPM flags instead of `create.ts` flags. It tries to parse, determines they do nothing, ignores them, and proceeds to pass in the other arguments.

image

See the top red circle where `--folder` and `--location` are actually dropped. Then further you can see how `location` actually parsed `iamfolder` due to guessing it's position from the command string

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,13 +57,19 @@ Runs all tests:
 
 - `npm test` (or the shorthand, `npm t`)
 
+## To create a new chain/flavor
+
+- `npm run create <name> chains`
+
+This will create a new folder at `src/chains/<name>` where `<name>` should be the flavor name (e.g. `ethereum`), which you then can [create packages under](#to-create-a-new-package).
+
 ## To create a new package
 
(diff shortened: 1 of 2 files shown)


davidmurdoch:
Ah, I see. I only ever tested `--location` followed by `--folder`.

seesemichaelj:
Closing in favor of #707

PR #701, opened 2021-01-05
chore: move chain packages to a subfolder

mikeseese:
This PR resolves #700:
- Moves `chains/<chain>` to `chains/<chain>/<chain>`
- Modifies `scripts/create.ts` to suppport subpackages of `chains/*`
- Moves related options to a subpackage `chains/<chain>/options` with NPM name `@ganache/<chain>-options`

PR #705, opened 2021-01-11
Ensure proper usage of yargs + npm `create` script with using `--`

seesemichaelj:
This PR is part of #700

See more details of why this PR is necessary in #704

**This PR is option 2 (using named, non-positional args, but requiring `--`**

--

Because of the reasons stated in #704, if you want to use non-positional/named arguments in an NPM script, you must use the `--` between the `npm run <script>` and the first flag.

This PR essentially re-introduces the changes I originally had in #701

### DIFF
--- a/scripts/create.ts
+++ b/scripts/create.ts
@@ -31,7 +31,7 @@ const chainLocations = getDirectories(join(__dirname, "../src/chains")).map(
 locations = locations.concat(chainLocations);
 const argv = yargs
   .command(
-    `$0 <name> --location [--folder]`,
+    `$0 <name> [options]`,
     `Create a new package in the given location with the provided name.`,
     yargs => {
       return yargs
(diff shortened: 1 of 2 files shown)


scripts/create.ts
```diff
@@ -40,28 +40,29 @@ const argv = yargs
...
+          demandOption: false
         });
     }
   )
   .demandCommand()
   .version(false)
   .help(false)
   .updateStrings({
-    "Positionals:": chalk.bold("Options"),
-    "Options:": ` `,
```
seesemichaelj, line 64: I'm not sure why you insisted on joining these two while still having positional and non-positional arguments (from the way the script was written, despite they were all being processed as positional anyways). Regardless, if you have two separate types of arguments, they should be separated to help out developers understand what is what. By joining these, you're removing contextual help/usage information for a little extra screen space. I encourage you keep this if you choose this PR option.

seesemichaelj:
Closing in favor of #707

PR #707, opened 2021-01-11
make create.ts use the original npm command's argv

davidmurdoch:


scripts/create.ts
```diff
@@ -29,37 +47,37 @@ const chainLocations = getDirectories(join(__dirname, "../src/chains")).map(
...
-        })
-        .option("folder", {
-          alias: "f",
-          describe: chalk`Optional override for the folder name for the package instead of using {dim <}name{dim >}.`
-        });
-    }
-  )
+
+const argv = yargs(getArgv())
+  .command(`${COMMAND_NAME} <name>`, "", yargs => {
```
seesemichaelj, line 52: Since we actually provide `.position("name")` I think this should actually be

```suggestion
  .command("$0", "", yargs => {
```

I also don't think we should have `create` there. That would be like running `npm run create create ...`

scripts/create.ts
```diff
@@ -29,37 +47,37 @@ const chainLocations = getDirectories(join(__dirname, "../src/chains")).map(
...
+        type: "string"
+      });
+  })
   .demandCommand()
   .version(false)
   .help(false)
   .updateStrings({
+    // a little hack just to join the "Positionals" section with the "Options" section, for brevity
     "Positionals:": chalk.bold("Options"),
     "Options:": ` `,
```
seesemichaelj, line 82: There's absolutely no reason to have brevity here, but you seem persistent about keeping it so this will be my last objection to mixing different types of arguments in the help message

scripts/create.ts
```diff
@@ -29,37 +47,37 @@ const chainLocations = getDirectories(join(__dirname, "../src/chains")).map(
...
+      )
+      .positional("name", {
+        // the spaces here are a hack to make this command description line up with the others in the help output
+        describe: `          The name for the new package.`,
+        type: "string",
+        demandOption: true
+      })
+      .option("location", {
+        alias: "l",
+        describe: `The location for the new package.`,
```
seesemichaelj, line 66:
```suggestion
        describe: `The location for the new package.`,
        type: "string",
```

scripts/create.ts
davidmurdoch: The `argv` array passed into to yargs, as parsed by (`getArgv`) is `["create"] ,"name", "--location" ...]`, so I'd need to update that to remove one more argument from the start of the parsed npm argv.

I put the command name here so it _isn't_ a catch all (`$0`), but I don't think it matters either way, really.

seesemichaelj: Oh, then it's fine as is

scripts/create.ts
davidmurdoch: Because of `choices: string[]` below, the resulting `yargs[...].argv.location` must be a string and is already cast to a TypeScript `string`. So adding the `type` here just makes the help text more verbose for little to no benefit (as it already lists out all valid values).

Is there another reason to add the type here?

seesemichaelj: No other reason, your explanation makes perfect sense 👍

seesemichaelj (approved):
LGTM, having that extra section to `CONTRIBUTING.md` would be nice to have. Either way, 👍

Issue #689, opened 2021-01-05
Add flavor abstraction to CLI


seesemichaelj:
This rearchitects `@ganache/cli` in the `next` branch to abstract away different flavors so that we can easily add new flavors.

mikeseese:
Resolved in #708

mikeseese:
Closing as this work is complete; just awaiting future code review

PR #708, opened 2021-01-11
feat: change CLI to dynamically generate args from flavor options

### LINKED ISSUE #690: Rename CLI options to align with provider/server options
The CLI has used snake-case (with inconsistencies) name styling for options where `@ganache/core` now consistently uses camel-case. For example, `chain.defaultBalance` instead of `default_balance_ether`

This issue changes the CLI options to be consistent with the provider/server options.

~@davidmurdoch do you want these to be back-compatible?~ This was answered on Zoom: yes

mikeseese:
This PR is part of #689 which changes `@ganache/ethereum-options` to include the necessary values/types to support the new CLI `src/args.ts` file which dynamically generates the associated `yargs.option()` calls

This PR also resolves #690


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "build": "lerna run build && shx cp -r ./src/packages/ganache/dist/web ./docs/assets/js/ganache",
-    "clean": "npx shx rm -rf node_modules && npx lerna clean -y && npm run tsc.clean",
+    "clean": "npm run tsc.clean && npx lerna clean -y && npx shx rm -rf node_modules",
```
mikeseese, line 11: I had to reverse these because the first command `npx shx rm -rf node_modules` removes a non-global installation of lerna, causing the following commands to fail

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -256,7 +256,7 @@ export default class Blockchain extends Emittery.Typed<
         let firstBlockTime: number;
         if (options.chain.time != null) {
           // If we were given a timestamp, use it instead of the `_currentTime`
-          const t = +options.chain.time;
+          const t = options.chain.time.getTime();
```
mikeseese, line 259: I changed this to `.getTime()` because I changed the type of `chain.time` to `Date` instead of `Date | number`. I'm not sure why there was a `| number`. It's not in either legacy documentation that a user can provide a UTC millisecond timestamp; both CLI and Core documentation states it should be a ISO 8601 string or a Date object (core only)

I also double checked, and the `ganache-core@2` does assume it's a Date:
- https://github.com/trufflesuite/ganache-core/blob/develop/lib/blockchain_double.js#L88-L90
- https://github.com/trufflesuite/ganache-core/blob/develop/lib/blockchain_double.js#L1251-L1255

package.json
davidmurdoch: lerna is run with `npx` here, so that shouldn't be a problem. There might be another issue with the way ganache is configured. hmmm

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: good catch

src/chains/ethereum/options/src/chain-options.ts
```diff
@@ -134,36 +136,66 @@ export type ChainConfig = {
...
   },
   hardfork: {
     normalize,
+    shortDescription: "Set the hardfork rules for the EVM.",
     default: () => "muirGlacier",
-    legacyName: "hardfork"
+    legacyName: "hardfork",
+    cliAliases: ["k", "hardfork"],
+    cliType: "string",
+    cliChoices: Object.keys(enumerate<Hardfork>())
```
davidmurdoch: nice.

src/chains/ethereum/options/src/chain-options.ts
```diff
@@ -134,36 +136,66 @@ export type ChainConfig = {
...
   networkId: {
     normalize,
+    shortDescription:
+      "The id of the network returned by the RPC method `net_version`.",
     default: () => Date.now(),
-    legacyName: "network_id"
+    defaultDescription:
+      "System time at process start or Network ID of forked blockchain if configured.",
+    legacyName: "network_id",
+    cliAliases: ["i", "networkId", "network_id"],
```
davidmurdoch: `"network_id"` shouldn't be a CLI alias, as it doesn't work for `ganache-cli@latest` as it is today.

package.json
mikeseese: All I know is running `npm run clean` would finally get to `npm run tsc.clean` and I would get `lerna not found`. I'm guessing this has been working for you because you have a globally installed version of lerna, which isn't a requirement of this repo.

After changing the order of this command, I can run `npm run clean` and choosing to not look into the issue further and call it done. I don't think this is a configuration issue, nor do I think I should have to reinstall lerna (because that's what npx will do if I don't have it) just to finish running the clean command lol

src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -71,17 +82,32 @@ const logger: Logger = { log: () => {} };
...
+    shortDescription: "Set to `true` to log EVM opcodes.",
     default: () => false,
-    legacyName: "debug"
+    legacyName: "debug",
+    cliType: "boolean"
   },
   logger: {
     normalize,
+    shortDescription:
+      " An object, like `console`, that implements a `log` function.",
```
davidmurdoch:
```suggestion
      "An object, like `console`, that implements a `log` function.",
```

src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -136,39 +141,80 @@ export type MinerConfig = {
...
-    default: () => Quantity.from(90_000)
+    shortDescription:
+      'Sets the _default_ transaction gas limit in WEI. Set to "estimate" to use an estimate (slows down transaction execution by 40%+).',
+    default: () => Quantity.from(90_000),
+    cliType: "string"
   },
   callGasLimit: {
     normalize: Quantity.from,
+    shortDescription:
+      "Sets the transaction gas limit in WEI for `eth_call` and eth_estimateGas` calls.",
```
davidmurdoch:
```suggestion
      "Sets the transaction gas limit in WEI for `eth_call` and `eth_estimateGas` calls.",
```

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -181,45 +190,106 @@ export type WalletConfig = {
...
+    legacyName: "accounts",
+    cliAliases: ["account"],
+    cliType: "array"
   },
   seed: {
     normalize,
-    default: () => randomAlphaNumericString(10, alea()),
-    legacyName: "seed"
+    shortDescription: "Seed to use to generate a mnemonic.",
+    default: config =>
```
davidmurdoch, line 236: I just realized that my `default(config)` implementation depends on the order the options are specified in _this_ Definitions object.

The users options will be parsed in the order they appear in `Object.keys(WalletOptions)`.

Which would result in `seed` being parsed before `deterministic`, ensuring deterministic is _never_ set to `true`.

I haven't tested, but if my hunch here is true, the easiest thing to do is to just move `deterministic` right before `seed` in this `WalletOptions` object. And slap a big ol' comment in here about how the ordering is required (and test ensuring the correct wallet options get applied would be a nice bonus :-) ).

There are other more "technically correct" ways of solving this, but all the ones I've thought of have their own downsides and limitations (like maintaining dependency trees or a double pass over the options).

package.json
davidmurdoch: I'm good with the change; it make sense... I just found it strange.


Only commenting this just in case we need to come back to this one day:

I don't have lerna installed globally, and `npm run clean` (`npx shx rm -rf node_modules && npx lerna clean -y && npm run tsc.clean`) looks to be working on my machine.


```console
david at david-kubuntu in ~/work/ganache-core on chore/separate-options-from-flavors-next [!$]
$ lerna
lerna: command not found

david at david-kubuntu in ~/work/ganache-core on chore/separate-options-from-flavors-next [!$]
$ npx shx rm -rf node_modules

david at david-kubuntu in ~/work/ganache-core on chore/separate-options-from-flavors-next [!$]
$ npm run clean

> root@ clean /home/david/work/ganache-core
> npx shx rm -rf node_modules && npx lerna clean -y && npm run tsc.clean

npx: installed 21 in 3.73s
npx: installed 766 in 34.004s
lerna notice cli v3.22.1
lerna info versioning independent
lerna info clean removing /home/david/work/ganache-core/src/chains/ethereum/ethereum/node_modules
lerna info clean removing /home/david/work/ganache-core/src/chains/ethereum/options/node_modules
lerna info clean removing /home/david/work/ganache-core/src/chains/ethereum/utils/node_modules
lerna info clean removing /home/david/work/ganache-core/src/chains/tezos/options/node_modules
lerna info clean removing /home/david/work/ganache-core/src/chains/tezos/tezos/node_modules
lerna info clean removing /home/david/work/ganache-core/src/packages/cli/node_modules
lerna info clean removing /home/david/work/ganache-core/src/packages/core/node_modules
lerna info clean removing /home/david/work/ganache-core/src/packages/flavors/node_modules
lerna info clean removing /home/david/work/ganache-core/src/packages/ganache/node_modules
lerna info clean removing /home/david/work/ganache-core/src/packages/options/node_modules
lerna info clean removing /home/david/work/ganache-core/src/packages/promise-queue/node_modules
lerna info clean removing /home/david/work/ganache-core/src/packages/utils/node_modules
lerna success clean finished

> root@ tsc.clean /home/david/work/ganache-core
> npx lerna exec -- npx shx rm -rf lib dist

npx: installed 766 in 31.316s
lerna notice cli v3.22.1
lerna info versioning independent
lerna info Executing command in 12 packages: "npx shx rm -rf lib dist"
npx: installed 21 in 3.513sact:shx: verb lock using /home/david/.npm/_locks/staging-0393b40e8a8e169d.lock for /home/david/.npm/_npx/936938/lib/node_modules/.staging
npx: installed 21 in 3.599sesh-package-json:shx: timing action:finalize Completed in 21ms
npx: installed 21 in 3.643s
npx: installed 21 in 3.54s
npx: installed 21 in 3.545s
npx: installed 21 in 3.464s
npx: installed 21 in 3.585s
npx: installed 21 in 3.268s
npx: installed 21 in 3.468s
npx: installed 21 in 3.529s
npx: installed 21 in 3.39s
npx: installed 21 in 3.337s
lerna success exec Executed command in 12 packages: "npx shx rm -rf lib dist"
```

davidmurdoch (commented):
This is just a partial review with some comments. I'll do another pass at it later.

src/chains/ethereum/options/src/chain-options.ts
mikeseese: fair, I partially was considering removing the gap of the confusion between core and cli. i'm fine with removing it though 👍

src/chains/ethereum/options/src/wallet-options.ts
mikeseese: Looking at the code in `packages/options/src/create.ts`, I concur with your hunch 👍 Thanks for the heads up!

mikeseese: I also manually tested and confirmed that the order does matter (your hunch is correct)

davidmurdoch (approved):

PR #709, opened 2021-01-11
chore: separate filecoin package into filecoin and filecoin-options

mikeseese:
This is part of #700 which implements the filecoin parts of #702


davidmurdoch (approved):
LGTM (didn't try compilng/running/building it)

PR #712, opened 2021-01-12
feat: add `Filecoin.ChainNotify` method and Filecoin support websockets and subscriptions

### LINKED ISSUE #692: Successfully use `truffle preserve` with CLI

mikeseese:
This PR is part of #694 (although most Lotus applications want to use WebSockets)

This supports the Chain Explorer feature in the https://github.com/filecoin-shipyard/filecoin-network-inspector example

image

---

This PR also resolves #692:

### DIFF
--- a/src/packages/core/src/servers/ws-server.ts
+++ b/src/packages/core/src/servers/ws-server.ts
@@ -52,7 +52,10 @@ export default class WebsocketServer {
         message: ArrayBuffer,
         isBinary: boolean
       ) => {
-        let payload: ReturnType<typeof connector.parse>;
+        // We have to use type any instead of ReturnType<typeof connector.parse>
+        // on `payload` because Typescript isn't smart enough to understand the
+        // ambiguity doesn't actually exist
+        let payload: any;
         const useBinary = autoBinary ? isBinary : (wsBinary as boolean);
         try {
           payload = connector.parse(Buffer.from(message));
--- a/src/chains/filecoin/filecoin/src/api.ts
+++ b/src/chains/filecoin/filecoin/src/api.ts
@@ -1,5 +1,5 @@
 //#region Imports
-import { types } from "@ganache/utils";
+import { types, Quantity, PromiEvent } from "@ganache/utils";
 import Blockchain from "./blockchain";
 import {
   StorageProposal,
@@ -7,19 +7,23 @@ import {
 } from "./things/storage-proposal";
 import { SerializedRootCID, RootCID } from "./things/root-cid";
 import { SerializedDeal } from "./things/deal";
-import { SerializedTipset } from "./things/tipset";
+import { SerializedTipset, Tipset } from "./things/tipset";
 import { SerializedAddress } from "./things/address";
 import { SerializedMiner } from "./things/miner";
 import {
   SerializedRetrievalOffer,
   RetrievalOffer
 } from "./things/retrieval-offer";
+import Emittery from "emittery";
+import { HeadChange, HeadChangeType } from "./things/head-change";
 
 const _blockchain = Symbol("blockchain");
 
 export default class FilecoinApi implements types.Api {
   readonly [index: string]: (...args: any) => Promise<any>;
 
+  readonly #getId = (id => () => Quantity.from(++id))(0);
+  readonly #subscriptions = new Map<string, Emittery.UnsubscribeFn>();
   private readonly [_blockchain]: Blockchain;
 
   constructor(blockchain: Blockchain) {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -16,7 +16,8 @@
     "reinstall": "npm run clean && npm install",
     "test": "lerna exec -- npm run test",
     "tsc": "lerna exec -- npm run tsc",
-    "tsc.clean": "npx lerna exec -- npx shx rm -rf lib dist"
+    "tsc.clean": "npx lerna exec -- npx shx rm -rf lib dist",
+    "start": "lerna exec --scope @ganache/cli -- node --inspect lib/src/cli.js"
```
davidmurdoch:
```suggestion

```

Let's remove this, since it will be in conflict with #721.

src/packages/core/src/options/server-options.ts
```diff
@@ -63,20 +63,24 @@ export const ServerOptions: Definitions<ServerConfig> = {
     normalize,
     shortDescription: "Enable a websocket server.",
     default: () => true,
-    legacyName: "ws"
+    legacyName: "ws",
+    cliType: "boolean",
+    cliAliases: []
```
davidmurdoch: The default for aliases after merging #721 will be `[]` (https://github.com/trufflesuite/ganache-core/pull/721/files#diff-2b2281ab4da879ef24d594d2dc8cbee929624f9b3e1438f88d20694aa0351490R94). So I think it is safe to remove this.

src/packages/core/src/options/server-options.ts
```diff
@@ -63,20 +63,24 @@ export const ServerOptions: Definitions<ServerConfig> = {
...
   },
   wsBinary: {
     normalize,
     shortDescription:
       "Whether or not websockets should response with binary data (ArrayBuffers) or strings.",
-    default: () => "auto"
+    default: () => "auto",
+    cliChoices: [true, false, "auto"] as any[]
```
davidmurdoch: :+1:

src/packages/core/src/options/server-options.ts
```diff
@@ -63,20 +63,24 @@ export const ServerOptions: Definitions<ServerConfig> = {
...
   },
   rpcEndpoint: {
     normalize,
     shortDescription:
       "Defines the endpoint route the HTTP and WebSocket servers will listen on.",
     default: () => "/",
-    defaultDescription: "'/rpc/v0' for Filecoin, '/' otherwise"
+    defaultDescription: "'/rpc/v0' for Filecoin, '/' otherwise",
```
davidmurdoch: Sorry, I must have missed where this `rpcEndpoint` was added previously. I guess there isn't really a better place to put it? :-/

If not, let's update default to take the `config`. IIRC, it should have a `flavor` property on it, so we can actually generate the right default via https://github.com/trufflesuite/ganache-core/blob/next/src/packages/options/src/create.ts#L45

If it doesn't have a flavor property... UI guess we need to look into how hard it is to add it in.

src/packages/core/src/servers/ws-server.ts
```diff
@@ -52,7 +52,10 @@ export default class WebsocketServer {
         message: ArrayBuffer,
         isBinary: boolean
       ) => {
-        let payload: ReturnType<typeof connector.parse>;
+        // We have to use type any instead of ReturnType<typeof connector.parse>
+        // on `payload` because Typescript isn't smart enough to understand the
+        // ambiguity doesn't actually exist
+        let payload: any;
```
davidmurdoch, line 58: :cry: 

Is the ambiguity caused by just having multiple chains? Or is it specific to the Filecoin API implementation?

davidmurdoch (changes requested):

src/packages/core/src/servers/ws-server.ts
mikeseese: It's ambiguity caused by having multiple chains, not specific to Filecoin. It's likely that typescript is smart enough with some more convoluted types lol, but I didn't take the appropriate time to learn how to do the complex types yall have set up, because the current types do not support it. I personally don't think it's worth the time for this small functionality

src/packages/core/src/options/server-options.ts
mikeseese: > Sorry, I must have missed where this rpcEndpoint was added previously.

And by no part of your fault! Apparently I missed making a PR where it's implemented that's supposed to be upstream of this branch. I think it was ambiguous until the `next` CLI implementation settled down since rebasing was an active target

> I guess there isn't really a better place to put it? :-/

I think this is the right place for it, but there is potential down the road that it would be moved out due to potential need to support more routes than just the main RPC endpoint. For now, I think this is a good place for it.

> If not, let's update default to take the config. IIRC, it should have a flavor property on it, so we can actually generate the right default via next/src/packages/options/src/create.ts#L45
> 
> If it doesn't have a flavor property... UI guess we need to look into how hard it is to add it in.

Ah! I'll look into this

package.json
mikeseese:
```suggestion
    "tsc.clean": "npx lerna exec -- npx shx rm -rf lib dist"
```

mikeseese: applied in another commit due to your suggestion leaving a trailing comma

src/packages/core/src/options/server-options.ts
mikeseese: Ok, the work for this is now in PR #726, which includes the preface to adding `rpcEndpoint` and the correction to use `config.flavor` for the default

mikeseese:

PR #713, opened 2021-01-14
fix: change `Filecoin.WalletBalance` to have correct denominations

mikeseese:
This PR is part of #694:
- Fixes the correct smallest denomination of FIL to be attoFIL instead of zeptoFIL
- I also went ahead and changed the base type of `Balance` to `bigint` instead of `string` and added serialization support

image


davidmurdoch (approved):
LGTM

mikeseese:

PR #714, opened 2021-01-14
feat: get `Filecoin.ClientRetrieve` to actually save a file

mikeseese:
This PR is part of #694. After uploading a file under `/market` in the network inspector example, there is an option (that is commented out in the example code) to download the file. `Filecoin.ClientRetrieve` is supposed to save the file to a location on disk provided by the user in the `ref` argument. This PR adds the functionality of using `ref` and saving the file to disk. This is an initial implementation and will likely need to be adjusted later as use cases expand.


src/chains/filecoin/filecoin/src/ipfs-server.ts
```diff
@@ -73,13 +43,29 @@ class IPFSServer {
...
+        Discovery: {
+          MDNS: {
+            Enabled: false
+          },
+          webRTCStar: {
+            Enabled: false
+          }
+        },
+        // API isn't in the types, but it's used by ipfs-http-server
+        // @ts-ignore
```
davidmurdoch, line 56: :+1:

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -188,11 +191,39 @@ export default class Blockchain extends Emittery.Typed<
   }
 
   private async getIPFSObjectSize(cid: string): Promise<number> {
-    let stat = await this.ipfsServer.node.object.stat(cid, {
+    let stat = await this.ipfsServer.node.object.stat(cid as any, {
```
davidmurdoch: I don't think `any` is needed here.

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -188,11 +191,39 @@ export default class Blockchain extends Emittery.Typed<
...
+  private async downloadFile(cid: string, ref: FileRef): Promise<void> {
+    const size = await this.getIPFSObjectSize(cid);
+    const content = new Uint8Array(size);
+    const chunks = this.ipfsServer.node.files.read(new IPFS_CID(cid), {
+      timeout: 500 // Enforce a timeout; otherwise will hang if CID not found
+    });
+
+    let index = 0;
+    for await (const chunk of chunks) {
+      content.set(chunk, index);
```
davidmurdoch: I'm a little worried that a large file could clobber the heap. Should we update this to pipe each chunk directly to a file?

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -188,11 +191,39 @@ export default class Blockchain extends Emittery.Typed<
...
+      index += chunk.byteLength;
+    }
+
+    const dirname = path.dirname(ref.path);
+    try {
+      if (!fs.existsSync(dirname)) {
+        await fs.promises.mkdir(dirname, { recursive: true });
+      }
+      await fs.promises.writeFile(ref.path, content.slice(0, index), "binary");
```
davidmurdoch: Disregard this comment if you refactor this to pipe the chunks directly to disk...why do we need `content.slice(0, index)`, here?

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -214,8 +245,8 @@ export default class Blockchain extends Emittery.Typed<
       state: DealState.Validating, // Not sure if this is right, but we'll start here
       message: "",
       provider: this.miner,
-      pieceCid: null,
-      size: size,
+      pieceCid: proposal.data.pieceCid,
+      size: proposal.data.pieceSize || size,
```
davidmurdoch, line 249: It looks like the `let size = await this.getIPFSObjectSize(proposal.data.root["/"].value);` line isn't necessary if `proposal.data.pieceSize` is set. Can you adjust so we don't fetch the `size` if we don't need it?

davidmurdoch (changes requested):
Looks like some mighty fine work here! :superhero_man: 

Left a few comments/requests.

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: Sure!

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: I will write directly to disk, but FWIW, the slice here is because we allocated more RAM in `content` than we needed. since `ipfs.object.stat().CumulativeSize` includes IPFS overhead of the DAG node storage. For example a 5 byte text file actually requires 13 bytes on IPFS

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: `proposal.data.pieceSize` is an optional parameter and I have seen examples that even have provided it to be `0` instead of undefined to indicate "the miner should calculate it". I can change the type to have `pieceSize?` (which may be a rabbit hole with the current way these types work), but we'd still need to fetch the size.

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: good catch; it isn't. it used to be with the older version of `ipfs` that we had

mikeseese:

PR #715, opened 2021-01-14
feat: add support for `Filecoin.StateMinerPower` method

mikeseese:
This PR is part of #694. It makes the necessary changes to support the `/miners` page on the network inspector example

image


mikeseese:

Issue #693, opened 2021-01-05
Successfully use Jim Pick's js-lotus-client-workshop with CLI


seesemichaelj:
https://github.com/filecoin-shipyard/js-lotus-client-workshop

PR #717, opened 2021-01-15
fix: update storage DealState enum from filecoin source

mikeseese:
This PR is part of #693. It updates the `DealState` enum for storage deals and adds more comments/context about how/where to update the enum in the future

### DIFF
--- a/src/chains/filecoin/filecoin/src/deal-state.ts
+++ b/src/chains/filecoin/filecoin/src/deal-state.ts
@@ -1,46 +1,127 @@
 // Note that the runtime number values of these enums match
-// *exactly* the number values of the same states on the Lotus server.
+// *exactly* the number values of the same states found at
+// go-fil-markets/storagemarket/dealstatus.go (https://git.io/JtJAS)
+// **DO NOT TRUST** other sources (i.e. lotus/lotuspond/front/src/Client.js
+// or js-lotus-client-workshop/src/08-deals/deal-list.js)
 // Don't reorganize unless you know what you're doing.
 
+// Updated to v1.1.1 (https://github.com/filecoin-project/go-fil-markets/blob/v1.1.1/storagemarket/dealstatus.go)
 export enum DealState {
+  // Unknown means the current status of a deal is undefined
   Unknown,
+
+  // ProposalNotFound is a status returned in responses when the deal itself cannot
+  // be located
   ProposalNotFound,
+
+  // ProposalRejected is returned by a StorageProvider when it chooses not to accept
+  // a DealProposal
   ProposalRejected,
+
+  // ProposalAccepted indicates an intent to accept a storage deal proposal
   ProposalAccepted,
+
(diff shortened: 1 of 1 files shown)


src/chains/filecoin/filecoin/src/deal-state.ts
```diff
@@ -1,46 +1,127 @@
 // Note that the runtime number values of these enums match
-// *exactly* the number values of the same states on the Lotus server.
+// *exactly* the number values of the same states found at
+// go-fil-markets/storagemarket/dealstatus.go (https://git.io/JtJAS)
+// **DO NOT TRUST** other sources (i.e. lotus/lotuspond/front/src/Client.js
+// or js-lotus-client-workshop/src/08-deals/deal-list.js)
 // Don't reorganize unless you know what you're doing.
```
davidmurdoch, line 6: Since this is less of an `enum` and more of a mapping, should we just convert this to a map?

Or at least explicitly type the integer values instead of relying on position? .e.g,

```
export enum DealState {
  // Unknown means the current status of a deal is undefined
  Unknown = 0,
  ```
  
This is just a question, not a request.

File: src/chains/filecoin/filecoin/src/deal-state.ts [hunk 2]
```diff
@@ -1,46 +1,127 @@
...
 
+// Updated to v1.1.1 (https://github.com/filecoin-project/go-fil-markets/blob/v1.1.1/storagemarket/dealstatus.go)
 export enum DealState {
+  // Unknown means the current status of a deal is undefined
```
davidmurdoch, line 10: If you do keep these as an enum instead of a map (see above comment), you can use JSDoc-style comments to have intellisense pick up these descriptions for us. e.g.,

image

instead of:

image

davidmurdoch (changes requested):
Really appreciate that you added all `DealState` descriptions in here! I love inline docs! :-)

No changes requested, just some suggestions.

src/chains/filecoin/filecoin/src/deal-state.ts
mikeseese: > Since this is less of an enum and more of a mapping, should we just convert this to a map?

This definitely is an enum rather than a map; I personally can't think of why this would be considered a map. It's implemented in golang as an ordered enum and it's logical purpose is an enum.

> Or at least explicitly type the integer values instead of relying on position? .e.g,

The thing is, that this code is essentially a direct port from the golang implementation. They don't use numbers in their implementation, so I favored not using them here to be consistent with their implementation.

src/chains/filecoin/filecoin/src/deal-state.ts
mikeseese: Would you prefer that we have the JSDoc, requiring further manipulation from copying this from the source or  leave it as is to be as close/consistent with the golang source?

mikeseese:

PR #719, opened 2021-01-15
chore: update ipfs server to have a more generic options setup

mikeseese:
This PR updates `IPFSServer` to properly consume options and use the host option. It also renames the directory used for IPFS files to not have the word `foo`.


mikeseese:

PR #721, opened 2021-01-15
chore: polish cli

davidmurdoch:


src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -190,46 +191,38 @@ export type WalletConfig = {
...
-        } else {
-          return rawInput as OptionsAccount[];
-        }
-      } else {
-        return [];
-      }
-    },
-    shortDescription:
-      "Array of Accounts. Each object should have a balance key with a hexadecimal value.",
+    normalize,
```
davidmurdoch, line 206: Because `Ganache.provider({wallet: {accounts: ["privatekey,initialbalance"]}})` should be *invalid* input I've separated out the logic that parses this cli input into its own method, `cliCoerce`.

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -190,46 +191,38 @@ export type WalletConfig = {
...
+    cliDescription: `Account data in the form \`<private_key>,<initial_balance>\`, can be specified multiple times. Note that private keys are 64 characters long and must be entered as an 0x-prefixed hex string. Balance can either be input as an integer, or as a 0x-prefixed hex string with either form specifying the initial balance in wei.`,
```
davidmurdoch, line 207: I'm actually not entirely sure about this super-verbose cli documentation now. :-/

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -190,46 +191,38 @@ export type WalletConfig = {
...
     legacyName: "accounts",
     cliAliases: ["account"],
-    cliType: "array"
+    cliType: "array:string",
```
davidmurdoch, line 210: Though somewhat undocumented, yargs can "understand" that a type can be an array of "type". So that's what this is doing (well... the string is parsed into `array: true, type: string` later, maybe just `cliIsArray: true, cliType: "string"` would be better?)

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,150 @@
...
-  FgRed: "\x1b[31m",
-  FgYellow: "\x1b[33m"
-};
+const wrapWidth = Math.min(120, yargs.terminalWidth());
+const NEED_HELP = "Need more help? Reach out to the Truffle community at";
+const COMMUNITY_LINK = "https://gitter.im/ConsenSys/truffle";
+
+const highlight = (t: string) =>
+  marked.parseInline(t).replace(/&lt;/g, "<").replace(/&gt;/g, ">");
+const center = (t: string) => " ".repeat((wrapWidth - t.length) >> 1) + t;
```
davidmurdoch: If it wasn't clear, this function just produces text that's been-left padded so it is center-aligned in the terminal window.

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,150 @@
...
-  Reset: "\x1b[0m",
-  FgRed: "\x1b[31m",
-  FgYellow: "\x1b[33m"
-};
+const wrapWidth = Math.min(120, yargs.terminalWidth());
+const NEED_HELP = "Need more help? Reach out to the Truffle community at";
+const COMMUNITY_LINK = "https://gitter.im/ConsenSys/truffle";
+
+const highlight = (t: string) =>
+  marked.parseInline(t).replace(/&lt;/g, "<").replace(/&gt;/g, ">");
```
davidmurdoch: `marked-terminal`, the rendered `marked` is using for terminal output, is a bit rough around the edges, as it still tries to HTML encode chars like `<` and `>`, despite being purpose-built for terminal output :facepalm:.

scripts/create.ts
```diff
@@ -109,7 +108,7 @@ process.stdout.write(`${COLORS.Reset}`);
   }
 
   // determines how many `../` are needed for package contents
-  const numDirectoriesAwayFromRoot = 3 + location.split(sep).length;
+  const numDirectoriesAwayFromRoot = 2 + location.split(sep).length;
```
davidmurdoch, line 111: not clue what changed to make this start pointing to the wrong directory. Maybe i'm crazy?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -175,10 +175,10 @@ export default class EthereumApi implements types.Api {
 
     const { chain } = options;
     const { initialAccounts } = (this.#wallet = wallet);
-    const coinbaseAddress = parseCoinbaseAddress(
+    const coinbaseAddress = (options.miner.coinbase = parseCoinbaseAddress(
```
davidmurdoch: `options.miner.coinbase` would still be a number, if that a number is what was passed in to the option, if we didn't fix it up here.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -553,7 +553,7 @@ export default class Blockchain extends Emittery.Typed<
     const genesis = new RuntimeBlock(
       rawBlockNumber,
       Quantity.from(BUFFER_32_ZERO),
-      this.#options.miner.coinbase as Address,
+      this.coinbase,
```
davidmurdoch, line 556: since we don't yet support updating any options after ganache has started, i'm just using the instance's `coinbase` here, just so it's obvious later.

src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -168,53 +169,36 @@ export const MinerOptions: Definitions<MinerConfig> = {
...
-          if (Number.isSafeInteger(index)) {
-            return index;
-          } else {
-            throw new Error(
-              `Invalid value for option miner.coinbase: ${rawType}`
-            );
-          }
-        }
-      }
+      return typeof rawType === "number" ? rawType : Address.from(rawType);
```
davidmurdoch, line 194: see comment a few lines down for why this function was reverted.

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
-      default: {
+      // since "ethereum" is the DefaultFlavor we don't need a `case` for it
+      case DefaultFlavor:
+        command = ["$0", flavor];
         defaultPort = 8545;
         break;
-      }
+      default:
+        command = flavor;
+        defaultPort = 8545;
```
davidmurdoch, line 135: I know this behavior differs from the previous. But this default case is should be impossible to hit anyway. /shrug

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
-          const group = `${category.charAt(0).toUpperCase()}${category.substr(
+          type GroupType = `${Capitalize<typeof category>}:`;
+          const group = `${category[0].toUpperCase()}${category.slice(
             1
-          )}:`;
-          const categoryObj = flavorDefaults[category] as Definitions<any>;
-          const options = Object.keys(categoryObj);
-          for (const option of options) {
+          )}:` as GroupType;
+          const categoryObj = (flavorDefaults[
```
davidmurdoch, line 148: I tried so hard to get this type to be what I wanted, but ultimately had to change it. I used `Definitions<Base.Config>` instead of `Definitions<any>` so the rest the types below would work.

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
             const optionObj = categoryObj[option];
             if (optionObj.disableInCLI !== true) {
-              const useAliases = typeof optionObj.cliAliases !== "undefined";
-              const alias = useAliases
-                ? optionObj.cliAliases
-                : (optionObj as any).legacyName;
+              let alias = optionObj.cliAliases;
 
-              let description = optionObj.shortDescription;
+              let description = highlight(optionObj.cliDescription || "");
```
davidmurdoch: Many of our descriptions were using markdown syntax... so why not render them as markeddown-ish in the terminal!!

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
               if (alias) {
-                description = `${description}\n${COLORS.Bold}${
-                  COLORS.FgYellow
-                }Deprecated aliases: ${
-                  Array.isArray(alias)
-                    ? alias.filter(a => a.length > 1).join(", ")
-                    : alias
-                }${COLORS.Reset}\n`;
+                description = chalk`${description}${EOL}{dim deprecated aliases: ${alias
```
davidmurdoch: I liked the deprecated alias being less noticeable, instead of bright bold yellow...fight me if you care! :-)

davidmurdoch: Oh, and I dropped the manual `COLORS` approach since marked-terminal uses `chalk` already

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
-                  defaultValue = [defaultValue];
-                }
-              }
+              const generateDefaultDescription = () => {
+                // default sometimes requires a config, so we set supply one
+                return (state[option] = optionObj.default
+                  ? optionObj.default(state).toString()
+                  : undefined);
+              };
+              const defaultDescription =
```
davidmurdoch: If we provide a defaultDescription, we use that, otherwise we use the CLI's default value.

We do NOT supply cli with an actual `default`, as the options fill be filled with their defaults by each provider anyway. It was causing some issues regarding coercing values via `normalize`, when `normalize` was set up to only deal with JS values (not the CLI types).

e.g., `"privateKey,balance"` shouldn't be allowed in a call to `Ganache.provider`, but is the only way for CLI to receive the value. There is an optional `cliCoerce` function now, that converts from the `cliType` to the `rawType` (the type `normalize` understands).

I hope it works. ha.

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
+                optionObj.defaultDescription || generateDefaultDescription();
 
-              flavorArgs = flavorArgs.option(`${category}.${option}`, {
+              // we need to specify the type of each array so yargs properly casts
+              // the types held within each array
+              const { cliType } = optionObj;
+              const array = cliType && cliType.startsWith("array:"); // e.g. array:string or array:number
+              const type = (array
```
davidmurdoch: takes `array:string` and turns into into `array: true, type: "string"`. Or just `"array": true, 
type: "array"` if we want yargs to cast args that looks like numbers to numbers for automatically.

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
+              };
+              flavorArgs = flavorArgs.option(`${category}.${option}`, options);
+
+              // create *hidden* deprecated aliases:
+              const aliasOptions = { hidden: true, ...options };
+              aliasOptions.alias = [`${category}.${option}`];
+              alias.forEach(a => {
+                // don't make hidden aliases for single-letter aliases
+                if (a.length === 1) return;
+                flavorArgs = flavorArgs.option(a, aliasOptions);
```
davidmurdoch: I ended up hiding the aliases after all.

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
               });
             }
           }
         }
 
+        // TODO: combine these cli options with core's `ServerOptions`
```
davidmurdoch: I noticed you noticed server's ServerOptions... I didn't feel like figuring that out yet.

src/packages/options/src/base.ts
```diff
@@ -1,3 +1,30 @@
+import { NoUnion } from "./types";
+type PrimitiveCliTypes = string | number | boolean;
+export type CliTypes =
```
davidmurdoch, line 3: :-D We'll fix these weird plural names later.

src/packages/options/src/definition.ts
```diff
@@ -23,26 +26,29 @@ export type InternalConfig<
...
+    readonly cliDescription: string;
     readonly disableInCLI?: boolean;
     readonly cliAliases?: string[];
-    readonly cliType?: "string" | "number" | "boolean" | "array";
     readonly cliChoices?: string[] | number[];
-  } & (void extends OptionHasDefault<C, N>
-    ? {}
+  } & (void extends OptionHasCliType<C, N>
+    ? {
+        readonly cliType?: CliTypeMap<CliTypes> | null;
```
davidmurdoch, line 38: If the `Definition` type does NOT have a `cliType`, allow all valid CliType(s) strings.

Otherwise require the specific string that matches it.

patches/ts-transformer-inline-file+0.1.1.patch
```diff
@@ -0,0 +1,47 @@
+diff --git a/node_modules/ts-transformer-inline-file/transformer.js b/node_modules/ts-transformer-inline-file/transformer.js
```
davidmurdoch, line 1: Ok, so webpacking without building first caused problems because `ts-loader` can't handle project references correctly and I had to disable the typeChecker for the final bundle. But then our transforms stopped working...because they rely on types.

By the time I figure that out, well... I had it figured out. So here is that file.

I'll probably just switch to requiring `tsc` to be run before `webpack` in the near future (which will allow us to use the super fast `esbuild`, since we should  just be compiling JS at that point). But I am committing this for now since I need to move on to other things.

scripts/link-ts-references.ts
```diff
@@ -129,7 +130,7 @@ function saveConfigs(configs: PackageInfo[]) {
   configs.forEach(({ modified, path, tsConfig }) => {
     if (modified) {
       const tsConfigFile = join(path, "tsconfig.json");
-      writeFileSync(tsConfigFile, JSON.stringify(tsConfig, null, 2));
+      writeFileSync(tsConfigFile, JSON5.stringify(tsConfig, null, 2));
```
davidmurdoch, line 134: This `link-ts-references.ts` script was previously clobbering comments in `tsconfig.json` files. This JSON5 library fixes it :mage_man:. (RIP old comments)

src/chains/tezos/options/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "../../../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "lib"
+    "outDir": "lib",
+    "composite": true
```
davidmurdoch, line 5: eventually this will be used in other packages... so let's include `composite: true` now, since it will be required later.

davidmurdoch (commented):
@seesemichaelj do you care to review my review?

src/chains/ethereum/options/src/wallet-options.ts
davidmurdoch: It looks like this:

image


I guess it isn't too bad.

patches/ts-transformer-inline-file+0.1.1.patch
mikeseese: > I'll probably just switch to requiring tsc to be run before webpack in the near future (which will allow us to use the super fast esbuild, since we should just be compiling JS at that point).

Ya, this is an obvious path forward here. No reason why we shouldn't be webpacking off of the `tsc` output. Since webpack is rarely called, I don't know why you wouldn't run `npm run tsc` first.

The current implementation you have is fine for now, but it's definitely prone for errors and harder to grok. But again, that's ok

> But I am committing this for now since I need to move on to other things.

👍

scripts/create.ts
mikeseese: The original logic broke when you changed my regex to detect number of slashes to using `.split()` which increased the `<location-expr>.length` to have 1 more, requiring you to subtract 1 from the constant. e.g. original algorithm had `ethereum` => 0 and `ethereum/ethereum` => 1 and the current one has `ethereum` => 1 and `ethereum/ethereum` => 2

Originally, the algorithm was `3 + number of slashes`, but you changed the algorithm to `3 + number of directories after "src"`, so it makes sense that it needs to be `2` instead

src/chains/ethereum/ethereum/src/api.ts
mikeseese: `options` should be immutable at this point since you're using `EthereumInternalOptions`. This is a hack (which is a slippery slope on a good design). I'm fine with this, but I don't think the design of this system was hoping for this.

Further, if you're going to continue with this approach, if you're going to set `options.miner.coinbase`, there's no reason to have a separate `coinbaseAddress` variable

src/chains/ethereum/options/src/index.ts
```diff
@@ -16,6 +16,7 @@ import {
   Options,
   OptionsConfig
 } from "@ganache/options";
+import { UnionToIntersection } from "@ganache/options/src/types";
```
mikeseese: you should stick to importing only from `@ganache/options` and make sure to export anything you need externally through the `main` export

src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -123,6 +123,7 @@ export type MinerConfig = {
       rawType: string | number;
       type: Address | number;
       hasDefault: true;
+      cliType: string | number;
```
mikeseese: You should omit `cliType` if you want multiple types or hide the type from the help message. It doesn't really help us to put it here since `rawType` is what captures the intent of what `miner.coinbase`, as it stands in this PR, expresses

src/packages/options/src/definition.ts
```diff
@@ -23,26 +26,29 @@ export type InternalConfig<
...
-  k: infer I
-) => void
-  ? I
-  : never;
-
 export type Definitions<C extends Base.Config> = {
   [N in OptionName<C>]: {
     readonly normalize: Normalize<C, N>;
-    readonly shortDescription: string;
+    readonly cliDescription: string;
```
mikeseese, line 32: FWIW, I used `short` instead of `cli` because that's more generic and supports other potential uses of a short description that is not specific to the CLI interface, but whatevs

src/packages/options/src/base.ts
mikeseese: why not fix them now? now is a perfect time to easily and quickly fix them before you merge the PR

i'll happily push a commit if you don't want to take the time to do it now?

src/chains/ethereum/options/src/wallet-options.ts
mikeseese: Let the users complain before you remove it. It's fine. It's extremely helpful to not have to go find the github page to find the documentation. verbose help is always more helpful. If you want the strings to be shorter, write them in a more concise manner that still gets the point across.

Users should almost always be able to know how to use the CLI from the CLI help usage. It infuriates me when a CLI usage forces me to use `man` to get the basic rundown of an argument.

src/chains/ethereum/options/src/wallet-options.ts
mikeseese: I think this syntax is better than separate variables. It's more concise and intuitive once you get the drift. I think separate variables can cause someone to somehow only read `cliType` and not think it's an array

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -115,6 +115,7 @@ export type WalletConfig = {
          */
         unlocked_accounts: Array<string | number>;
       };
+      cliType: (string | number)[];
```
mikeseese: This is kind of weird. yargs will just give us a `string[]` since you only specified `cliType: "array"` and we'll have to parse strings into numbers regardless. I think this should be `string[]`

src/packages/cli/src/args.ts
davidmurdoch: I refactored this since I left this comment.

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
-    const commandAliases = flavor === DefaultFlavor ? ["$0", flavor] : flavor;
+  let flavor: keyof typeof DefaultOptionsByName;
+  for (flavor in DefaultOptionsByName) {
+    const flavorDefaults = DefaultOptionsByName[flavor];
+    let command: Command;
     let defaultPort: number;
     switch (flavor) {
-      case EthereumFlavorName:
-      default: {
+      // since "ethereum" is the DefaultFlavor we don't need a `case` for it
```
mikeseese, line 128: sure, i dont think `DefaultFlavor` will ever not be `EthereumFlavorName`, but when you'll have:

``` typescript
switch (flavor) {
  case FilecoinFlavorName:
    ...
  case TezosFlavorName:
    ...
  case DefaultFlavor:
    ...
}
```

It just is easier to grok if it was `EthereumFlavorName`. but i don't really care either way

EDIT: but now I see why `DefaultFlavor` makes sense, since you're assigning the default command to it. 🤷

src/packages/cli/src/args.ts
mikeseese: Ya I was also having issues trying to get the type to work, but sounds good

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
-                  defaultValue = [];
-                } else {
-                  defaultValue = [defaultValue];
-                }
-              }
+              const generateDefaultDescription = () => {
+                // default sometimes requires a config, so we set supply one
+                return (state[option] = optionObj.default
+                  ? optionObj.default(state).toString()
+                  : undefined);
```
mikeseese: neat workaround to me just ignoring the ones that required a config!

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
+                optionObj.defaultDescription || generateDefaultDescription();
 
-              flavorArgs = flavorArgs.option(`${category}.${option}`, {
+              // we need to specify the type of each array so yargs properly casts
+              // the types held within each array
+              const { cliType } = optionObj;
+              const array = cliType && cliType.startsWith("array:"); // e.g. array:string or array:number
+              const type = (array
+                ? cliType.slice(6) // remove the "array:" part
+                : cliType) as YargsPrimitiveCliTypeStrings;
```
mikeseese: does yargs handle `null` appropriately for `type` (aka treats it as it wasn't provided)? i'd probably do something like `: (cliType || undefined)` to force `type` to be `undefined` if it was `null`

src/packages/cli/src/args.ts
```diff
@@ -1,130 +1,168 @@
...
+  renderer: new TerminalRenderer({
+    codespan: chalk.hex(TruffleColors.porsche),
+    // Disable `unescape` since doesn't work for everything (we just do it ourselves)
+    unescape: false
+  })
+});
+
+const wrapWidth = Math.min(120, yargs.terminalWidth());
+const NEED_HELP = "Need more help? Reach out to the Truffle community at";
+const COMMUNITY_LINK = "https://gitter.im/ConsenSys/truffle";
```
mikeseese, line 26: yall aren't pushing for spectrum usage??? :(

src/packages/cli/src/args.ts
mikeseese: it looks better this way!

src/packages/cli/src/args.ts
mikeseese: I actually implemented supporting those somewhere in a filecoin PR which I can bring out into a PR against next

mikeseese (changes requested):

src/packages/options/src/base.ts
davidmurdoch: I probably do this awkward plural thing a lot, so i was thinking about fixing all instances of the weirdness.

scripts/create.ts
davidmurdoch: I only changed it because the previous version didn't support Window's path separators. I probably should have tested my change before committing. :ha: :grimacing:

src/chains/ethereum/options/src/index.ts
davidmurdoch: Gah, totally missed this auto import! Good catch!

src/chains/ethereum/options/src/miner-options.ts
davidmurdoch: A though: Yargs supports multiple types, but only if you don't specify a type... which makes me wonder if it will treat `--option true` as a string or a boolean.

Also, agreed! This type is redundant.

src/packages/cli/src/args.ts
davidmurdoch: I monitor gitter sometimes, and there are 5916 people in gitter, and only 306 in spectrum. ¯\\_(ツ)_/¯

src/packages/cli/src/args.ts
davidmurdoch: It does handle it. It _only_ checks if type is a valid types (boolean, string, number, etc), `null,` `undefined`, "this is a shoe", etc, would be ignored.

src/packages/cli/src/args.ts
```diff
@@ -177,5 +180,22 @@ export default function (version: string, isDocker: boolean) {
...
+  for (let key in parsedArgs) {
+    const parts = key.split(".");
+    // only copy namespaced keys
+    if (parts.length > 1) {
+      const last = parts.pop();
+      const obj = parts.reduce((acc: any, part: string) => {
+        // build the obj path (e..g, `finalArgs: {wallet: {}}`)
+        return (acc[part] = acc[part] || {});
+      }, finalArgs);
+      obj[last] = parsedArgs[key];
```
mikeseese: FWIW, this is a bit non-intuitive

I would have written something like

```suggestion
      const [ optionCategory, optionName ] = parts;
      if (!finalArgs[optionsCategory]) {
        finalArgs[optionsCategory] = {};
      }
      finalArgs[optionsCategory][optionName] = parsedArgs[key];
```

Up to you how you want to portray this logic

davidmurdoch: Mine was just supporting a possible future where there are more than one "." in a key name. A contrived example: `"server.ipfs.port"` would create `server:{ ipfs: {port: 1234} }`.

So, the logic between the two is a little different; yours only takes into account what is possible with our current options... which is probably okay, since tons of other internals would have to be rewritten to support that future.

So...I'll switch to yours. :-) Commit incoming!

src/packages/colors/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "lib",
+    "composite": true
+  },
+  "include": ["src"]
```
mikeseese: this is missing the index ts file and is not compiling it

```suggestion
  "include": ["src", "index.ts"]
```

src/packages/cli/src/args.ts
davidmurdoch: updated: https://github.com/trufflesuite/ganache-core/pull/721/files#diff-2b2281ab4da879ef24d594d2dc8cbee929624f9b3e1438f88d20694aa0351490R184-R200

src/packages/options/src/definition.ts
davidmurdoch: I figured, but since some of these descriptions are specific to CLI usage now (like `wallet.accounts`) I just wanted to make it explicit.

src/chains/ethereum/options/src/wallet-options.ts
davidmurdoch: TIL: a yargs option with `type: "array"` causes it to try to coerce args to a number, but if that fails, it will return a string. So `0x132456789...ABCDEF` results in a giant number! (i've fixed this in the latest code)

src/packages/cli/src/args.ts
```diff
@@ -33,7 +33,13 @@ function unescapeEntities(html: string) {
     .replace(/&#39;/g, "'");
 }
 const highlight = (t: string) => unescapeEntities(marked.parseInline(t));
-const center = (t: string) => " ".repeat((wrapWidth - t.length) >> 1) + t;
+const center = (str: string) =>
+  wrapWidth ? " ".repeat((wrapWidth - str.length) >> 1) + str : str;
```
mikeseese: This won't work if `str.length > wrapWidth`. This probably should be

```suggestion
const center = (str: string) => " ".repeat(Math.max(0, (wrapWidth - str.length) >> 1)) + str;
```

mikeseese (changes requested):

src/packages/cli/src/args.ts
davidmurdoch: good catch. fixed in [`17196cd` (#721)](https://github.com/trufflesuite/ganache-core/pull/721/commits/17196cdab81f971476dcdbaa84bee79975b7309a)

got rid of the bitshift math since this isn't a 1 liner anymore

mikeseese (approved):
LGTM

PR #722, opened 2021-01-15
fix: update Filecoin interface types

### LINKED ISSUE #720: Ensure Filecoin types are compliant with the current spec
While working on the Meme Marketplace investigation, I found some types that were wrong in the `BlockHeader` (we call it `Block`).  This issue is to scrub our current list of types and ensure they match the specifications

mikeseese:
This PR resolves #720 by updating all of the Filecoin API types to their correct names and added missing values


mikeseese:

PR #723, opened 2021-01-16
feat: add server options to cli and default logging

mikeseese:
This PR adds the `@ganache/core` `server` options to the CLI. It also changes the default logger to be stdout as stated in the JSDoc

### DIFF
--- a/src/packages/cli/src/args.ts
+++ b/src/packages/cli/src/args.ts
@@ -6,6 +6,7 @@ import {
   Definitions,
   YargsPrimitiveCliTypeStrings
 } from "@ganache/options";
+import { serverDefaults } from "@ganache/core";
 import { Command, Argv } from "./types";
 import chalk from "chalk";
 import { EOL } from "os";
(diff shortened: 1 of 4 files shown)


src/packages/cli/src/args.ts
```diff
@@ -86,64 +151,20 @@ export default function (version: string, isDocker: boolean) {
...
-              flavorArgs = flavorArgs.option(key, options);
-            }
+            processOption(state, category, group, option, optionObj, flavorArgs);
           }
         }
 
-        // TODO: combine these cli options with core's `ServerOptions`
+        const state = {};
+        const categoryObj = serverDefaults.server;
+        const options = Object.keys(categoryObj);
```
davidmurdoch: No need to get all the keys first since `categoryObj` is already iterable via `for(const option in categoryObj) {`. I'm not too worried about inherited properties showing up (which `Object.keys` omits).

davidmurdoch (changes requested):
Just one little change requested. Thanks!

davidmurdoch (approved):

PR #725, opened 2021-01-18
fix: ignore folders that are packages from another branch

mikeseese:
If you checkout a branch that doesn't include a newly introduced package from your last branch (which you had ran `npm run tsc`), the `link-ts-references.ts` script erroneously picks it up as a folder without the necessary package files. This PR ignores folders that don't have a `package.json`


davidmurdoch (approved):
LGTM!

PR #726, opened 2021-01-18
feat: add filecoin to dynamic CLI

mikeseese:
This PR is the filecoin-specific changes to support the CLI. This PR is part of #689

### DIFF
--- a/src/chains/filecoin/filecoin/src/provider.ts
+++ b/src/chains/filecoin/filecoin/src/provider.ts
@@ -7,8 +7,10 @@ import { Schema } from "@filecoin-shipyard/lotus-client-schema";
 import Blockchain from "./blockchain";
 import {
   FilecoinOptionsConfig,
-  FilecoinProviderOptions
+  FilecoinProviderOptions,
+  FilecoinInternalOptions
 } from "@ganache/filecoin-options";
+import cloneDeep from "lodash.clonedeep";
 
 // Meant to mimic this provider:
 // https://github.com/filecoin-shipyard/js-lotus-client-provider-browser
@@ -16,7 +18,7 @@ export default class FilecoinProvider
   extends Emittery.Typed<undefined, "ready">
   // Do I actually need this? `types.Provider` doesn't actually define anything behavior
   implements types.Provider<FilecoinApi> {
-  #options: FilecoinProviderOptions;
+  #options: FilecoinInternalOptions;
   #api: FilecoinApi;
   #executor: utils.Executor;
 
@@ -42,6 +44,29 @@ export default class FilecoinProvider
     });
 
     this.#api = new FilecoinApi(this.blockchain);
+  }
+
+  /**
+   * Returns the options, including defaults and generated, used to start Ganache.
+   */
+  public getOptions() {
+    return cloneDeep(this.#options);
+  }
+
+  /**
+   * Returns the unlocked accounts
+   */
+  public getInitialAccounts() {
+    const accounts: Record<
+      string,
+      { unlocked: boolean; secretKey: string; balance: number }
+    > = {};
+    accounts[this.blockchain.address.serialize()] = {
+      unlocked: true,
+      secretKey: this.blockchain.address.privateKey,
+      balance: this.blockchain.balance.toFIL()
+    };
+    return accounts;
   }
 
   async connect() {
--- a/src/chains/filecoin/filecoin/src/provider.ts
+++ b/src/chains/filecoin/filecoin/src/provider.ts
@@ -59,12 +59,12 @@ export default class FilecoinProvider
   public getInitialAccounts() {
     const accounts: Record<
       string,
-      { unlocked: boolean; secretKey: string; balance: number }
+      { unlocked: boolean; secretKey: string; balance: bigint }
     > = {};
     accounts[this.blockchain.address.serialize()] = {
       unlocked: true,
       secretKey: this.blockchain.address.privateKey,
-      balance: this.blockchain.balance.toFIL()
+      balance: BigInt(this.blockchain.balance.value)
     };
     return accounts;
   }
--- a/src/chains/filecoin/options/src/miner-options.ts
+++ b/src/chains/filecoin/options/src/miner-options.ts
@@ -18,6 +18,12 @@ export type MinerConfig = {
       hasDefault: true;
     };
 
+    /**
+     * Internal flag to determine if ganache should automine/instamine.
+     * Is set/overridden by the miner.blockTime option.
+     *
+     * @default true
+     */
     automining: {
       type: boolean;
       rawType: boolean;
@@ -29,10 +35,16 @@ export type MinerConfig = {
 export const MinerOptions: Definitions<MinerConfig> = {
   blockTime: {
     normalize,
-    default: () => 0
+    shortDescription:
+      'Sets the `blockTime` in seconds for automatic mining. A blockTime of `0`  enables "instamine mode", where new executable transactions will be mined instantly.',
+    default: () => 0,
+    cliType: "number"
   },
   automining: {
     normalize,
+    shortDescription:
+      "Internal flag to determine if ganache should automine/instamine.",
+    disableInCLI: true,
     default: () => true
   }
 };
--- a/src/chains/filecoin/options/src/miner-options.ts
+++ b/src/chains/filecoin/options/src/miner-options.ts
@@ -35,14 +35,14 @@ export type MinerConfig = {
 export const MinerOptions: Definitions<MinerConfig> = {
   blockTime: {
     normalize,
-    shortDescription:
+    cliDescription:
       'Sets the `blockTime` in seconds for automatic mining. A blockTime of `0`  enables "instamine mode", where new executable transactions will be mined instantly.',
     default: () => 0,
     cliType: "number"
   },
   automining: {
     normalize,
-    shortDescription:
+    cliDescription:
       "Internal flag to determine if ganache should automine/instamine.",
     disableInCLI: true,
     default: () => true
--- a/src/chains/filecoin/options/src/miner-options.ts
+++ b/src/chains/filecoin/options/src/miner-options.ts
@@ -17,34 +17,21 @@ export type MinerConfig = {
       type: number;
       hasDefault: true;
     };
-
-    /**
-     * Internal flag to determine if ganache should automine/instamine.
-     * Is set/overridden by the miner.blockTime option.
-     *
-     * @default true
-     */
-    automining: {
-      type: boolean;
-      rawType: boolean;
-      hasDefault: true;
-    };
   };
 };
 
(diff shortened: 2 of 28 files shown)


src/chains/filecoin/options/src/miner-options.ts
```diff
@@ -29,10 +35,16 @@ export type MinerConfig = {
...
+    cliDescription:
+      'Sets the `blockTime` in seconds for automatic mining. A blockTime of `0`  enables "instamine mode", where new executable transactions will be mined instantly.',
+    default: () => 0,
+    cliType: "number"
   },
   automining: {
     normalize,
+    cliDescription:
+      "Internal flag to determine if ganache should automine/instamine.",
+    disableInCLI: true,
```
davidmurdoch: Since `automining` is the same as `blockTime: 0`, it is not exposed to the cli, and we don't have anything like this on the ethereum side should we:

 * copy it to Ethereum (and expose it to the CLI) since it is more explicit (probably renamed to `automine`)
 * remove it from Filecoin, since it is redundant
 * leave it as is
 
 ?
 
 Thoughts?

src/chains/filecoin/options/src/wallet-options.ts
```diff
@@ -36,8 +44,23 @@ export type WalletConfig = {
 };
 
 export const WalletOptions: Definitions<WalletConfig> = {
```
davidmurdoch, line 47: Now that #732 is merged in, we need to add 

```
   exclusiveGroups: [
    ["deterministic", "seed"]
  ];
```

to the `WalletConfig`

as well as fill in the `conflicts` arrays on `WalletOptions` (example: https://github.com/trufflesuite/ganache-core/pull/732/files#diff-b465f04541f0c5e2ad842f50433df5af81c1bdd98b7921c558dbd30d635ca25dR247)

src/chains/filecoin/options/src/wallet-options.ts
```diff
@@ -36,8 +44,23 @@ export type WalletConfig = {
...
+    cliType: "boolean"
+  },
   seed: {
     normalize,
-    default: () => randomAlphaNumericString(10, alea())
+    cliDescription: "Seed to use to generate a mnemonic.",
+    // The order of the options matter here! `wallet.deterministic`
+    // needs to be prior to `wallet.seed` for `config.deterministic`
+    // below to be set correctly
+    default: config =>
```
davidmurdoch, line 62: Ethereum options uses `-s` as a shorthand for `--wallet.seed`, is there any reason not to copy that over to Filecoin?

src/packages/cli/package.json
```diff
@@ -55,6 +55,7 @@
     "@ganache/colors": "^0.1.0",
     "@ganache/ethereum": "0.1.0",
     "@ganache/ethereum-options": "0.1.0",
+    "@ganache/filecoin": "0.1.0",
```
davidmurdoch, line 58: Since we don't want filecoin shipping with cli (and ganache) I think this should be a `peerDependency`, and the version should _probably_ be `"^0.1.0"`. This gets into the "fancy downloader" convo, so maybe it's time we have that.

Perhaps we should add the download into `@ganache/core`'s `connector-loader.ts`? I figure `@ganache/core` would also need to detect that a flavor is _valid_ but not installed; of course, when used programmatically it wouldn't provide an interactive installation, but it could still perform the same error logic.

Now that I've typed all that... for now we could just show a helpful error message if the flavor dependency isn't already installed. e.g., 'The "filecoin" flavor of ganache is not installed. Install it by running `npm install @ganache/filecoin`' (obviously it's a little more complex if we want to determine global installation, package manager used, etc)

image

src/chains/filecoin/filecoin/src/provider.ts
```diff
@@ -44,6 +46,29 @@ export default class FilecoinProvider
...
+   */
+  public getInitialAccounts() {
+    const accounts: Record<
+      string,
+      { unlocked: boolean; secretKey: string; balance: number }
+    > = {};
+    accounts[this.blockchain.address.serialize()] = {
+      unlocked: true,
+      secretKey: this.blockchain.address.privateKey,
+      balance: this.blockchain.balance.toFIL()
```
davidmurdoch: Could `toFIL`'s return value overflow `MAX_SAFE_INTEGER`? Or is `FIL` not the minimum denomination? Ethereum ganache returns WEI; if filecoin has a similar denomination paradigm I think it returns the smallest unit.

src/packages/cli/src/initialize/filecoin.ts
```diff
@@ -0,0 +1,42 @@
...
+  addresses.forEach(function (address, index) {
+    console.log(`(${index}) ${accounts[address].secretKey}`);
+  });
+
+  console.log("");
+  console.log(
+    `Lotus API started on ${serverSettings.host}:${serverSettings.port}`
+  );
+  console.log(
+    `IPFS  API started on ${liveOptions.chain.ipfsHost}:${liveOptions.chain.ipfsPort}`
```
davidmurdoch: How about `{thing} RPC listening on ${tech.host}:${tech.port}`?

src/packages/core/src/options/server-options.ts
```diff
@@ -49,5 +60,20 @@ export const ServerOptions: Definitions<ServerConfig> = {
...
+    default: config => {
+      switch (config.flavor) {
+        case FilecoinFlavorName:
+          return "/rpc/v0";
+        case DefaultFlavor:
+        default:
+          return "/";
+      }
+    },
+    defaultDescription: "'/rpc/v0' for Filecoin, '/' otherwise"
```
davidmurdoch:
```suggestion
    defaultDescription: '"/" (Ethereum), "/rpc/v0" (Filecoin)'
```

Reason: I like to put ethereum things first. Also, shortening it because tezos's endpoint will be added in the future.

src/packages/core/src/servers/http-server.ts
```diff
@@ -84,13 +85,21 @@ function sendResponse(
...
+    app: TemplatedApp,
+    connector: Connector,
+    options: HttpServerOptions
+  ) {
     this.#connector = connector;
 
     // JSON-RPC routes...
-    app.post("/", this.#handlePost).options("/", this.#handleOptions);
+    app
+      .post(options.rpcEndpoint || "/", this.#handlePost)
```
davidmurdoch: If `rpcEndpoint` must have a value at this point.

I _think_ `ServerOptions` just need to implement and export the equiv of `EthereumInternalOptions`, if it doesn't already, which represents what gets returned after normalization.

See https://github.com/trufflesuite/ganache-core/blob/8a7ba780a664e32b2210e32c3bfebe359e7cfc37/src/chains/ethereum/options/src/index.ts#L51-L53

(or, if this and ws-server.ts are the only place that needs it you could probably get away with some version of `export type HttpServerOptions = Required<Pick<ServerOptions["server"], "rpcEndpoint">>`.

src/packages/core/src/servers/ws-server.ts
```diff
@@ -32,7 +35,7 @@ export default class WebsocketServer {
     const connections = this.#connections;
     const wsBinary = options.wsBinary;
     const autoBinary = wsBinary === "auto";
-    app.ws("/", {
+    app.ws(options.rpcEndpoint || "/", {
```
davidmurdoch: Same here RE: `rpcEndpoint` should already have a value.

src/packages/flavors/package.json
```diff
@@ -43,6 +43,8 @@
   "dependencies": {
     "@ganache/ethereum": "^0.1.0",
     "@ganache/ethereum-options": "0.1.0",
+    "@ganache/filecoin": "0.1.0",
```
davidmurdoch, line 46: This should also probably be a miner-semver `peerDependency` here as well (`tezos` technically needs to be moved as well)

src/packages/options/src/create.ts
```diff
@@ -42,7 +43,10 @@ function fill(defaults: any, options: any, target: any, namespace: any) {
         if (value !== undefined) {
           config[key] = propDefinition.normalize(value);
         } else if (hasOwn(propDefinition, "default")) {
-          config[key] = propDefinition.default(config);
+          config[key] = propDefinition.default({
```
davidmurdoch: A clever solution! Instead of destructuring the entire config for each option (makes this fn `O(n log n)`, I think?, albeit n is very small) can we have the `default` function take a second parameter for `flavor: string` instead?

(I initially thought it'd be nice to have it take in a `rootConfig`... but we'd be in circular dependency hell if we tried to get valid typings for that!)

davidmurdoch (changes requested):

src/chains/filecoin/options/src/miner-options.ts
mikeseese: I think we should remove it from Filecoin. I understand what Tim was going for, but I did cringe when seeing the redundant option. I'll remove it!

src/chains/filecoin/options/src/wallet-options.ts
mikeseese: No reason not to! I'll add it

src/packages/cli/package.json
mikeseese: > This gets into the "fancy downloader" convo, so maybe it's time we have that.

Because of this, I'm favoring that we leave this as is *until* we're actually considering merging this into `develop`. That way I can continue working without any fancy stuff. We can determine the delivery mechanism and how I need to support that later. I don't think we should work any of this part until we're closer to merging this into `develop` (probably early March)

> and the version should probably be `"^0.1.0"`

When should versions be pinned and when shouldn't they be? This just did whatever the default `lerna add` behavior was, but I am trying to pin everything new I add due to [the contribution guidelines](https://github.com/trufflesuite/ganache-core/blob/develop/CONTRIBUTING.md#L164)

src/chains/filecoin/filecoin/src/provider.ts
mikeseese: `FIL` is not the minimum denomination, `attoFIL` is (10^-18). I can change this to return the minimum denomination

src/packages/flavors/package.json
mikeseese: You can move `tezos` on `develop` if you'd like. My response from https://github.com/trufflesuite/ganache-core/pull/726#discussion_r561306779 applies here too

mikeseese:

src/packages/cli/package.json
mikeseese: FYI I went ahead and added initial support for this as it's necessary for some of my future PRs to be acceptable for you (otherwise I'd have to turn off all tests with node versions less than 12 rather than just filecoin tests)

I added the logic in `@ganache/flavors`. I initially was thinking of putting it in `connector-loader.ts` like you, but something felt better about putting it in `flavors`. We can definitely discuss more whenever we review that work

PR #738, opened 2021-01-19
chore: add "push" action to "build all" on `develop`

mikeseese:
This PR adds build jobs for the remaining parts of the build matrix for `push` events to `develop`

Here is an example of what it would look like: https://github.com/trufflesuite/ganache-core/actions/runs/497178739


mikeseese:
@davidmurdoch This is ready for review now

davidmurdoch (approved):

PR #755, opened 2021-01-26
feat: add support for `debug_storageRangeAt` RPC

### LINKED ISSUE #682: implement `debug_storageRangeAt`
Internal discussion: https://trufflesuite.slack.com/archives/CB2BNGNDP/p1608264304095500

eshaben:
Resolves #682


eshaben:
Need to clean up the code (copy pasta from `debug_traceTransaction`), clean up/add typings, add annotations for documentation in api, and probably more... but the logic is there, this is close.

eshaben:
Hahaha... what my last comment said but hopefully for real this time

eshaben:
Not the prettiest way of implementing this rpc method but it gets the job done 🤷 

Basically we are storing the raw, un-hashed version of storage keys along with their hashed keys so we can return both hashed and raw keys to the user for the `debug_storageRangeAt` method. Then after we replay the transaction, we use the trie to create a read stream and get all of our storage keys from the transaction. Then we sort the keys before we can filter through the ones we need, using the hashed key to get the raw key from the db, and finally return the desired range of storage data.

Findings: geth stores their hashed keys along with the raw, un-hashed keys. geth also sorts their keys after getting them out of the database (see here: https://gist.github.com/eshaben/ae99036dc15a3784ed3ea50856cd733d -- this is the ordering of keys returned from the trie read stream) 🤷 

This is ready for initial review!

If anyone wants to test against geth, I used the following command to start up a geth console: `geth --rpc --rpccorsdomain="https://remix.ethereum.org" --rpcapi web3,eth,debug,personal,net --vmdebug --datadir test-chain-dir --dev console`
☝️ this command requires you create a directory called `test-chain-dir` for storing chaindata. 

Then I went to remix and deployed my contracts and sent transactions. After sending transactions I went back to geth console and tried to use `debug_storageRangeAt` method, like so...
```
debug.storageRangeAt(
"0x29a4daa5e2a33843a5967e7f6014f95862772fa6077afeddeebc9b85e966dbda", // blockHash
 0, // txIndex
"0xe33983B0ab2E5987e6b88E0a860E859061F08C68", // contractAddress
"0x405787fa12a823e0f2b7631cc41b3ba8828b3321ca811111fa75cd3aa3bb5ace", // startKey
 2 // maxResults
);
```

If you have any questions or want help configuring test environment for this, let me know!

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1846,6 +1846,35 @@ export default class EthereumApi implements types.Api {
...
+   * return storage data given a starting key and max number of entries to return.
+   *
+   * @param blockHash
+   * @param txIndex
+   * @param contractAddress
+   * @param startKey
+   * @param maxResult
+   * @returns returns a storage object with the keys being keccak-256 hashes of the storage keys,
+   * and the values being the raw, unhashed key and value for that specific storage slot. Also
+   * retuns a next key which is the keccak-256 hash of the next key in storage for continuous downloading.
```
```suggestion
   * returns a next key which is the keccak-256 hash of the next key in storage for continuous downloading.
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -365,6 +370,11 @@ export default class Blockchain extends Emittery.Typed<
         );
       });
 
+      // save storage keys to the database one at a time
```
```suggestion
      // save storage keys to the database
```

Because this loop happens within the `batch` callback, teeeecchhhhnically they'll all get saved at the same time :-)

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1121,6 +1132,152 @@ export default class Blockchain extends Emittery.Typed<
...
+   * Returns a contract's storage given a starting key and max number of
+   * entries to return.
+   *
+   * Strategy:
+   *
+   *  1. Get transaction hash
+   *  2. Replay transaction
+   *  3. Create read stream to get the storage keys from the transaction
+   *  4. Sort storage keys
+   *  5. Filter, assemble, and send storage data for the given range back
```
davidmurdoch: :heart:

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1121,6 +1132,152 @@ export default class Blockchain extends Emittery.Typed<
...
+    // replay the transaction
+    try {
+      await this.traceTransaction(transactionHash, {
+        disableMemory: true,
+        disableStack: false,
+        disableStorage: false
+      });
+    } catch (e) {
+      throw new Error(e);
+    }
```
davidmurdoch: `traceTransaction` shouldn't (:sweat_smile:) have any side effects, so this bit can probably be removed.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1846,6 +1846,35 @@ export default class EthereumApi implements types.Api {
...
 
+  /**
+   * Attempts to replay the transaction as it was executed on the network and
+   * return storage data given a starting key and max number of entries to return.
+   *
+   * @param blockHash
+   * @param txIndex
+   * @param contractAddress
+   * @param startKey
+   * @param maxResult
```
davidmurdoch: Can you add descriptions for each of these parameters?

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -193,6 +204,10 @@ export default class Miner extends Emittery.Typed<
           options.extraData,
           []
         );
+        const finalizedBlockData = {
+          ...serializedBlockData,
+          storageKeys
+        };
```
davidmurdoch: Let's just modify `runtimeBlock.finalize` to take in `storageKeys` as a param, and then have it output `{block: Block, serialized: Buffer, storageKeys: Map<Buffer, Buffer>}`.

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -209,6 +224,41 @@ export default class Miner extends Emittery.Typed<
...
+      ) => {
+        const stack: ITraceData[] = [];
+        for (const stackItem of event.stack) {
+          stack.push(TraceData.from(stackItem.toArrayLike(Buffer)));
+        }
+
+        if (event.opcode.name === "SSTORE") {
+          const key = stack[stack.length - 1];
+          const hashedKey = keccak(key.toBuffer());
+          storageKeys.set(hashedKey, key.toBuffer());
```
davidmurdoch, line 236: Because Map's keys don't compare Buffer data, but instead they compare the Buffer _instance_, you could end up with multiple keys that represent the same data. The key probably needs to be converted to a string here.

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -209,6 +224,41 @@ export default class Miner extends Emittery.Typed<
...
+        event: StepEvent,
+        next: (error?: any, cb?: any) => void
+      ) => {
+        const stack: ITraceData[] = [];
+        for (const stackItem of event.stack) {
+          stack.push(TraceData.from(stackItem.toArrayLike(Buffer)));
+        }
+
+        if (event.opcode.name === "SSTORE") {
+          const key = stack[stack.length - 1];
```
davidmurdoch: It looks like we only every need the last item from the stack. If `event.stack` is an actually just an array we can skip the `for` loop and do `const key = event.stack[event.stack.length - 1].toArrayLike(Buffer)`

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -25,13 +25,14 @@ import { promisify } from "util";
...
 import Manager from "./data-managers/manager";
-import { encode as rlpEncode } from "rlp";
+import { encode as rlpEncode, decode as rlpDecode } from "rlp";
 import Common from "ethereumjs-common";
 import VM from "ethereumjs-vm";
 import BlockLogManager from "./data-managers/blocklog-manager";
 import { EVMResult } from "ethereumjs-vm/dist/evm/evm";
 import { VmError, ERROR } from "ethereumjs-vm/dist/exceptions";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
+import { keccak } from "@ganache/utils/src/utils/keccak";
```
davidmurdoch: VS Code like to automatically write these imports, but webpack seems to choke on them. To make it work add `keccak` to the destructuring of `utils` on line 37, i.e., 

```
const {
  BUFFER_EMPTY,
  RPCQUANTITY_EMPTY,
  BUFFER_32_ZERO,
  BUFFER_256_ZERO,
  RPCQUANTITY_ZERO,
  keccak
} = utils;
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1121,6 +1132,446 @@ export default class Blockchain extends Emittery.Typed<
...
+        keys,
+        {
+          disableMemory: true,
+          disableStack: false,
+          disableStorage: false
+        }
+      );
+      result.storage = storage;
+    } catch (e) {
+      throw new Error(e);
```
davidmurdoch: No need for this try/catch

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -0,0 +1,487 @@
...
+          }
+        };
+
+        assert.deepStrictEqual(result.storage, storage);
+        assert.strictEqual(result.nextKey, null);
+      });
+
+      it("should return account doesn't exist error when debugging a deployment transaction", async () => {
+        try {
+          await provider.send("debug_storageRangeAt", [
```
davidmurdoch: Use `assert.rejects` instead of try/catch here

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -0,0 +1,487 @@
...
+            value:
+              "0x0000000000000000000000000000000000000000000000000000000000000001"
+          }
+        };
+
+        assert.deepStrictEqual(result.storage, storage);
+        assert.strictEqual(result.nextKey, null);
+      });
+
+      it("should return correct storage given different transaction indexes", async () => {
```
davidmurdoch, line 263: Because this test modifies the storage the previous tests rely on can we wrap this test in `evm_snapshot` and `evm_revert`? Or maybe move it to it's own `describe`? While it's fine now, if someone adds a similar test to the ones above right below this one, the results will be different depending on if it is run alone (via `it.only`) or with all others).

If you prefer the `evm_snapshot` way, just make sure the `evm_revert` is called within a `finally` block:

```javascript
const snapshotId = await provider.send("evm_snapshot");

try {
  // ...
} finally {
  await provider.send("evm_revert");
}
```

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -0,0 +1,487 @@
...
+
+        const txReceipt1 = await provider.send("eth_getTransactionReceipt", [
+          tx1
+        ]);
+        const txReceipt2 = await provider.send("eth_getTransactionReceipt", [
+          tx2
+        ]);
+        const txReceipt3 = await provider.send("eth_getTransactionReceipt", [
+          tx3
+        ]);
```
davidmurdoch: a slightly neater way:

```suggestion
        const [txReceipt1, txReceipt2, txReceipt3] = await Promise.all([
          provider.send("eth_getTransactionReceipt", [tx1]),
          provider.send("eth_getTransactionReceipt", [tx2]),
          provider.send("eth_getTransactionReceipt", [tx3])
        ]);
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1121,6 +1132,446 @@ export default class Blockchain extends Emittery.Typed<
...
+              // assign nextKey and remove it from array of keys
+              const nextKey = keys.pop();
+              result.nextKey = Data.from(nextKey).toJSON();
+            }
+
+            resolve(keys);
+          });
+      });
+    };
+    await getStorageKeys();
```
davidmurdoch: Can you move the `let keys: Buffer[] = [];` declaration inside the `getStorageKeys` function and then:

```suggestion
    const keys = await getStorageKeys();
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1121,6 +1132,446 @@ export default class Blockchain extends Emittery.Typed<
...
+      Buffer /*codeHash*/
+    ])[2];
+
+    // use the addresses storage trie to get relevant hashed keys
+    let keys: Buffer[] = [];
+    const getStorageKeys = () => {
+      return new Promise((resolve, reject) => {
+        trie
+          .createReadStream()
+          .on("data", async data => {
```
```suggestion
          .on("data", data => {
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1121,6 +1132,446 @@ export default class Blockchain extends Emittery.Typed<
...
+            keys.push(data.key);
+          })
+          .on("end", async () => {
```
```suggestion
          .on("end", () => {
```

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -209,6 +224,41 @@ export default class Miner extends Emittery.Typed<
...
+        gasLeft: BN;
+        memory: Array<number>; // Not officially sure the type. Not a buffer or uint8array
+        stack: Array<BN>;
+        depth: number;
+        opcode: {
+          name: string;
+        };
+        pc: number;
+        address: Buffer;
+      };
```
davidmurdoch: Maybe we can move this type from here (and from the `traceTransaction`) to a common place, maybe `@ganache/ethereum-utils` can export this type?

davidmurdoch (changes requested):
Looking good and loving the new tests! Made some comments and suggestions. I'll do another review once you've done the clean up you mentioned.

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -15,6 +17,7 @@ import VM from "ethereumjs-vm";
 import { encode as rlpEncode } from "rlp";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
 import replaceFromHeap from "./replace-from-heap";
+import { keccak } from "@ganache/utils/src/utils/keccak";
 const { BUFFER_EMPTY, BUFFER_256_ZERO } = utils;
```
davidmurdoch:
```suggestion
const { BUFFER_EMPTY, BUFFER_256_ZERO, keccak } = utils;
```

davidmurdoch (changes requested):
Only one more little quick fix and I think it's good to go! Awesome work @eshaben !

PR #757, opened 2021-01-28
chore: run from Lerna root to fix paths

mikeseese:
Typescript errors show file paths relative to the CWD. When using Lerna, the CWD is the package directory, not the Lerna root directory. This PR uses the Lerna-injected environment variable `LERNA_ROOT_PATH` to run the specified command from the lerna root directory so that filepaths are relative to there.

This PR also introduces a `lerna-mocha.ts` script which standardizes most of the mocha args

This PR only applies these changes for `tsc` and `test` commands, but may want to be expanded into other commands


scripts/lerna-ttsc.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/env bash
+
+if [ "${LERNA_ROOT_PATH}" != "" ]; then
+  PACKAGE_DIR=$(pwd)
+  cd ${LERNA_ROOT_PATH}
+  if [[ $# == 0 ]]; then
+    ttsc --project ${PACKAGE_DIR}/tsconfig.json
+  else
+    ttsc $@
```
mikeseese: This should allow you to still do something like `../../../../scripts/lerna-ttsc.sh --project $(pwd)/other.tsconfig.json`

mikeseese:
This isn't ready yet; it doesn't fix it for tests

davidmurdoch:
In lieu of these scripts, I wonder if we can do something like `lerna exec -- cross-env-shell cd $LERNA_ROOT_PATH \\&\\& ttsc --project \\$PWD`, but make it work also on Windows. I know `cross-env` works some magic for windows env vars. I haven't tested this script, so it is surely wrong... but the idea is mostly same, I think?

mikeseese:
What's wrong with these scripts though? If it works and it's a perfectly viable solution, then let's use it?

davidmurdoch:
> What's wrong with these scripts though? If it works and it's a perfectly viable solution, then let's use it?

This PR changes 17 files, but if we can just change two lines in one file to do the same thing I think it'd be worth it. Again, I haven't tested out my idea, which probably means it is trash; I'm just suggesting that it might be worth investigating.

mikeseese:
Don't mind the `bar` script name; it's just a placeholder.

`lerna exec -- cross-env-shell cd $LERNA_ROOT_PATH \\&\\& ttsc --project \\$PWD`

image

No cigar. Tried seeing what `\\$PWD` resolves to:

`lerna exec -- cross-env-shell cd $LERNA_ROOT_PATH \\&\\& echo \\$PWD`

image

Which is what we want, but if we got the output from the first command and the output of the second command, it seems that `cross-env-shell` is removing any `cd` changes from the first command (which is not default bash behavior), which is a deal breaker using `cross-env-shell`

I looked at the `cross-env` examples, which would point us to a wrapping quotes around the full command, but `lerna exec -- cross-env-shell PACKAGE_DIR=$PWD \"cd \\$LERNA_ROOT_PATH && tsc --project \\$PWD\"` provides the same thing.

FWIW, their example `cross-env-shell GREETING=Hi TEST=Joe \"echo $GREETING && echo $TEST\"` (changed NAME to TEST since $NAME is equivalent to $(hostname)) doesnt work for me:

image

Apparently `cross-env` went into maintenance mode and isn't accepting issues as of December.

I'm guessing this usage may or may not be possible with the current design of `cross-env`. I personally don't think it's worth more of my time to investigate the nuances of how cross-env parses the command and how NodeJS's `spawn` command (since that's what `cross-env` uses) would use the parsed command. If you think it's worth the extra time, I can do it. I honestly think this PR's design allows for a pretty nice devex interface of control and understanding what's getting executed.

mikeseese:
I did try updating `cross-env` to the latest version `7.0.3` (we're on `7.0.2`), and it didn't work; though I didn't expect it to since the release notes seemed to be unrelated

mikeseese:
Alright @davidmurdoch, looks like windows tests are being picked up now

mikeseese:
from @davidmurdoch [on slack](https://trufflesuite.slack.com/archives/C014QAVA0KA/p1613504640330700?thread_ts=1613502402.328500&cid=C014QAVA0KA):

>This isn't coming from just me on this one, btw. Though it was ultimately the deciding factor.
>
>This issue was the additional maintenance of these scripts, as well as the use of magic strings and the centralized mocha params. But ultimately, its that this actually (mostly*) only affects Windows devs, as typescript outputs unix paths, which vscode doesn't seem to do partial path matches for (./src/index.ts doesn't seem to search for .\src\packages\core\src\index.ts). So, this could probably be considered a typescript bug... and working around it isn't a truffle priority.
>
>Additionally, npm v7 is here, and if we don't switch to that, we're considering switching to yarn, each of which would likely make these solution obsolete anyway.
>
>So while it's useful for vscode power users, it's just not something we have prioritized because it is just the two of us that this is affecting right now.
>
>\* it's annoying on Linux too, but because it matches multiple files sometimes.

PR #765, opened 2021-02-03
Add cross-env as a dev dependency to the CLI

tcoulter:
Add cross-env as a dev dependency so it doesn't need to be installed globally when running npm start


tcoulter:
Closing. Instead, run `npm start` from the root directory.

PR #766, opened 2021-02-04
feat: throw exception when conflicting options are passed to the provider

tcoulter:


src/packages/core/tests/connector.test.ts
```diff
@@ -31,6 +31,27 @@ describe("connector", () => {
     });
   });
 
+  it("errors when conflicting options are passed to the provider", async () => {
+    let error: Error;
+
+    try {
+      const p = Ganache.provider({
```
davidmurdoch: Let's use [`assert.throws`](https://nodejs.org/api/assert.html#assert_assert_throws_fn_error_message) here.

src/packages/core/tests/connector.test.ts
```diff
@@ -31,6 +31,27 @@ describe("connector", () => {
...
+        wallet: {
+          deterministic: true,
+          seed: "123"
+        } as Object // "as Object" lets us get around ts typechecking during compilation
```
davidmurdoch: :+1:

src/packages/core/src/connector-loader.ts
```diff
@@ -26,6 +31,41 @@ export default {
...
+            ) {
+              throw new Error(
+                `Options ${category}.${option} and ${category}.${conflictingKey} are mutually exclusive`
+              );
+            }
+          }
+        }
+      }
+    }
+
```
davidmurdoch: I think this logic should be part of options normalization, as we already loop over everything there.

Here is where this magic happens: https://github.com/trufflesuite/ganache-core/blob/517795846cac9cf1ce14ff4855ae9457ea6f3f59/src/packages/options/src/create.ts#L25
(the "`: any`"s here suck, but TS was hitting its max depth when I had things typed :-/)

The gist of that function (I really should just document this in the code :grimacing:):

 * `defaults` are what we use if there are no user supplied options.
 * `options` is the _user_ supplied options.
 * `target` is what has been normalized so far
 * `namespace` is AKA "group" (miner, chain, wallet, etc)
 
It loops over each property in `default`, and uses then `normalizes` the corresponding `property` in `options`, if present, or grabs a default value.
 
You could perform a "conflict check" right before the calls to `propDefinition.normalize` _OR_ you could do it in the properties own `normalize` function, which I think would be ideal.

If you'd like it done in each options' `normalize` function (which I think I'd prefer) you'd need to modify a few things in the default options, types, and cli...

Here is an example default option:

https://github.com/trufflesuite/ganache-core/blob/517795846cac9cf1ce14ff4855ae9457ea6f3f59/src/chains/ethereum/options/src/wallet-options.ts#L225-L230 

 * You'd change the call to `normalize` (over in `options/src/create.ts`) to include the `namespaceOptions` as an argument
 * You'd update the `normalize` function for options that may have conflicts to check the `namespaceOptions` for conflicts, error if so. You could create a little helper function in  `ethereum/options/src/helpers.ts` to do the check.
 * remove the `conflicts` array from each default option
 * remove the type definition for `conflicts` in the [`Definitions` type](https://github.com/trufflesuite/ganache-core/blob/517795846cac9cf1ce14ff4855ae9457ea6f3f59/src/packages/options/src/definition.ts#L39-L44).
 * disable conflict checking yargs within `cli.ts`, as there seems to be no sense in checking it twice.
 
 Let me know if you'd like to sync on this!

src/packages/core/tests/connector.test.ts
```diff
@@ -31,6 +31,27 @@ describe("connector", () => {
     });
   });
 
+  it("errors when conflicting options are passed to the provider", async () => {
```
davidmurdoch: If we end up moving the conflicting options checking logic to the options package, let's move these tests there, too!

davidmurdoch (changes requested):
Thanks, Tim! I've made some suggestions for changes, as I think this logic should be part of the options package. Let me know if you want to sync on this!

src/packages/core/src/connector-loader.ts
tcoulter: Not against moving it, but did want to discuss this: 

> disable conflict checking yargs within cli.ts, as there seems to be no sense in checking it twice.

This was my original thinking too, but I've since changed my mind after diving into it. The difference between the code above and the code in `cli.js` is that `cli.js` throws the conflict checking off to `yargs` while the code above performs the error itself. It looks as though `yargs` has some extra magic that can detect the conflict while both A) disambiguating old args vs. new args, and B) translating the error message to match the arg that the user put in (whether an old arg or new arg). Given that the CLI accepts both, I think it's very valuable to have that disambiguation and translation in the error message _while also_ checking for conflicts upstream (the code above) to catch when users send conflicting args while only using the library. Thoughts? Granted, what I did was a straight copy/paste, which may be okay, but there is some code sharing we can do. 

Per your other suggestions, I'm not sure I follow completely, and so will need a sync. That said, removing the `conflicts` arrays in the `normalize` strategy sounds like a step backward in readability, at least from my recent "I just read the code for the first time" perspective. I found it relatively easy to find that conflict definition and how that array was used; it seems messy to me to convert that to `n` number of calls to a helper within each option's `normalize` method, where checking for conflicts somewhat overloads the term "normalize". 

Either way, my time is booked with a lot of meetings this week (we'll see about next week). If this is something you think you can knock out of the park fast (whichever direction you choose!) feel free to pick it up.

davidmurdoch: Looks like @eggplantzzz will be taking this one on. I'll sync up with him. Thanks!

src/packages/options/src/create.ts
```diff
@@ -22,40 +22,85 @@ export type Defaults<O extends Options> = {
...
-        config[key] = propDefinition.normalize(namespaceOptions[key]);
+        checkForConflicts(
+          key,
+          value,
+          suppliedOptions,
+          propDefinition.conflicts
+        );
+        const normalized = propDefinition.normalize(namespaceOptions[key]);
+        config[key] = normalized;
+        suppliedOptions.set(key, { inputName: key, normalized });
```
davidmurdoch: Let's format `inputName` as `${namespace}.${key}` so the option name in the error message appears as `wallet.accounts` instead of just `accounts`.

src/packages/options/src/create.ts
```diff
@@ -22,40 +22,85 @@ export type Defaults<O extends Options> = {
   [K in keyof O]: Definitions<O[K]>;
 };
 
+const checkForConflicts = (
+  inputOptionName: string,
+  inputValue: unknown,
```
davidmurdoch: Looks like we don't need `inputValue`

src/packages/options/src/create.ts
```diff
@@ -22,40 +22,85 @@ export type Defaults<O extends Options> = {
...
+};
+
 function fill(defaults: any, options: any, target: any, namespace: any) {
   const def = defaults[namespace];
   const config = (target[namespace] = target[namespace] || {});
 
+  const suppliedOptions = new Map<
+    string,
+    { inputName: string; normalized: unknown }
+  >();
```
davidmurdoch: Since we don't use the `normalized` value anywhere, we can simplify this to just 

```
const suppliedOptions = new Map<string, string>();
```

src/packages/options/src/create.ts
```diff
@@ -22,40 +22,85 @@ export type Defaults<O extends Options> = {
...
+  suppliedOptions: Map<string, { inputName: string; normalized: unknown }>,
+  conflicts?: string[]
+) => {
+  if (!conflicts) return;
+  for (const conflict of conflicts) {
+    if (suppliedOptions.has(conflict)) {
+      throw new Error(
+        `Values for both "${inputOptionName}" and ` +
+          `"${suppliedOptions.get(conflict).inputName}" cannot ` +
+          `be specified. They are mutually exclusive.`
```
davidmurdoch:
```suggestion
        `Values for both "${inputOptionName}" and ` +
          `"${suppliedOptions.get(conflict).inputName}" cannot ` +
          `be specified; they are mutually exclusive.`
```

davidmurdoch (changes requested):
Looking really good! I left a couple of some thoughts for some quick changes.

src/packages/options/src/create.ts
eggplantzzz: Will we ever want to preserve what they input and the value? It could be that later on there could be conflicts related to certain values for different options? Or perhaps this is very premature...

eggplantzzz: So anyways, if we change the format of the map will look like `<namespace>.<inputName> => value`? Or `<namespace>.<canonicalName> => <inputName>`? Hmmm...

davidmurdoch: I don't think we need to store the `value` since we aren't going to do anything with it just yet. I'm a strong proponent of YAGNI (You Aren't Going to Need It)

src/packages/options/src/create.ts
davidmurdoch: Alternatively, we could just just pass `namespace` to `checkForConflicts` and have it do the string concatenation if it needs it.

src/packages/options/src/create.ts
eggplantzzz: Yes, I'm also into the YAGNI

davidmurdoch (approved):

PR #767, opened 2021-02-04
feat: add filecoin to dynamic cli

mikeseese:
This PR is a continuation of the work that was done in #726, but it includes more. Here's the full list of this PR's intention:
- Add the Filecoin package to the CLI interface properly
  - This includes keeping the `@ganache/filecoin` package as a `peerDependency` for `@ganache/flavors` and failing if the package cannot be found
  - Due to the above, a `link.filecoin` script was added to the root package which is meant as a development helper to link the filecoin package for usage
  - This peer dependency architecture also required me to provide separate type declarations to prevent a circular dependency. I have a strong feeling @davidmurdoch will not be a fan of this, but I believe this is the only setup that will work *in production*. I could be wrong and am open to alternatives. I provided a root script `tsc.filecoin.declarations` to allow for quick ease and file cleaning via `pretty-quick`
  - Neither of these two things (root linking script and type declaration generation) are currently automated (meaning the developer has to call them when they need it). I'm not sure if/who you want this in your development process, so I left it out as it's a trivial thing to setup later. I can do this if you let me know what the desired dev experience should be.
- Preliminary reorganization of tests in `@ganache/filecoin`. I originally intended to fix tests, but I found myself spending too much time fixing tests that were no longer fully applicable due to large overhauls in `@ganache/filecoin/src/things`. I didn't want to spend much time getting the tests to pass in this state. You can see tests are passing in the [subsequent branch](https://github.com/trufflesuite/ganache-core/commits/feat/filecoin-chainnotify). I strongly encourage we do not fixate on tests that are [failing prior to this branch](https://github.com/trufflesuite/ganache-core/commits/filecoin) when this branch does not alter the behavior behind those tests.
- Some other minor housekeeping stuff (some code styling, etc)


src/chains/filecoin/filecoin/package.json
```diff
@@ -27,9 +31,10 @@
     "docs.build": "rm -rf ./lib/docs ./lib/api.json && npm run docs.typedoc",
     "docs.typedoc": "typedoc --options ./typedoc.json --readme ./README.md --out ../../../docs/typedoc --json ../../../docs/typedoc/api.json src/api.ts",
     "docs.preview": "ws --open --port 3012 --directory ../../../",
-    "tsc": "ttsc",
+    "tsc": "ts-node ../../../../scripts/lerna-run.ts ttsc --project _PACKAGEDIR_/tsconfig.json",
+    "tsc.declarations": "ts-node ../../../../scripts/lerna-run.ts ttsc --project _PACKAGEDIR_/tsconfig.declarations.json",
     "test": "nyc --reporter lcov npm run mocha",
-    "mocha": "cross-env TS_NODE_COMPILER=ttypescript TS_NODE_FILES=true mocha --exit --check-leaks --throw-deprecation --trace-warnings --require ts-node/register --timeout 10000 'tests/**/*.test.ts'"
+    "mocha": "cross-env TS_NODE_COMPILER=ttypescript TS_NODE_FILES=true ts-node ../../../../scripts/lerna-mocha.ts --require ts-node/register '_PACKAGEDIR_/tests/**/*.test.ts'"
```
mikeseese, line 37: All of the `filecoin` branches are currently based off of #757 as it helps my development process. This adheres to that if it is accepted. It will be changed if #757 is rejected.

mikeseese:
I'm closing this PR because @tcoulter wants PRs that have passing tests. Instead we are creating a PR with the next branch (`feat/filecoin-chainnotify`) that targets `filecoin` (so it'll also include this work)

PR #768, opened 2021-02-04
feat: add filecoin to CLI, `Filecoin.ChainNotify`  method, and Filecoin support for websockets/subscriptions

mikeseese:
This PR is a continuation of both #767 (which was a continuation of #726) and #712. Here is what this PR adds:

- Add the Filecoin package to the CLI interface properly
  - This includes keeping the `@ganache/filecoin` package as a `peerDependency` for `@ganache/flavors` and failing if the package cannot be found
  - This peer dependency architecture required me to provide separate type declarations to prevent a circular dependency. I have a strong feeling @davidmurdoch will not be a fan of this, but I believe this is the only setup that will work *in production*. I could be wrong and am open to alternatives. I provided a root script `tsc.filecoin.declarations` to allow for quick ease and file cleaning via `pretty-quick`
  - Neither of these two things (root linking script and type declaration generation) are currently automated (meaning the developer has to call them when they need it). I'm not sure if/who you want this in your development process, so I left it out as it's a trivial thing to setup later. I can do this if you let me know what the desired dev experience should be.
- Preliminary reorganization of tests in `@ganache/filecoin`. This branch also fixes existing tests.
- Some other minor housekeeping stuff (some code styling, etc)
- Supports the Chain Explorer feature in the https://github.com/filecoin-shipyard/filecoin-network-inspector example
- Adds support for `truffle preserve`
- Corrects all of the Filecoin types to be consistent with the Filecoin API v1.4.0

It working with the Chain Explorer:
image

It working with `truffle preserve`:


.github/workflows/pr.yml
```diff
@@ -48,6 +48,7 @@ jobs:
         if: startsWith(matrix.os, 'windows-')
         run: npm config set msvs_version 2015
       - run: npm ci
+      - run: npm run tsc
```
davidmurdoch, line 51: :+1:

scripts/lerna-ttsc.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/env bash
```
davidmurdoch: is this file used anywhere?

src/chains/ethereum/options/src/index.ts
```diff
@@ -18,14 +17,22 @@ import {
...
-export type EthereumOptions = {
+type EthereumConfig = {
   chain: ChainConfig;
   database: DatabaseConfig;
   logging: LoggingConfig;
   miner: MinerConfig;
   wallet: WalletConfig;
 };
 
+export const EthereumDefaults: Defaults<EthereumConfig> = {
```
davidmurdoch, line 28: Any reason for the uppercase "E" here? Title casing is typically used for constructors elsewhere in the codebase.

src/chains/ethereum/options/src/index.ts
```diff
@@ -44,27 +51,15 @@ export type EthereumLegacyOptions = Partial<
...
-    [K in keyof EthereumOptions]: ExternalConfig<EthereumOptions[K]>;
+    [K in keyof EthereumConfig]: ExternalConfig<EthereumConfig[K]>;
   }
 >;
 
 export type EthereumInternalOptions = {
-  [K in keyof EthereumOptions]: InternalConfig<EthereumOptions[K]>;
-};
-
-export type EthereumDefaults = {
```
davidmurdoch, line 55: :+1:

src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -140,7 +140,13 @@ export type MinerConfig = {
 
 export const MinerOptions: Definitions<MinerConfig> = {
   blockTime: {
-    normalize,
+    normalize: rawInput => {
+      if (rawInput < 0) {
```
davidmurdoch, line 144: Interestingly, we used to accept negative block times and would try to mine them with a negative time out.

The current code requires that blockTime be greater than 0, otherwise it ignores the blockTime  value. I wonder if this check will break any work flows that accidentally mined on a negative interval.

I'm okay with breaking those without official documenting this as a breaking change since it is self documenting via the obvious error message, so :+1:

src/chains/filecoin/filecoin/package.json
```diff
@@ -5,6 +5,10 @@
   "author": "Tim Coulter",
   "homepage": "https://github.com/trufflesuite/ganache-core/tree/develop/src/filecoin#readme",
   "license": "MIT",
+  "engines": {
+    "node": ">=12.13.0 <=14.14.0",
```
davidmurdoch, line 9: :+1: 

also, :sad-panda:

src/chains/ethereum/ethereum/tests/api/debug/debug.test.ts
```diff
@@ -311,6 +311,6 @@ describe("api", () => {
       }
 
       assert.strictEqual(countMap.size, expectedObjectsInFinalTrace);
-    });
+    }).timeout(5000);
```
davidmurdoch, line 314: Was there a performance regression causing this to require an increase?

src/chains/filecoin/filecoin/src/api.ts
```diff
@@ -99,8 +186,8 @@ export default class FilecoinApi implements types.Api {
     return {};
   }
 
-  async "Filecoin.GanacheMineTipset"(): Promise<SerializedTipset> {
-    await this[_blockchain].mineTipset();
-    return this[_blockchain].latestTipset().serialize();
+  async "Ganache.MineTipset"(): Promise<SerializedTipset> {
```
davidmurdoch, line 189: @seesemichaelj @tcoulter should we bikeshed what what the namespace for custom RPC methods should be?

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -1,99 +1,105 @@
...
+  tipset: Tipset;
 };
 
 export default class Blockchain extends Emittery.Typed<
   BlockchainEvents,
   keyof BlockchainEvents
 > {
   readonly tipsets: Array<Tipset> = [];
-  readonly miner: Miner;
+  readonly miner: string; // using string until we can support more address types in Address
```
davidmurdoch, line 29: Should we change this comment to a `// TODO` so we can easily search for it later?

src/packages/utils/src/utils/random-number-generator.ts
```diff
@@ -0,0 +1,37 @@
...
+    } else {
+      this.rng = Math.random;
+    }
+  }
+
+  getNumber(range: number = 1): number {
+    // I believe this check may be a tiny bit faster than
+    // always multiplying by 1
+    if (range !== 1) {
+      return this.rng() * range;
```
davidmurdoch: While reading this I noticed that the range of number is exclusive of the `range` value as the rng outputs values from `[0 to 1)` (0 inclusive, 1 exclusive), so when multiplied by `255` the max value is `254.9999...`. `Buffer.from` will automatically case these values to an integer by lopping of the decimal (`Buffer.from([254.999])` -> `[fe]`).

The implementation in ethereum has the same bug. I'll fix the little purpose-built RNG used over there; probably by increasing the upper range by 1. I dunno, haven't thought about it yet.

src/packages/options/src/definition.ts
```diff
@@ -57,7 +57,11 @@ export type Definitions<C extends Base.Config> = {
     (void extends OptionHasDefault<C, N>
       ? {}
       : {
-          readonly default: (config: InternalConfig<C>) => OptionType<C, N>;
+          // using type string for flavor to prevent circular dependency
```
davidmurdoch, line 60: :+1:

davidmurdoch (changes requested):
This is a partial review. I just want to get this feedback to you sooner rather than later.

scripts/lerna-ttsc.sh
mikeseese: No, it's a remnant of my old approach for #757, it somehow snuck in with the various rebases. Good catch!

src/chains/ethereum/options/src/index.ts
mikeseese: This was a personal choice that it felt that `ethereumDefaults` was out of place and inconsistent with the several other exports here. Further, there aren't many examples of the codebase of an exported const object's naming. outside of these default objects, you have a `params` in `@ganache/ethereum-utils` and 4 `WeakMap` variables in `json-rpc-base-types.ts`. So there wasn't really a precedent set for the styling here other than these default objects (seems that I forgot one in `@ganache/core/src/options`).

I can change them all back to be camelcase, but camelcase const objects feel mutable and modular rather than global/immutable

---

While I still stand by the above, I remembered that the driving change of this happened when changing `@ganache/flavors/index.ts` which has `export const DefaultOptionsByName`. So since I prefer the title casing here, I switched those to adhere to that rather than switching it to camelcasing

src/chains/ethereum/ethereum/tests/api/debug/debug.test.ts
mikeseese: No; this isn't needed on local development. However, I pinpointed these tests to be the ones that very consistently went past the 2000ms timeout on github CI for nodejs 12+ versions. I could try to find examples of those tests, but it seemed they take ~3.5 seconds on CI not infrequently

src/chains/filecoin/filecoin/src/api.ts
mikeseese: I will make my one remark, but you can tell me whatever you want it to be; I'm not going to spend more time bikeshedding about this.

First, using the `Filecoin.` prefix here for ganache specific stuff is misleading. It makes it seem like it's part of the Filecoin schema, but it's not. The schema doesn't even have those prefixes; it's clients/nodes that adhere to the schema that actually add/remove the prefix. You can see the schema [here](https://github.com/filecoin-shipyard/js-lotus-client-schema) or the actual methods [here](https://github.com/filecoin-shipyard/js-lotus-client-schema/blob/master/mainnet/index.js) doesn't actually refer to the `Filecoin.` prefix; it's part of the API specification rather than schema specification. This throws out the idea that we should put our custom messages in the same namespace as the official schema.

Then there's what ganache already does. Ethereum ganache uses `evm_mine` for the equivalent here. I think the `evm_` prefix here is misleading to begin with. It makes it seem like this is a standardized method, but it's actually only implemented anywhere with Ganache. Further, we aren't implementing the VM here, so we can't do something `FVM.` without being misleading. So that throws out the window of "what is existing ganache behavior".

Since I believe it's important to be consistent at the local areas rather than global (it's more important to be consistent with the rest of the Filecoin api than it is to be consistent with the rest of the Ganache codebase) because thats what developers are familiar with, `Ganache.` felt like very apt namespace here. Using `Truffle` or `TruffleSuite` would both either be confusing or nonintuitive. I also think it's important that there *is* a namespace rather than no namespace, because the Filecoin API clearly wants to create namespaces.

With that all said, I'm happy for yall to tell me whatever you want, but that's the reasoning here.

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: Are you okay with keeping it as it is if my next task is to correct it? (I probably will finish supporting it before you respond to this)

src/packages/utils/src/utils/random-number-generator.ts
mikeseese: I guess `range` is a misnomer here because of that issue; perhaps it should be named `upperExclusiveBound`?

I believe that we shouldn't try to manipulate it to output a `[0 to 1]` set of values, but rather change the consuming code if it needs to make sure the upper bound is met sometimes. For example, if we're using `Buffer.from([rng.getNumber(255)])` to generate an address which we know that we sometimes want `ff`, then we can do `Buffer.from([Math.round(rng.getNumber(255))])` which is safe because we know `Buffer.from` is already going to truncate it, so that is perfectly valid code

Otherwise we can add a method here for `getRoundedInteger(range: number) : number` which calls the `Math.round` for the tool. However, we can't do this in `getNumber` because it's hard to know what this random number is being used for, and whether or not it's range is large enough to be rounded

An interesting issue that I didn't consider. Let me know if/how/when you'd like this to be changed; it's likely fine for the uses that I'm using it for, but it could be misleading for new devs stumbling on it

src/chains/ethereum/options/src/index.ts
davidmurdoch: Probably just a case of me not being consistent. I don't really think of this particular value as a "constant", but I guess we could think of it that way. For simple constants I've been using UPPERCASE (See constants.ts in the utils package).

You can leave this one as is and Resolve this conversastion.

src/chains/filecoin/filecoin/src/api.ts
davidmurdoch: I've always thought of `evm_` as our way of saying "this is a custom namespace, but we invite other tools to copy us". Whereas had we done `ganache_` it would encourage a divide in the standardization of the tools.

@tcoulter, any thoughts?

src/chains/filecoin/filecoin/src/blockchain.ts
davidmurdoch: Yes, but only if show me the commit first :-p

src/packages/utils/src/utils/random-number-generator.ts
davidmurdoch: Here is my super simple fix: https://github.com/trufflesuite/ganache-core/pull/770

I think just renaming it `upperExclusiveBound` would be fine here. Then just update `getBuffer`'s `range` to be `256`. I think that'll work ok here. We don't need a perfectly random distribution, but i'd like to be able to get all numbers between 0 and 255, inclusive of both.

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: I was not faster than you; dinner beat me to the punch haha

Here it is: https://github.com/trufflesuite/ganache-core/commit/b67757aab542868fa8e5f00dac21b8a1bda21e63

src/chains/filecoin/filecoin/src/api.ts
tcoulter: Couple thoughts: 

- All good points @seesemichaelj.
- So far, I have no strong preferences. But do have some thoughts, below. 
- I'm open to a prefix that's less branded, per @davidmurdoch's comment. @davidmurdoch, do you have a suggestion for what this could be? Maybe `Miner.MineTipset`? I forget though if the `Miner` namespace is already used. Other prefixes: `Debug`. `Test`. `Develop`. `Dev`? 
- Separately, does using a namespace other than `Filecoin` make the method harder to call from convenience libraries like `lotus-client-rpc`? If so, I'd lean toward keeping the `Filecoin` prefix.

mikeseese: You cannot execute this method from `lotus-client-rpc` because that library only uses methods from the schema. Either filecoin would need to add this to the official schema somewhere (I'm doubtful since this method makes no sense outside of ganache) or it would need to be added separately. `filecoin.js` would need to add it as well regardless and doesnt make a difference to them

FWIW, `evm_` didn't get picked up by the community and `evm_mine` is still, several years later, exclusively a Ganache method. I honestly don't see it any different here, and as a dev I'd rather know that this is specific to ganache

src/chains/filecoin/filecoin/src/api.ts
```diff
@@ -1,79 +1,166 @@
...
+      return Promise.resolve(true);
+    } else {
+      return Promise.resolve(false);
+    }
   }
 
-  async "Filecoin.StateListMiners"(): Promise<Array<SerializedMiner>> {
-    return [this[_blockchain].miner.serialize()];
+  async "Filecoin.StateListMiners"(): Promise<Array<string>> {
+    return [this.#blockchain.miner];
```
tcoulter, line 129: Spidey sense (meaning I could be totally wrong): Is serialize() still needed here?

src/chains/filecoin/filecoin/src/provider.ts
```diff
@@ -44,28 +42,74 @@ export default class FilecoinProvider
...
+  async send(payload: JsonRpc.Request<FilecoinApi>) {
+    const result = await this._requestRaw(payload);
+    return result.value;
   }
 
-  async send<Method extends keyof FilecoinApi = keyof FilecoinApi>(
+  async _requestRaw<Method extends keyof FilecoinApi = keyof FilecoinApi>(
     payload: JsonRpc.Request<FilecoinApi>
   ) {
     // I'm not entirely sure why I need the `as [string]`... but it seems to work.
```
tcoulter: Minor: Looks like this comment no longer applies? Perhaps it's worth documenting the `any` instead.

src/chains/filecoin/filecoin/src/schema.ts
```diff
@@ -1,18 +1,28 @@
...
+  ...LotusSchema.mainnet.storageMiner.methods,
+  ...LotusSchema.mainnet.gatewayApi.methods,
+  ...LotusSchema.mainnet.walletApi.methods,
+  ...LotusSchema.mainnet.workerApi.methods
+};
+
 // Use the FilecoinAPI to create a schema object representing the functions supported.
 for (const methodName of Object.getOwnPropertyNames(FilecoinApi.prototype)) {
-  if (methodName == "constructor") {
+  if (!methodName.startsWith("Filecoin.")) {
```
tcoulter, line 18: It looks like this Schema won't include methods with the prefix `Ganache`, assuming we continue with that prefix (see earlier discussion). Will this need to be changed if we change the prefix?

src/chains/filecoin/filecoin/src/things/beacon-entry.ts
```diff
@@ -13,33 +15,49 @@ interface BeaconEntryConfig {
...
     data: {
-      type: string;
+      type: Buffer;
       serializedType: string;
       serializedName: "Data";
     };
   };
 }
 
+type C = BeaconEntryConfig;
```
tcoulter: Is `C` used here because you didn't want to type out `BeaconEntryConfig` below? If so, please change back to `BeaconEntryConfig` for readability. 

As background, my initial impression was that `BeaconEntry` was templated (e.g., `class BeaconEntry<C extends BaseConfig>`), which turned out not to be the case.

src/chains/filecoin/filecoin/src/things/channel-id.ts
```diff
@@ -0,0 +1,70 @@
...
+    };
+    id: {
+      type: number;
+      serializedType: number;
+      serializedName: "ID";
+    };
+  };
+};
+
+type C = ChannelIDConfig;
```
tcoulter: Same comment as above. Is this for speedy typing?

src/chains/filecoin/filecoin/src/things/data-transfer-channel.ts
```diff
@@ -0,0 +1,146 @@
...
+    };
+    transferred: {
+      type: number;
+      serializedType: number;
+      serializedName: "Transferred";
+    };
+  };
+};
+
+type C = DataTransferChannelConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/deal-info.ts
```diff
@@ -0,0 +1,232 @@
...
+    };
+    dataTransfer: {
+      type: DataTransferChannel;
+      serializedType: SerializedDataTransferChannel;
+      serializedName: "DataTransfer";
+    };
+  };
+};
+
+type C = DealInfoConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/election-proof.ts
```diff
@@ -5,33 +5,59 @@ import {
...
     vrfProof: {
-      type: string;
+      type: Buffer;
       serializedType: string;
       serializedName: "VRFProof";
     };
   };
 }
 
+type C = ElectionProofConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/head-change.ts
```diff
@@ -0,0 +1,68 @@
...
+    };
+    val: {
+      type: Tipset;
+      serializedType: SerializedTipset;
+      serializedName: "Val";
+    };
+  };
+}
+
+type C = HeadChangeConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/post-proof.ts
```diff
@@ -0,0 +1,61 @@
...
+    };
+    proofBytes: {
+      type: Buffer;
+      serializedType: string;
+      serializedName: "ProofBytes";
+    };
+  };
+}
+
+type C = PoStProofConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/query-offer.ts
```diff
@@ -0,0 +1,164 @@
...
+    };
+    minerPeer: {
+      type: RetrievalPeer;
+      serializedType: SerializedRetrievalPeer;
+      serializedName: "MinerPeer";
+    };
+  };
+};
+
+type C = QueryOfferConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/retrieval-order.ts
```diff
@@ -0,0 +1,164 @@
...
+    };
+    minerPeer: {
+      type: RetrievalPeer;
+      serializedType: SerializedRetrievalPeer;
+      serializedName: "MinerPeer";
+    };
+  };
+};
+
+type C = RetrievalOrderConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/retrieval-peer.ts
```diff
@@ -0,0 +1,73 @@
...
+    };
+    pieceCID: {
+      type: RootCID;
+      serializedType: SerializedRootCID;
+      serializedName: "PieceCID";
+    };
+  };
+};
+
+type C = RetrievalPeerConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/root-cid.ts
```diff
@@ -8,35 +8,44 @@ import {
...
       serializedName: "/";
     };
   };
 }
 
-class RootCID
-  extends SerializableObject<RootCIDConfig>
-  implements DeserializedObject<RootCIDConfig> {
-  get config(): Definitions<RootCIDConfig> {
+type C = RootCIDConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/signature.ts
```diff
@@ -0,0 +1,62 @@
...
+    };
+    data: {
+      type: Buffer;
+      serializedType: string;
+      serializedName: "Data";
+    };
+  };
+}
+
+type C = SignatureConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/start-deal-params.ts
```diff
@@ -0,0 +1,164 @@
...
+    };
+    verifiedDeal: {
+      type: boolean;
+      serializedType: boolean;
+      serializedName: "VerifiedDeal";
+    };
+  };
+};
+
+type C = StartDealParamsConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/storage-market-data-ref.ts
```diff
@@ -0,0 +1,88 @@
...
+    };
+    pieceSize: {
+      type: 0;
+      serializedType: 0;
+      serializedName: "PieceSize";
+    };
+  };
+};
+
+type C = StorageMarketDataRefConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/ticket.ts
```diff
@@ -5,32 +5,45 @@ import {
...
       serializedName: "VRFProof";
     };
   };
 }
 
-class Ticket
-  extends SerializableObject<TicketConfig>
-  implements DeserializedObject<TicketConfig> {
-  get config(): Definitions<TicketConfig> {
+type C = TicketConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/src/things/tipset.ts
```diff
@@ -27,51 +29,63 @@ interface TipsetConfig {
   };
 }
 
-class Tipset
-  extends SerializableObject<TipsetConfig>
-  implements DeserializedObject<TipsetConfig> {
-  get config(): Definitions<TipsetConfig> {
+type C = TipsetConfig;
```
tcoulter: Same as above.

src/chains/filecoin/filecoin/tests/blockchain/blockchain.test.ts
```diff
@@ -230,26 +265,49 @@ describe("Blockchain", () => {
...
+      if (blockchain) {
+        await blockchain.stop();
+      }
+    });
+
+    it("creates the expected address from seed", async () => {
       blockchain = new Blockchain(
         FilecoinOptionsConfig.normalize({
           wallet: {
             seed: "tim is a swell guy"
```
tcoulter, line 281: Nice of you to leave this in there. ;)

tcoulter (changes requested):
Lots of awesome work here, and it's great to see this progressing. Aside, it's super cool you got Websockets in there. 

I made a handful of comments. The majority of them are related to code style for one specific pattern you use within the code. Would love for you to change that. The rest are questions. 

Thanks again!

src/chains/filecoin/filecoin/src/api.ts
mikeseese: No because `miner` was changed to type `string` in this PR. If you saw, I removed the `Miner` type you had before because it's not a real Filecoin API type. ID protocol address support was added in a distant branch last night to change the behavior of this: https://github.com/trufflesuite/ganache-core/commit/b67757aab542868fa8e5f00dac21b8a1bda21e63 where `miner` becomes type `Address`

src/chains/filecoin/filecoin/src/schema.ts
mikeseese: > It looks like this Schema won't include methods with the prefix Ganache, assuming we continue with that prefix (see earlier discussion).

That is correct. The way I interpreted your intended purpose of this object was to initialize `lotus-client-rpc`, which [prefixes](https://github.com/filecoin-shipyard/js-lotus-client-rpc/blob/master/index.js#L38) all methods with `Filecoin.`.

If we wanted `lotus-client-rpc` to actually be able to call the mine tipset function, we would need to change the prefix to `Filecoin.` despite it being part of the main namespace. We can also alternatively change `lotus-client-rpc` as I mentioned in https://github.com/trufflesuite/ganache-core/pull/768#discussion_r571204685 which would require a change here

However, this function will not need to change regardless if we change the prefix if the intention is to use it with initializing `lotus-client-rpc` unless we change `lotus-client-rpc`. This function just tells that library which `Filecoin.` methods are available, since it will only execute those methods due to the hardcoded prefix in their code.

src/chains/filecoin/filecoin/src/api.ts
mikeseese: A comment @tcoulter made on another piece of code made me realize that we can have this method work with the current `lotus-client-rpc` if it has the `Filecoin.` prefix. This is due to `lotus-client-rpc` [hardcoding](https://github.com/filecoin-shipyard/js-lotus-client-rpc/blob/master/index.js#L38) the `Filecoin.` prefix

However, I have every belief that `lotus-client-rpc` was designed around the fact that there is currently only one namespace. A very simple PR over there would enable `Ganache.MineTipset` to be callable from `provider.ganacheMineTipset()` or `provider.mineTipset()` depending on Protocol Labs' preference.

src/chains/filecoin/filecoin/src/things/beacon-entry.ts
mikeseese: > Is C used here because you didn't want to type out BeaconEntryConfig below? If so, please change back to BeaconEntryConfig for readability.

Yes, it helps reduce a bit of time but I can change it back. I actually considered it more readable with it typed out because every single class is the same way. It ended up being noise/clutter for the some 25+ types in the latest branch that follow the exact same pattern. Let me know if you still want me to change them back to using the verbose type.

src/chains/filecoin/filecoin/src/things/channel-id.ts
mikeseese: I'm going to resolve all of these extra comments; I'll be sure to make changes to each one if you decide you still want the change.

src/chains/filecoin/filecoin/src/api.ts
tcoulter: Sounds good!

src/chains/filecoin/filecoin/src/things/beacon-entry.ts
tcoulter: I know this sucks man but would love for you to change them all. I reviewed it with @gnidan and @davidmurdoch just to make sure I wasn't being a stickler, and they agree the `C` syntax is confusing when not templated. Thanks!

src/chains/filecoin/filecoin/src/api.ts
tcoulter: Good points. 

Unless you have strong feelings @davidmurdoch I'm happy to lave it `Ganache.<method>`.

src/chains/filecoin/filecoin/src/things/beacon-entry.ts
mikeseese: No problem

src/chains/filecoin/filecoin/src/api.ts
davidmurdoch: Thanks Tim. Ship it!

src/chains/filecoin/filecoin/src/schema.ts
tcoulter: All good for leaving it how it is. Thanks @seesemichaelj, and sorry for the delay.

mikeseese:
@tcoulter approved this [on Slack](https://trufflesuite.slack.com/archives/C014QAVA0KA/p1613077582273500?thread_ts=1612808486.218100&cid=C014QAVA0KA)

PR #770, opened 2021-02-05
fix(ethereum): ensure randomBytes can return 0-255

davidmurdoch:
Previously, it returned only bytes 0 - 254.

### DIFF
--- a/src/chains/ethereum/options/src/wallet-options.ts
+++ b/src/chains/ethereum/options/src/wallet-options.ts
@@ -6,7 +6,13 @@ import { Definitions, DeterministicSeedPhrase } from "@ganache/options";
 
 const { alea } = seedrandom;
 
-function randomBytes(length: number, rng: () => number) {
+/**
+ * WARNING: to maintain compatibility with ganache v2 this RNG only generates
+ * numbers from 0-254 instead of 0-255! Hence the name, `notVeryRandomBytes`
+ * @param length
+ * @param rng
+ */
+function notVeryRandomBytes(length: number, rng: () => number) {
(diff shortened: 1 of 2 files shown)


mikeseese (approved):
LGTM

mikeseese:
Do note this may change the behavior of `--deterministic` or `--seed`! It did for me (but it's ok because I haven't released)

davidmurdoch:
> Do note this may change the behavior of --deterministic or --seed! It did for me (but it's ok because I haven't released)

BREAK ALL THE THINGS!

davidmurdoch:
> Do note this may change the behavior of --deterministic or --seed! It did for me (but it's ok because I haven't released)

Ah, wait... crap. While this function doesn't matter for the start up accounts... there is another function that does have the same bug, even in version 2, and that WILL change things. hmmmmm

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -6,7 +6,13 @@ import { Definitions, DeterministicSeedPhrase } from "@ganache/options";
...
 const { alea } = seedrandom;
 
-function randomBytes(length: number, rng: () => number) {
+/**
+ * WARNING: to maintain compatibility with ganache v2 this RNG only generates
+ * numbers from 0-254 instead of 0-255! Hence the name, `notVeryRandomBytes`
+ * @param length
+ * @param rng
+ */
+function notVeryRandomBytes(length: number, rng: () => number) {
```
davidmurdoch, line 15: @seesemichaelj can I get a 👍 if you approve of this new commit and message?

mikeseese (approved):

PR #771, opened 2021-02-05
fix: set `difficulty` default to 1 and add `totalDifficulty`

eggplantzzz:
Currently Ganache sets all block headers to have `difficulty` and `totalDifficulty` hard-coded to 0. This PR causes Ganache to set the difficulty of each block (by default) to 1 (the user can specify their own block difficulty on the command line). The `totalDifficulty` is also now calculated based on the previous block's `totalDifficulty` and the block difficulty.

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -456,7 +456,8 @@ export default class Blockchain extends Emittery.Typed<
       previousBlock.hash(),
       this.coinbase,
       this.#options.miner.blockGasLimit.toBuffer(),
-      Quantity.from(timestamp == null ? this.#currentTime() : timestamp)
+      Quantity.from(timestamp == null ? this.#currentTime() : timestamp),
+      previousHeader.difficulty
     );
   };
 
@@ -555,7 +556,8 @@ export default class Blockchain extends Emittery.Typed<
       Quantity.from(BUFFER_32_ZERO),
       this.coinbase,
       blockGasLimit.toBuffer(),
-      Quantity.from(timestamp)
+      Quantity.from(timestamp),
+      this.#options.miner.difficulty
     );
 
     // store the genesis block in the database
@@ -897,7 +899,8 @@ export default class Blockchain extends Emittery.Typed<
       parentBlock.header.miner,
       parentBlock.header.gasLimit.toBuffer(),
       // make sure we use the same timestamp as the target block
-      targetBlock.header.timestamp
+      targetBlock.header.timestamp,
+      parentBlock.header.difficulty
     ) as RuntimeBlock & { uncleHeaders: []; transactions: Transaction[] };
     newBlock.transactions = [];
     newBlock.uncleHeaders = [];
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -457,7 +457,7 @@ export default class Blockchain extends Emittery.Typed<
       this.coinbase,
       this.#options.miner.blockGasLimit.toBuffer(),
       Quantity.from(timestamp == null ? this.#currentTime() : timestamp),
-      previousHeader.difficulty
+      this.#options.miner.difficulty
     );
   };
 
@@ -900,7 +900,7 @@ export default class Blockchain extends Emittery.Typed<
       parentBlock.header.gasLimit.toBuffer(),
       // make sure we use the same timestamp as the target block
       targetBlock.header.timestamp,
-      parentBlock.header.difficulty
+      this.#options.miner.difficulty
     ) as RuntimeBlock & { uncleHeaders: []; transactions: Transaction[] };
     newBlock.transactions = [];
     newBlock.uncleHeaders = [];
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -457,7 +457,8 @@ export default class Blockchain extends Emittery.Typed<
       this.coinbase,
       this.#options.miner.blockGasLimit.toBuffer(),
       Quantity.from(timestamp == null ? this.#currentTime() : timestamp),
-      this.#options.miner.difficulty
+      this.#options.miner.difficulty,
+      previousBlock.header.totalDifficulty
     );
   };
 
@@ -557,7 +558,8 @@ export default class Blockchain extends Emittery.Typed<
       this.coinbase,
       blockGasLimit.toBuffer(),
       Quantity.from(timestamp),
-      this.#options.miner.difficulty
+      this.#options.miner.difficulty,
+      Quantity.from(0) // we start the totalDifficulty at 0
     );
 
     // store the genesis block in the database
@@ -900,7 +902,8 @@ export default class Blockchain extends Emittery.Typed<
       parentBlock.header.gasLimit.toBuffer(),
       // make sure we use the same timestamp as the target block
       targetBlock.header.timestamp,
-      this.#options.miner.difficulty
+      this.#options.miner.difficulty,
+      parentBlock.header.totalDifficulty
     ) as RuntimeBlock & { uncleHeaders: []; transactions: Transaction[] };
     newBlock.transactions = [];
     newBlock.uncleHeaders = [];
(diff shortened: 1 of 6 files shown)


src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -43,6 +43,16 @@ export type MinerConfig = {
       };
     };
 
+    /**
+     * Sets the block difficulty
+     * @default 1
```
davidmurdoch, line 49:
```suggestion

     * @default 1
```

src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -173,6 +183,12 @@ export const MinerOptions: Definitions<MinerConfig> = {
     default: () => Quantity.from(90_000),
     cliType: "string"
   },
+  difficulty: {
+    normalize: Quantity.from,
+    cliDescription: "Sets the block difficulty for the network.",
+    default: () => Quantity.from(1),
```
davidmurdoch: Can you use `RPCQUANTITY_ONE` instead of `Quantity.from(1)`? You can get it from `import { utils } from "@ganache/utils"` here?

src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -173,6 +183,12 @@ export const MinerOptions: Definitions<MinerConfig> = {
     default: () => Quantity.from(90_000),
     cliType: "string"
   },
+  difficulty: {
+    normalize: Quantity.from,
+    cliDescription: "Sets the block difficulty for the network.",
```
davidmurdoch:
```suggestion
    cliDescription: "Sets the block difficulty.",
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -555,7 +557,9 @@ export default class Blockchain extends Emittery.Typed<
       Quantity.from(BUFFER_32_ZERO),
       this.coinbase,
       blockGasLimit.toBuffer(),
-      Quantity.from(timestamp)
+      Quantity.from(timestamp),
+      this.#options.miner.difficulty,
+      Quantity.from(0) // we start the totalDifficulty at 0
```
davidmurdoch: Use 
```suggestion
      RPCQUANTITY_ZERO // we start the totalDifficulty at 0
```

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -0,0 +1,35 @@
+import assert from "assert";
```
davidmurdoch, line 1: Can you add a tests that change the starting default difficulty.

Also, lets test the increasing `totalDifficulty` (both with and without the default difficulty).

davidmurdoch (changes requested):
Really Awesome work. I committed a change, if you wouldn't mind reviewing it.

The gist of the change I made was because:

 * `totalDifficulty` shouldn't be included in block size calculations
 * `totalDifficulty` shouldn't be included in the block hash (which it previously was, because it was included in the `raw` array)

It seems to work now, and as a bonus, I figured out why our block size calculations were always off a little (PR for that coming soon!)

I also requested some more tests :-D

Thanks

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
eggplantzzz: sounds good to me

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+// helper to mine an arbitrary number of blocks using dummy transactions
+const mineBlocks = async (number: number, provider: EthereumProvider) => {
+  const [from] = await provider.send("eth_accounts");
+  for (let x = number; x > 0; x--) {
+    await provider.send("eth_subscribe", ["newHeads"]);
+
+    await provider.send("eth_sendTransaction", [{ from, to: from }]);
+
+    await provider.once("message");
```
davidmurdoch: If you don't need the transaction itself, you can do `await provider.send("evm_mine")` instead.

Oh, actually... now that #793 is merged (as of yesterday), you can do `await provider.send("evm_mine", [{blocks: number}])`, no (backwards :smile:) loop needed!

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -0,0 +1,99 @@
...
+          });
+          afterEach(async () => {
+            provider && (await provider.disconnect());
+          });
+
+          it("equals the block difficulty for the genesis block", async () => {
+            const block = await provider.send("eth_getBlockByNumber", ["0x0"]);
+            assert.strictEqual(
+              block.totalDifficulty,
+              `0x${DEFAULT_DIFFICULTY}`
```
davidmurdoch: If `DEFAULT_DIFFICULTY` is ever greater than 9 in the future this test will fail.

Other ways of solving this:

```
assert.strictEqual(
  block.totalDifficulty,
  Quantity.from(DEFAULT_DIFFICULTY).toString()
);
```

or 

```
assert.strictEqual(
  parseInt(block.totalDifficulty, 16),
  DEFAULT_DIFFICULTY
);
```

or

```
assert.strictEqual(
  BigInt(block.totalDifficulty),
  DEFAULT_DIFFICULTY // change DEFAULT_DIFFICULTY to a BigInt
);
```

and lots of others, I'm sure.

davidmurdoch: Can you also update other places that use the `0x{number}` pattern in these tests? Thanks!

eggplantzzz: Sorry, I meant to update these but must have overlooked them. Updated!

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
eggplantzzz: Oh cool! I'll merge in `develop` and update this!

eggplantzzz: I like deleting code.

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -0,0 +1,93 @@
...
+            assert.strictEqual(
+              block.totalDifficulty,
+              `0x${DEFAULT_DIFFICULTY.toString(16)}`
+            );
+          });
+          it("equals the sum of the difficulty of all blocks (hex)", async () => {
+            const numberOfBlocksToMine = Math.floor(Math.random() * 10 + 1);
+            await provider.send("evm_mine", [{ blocks: numberOfBlocksToMine }]);
+            const block = await provider.send("eth_getBlockByNumber", [
+              `0x${numberOfBlocksToMine}`
```
davidmurdoch: When `numberOfBlocksToMine` to mine is `10` it sends `"0x10"`, which is `16`

```suggestion
              `0x${numberOfBlocksToMine.toString(16)}`
```

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -0,0 +1,93 @@
...
+            assert.strictEqual(
+              block.totalDifficulty,
+              `0x${difficulty.toString(16)}`
+            );
+          });
+          it("equals the sum of the difficulty of all blocks (hex)", async () => {
+            const numberOfBlocksToMine = Math.floor(Math.random() * 10 + 1);
+            await provider.send("evm_mine", [{ blocks: numberOfBlocksToMine }]);
+            const block = await provider.send("eth_getBlockByNumber", [
+              `0x${numberOfBlocksToMine}`
```
davidmurdoch:
```suggestion
              `0x${numberOfBlocksToMine.toString(16)}`
```

davidmurdoch (changes requested):
Two more little "suggestions" and it's good to go!

PR #774, opened 2021-02-05
fix: handle exiting better for CLI

### LINKED ISSUE #773: Ganache 7 CLI doesn't respond correctly to Ctrl+C
When running Ganache 7's CLI, `Ctrl+C` doesn't work until after `Ctrl-D` is pressed (or put differently, I could only use `Ctrl+D` to reliably close the application). 

I dug into this a bit and added some logging to the top of CLI's `closehandler` method. Looks like the first press of `Ctrl+C` changes the server status from `1` to `4`, but does not completely shut down the process. 

image

To reproduce: 

1. Run the command line via `npm start` or the latest npm release (`npm install -g ganache@internal` then `ganache`)
2. Wait for server to completely start
3. Press `Ctrl+C`. Notice that the application doesn't close. You can press it any number of times and there doesn't seem to be any effect. 
4. Press `Ctrl+D`. At this point I received two behaviors. Either the process would immediately exit, or it would print a new line, after which `Ctrl+C` would work correctly. 

Note that in these reproduction steps I had not sent a request to the server. Also, this appears to be chain agnostic, as the same behavior occurs when specifying the `filecoin` flavor as well (when used in the proper branch).

mikeseese:
This PR fixes #773. I'm not sure what's different about windows here, but a couple of things that ensures things close (and reasonably gracefully):
1. Changing `process.exitCode = 0` to `process.exit(0)`. The [former](https://nodejs.org/dist/latest-v14.x/docs/api/process.html#process_process_exitcode) does not exit; it just sets the code for whenever the process does exit. However, the process would only exit if the server's `listen` (or some other dangling open function handle) all close out. We shouldn't wait for this, and apparently on Windows it does not close. This fixes the issue.
1. I also added a force quit option here if for some reason `await server.close()` takes a really long time. This was my original though of the bug, so I implemented this, only to find out that `await server.close()` was not the issue. I left it in for good measure. Do note I verified that [node uses 'Ctrl+C' terminology on Mac as well](https://github.com/nodejs/node/blob/master/lib/repl.js#L778)

### DIFF
--- a/src/packages/cli/src/cli.ts
+++ b/src/packages/cli/src/cli.ts
@@ -48,6 +48,7 @@ const server = Ganache.server(argv);
 console.log(detailedVersion);
 
 let started = false;
+let shuttingDown = false;
 process.on("uncaughtException", function (e) {
   if (started) {
     logAndForceExit([e], 1);
@@ -69,12 +70,20 @@ if (process.platform === "win32") {
 }
 
 const closeHandler = async function () {
+  if (shuttingDown) {
+    logAndForceExit(["User requested to force quit."], 1);
+    return; // should not get here because of process.exit() but just in case
+  }
+
   try {
     // graceful shutdown
     if (server.status === 1) {
+      console.log(
+        "Attempting to gracefully shut down Ganache. Press Ctrl+C again to force quit."
+      );
       await server.close();
     }
-    process.exitCode = 0;
+    process.exit(0);
   } catch (err) {
     logAndForceExit(
       [


davidmurdoch:
This bug is sort of my fault. As I changed it from process.exit() to exitCode=0 due to this comment in the node docs:

> In most situations, it is not actually necessary to call process.exit() explicitly. The Node.js process will exit on its own if there is no additional work pending in the event loop. The process.exitCode property can be set to tell the process which exit code to use when the process exits gracefully.

My chance may have just been highlighting an issue we have when shutting down on windows: something is keeping the event loop alive. I think we should fix (or at least investigate) that bug, instead of covering it up. My hunch is that the bug is in uWebsockets, so maybe we should still ship with this PR's fix, as that upstream fix could take some time.

We will have someone from truffle look in to this event loop issue more closely soon.

mikeseese:
Sounds good. In the meantime, this will force @tcoulter to have to kill node processes in the task manager manually to be able to test the CLI. I believe that this needs to be merged until someone has the time to look into it and change it back to setting the exit code later

At a minimum, we should merge everything except the `process.exit()` change so @tcoulter has an easy way to kill the process in the meantime

mikeseese:
Closing in favor of another solution

PR #777, opened 2021-02-08
chore: enabling building from root

davidmurdoch:
Each project extends the root tsconfig-base.json now. This file is now our centralized common compiler options.

We now have a "solution" tsconfig.json file that simply has a reference to our "leaf-node" project, `ganache`, with `"files"` set to an empty array (otherwise the solution file will cause double compilation of files).

This presents a simple entry point; e.g. we simply run `tsc --build` to build all endpoints because we list our "subproject" (ganache) as a reference in our "solution" `tsconfig.json`

You can find examples of this pattern in the typescript repo: [src/tsconfig_base.json](https://github.com/microsoft/TypeScript/blob/f2bcb2101be94a340ca88111aeb67c1474b30f2c/src/tsconfig-base.json), [src/tsconfig.json](https://github.com/microsoft/TypeScript/blob/f2bcb2101be94a340ca88111aeb67c1474b30f2c/src/tsconfig.json), and [src/tsc/tsconfig.json](https://github.com/microsoft/TypeScript/blob/f2bcb2101be94a340ca88111aeb67c1474b30f2c/src/tsc/tsconfig.json).

TypeScript documentation outlines this project structure here: https://www.typescriptlang.org/docs/handbook/project-references.html#overall-structure


.gitattributes
```diff
@@ -1,3 +1,4 @@
 docs/assets/launch.json linguist-language=JSON5
+tsconfig-base.json linguist-language=JSON5
```
davidmurdoch, line 2: This is just to tell github not to freak out over comments in the tsconfig-base.json file

package.json
```diff
@@ -7,16 +7,16 @@
...
-    "build": "lerna run build && shx cp -r ./src/packages/ganache/dist/web ./docs/assets/js/ganache",
+    "build": "lerna run build && shx cp -r src/packages/ganache/dist/web docs/assets/js/ganache",
     "clean": "npm run tsc.clean && npx lerna clean -y && npx shx rm -rf node_modules",
-    "create": "ts-node ./scripts/create",
-    "docs.build": "lerna run docs.build && ts-node ./scripts/build-docs/",
+    "create": "ts-node scripts/create",
+    "docs.build": "lerna run docs.build && ts-node scripts/build-docs/",
     "docs.preview": "lerna run docs.preview",
-    "postinstall": "lerna bootstrap && npx patch-package && ts-node ./scripts/link-ts-references.ts && ts-node ./scripts/postinstall",
+    "postinstall": "lerna bootstrap && npx patch-package && ts-node scripts/link-ts-references.ts && npm run tsc && ts-node scripts/postinstall",
```
davidmurdoch, line 15: Added `npm run tsc` is part of the postinstall, as it is a requirement for getting typescript references to work.

package.json
```diff
@@ -7,16 +7,16 @@
...
     "reinstall": "npm run clean && npm install",
     "start": "lerna exec --loglevel=silent --scope ganache -- npm run start --silent -- ",
     "test": "lerna exec -- npm run test",
-    "tsc": "lerna exec -- npm run tsc",
+    "tsc": "ttsc --build src",
```
davidmurdoch, line 19: the magic!

scripts/create.ts
```diff
@@ -108,8 +108,8 @@ process.stdout.write(`${COLORS.Reset}`);
   }
 
   // determines how many `../` are needed for package contents
-  const numDirectoriesAwayFromRoot = 2 + location.split(sep).length;
-  const relativePathToRoot = "../".repeat(numDirectoriesAwayFromRoot);
+  const numDirectoriesAwayFromSrc = 1 + location.split(sep).length;
```
davidmurdoch, line 114: the base config was moved to `src`

scripts/create.ts
```diff
@@ -212,6 +215,7 @@ describe("${packageName}", () => {
     const tests = join(dir, "tests");
     const src = join(dir, "src");
 
+    //@ts-ignore
```
davidmurdoch, line 218: The scripts are all run without a reference tsconfig, since those settings don't really matter so much... but ts complained here so I added these.

src/packages/flavors/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../../../tsconfig.json",
+  "extends": "../../tsconfig-base.json",
```
mikeseese, line 2: you missed `flavors/package.json`

src/packages/options/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../../../tsconfig.json",
+  "extends": "../../tsconfig-base.json",
```
mikeseese, line 2: you missed `options/package.json`

src/packages/utils/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../../../tsconfig.json",
+  "extends": "../../tsconfig-base.json",
```
mikeseese, line 2: you missed `utils/package.json`

src/tsconfig-base.json
```diff
@@ -16,7 +16,7 @@
     "noImplicitAny": false,
     "newLine": "lf",
     "lib": ["ES2020"],
-    "typeRoots": ["./node_modules/@types", "src/@types"],
+    "typeRoots": ["../node_modules/@types"],
```
mikeseese: I believe `typeRoots` is redundant now that you're not referencing something other than a `**/node_modules/@types`: https://www.typescriptlang.org/tsconfig#typeRoots

mikeseese (changes requested):
Some minor things. It may also be good to add something in `CONTRIBUTING.md` explaining that `npm run tsc` will only compile dependencies/subdependencies of `packages/ganache`. This is primarily useful for any newcomers adding a new package that hasn't yet hooked it all up

Alternatively, you can add all the leaf packages to the root `tsconfig.json`'s `references`. It shouldn't compile them more than once, but it would create a design that suggests users need to add it there for their package to compile (despite, yes only having `packages/ganache` would do anything that mattered for release at the very moment, but not always so in development)

src/tsconfig-base.json
davidmurdoch: appears you are correct! thanks!!!

davidmurdoch:
> Some minor things. It may also be good to add something in CONTRIBUTING.md explaining that npm run tsc will only compile dependencies/subdependencies of packages/ganache. This is primarily useful for any newcomers adding a new package that hasn't yet hooked it all up

@seesemichaelj Ah, that's a really good point. I've added a note to `CONTRIBUTING.md`.

mikeseese (approved):

PR #779, opened 2021-02-09
feat: log warning on legacy-style eth_getTransactionReceipt calls

eggplantzzz:
In Ganache v7, instamining will be enabled by default but implemented in a different way than v6.

In v6, a transaction is sent, Ganache mines the transaction, and then returns the transaction hash. This enables the user to immediately request the transaction receipt via `eth_getTransactionReceipt`. This is, however, not how nodes in the wild behave. In the wild, the transaction hash is returned to the user and then the user will need to wait until the transaction is mined before requesting the receipt. If they request the receipt before it is mined, they will receive a `null` response.

For v7, this behavior is modified to more accurately reflect what a user will experience from a node in the wild. This may be problematic for some folks upgrading as tests will probably break etc. In order to help them upgrade with more ease, a warning message is implemented here to help them understand the situation.

### DIFF
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -1201,16 +1201,19 @@ export default class EthereumApi implements types.Api {
     ]);
     if (transaction) {
       return receipt.toJSON(block, transaction);
-    } else if (
-      this.#options.miner.blockTime <= 0 &&
-      this.#options.miner.legacyInstamine !== true &&
+    }
+
+    const options = this.#options;
+    if (
+      options.miner.blockTime <= 0 &&
+      options.miner.legacyInstamine !== true &&
       this.#blockchain.isStarted()
     ) {
       // if we are performing non-legacy instamining, then check to see if the
       // transaction is pending so as to warn about the v7 breaking change
       const tx = this.#blockchain.transactions.transactionPool.find(txHash);
       if (tx != null) {
-        this.#options.logging.logger.log(
+        options.logging.logger.log(
           " > Ganache `eth_getTransactionReceipt` notice: the transaction with hash\n" +
             ` > \`${txHash.toString()}\` has not\n` +
             " > yet been mined. See https://trfl.co/v7-instamine for additional information."
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -1202,6 +1202,17 @@ export default class EthereumApi implements types.Api {
     if (transaction) {
       return receipt.toJSON(block, transaction);
     } else {
+      // check to see if the transaction is pending
+      const tx = this.#blockchain.transactions.transactionPool.find(txHash);
+      if (tx != null) {
+        this.#options.logging.logger.log(
+          " > Ganache `eth_getTransactionReceipt` warning: the requested transaction has not yet been mined."
+        );
+        this.#options.logging.logger.log(
+          " > After it has been mined you will be able to obtain the receipt. See https://trfl.co/v7-instamine for more details."
+        );
+      }
+
       return null;
     }
   }
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -1203,14 +1203,14 @@ export default class EthereumApi implements types.Api {
       return receipt.toJSON(block, transaction);
     }
 
+    // if we are performing non-legacy instamining, then check to see if the
+    // transaction is pending so as to warn about the v7 breaking change
     const options = this.#options;
     if (
       options.miner.blockTime <= 0 &&
       options.miner.legacyInstamine !== true &&
       this.#blockchain.isStarted()
     ) {
-      // if we are performing non-legacy instamining, then check to see if the
-      // transaction is pending so as to warn about the v7 breaking change
       const tx = this.#blockchain.transactions.transactionPool.find(txHash);
       if (tx != null) {
         options.logging.logger.log(
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -1201,20 +1201,23 @@ export default class EthereumApi implements types.Api {
     ]);
     if (transaction) {
       return receipt.toJSON(block, transaction);
-    } else {
-      // check to see if the transaction is pending
+    } else if (
+      this.#options.miner.blockTime <= 0 &&
+      this.#options.miner.legacyInstamine !== true &&
+      this.#blockchain.isStarted()
+    ) {
+      // if we are performing non-legacy instamining, then check to see if the
+      // transaction is pending so as to warn about the v7 breaking change
       const tx = this.#blockchain.transactions.transactionPool.find(txHash);
       if (tx != null) {
         this.#options.logging.logger.log(
-          " > Ganache `eth_getTransactionReceipt` warning: the requested transaction has not yet been mined."
-        );
-        this.#options.logging.logger.log(
-          " > After it has been mined you will be able to obtain the receipt. See https://trfl.co/v7-instamine for more details."
+          " > Ganache `eth_getTransactionReceipt` notice: the transaction with hash\n" +
+            ` > \`${txHash.toString()}\` has not\n` +
+            " > yet been mined. See https://trfl.co/v7-instamine for additional information."
         );
       }
-
-      return null;
     }
+    return null;
   }
 
   /**
--- a/src/chains/ethereum/ethereum/src/provider.ts
+++ b/src/chains/ethereum/ethereum/src/provider.ts
@@ -266,6 +266,8 @@ export default class EthereumProvider
             : JSON.stringify(params, null, 2).split("\n").join("\n   > ")
         }`
       );
+    } else {
+      options.logging.logger.log(method);
     }
   };
 
--- a/src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
@@ -6,7 +6,7 @@ describe("api", () => {
   describe("eth", () => {
     describe("getTransactionReceipt", () => {
       let provider: EthereumProvider;
-      let logger;
+      let logger, from;
 
       beforeEach(async () => {
         // create a logger to test output
@@ -22,6 +22,7 @@ describe("api", () => {
           }
         };
         provider = await getProvider({ logging: { logger } });
+        [from] = await provider.send("eth_accounts");
       });
 
       afterEach(() => {
@@ -29,14 +30,10 @@ describe("api", () => {
       });
 
       it("returns the receipt for the transaction", async () => {
-        const [from] = await provider.send("eth_accounts");
         await provider.send("eth_subscribe", ["newHeads"]);
 
         const hash = await provider.send("eth_sendTransaction", [
-          {
-            from,
-            to: from
-          }
+          { from, to: from }
         ]);
 
         // wait for the tx to be mined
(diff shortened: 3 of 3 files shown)


src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1202,6 +1202,17 @@ export default class EthereumApi implements types.Api {
     if (transaction) {
       return receipt.toJSON(block, transaction);
     } else {
+      // check to see if the transaction is pending
+      const tx = this.#blockchain.transactions.transactionPool.find(txHash);
```
eggplantzzz, line 1214: Do we want to only do this when `instamine` is enabled? It might actually be helpful in the general case but that is kind of starting a new pattern of behavior.

davidmurdoch: Yes, let's only perform this check if `this.#options.miner.blockTime <= 0 && this.#options.miner.legacyInstamine !== true && this.#blockchain.isStarted()` (in human words: we're instamining, but not legacy instamining, and the chain is not paused).

Additionally, can you cache the result of the `this.#options` lookup. These properly lookups for private (#options) fields is actually a bit on the expensive side (as each one currently get transpiled into multiple function calls), so I try to cache them whenever there are multiple property accesses in a single function.

Once these changes are in let's add some more tests to ensure these checks are working correctly.

To write a test that pauses the blockchain you just need to do `await provider.send("miner_stop", []);`. Oh, and let me know if you have questions as why this additional condition is needed.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1202,6 +1202,17 @@ export default class EthereumApi implements types.Api {
...
+      if (tx != null) {
+        this.#options.logging.logger.log(
+          " > Ganache `eth_getTransactionReceipt` warning: the requested transaction has not yet been mined."
+        );
+        this.#options.logging.logger.log(
+          " > After it has been mined you will be able to obtain the receipt. See https://trfl.co/v7-instamine for more details."
+        );
```
davidmurdoch: Let's bikeshed the message. My suggestion:

```suggestion
        this.#options.logging.logger.log(
          " > Ganache `eth_getTransactionReceipt` notice: the transaction with hash\n" +
            ` > \`${txHash.toString()}\` has not\n` +
            " > yet been mined. See https://trfl.co/v7-instamine for additional information."
);

```

renders as:

```console
 > Ganache `eth_getTransactionReceipt` notice: the transaction with hash
 > `0xf56ad903c7dcd10f1d108c0defc47125fc2909d306257362b001981e19bd257a` has not
 > yet been mined. See https://trfl.co/v7-instamine for additional information.
```

Additional notes:
 * I've shortened the overall message by removing the "After it has been mined" part, since we are already linking them out to additional information.
 * I've limited the length to 80 characters per line (ideally we'd measure the terminal width and handle wrapping that way... but since we don't have anything like that in place at the moment I've wrapped it manually)
 * I'm logging all lines in one call. The reason to do this is that on Windows calls to `console.log` are _asynchronous_ and can cause output to be "unexpectedly interleaved" https://nodejs.org/api/process.html#process_a_note_on_process_i_o (I haven't been really been enforcing this in the rest of the code-base, but really should be)

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -266,6 +266,8 @@ export default class EthereumProvider
             : JSON.stringify(params, null, 2).split("\n").join("\n   > ")
         }`
       );
+    } else {
+      options.logging.logger.log(method);
```
davidmurdoch, line 270: :+1:

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      beforeEach(async () => {
+        // create a logger to test output
+        logger = {
+          clearLoggedStuff: function () {
+            this.loggedStuff = "";
+          },
+          loggedStuff: "",
+          log: function (message) {
+            if (message) {
+              this.loggedStuff = this.loggedStuff.concat(message);
```
davidmurdoch: MDN recommends using `+=` instead of concat. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#performance


```suggestion
              this.loggedStuff += message;
```

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+            from,
+            to: from
+          }
+        ]);
+
+        // do not wait for the tx to be mined which will create a warning
+        await provider.send("eth_getTransactionReceipt", [hash]);
+        assert(
+          logger.loggedStuff.includes(
+            "Ganache `eth_getTransactionReceipt` warning"
```
davidmurdoch: If we end up changing the message above don't forget to update the test! :-)

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        const hash = await provider.send("eth_sendTransaction", [
+          {
+            from,
+            to: from
+          }
+        ]);
+
+        // do not wait for the tx to be mined which will create a warning
+        await provider.send("eth_getTransactionReceipt", [hash]);
```
davidmurdoch: Let's also check to ensure we don't return the receipt here (this test could be useful on it's own once we implement our in-memory receipt cache)

```suggestion
        const result = await provider.send("eth_getTransactionReceipt", [hash]);
        assert.strictEqual(result, null);
```

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+          hash
+        ]);
+        assert(receipt);
+      });
+
+      it("returns null if the transaction does not exist", async () => {
+        const result = await provider.send("eth_getTransactionReceipt", [
+          "0x0"
+        ]);
+        assert.strictEqual(result, null);
```
davidmurdoch, line 56: Make sure we don't accidentally log the message when we shouldn't.

```suggestion
        assert.strictEqual(result, null);

        assert(
          !logger.loggedStuff.includes(
            "Ganache `eth_getTransactionReceipt` warning"
          )
        );
```

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+            to: from
+          }
+        ]);
+
+        // wait for the tx to be mined
+        await provider.once("message");
+        const receipt = await provider.send("eth_getTransactionReceipt", [
+          hash
+        ]);
+        assert(receipt);
```
davidmurdoch, line 44: Make sure we don't accidentally log the message when we shouldn't.

```suggestion
        assert(receipt);
        
        assert(
          !logger.loggedStuff.includes(
            "Ganache `eth_getTransactionReceipt` warning"
          )
        );
```

davidmurdoch: Note: I didn't test this suggestion! :sweat:

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
davidmurdoch: Note: I didn't test this suggestion! :sweat:

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+          loggedStuff: "",
+          log: function (message) {
+            if (message) {
+              this.loggedStuff = this.loggedStuff.concat(message);
+            }
+          }
+        };
+        provider = await getProvider({ logging: { logger } });
+      });
+
```
davidmurdoch, line 31: I made suggestions to check the logs in each test below, so this change is probably useful:

```suggestion

      afterEach(() => {
        logger.clearLoggedStuff();
      });
      
```

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      it("returns null if the transaction does not exist", async () => {
+        const result = await provider.send("eth_getTransactionReceipt", [
+          "0x0"
+        ]);
+        assert.strictEqual(result, null);
+      });
+
+      it("logs a warning if the transaction hasn't been mined yet", async () => {
+        logger.clearLoggedStuff();
```
davidmurdoch: I made a suggestion above to add an `afterEach` that does this for us. If you accept the suggestion you should accept this one too!

```suggestion
```

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
```diff
@@ -0,0 +1,74 @@
+import getProvider from "../../helpers/getProvider";
+import assert from "assert";
+import EthereumProvider from "../../../src/provider";
+
+describe("api", () => {
+  describe("eth", () => {
+    describe("getTransactionReceipt", () => {
+      let provider: EthereumProvider;
```
davidmurdoch, line 8: Eventually we'll port more `eth_getTransactionReceipt` tests over, so let's add another `describe` block here. Something like

```typescript
describe("legacy instamine detection and warning", () => {
  let provider: EthereumProvider;
  // ...
});
  ```

File: src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts [hunk 12]
```diff
@@ -0,0 +1,74 @@
...
+      let logger;
+
+      beforeEach(async () => {
+        // create a logger to test output
+        logger = {
+          clearLoggedStuff: function () {
+            this.loggedStuff = "";
```
davidmurdoch, line 15: I always forget about this `this` magic! Love it!

davidmurdoch (changes requested):
Really great work on this! Thanks for not being afraid to just dive in to these ganache internals. I'm made some suggestions and requested some changes. We should probably talk about some of them... though I know you aren't a huge fan of bikeshedding... so i can always find someone else to be pedantic with if you'd prefer :grin:!

eggplantzzz:
Do note @davidmurdoch that I am not against bikeshedding in general. But when I don't feel strongly, I just let others that do decide :)

src/chains/ethereum/ethereum/tests/api/eth/getTransactionReceipt.test.ts
eggplantzzz: sounds good!

src/chains/ethereum/ethereum/src/api.ts
eggplantzzz: I'm ok with your suggestion. I'll make note of the Windows mention: I don't know if it matters but should we use `OS.EOL` instead of `\n`?

davidmurdoch: Ah, good question, Looks like Node.js uses `\n` for all `console.log`s (https://github.com/nodejs/node/blob/master/lib/internal/console/constructor.js#L276). Someone tried to "fix it", and it was closed as "won't fix": https://github.com/nodejs/node/issues/5038#issuecomment-179516030

So just `\n` is "safe". TIL!

eggplantzzz: super!

davidmurdoch (approved):

PR #783, opened 2021-02-12
WIP polyfill for uWebSocket.js

davidmurdoch:
Let's not delete this one for a while as we'll probably find some case where we'll need it for whatever reason.

This isn't a complete polyfill; it is only the stuff we actually use, er, well... it doesn't even cover that.

 * There is no websocket support yet, I was thinking we could shim `ws` in here, but the `upgrade` -> `connection` path gets weird. Doable still, but just a little weird.
 * HTTP test don't even pass (it looks like the node HTTP server doesn't actually close right away, so tests sometimes use a previous tests server /shrug)


mikeseese:
Closing in favor of a PR authored by me

PR #785, opened 2021-02-12
fix(cli): enable SIGINT to shutdown cli on Windows

davidmurdoch:
@seesemichaelj I liked the additional `console.log`gin you were outputting in one of your recent closed PRs, but didn't take time to find it. I'd appreciate your thoughts on these.

Here is normal graceful shutdown:

```console
Chain Id
==================
1337

Listening on 127.0.0.1:8545
Received shutdown signal: SIGINT
Shutting down…
Server has been shut down
~\Documents\work\ganache-core-copy [fix/windows-sigint +0 ~2 -0 !]> 
```

and if the user shuts down while ganache is starting:

```console
Ganache CLI v0.1.0 (ganache-core: 0.1.0)
Starting RPC server
Received shutdown signal: SIGINT
Server is currently starting; waiting…
Shutting down…
Server has been shut down
~\Documents\work\ganache-core-copy [fix/windows-sigint +0 ~2 -0 !]> 
```

```console
Chain Id
==================
1337

Listening on 127.0.0.1:8545
Received shutdown signal: SIGINT
Shutting down…

Received an error while attempting to close the server: 
{an error}
~\Documents\work\ganache-core-copy [fix/windows-sigint +0 ~2 -0 !]> 
```

### DIFF
--- a/src/packages/cli/index.ts
+++ b/src/packages/cli/index.ts
@@ -1,3 +1,3 @@
 import Ganache from "@ganache/core";
-export { ServerOptions, ProviderOptions } from "@ganache/core";
+export { ServerOptions, ProviderOptions, Status } from "@ganache/core";
 export default Ganache;
--- a/src/packages/cli/src/cli.ts
+++ b/src/packages/cli/src/cli.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 import Readline from "readline";
-import Ganache, { Status } from "../index";
+import Ganache, { Status } from "@ganache/core";
 import { $INLINE_JSON } from "ts-transformer-inline-file";
 import args from "./args";
 import {
--- a/src/packages/cli/src/cli.ts
+++ b/src/packages/cli/src/cli.ts
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
-import Ganache from "../index";
+import Readline from "readline";
+import Ganache, { Status } from "../index";
 import { $INLINE_JSON } from "ts-transformer-inline-file";
 import args from "./args";
 import {
@@ -20,7 +21,7 @@ const logAndForceExit = (messages: any[], exitCode = 0) => {
     (process.stdout as any)._handle.setBlocking(true);
   }
   try {
-    messages.forEach(console.log);
+    messages.forEach(message => console.log(message));
   } catch (e) {
     console.log(e);
   }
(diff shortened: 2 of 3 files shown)


src/packages/cli/src/cli.ts
```diff
@@ -20,7 +21,7 @@ const logAndForceExit = (messages: any[], exitCode = 0) => {
     (process.stdout as any)._handle.setBlocking(true);
   }
   try {
-    messages.forEach(console.log);
+    messages.forEach(message => console.log(message));
```
davidmurdoch, line 24: This was logging the message, the index, and the array for each message in `messages`. I fixed that so it only logs the message now.

src/packages/cli/index.ts
```diff
@@ -1,3 +1,3 @@
 import Ganache from "@ganache/core";
-export { ServerOptions, ProviderOptions } from "@ganache/core";
+export { ServerOptions, ProviderOptions, Status } from "@ganache/core";
```
davidmurdoch, line 2: Exporting as I wanted to use the server's actual `Status` flags in cli instead of hardcoding the status values there.

src/packages/cli/src/cli.ts
```diff
@@ -56,45 +57,68 @@ process.on("uncaughtException", function (e) {
...
+    switch (server.status) {
+      case Status.opening:
+        receivedShutdownSignal = true;
+        console.log("Server is currently starting; waiting…");
+        return;
+      case Status.open:
+        console.log("Shutting down…");
+        await server.close();
+        console.log("Server has been shut down");
+        break;
```
davidmurdoch, line 77: If we are currently opening it's best to wait until we're are fully opened before shutting down to avoid the possibility of DB corruption (or a db-induced SEGFAULT, which can happen!).

Thinking about it more, we actually do already guard against this here: https://github.com/trufflesuite/ganache-core/blob/8a7ba780a664e32b2210e32c3bfebe359e7cfc37/src/chains/ethereum/ethereum/src/blockchain.ts#L1128-L1134

So maybe this whole `Status.opening` case should be removed and we just let the internals take care of the problem. Yeah, I'm 95% sure external layers (like the CLI) shouldn't be concerned with safe-guarding the internals of the provider. I'd appreciate some other thoughts on this.

src/packages/cli/src/cli.ts
```diff
@@ -56,45 +57,68 @@ process.on("uncaughtException", function (e) {
...
+  const rl = (require("readline") as typeof Readline)
+    .createInterface({
+      input: process.stdin,
+      output: process.stdout
+    })
+    .on("SIGINT", () => {
+      // we must "close" the RL interface otherwise the process will think we
+      // are still listening
+      // https://nodejs.org/api/readline.html#readline_event_sigint
+      rl.close();
```
davidmurdoch, line 105: This is the actual fix to windows SIGINT handing problem.

src/packages/cli/src/cli.ts
```diff
@@ -56,45 +57,68 @@ process.on("uncaughtException", function (e) {
...
+process.on("SIGTERM", handleSignal);
+process.on("SIGHUP", handleSignal);
 
 async function startGanache(err: Error) {
   if (err) {
     console.log(err);
     process.exitCode = 1;
     return;
+  } else if (receivedShutdownSignal) {
+    closeHandler();
```
davidmurdoch, line 120: If we remove the `Status.opening` switch case I commented on above, this section will need to be removed too.

Just for completeness sake, removing this will (probably?) result in the line `Listening on 127.0.0.1:8545` being logged to the console, whereas it would not have appeared before.

src/packages/cli/src/cli.ts
```diff
@@ -106,5 +130,5 @@ async function startGanache(err: Error) {
     }
   }
 }
-
+console.log("Starting RPC server");
```
davidmurdoch, line 133: Is there any reason not to log more things on start up? Maybe this sort of thing might be annoying/unsightly since we don't currently have a way of setting a log level (I'm sure add `debug` support in "the future").

src/packages/core/index.ts
```diff
@@ -2,6 +2,7 @@ import ConnectorLoader from "./src/connector-loader";
 import { ProviderOptions, ServerOptions } from "./src/options";
 import Server from "./src/server";
 
+export { Status } from "./src/server";
```
davidmurdoch, line 5: Exporting as I wanted to use the server's actual Status flags in cli instead of hardcoding the status values there.

src/packages/cli/src/cli.ts
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
-import Ganache from "../index";
+import Readline from "readline";
+import Ganache, { Status } from "../index";
```
mikeseese: I feel like this should be

```suggestion
import Ganache, { Status } from "@ganache/core";
```

`cli/index.ts` doesn't do any magic, and there's no precedent to import already exported versions of a dependency

src/packages/cli/src/cli.ts
mikeseese: I think either method is fine. I don't see an issue with extra safe guards. The logic you have also gives extra logging so the user knows what's happening. I prefer the way you have it actually

src/packages/cli/src/cli.ts
mikeseese: I'm not sure if there's a reason to add more logging *here*. We log more in each of the flavor initializations. I'm not sure if this log is necessary, but I also don't see the harm in having it. A CLI doesn't have to be silent if it doesn't have verbosity settings; usually it's pretty verbose if the settings don't exist.

mikeseese (changes requested):

mikeseese (approved):

PR #791, opened 2021-02-15
feat: get Filecoin.ClientRetrieve to actually save a file and update IPFS

mikeseese:
This PR is a continuation of #714 and #719 (I have included IPFS updating in this PR as it was a good combination to reduce the number of PRs while still making sure they're not too big). The main difference in this PR is further maturation, technical correctness, and make sure there are sufficient tests.

After uploading a file under `/market` in the network inspector example, there is an option (that is commented out in the example code) to download the file. `Filecoin.ClientRetrieve` is supposed to save the file to a location on disk provided by the user in the `ref` argument. This PR adds the functionality of using `ref` and saving the file to disk. This is an initial implementation and will likely need to be adjusted later as use cases expand.

We also update IPFS to the latest version. `ipfs-http-server` depends on the latest https://github.com/hapijs/hapi which uses `node@12+` only features (private fields in JS). This means we can no longer run Filecoin tests on `node@10|11`; this PR updates that as well.


.github/workflows/push.yml
```diff
@@ -26,6 +26,10 @@ jobs:
       - run: npm test
         env:
           FORCE_COLOR: 1
+      - run: npm run test.filecoin
+        if: ${{ startsWith(matrix.node, '12.') || startsWith(matrix.node, '13.') || startsWith(matrix.node, '14.') }}
```
davidmurdoch: Can we just check for `!startsWith(matrix.node, '10.')` instead? That way, when we add node 15, and other future versions, these tests will automatically be included (we might even add in `latest` and `current` tested versions, if they add support: https://github.com/actions/setup-node/pull/58).

package.json
```diff
@@ -15,7 +15,8 @@
     "postinstall": "lerna bootstrap && npx patch-package && ts-node scripts/link-ts-references.ts && npm run tsc && ts-node scripts/postinstall",
     "reinstall": "npm run clean && npm install",
     "start": "lerna exec --loglevel=silent --scope ganache -- npm run start --silent -- ",
-    "test": "lerna exec -- npm run test",
+    "test": "lerna exec --ignore @ganache/filecoin* -- npm run test",
+    "test.filecoin": "npx lerna exec --scope @ganache/filecoin* -- npm run test",
```
davidmurdoch: Why are these changes needed?

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -213,7 +218,43 @@ export default class Blockchain extends Emittery.Typed<
...
+      );
+    }
+
+    const chunks = this.ipfsServer.node.files.read(new IPFS_CID(cid), {
+      timeout: 500 // Enforce a timeout; otherwise will hang if CID not found
+    });
+
+    for await (const chunk of chunks) {
+      try {
+        await fs.promises.appendFile(ref.path, chunk, "binary");
```
davidmurdoch: I have a couple of thoughts on this:

1. what happens if a file transfer fails part way through, do we leave a partially downloaded file at `ref.path`? Do we care if we do? I could see this happening if someone is playing with a large file, and then they kill our process while it is transferring. If we do care, should we instead write to a tmp file first then move that the ile to `ref.path` :grimacing:  (there are additional gotchas if we tried this, as moving files across a partition is slow and tricky)?
2. Could using a WriteStream instead of calling `appendFile` multiple times be better on performance? I'm concerned with open/closing a file many times on performance, as well as [Windows locking directories/files on newly created files](https://github.com/isaacs/node-graceful-fs/blob/24f88fd7f37829d34ba5df5edfd7fe20df55d74a/polyfills.js#L87-L89).
3. we could also skip the directory creation steps, only falling back to it if the initial write to the file fails.

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -235,8 +276,8 @@ export default class Blockchain extends Emittery.Typed<
       state: StorageDealStatus.Validating, // Not sure if this is right, but we'll start here
       message: "",
       provider: this.miner,
-      pieceCid: null,
-      size: size,
+      pieceCid: proposal.data.pieceCid,
+      size: proposal.data.pieceSize || size,
```
davidmurdoch: Can we only fetch `size` (line 682) if we are going to use it?

basically: delete line 262 then:

```typeScript
proposal.data.pieceSize || await this.getIPFSObjectSize(proposal.data.root.root.value)
```

If not, could we parallelized the calls to `getIPFSObjectSize` and `signProposal`-> `dagCBOR.util.cid`?

davidmurdoch (changes requested):
Looking good! Again, this is a high-level PR with some suggestions/comments. Let me know if you want to hop on zoom real quick to discuss anything.

.github/workflows/push.yml
mikeseese: It would need to be `${{ !startsWith(matrix.node, '10.') && !startsWith(matrix.node, '11.') }}` but yes, we can do that

package.json
mikeseese: Due to these requirements:
- Run all non-Filecoin tests on all NodeJS versions
- Only run Filecoin tests on NodeJS 12+ versions

That was realized where we can separate this into two scripts, hence these changes. Without these changes `npm run test` would fail on versions 12 and under.

If you would prefer that we don't use package.json scripts in the Github CI and manually call `lerna exec --ignore @ganache/filecoin* -- npm run test` instead of `npm run test` (and the same for filecoin), we can make those changes. I figured these changes were a more friendly dev experience (for me, others, grokability in the CI workflow documentation, etc) to clearly delineate the differences needed for CI to also be easily translatable in the NPM scripts. This allows developers, like yourself, that are focusing on supporting v10 to keep your node environment at v10 and still run `npm run test` and not have to bother with the filecoin changes.

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: 1. Wouldn't this just cause a data file corruption, which is hypothetically implied if you stop a service in the middle of saving data? I feel like we don't care about this scenario. The user may be equally confused if they stop the service and the file isn't there at all. If we really wanted to prevent confusion, the proper way is to scope creep (because it hasn't been used in any of the Filecoin dapps I've demoed and not in the support plan) and implement the subscription method `Filecoin.ClientRetrieveWithEvents` which dapps/users can monitor the progress of the retrieval and know whether or not the file was completely retrieved before shutting it down (or see that it was only partially retrieved when they can concur the file only partially exists).
2. It could help on performance for very large files. It could also open up a new issue related to partial downloads as what happens if the stream doesn't properly close. We can make this change if you want it; I personally don't think it's necessary at this time.
3. I consider the directory creation a `no-op` in terms of performance as it doesn't really affect Big O. I think the code is more readable from a logical stand point as is: first make sure the directory exists, then touch the file, then fill the file in chunks. However we can make this change if you really want it.

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: Ya I don't see why we can't fetch `size` only if we're going to use it

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: > If we do care, should we instead write to a tmp file first then move that the ile to ref.path 😬 (there are additional gotchas if we tried this, as moving files across a partition is slow and tricky)?

After thinking more, I believe this just reintroduces the problem. Moving files in some cases (i.e. across partitions as you mentioned) is the same as copying and deleting, rather than just changing the inode location, in which you would basically be doubling the file transfer ops. If we wanted to do this, we could follow the pattern browsers use during download and put the file in the same directory, but give it a suffix or prefix saying it's a partial file and rename it when it's complete

davidmurdoch: 1. It wouldn't cause corruption for many file types, the files would just be missing some information at the end (text files, some databases, some video and image formats, encrypted files, etc). An example of a popular implementation (not saying it's exactly right for us): Chrome names the file `{filename}.{ext}.crdownload` until it is done downloading, but that's likely just used for chrome's download manager pause/resume stuff. I'm okay just keeping the partially downloaded file if you are.
2. Not sure what you mean by `the stream doesn't partially close`. We just need to make sure we close it, no? Windows antivirus locking the file is still a concern I have here.
3. I agree on your point about readability, but also know that teeecchhnniiically it's not "correct", since it is still possible for the directory (and that empty file) to be deleted before we open a handle to the file. Oh, and I think `appendFile` creates the file (but not the directory) already.

mikeseese: > I'm okay just keeping the partially downloaded file if you are.

I'm personally fine with this; I don't think it's really an issue until it's reported as an issue. I the file suffix + rename on completion is a quick win if you'd prefer that. I don't think it's necessary to implement the extra method unless a user explicitly is asking for it.

> Not sure what you mean by the stream doesn't partially close. We just need to make sure we close it, no?

I was worried we may keep a lock on the file if our process gets force quit by the user, but after more thought I don't think is an issue.

> I agree on your point about readability, but also know that teeecchhnniiically it's not "correct", since it is still possible for the directory (and that empty file) to be deleted before we open a handle to the file.

I didn't consider this possibility. Your solution is also not technically correct because we would just pick up where we were at and have the first N bytes missing.

Using a write stream would keep the file open (and prevent the directory/file) from being closed while we're writing. I believe we keep the logic here, but add a write stream instead of append file. This should resolve both items 2 and 3

package.json
davidmurdoch: But now `npm test` doesn't run all tests that could affect filecoin, and I'd like for devs to be able to give a quick smoke test locally.

I'm going to chat internally about this a little more and get back to you!

mikeseese: Ok, do note this decision was made with the `CONTRIBUTING.md` guidelines still telling devs to use `node@10.7.0` and that you were going to figure out whatever the base nodejs version because of this whole "minimum v12 for filecoin". It felt that requirement for v12+ was the oddball here, not running in a lesser environment. happy change it so we have `test` and `test.v10_11` 🤷. whatever you want!

mikeseese: We could also make the `test` script more complicated to detect your nodejs version and change what gets ran

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: I went ahead and implemented the `.partial` suffix as well as the WriteStream. I believe these changes should resolve each of your concerns to some degree. Let me know if it's not enough!

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -255,6 +270,10 @@ export default class Blockchain extends Emittery.Typed<
         );
       }
     }
+
+    fileStream.close();
```
davidmurdoch: I think this just needs to be a `finally` block in order to get the :+1: from me!

package.json
davidmurdoch: Most of us don't like the idea of adding chain specific scripts to the root package.json, but no one came up with better alternatives.

My favorite idea is to make the filecoin package decide if its tests get run or not. Do you have any thoughts on a relatively "clean" way of doing that? Or can you think of any big downsides to doing it that way?

mikeseese: The cleanest way I could think of is yet another TS script that will check the current running NodeJS version; I couldn't find an environment variable for it, and not sure if checking an env variable in the command directly would even be windows friendly

The TS script could be a `is-atleast-node-12.ts` and has an exit code of 0 if it is and 1 otherwise and we can then use it like `ts-node is-atleast-node-12.ts && npm run test`. There could also be a whole dedicated TS script `test.ts` that checks the node version and then kicks off mocha as a sub process

davidmurdoch: From the outside it should look like the tests passed when run via node 10 and 11, which might be the tricky part here.

(You could determine the node version via `node -e 'process.exit(parseInt(process.version.slice(1)) < 12 ? 1 : 0)'` (untested).

Note that I don't care if we stuff a bunch of npm scripts into filecoin's package.json. So if you could think of a cross-platform way of doing that, I think we can originate things so it's maintainable.

mikeseese: > node -e 'process.exit(parseInt(process.version.slice(1)) < 12 ? 1 : 0)

This was the plan, but I was going to put it in a TS script because it is substantially cleaner than a long one-liner that hides in a json file. I can one-liner it if you'd prefer

davidmurdoch: No, a file would be better, I think.

scripts/skip-if-not-node-12.ts
```diff
@@ -0,0 +1,23 @@
+// To use this script, run `ts-node path/to/skip-if-not-node-12.ts || <command>`
```
davidmurdoch: rename to `scripts/skip-if-less-than-node-12.ts`?

davidmurdoch (approved):
Probably still need @tcoulter's Approval, as I haven't checked the Filecoin-specific bits whatsoever :-)

tcoulter (approved):
All good @seesemichaelj! Great work!

PR #798, opened 2021-02-18
perf: reduce size of published ganache bundle

### LINKED ISSUE #744: published ganache bundle is very large
we ship separate both cli and a core in one package, but cli should just use the bundled core as is.

The browser bundle is also a bit chunky, should this be shipped as a separate package? Or could we somehow make it use core (probably not, since it needs to polyfill many things)?

honestbonsai:
Resolves #744 

This PR removes the native node modules from `dist/cli`, which reduces the total `dist/cli` size from 29M  to 3.6M

TODO
1. ~Test `cli`~: `node ./src/packages/ganache/dist/cli/ganache.min.js`
2. ~Test `node`~
    ```
    // test.js
    const ganache = require("./src/packages/ganache/dist/node/ganache.min.js");
    console.log(ganache); // {server: ..., provider: ...}
    ```
3. ~Test `browser`~
    ```
    npm run build
    npm run docs.preview
    // Look at the interactive console, run some commands
    // const accounts = await provider.request({ method: "eth_accounts", params: []})
    // console.log(accounts)
    ```
4. ~Clean up, make things easier to understand~
5. ~Consider removing `cli` exports~ not doing


src/packages/ganache/webpack/webpack.cli.config.ts
```diff
@@ -9,7 +9,21 @@ const config: webpack.Configuration = merge({}, base, {
...
+    "keccak",
+    {
+      "@ganache/core": path.resolve(
+        __dirname,
+        "../",
+        "dist",
+        "node",
+        "ganache.min.js"
+      )
+    }
```
davidmurdoch: I think this will need to be something like `@ganache/core": path.join("../", "node", "ganache.min.js")`, otherwise your absolute path with be what is used to resolve `@ganache/core` in the bundle.

src/packages/ganache/webpack/webpack.common.config.ts
```diff
@@ -34,7 +34,6 @@ const base: webpack.Configuration = {
   output: {
     filename: "ganache.min.js",
     library: "Ganache",
-    libraryExport: "default",
```
davidmurdoch, line 39: Any thoughts on if this change has an effect elsewhere?

src/packages/ganache/index.ts
```diff
@@ -1,4 +1,12 @@
-import Ganache from "@ganache/cli";
+/*
+ * This file is the entry point for the resultant bundle dist/node/ganache.min.js
+ * dist/cli/ganache.min.js will then point dist/node/ganache.min.js
```
davidmurdoch:
```suggestion
 * dist/cli/ganache.min.js will then point to dist/node/ganache.min.js
```

src/packages/cli/package.json
```diff
@@ -48,10 +48,9 @@
     "tooling",
     "truffle"
   ],
-  "devDependencies": {
-    "@ganache/core": "0.1.0"
-  },
+  "devDependencies": {},
```
davidmurdoch:
```suggestion
```

davidmurdoch (changes requested):
Awesome work, @heyse! This is huge! (that was a bundle-size joke)

I've left some suggestions and comments. Let me know if you want to hop on zoom to review

src/packages/ganache/webpack/webpack.common.config.ts
honestbonsai: The only change I  noticed is that it manages to export the named exports in addition to the default export. I think if this is set to `default`, it only exports the default export.

honestbonsai: If we want to be really safe, I can isolate this change to just `webpack.node.config`

src/packages/ganache/webpack/webpack.cli.config.ts
honestbonsai: Hmm isn't `resolve` ok? It's basically like calling `cd __dirname` then the rest of the args are the constructed path you want to go to from that initial `cd`.

The output path above uses `resolve` as well.

davidmurdoch: Doing this writes _your_ path into the final bundle. So it will work on your machine, in that specific location, but not in another other machine or location.

src/packages/ganache/webpack/webpack.common.config.ts
davidmurdoch: Ah, ok.

Can you run some smoke tests locally to ensure we can still import and `require` both the `node` and `cli` bundles?

basically:

```javascript
// index.js
const ganache = require("{path-to-file}/ganache.min.js");
console.log(ganache) // should contain `provider` and `server` (`default` show up here, too)
```

```typescript
// index.ts, with no/default tsconfig.json (i.e., `esModuleInterop` is set to `false`)
import ganache from "{path-to-file}/ganache.min.js";
console.log(ganache) // should contain `provider` and `server`
```

davidmurdoch: important Note: I have NOT checked if this is how it behaved prior to your changes, but it's how I'd like it to behave, if possible. :-)

src/packages/ganache/webpack/webpack.cli.config.ts
honestbonsai: Cool, looks like that works

src/packages/ganache/webpack/webpack.common.config.ts
honestbonsai: Case 1

```
 node test.js
{ serverDefaults: [Getter],
  default:
   { server: [Function: server], provider: [Function: provider] } }
(node:11690) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

Case 2: `esModuleInterop` is set to `false`

```
npx ts-node test-2.ts
{ serverDefaults: [Getter],
  default:
   { server: [Function: server], provider: [Function: provider] } }
(node:14092) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

honestbonsai: Note I had to export `serverDefaults` to get `cli` working

davidmurdoch: Ah ok, so we might need to export `server` and `provider` in addition to the default; and probably export `serverDefaults` as part of `server` (maybe call it something like `server.defaults`). Are you up for making these changes?

honestbonsai: > Ah ok, so we might need to export `server` and `provider` in addition to the default; and probably export `serverDefaults` as part of `server` (maybe call it something like `server.defaults`). Are you up for making these changes?

Sure, but I don't really get why `serverDefaults` needs to be a part of `server`, as `cli` is expecting it to be a named export from `@ganache-core`

davidmurdoch: I don't think it really makes sense to have it exposed to the public interface at the level it is now. I'd really prefer if it wasn't exposed to users at all, but that'd likely be a lot more work.

davidmurdoch: I wonder if we can easily move `serverDefaults` from `packages/core` over to `packages/cli`; that way we won't need to expose it at all!

honestbonsai: > I wonder if we can easily move `serverDefaults` from `packages/core` over to `packages/cli`; that way we won't need to expose it at all!

I don't think this is possible as `core` uses it in `src/packages/core/src/options/index.ts`

honestbonsai: > important Note: I have NOT checked if this is how it behaved prior to your changes, but it's how I'd like it to behave, if possible. :-)

fyi this is how it is currently in `develop`

```
node test.js 
{ server: [Function: server], provider: [Function: provider] }
```

honestbonsai: After exported a named server and provider and patching the inferred type issue

```
node test.js 
{ serverDefaults: [Getter],
  server: [Function: server],
  provider: [Function: provider],
  default:
   { server: [Function: server], provider: [Function: provider] } }
```

PR #809, opened 2021-02-26
perf: optimize `debug_storageRangeAt`

davidmurdoch:


src/chains/ethereum/ethereum/src/@types/merkle-patricia-tree/index.d.ts
```diff
@@ -20,13 +20,17 @@ declare module "merkle-patricia-tree" {
     constructor(db: Database, root: Buffer);
     get(key: LargeNumber, cb: Callback<Buffer | null>): void;
     put(key: LargeNumber, value: LargeNumber, cb: Callback<never>): void;
-    copy(): Trie;
+    copy(): CheckpointTrie;
```
davidmurdoch, line 23: Just fixing some types

src/chains/ethereum/ethereum/src/@types/merkle-patricia-tree/secure.d.ts
```diff
@@ -0,0 +1,11 @@
+declare module "merkle-patricia-tree/secure" {
```
davidmurdoch, line 1: Just adding a type for SecureTrie

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -995,42 +999,42 @@ export default class Blockchain extends Emittery.Typed<
...
-    const transactionHash = Data.from(transaction.hash()).toString();
+    const transactionHash = transaction.hash();
+    let txHashCurrentlyProcessing: Buffer = null;
 
-    const beforeTxListener = async (tx: Transaction) => {
-      txHashCurrentlyProcessing = Data.from(tx.hash()).toString();
-      if (txHashCurrentlyProcessing == transactionHash) {
+    const beforeTxListener = (tx: Transaction) => {
+      txHashCurrentlyProcessing = tx.hash();
+      if (txHashCurrentlyProcessing.equals(transactionHash)) {
```
davidmurdoch, line 1008: Comparing buffers is faster than converting them to a string then comparing strings.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -995,42 +999,42 @@ export default class Blockchain extends Emittery.Typed<
...
-              Data.from(key).toString()
-            );
-
-            vm.stateManager.getContractStorage(
-              Address.from(contractAddress).toBuffer(),
-              rawKey,
-              (err: Error, result: Buffer) => {
-                if (err) {
-                  throw err;
+          const database = this.#database;
```
davidmurdoch, line 1010: For reasons, `this.#database` is slow once typescript compiles things, so I like to cache the lookup.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -995,42 +999,42 @@ export default class Blockchain extends Emittery.Typed<
...
+          return Promise.all(
```
davidmurdoch, line 1011: The `beforeTx` event emitter awaits the returned promise. I didn't find any race condition issues in our tests, but this seemed like a place they could crop up, so I thought it'd be good to "fix" it anyway.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -995,42 +999,42 @@ export default class Blockchain extends Emittery.Typed<
...
+            keys.map(async key => {
+              // get the raw key using the hashed key
+              const rawKey: Buffer = await database.storageKeys.get(key);
```
davidmurdoch, line 1014: I've changed the way the keys are stored so we don't need to convert to a string any more.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -995,42 +999,42 @@ export default class Blockchain extends Emittery.Typed<
...
+                rawKey,
+                (err: Error, result: Buffer) => {
+                  if (err) {
+                    throw err;
+                  }
+
+                  storage[Data.from(key, key.length).toString()] = {
+                    key: Data.from(rawKey, rawKey.length),
+                    value: Data.from(result, 32)
+                  };
```
davidmurdoch, line 1027: `Data` is pretty cool in that when it is JSON stringified it gets converted to RPC DATA. So there is no need to convert `.toJSON()` here.

By default, data remove leading `0`s, so if we want to keep them, which I believe we do in these cases, we need to specify the length.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1254,10 +1248,8 @@ export default class Blockchain extends Emittery.Typed<
     );
 
     // get the contractAddress account storage trie
-    const addressDataPromise = this.getFromTrie(
-      trie,
-      Address.from(contractAddress).toBuffer()
-    );
+    const contractAddressBuffer = Address.from(contractAddress).toBuffer();
```
davidmurdoch, line 1252: We use this buffer a few times, so let's compute it just once.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1274,42 +1266,49 @@ export default class Blockchain extends Emittery.Typed<
...
-              }
-            });
+      return new Promise<T>((resolve, reject) => {
+        const startKeyBuffer = Data.from(startKey).toBuffer();
+        const compare = (a: Buffer, b: Buffer) => a.compare(b) < 0;
+
+        const keys: Buffer[] = [];
+        const handleData = ({ key }) => {
+          // ignore anything that comes before our starting point
+          if (startKeyBuffer.compare(key) > 0) return;
```
davidmurdoch, line 1277: If the key comes _before_ our start key we can just ignore it. No need to sort data we know we'll throw away!

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1274,42 +1266,49 @@ export default class Blockchain extends Emittery.Typed<
...
+
+          // #4 - sort and filter keys
+          // insert the key exactly where it needs to go in the array
+          const position = findInsertPosition(keys, key, compare);
+          // ignore if the value couldn't possibly be relevant
+          if (position > maxResult) return;
+          keys.splice(position, 0, key);
```
davidmurdoch, line 1284: `findInsertPosition` does a binary search on a _sorted_ array to find the optimal index to insert a new item. We do NOT need to worry about this being a stable sort, since we are sorting the keys, and there will be only one of each key.

There _are_ other data structures better suited for fast insertion, but ultimately we'll need to convert it into an array anyway, so my _guess_ is that an array would end up winning out in practical use anyway.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1274,42 +1266,49 @@ export default class Blockchain extends Emittery.Typed<
...
 
-            // only take the maximum number of entries requested
-            keys = keys.slice(0, maxResult + 1);
-            if (keys.length > maxResult) {
-              // assign nextKey and remove it from array of keys
-              const nextKey = keys.pop();
-              result.nextKey = Data.from(nextKey).toJSON();
-            }
+        const handleEnd = () => {
+          if (keys.length > maxResult) {
```
davidmurdoch, line 1288: If we have more data than we want to return we need to trim it down, and grab the `nextKey` as well.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1274,42 +1266,49 @@ export default class Blockchain extends Emittery.Typed<
...
+              keys,
+              nextKey: null
+            });
+          }
+        };
 
-            resolve(keys);
-          });
+        const rs = storageTrie.createReadStream();
+        rs.on("data", handleData).on("error", reject).on("end", handleEnd);
```
davidmurdoch, line 1305: In case you miss it... I added an error handler here.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1274,42 +1266,49 @@ export default class Blockchain extends Emittery.Typed<
...
+            resolve({
+              keys,
+              nextKey: null
+            });
+          }
+        };
 
-            resolve(keys);
-          });
+        const rs = storageTrie.createReadStream();
```
davidmurdoch, line 1304: We can't get keys to come back already ordered because the merkle patricia tree library we use parallelizes up to 500 database requests at once.

This just means we have to sort client side, which could mean performance takes a hit.

But generally I think callers to `debug_storageRangeAt` will attempt to consume all available storage in one go, which means we would have had to load all the data anyway.

Where we'd take a hit is when we only need to return a small amount of storage. In that case we still have to load all of it.

I think turning the priority queue it uses internally into a FIFO queue would actually be very complicated, because it resolves both branches and leaves, and it'd have to ensure the proper ordering there. Not positive though.

Since that's not going to happen any time soon, the next best thing is to just insert them in order as we get them, instead of a big oll sort at the end.

This will suck a bit more after we get the in-memory cache in place. We won't need any parallelization in that case; and it'll actually just slow everything down (it'll still be faster than hitting leveldb, of course).

Thanks for watching. Don't forget to smash that like button!

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1318,7 +1317,7 @@ export default class Blockchain extends Emittery.Typed<
     // get storage data given a set of keys
     const options = {
       disableMemory: true,
-      disableStack: false,
+      disableStack: true,
```
davidmurdoch, line 1321: Didn't seem like we needed to trace the stack, and the tests still passed. I didn't look into this any further. :sweat_smile:

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -49,7 +49,10 @@ export default class Database extends Emittery {
   }
 
   #initialize = async () => {
-    const levelupOptions: any = { valueEncoding: "binary" };
+    const levelupOptions: any = {
+      keyEncoding: "binary",
```
davidmurdoch, line 53: This makes it so that we don't have to stringify our keys, which are almost always, if not always, already a Buffer.

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -231,9 +232,9 @@ export default class Miner extends Emittery.Typed<
         if (event.opcode.name === "SSTORE") {
           const key = TraceData.from(
             event.stack[event.stack.length - 1].toArrayLike(Buffer)
-          );
-          const hashedKey = Data.from(keccak(key.toBuffer())).toString();
-          storageKeys.set(hashedKey, key.toBuffer());
+          ).toBuffer();
+          const hashedKey = keccak(key);
+          storageKeys.set(hashedKey.toString(), { key, hashedKey });
```
davidmurdoch, line 237: Trading some memory for CPU. Previously the consumer of this data would convert the Map's keys (`hashedKey.toString()`) back into Buffers before saving them to the database. Now it just reads the Buffers stored in the values of the Map.

src/packages/utils/src/utils/find-insert-position.ts
```diff
@@ -0,0 +1,49 @@
...
+    offset += step;
+    if (!comp(val, array[offset])) {
+      first = ++offset;
+      count -= step + 1;
+    } else {
+      count = step;
+      offset = first;
+    }
+  }
+  return first;
```
davidmurdoch: Binary searches a sorted array for the sorted insert position.

This implementation was ported from the C++ stdlib, http://www.cplusplus.com/reference/algorithm/upper_bound/.

src/chains/ethereum/ethereum/src/blockchain.ts
cds-amal: How did you measure the performance difference @davidmurdoch ?

davidmurdoch: Don't need to. Just look at what typescript generates for our ES target. It's a function call, a Map `has` call, a branch, and a Map `get` call, IIRC.

I'd love to generate multiple builds so we can ship without these sorts of polyfills. I like the "true" private fields the syntax provides, and don't think I want to remove their use all together.

Happy to have your review on this!

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1274,42 +1266,50 @@ export default class Blockchain extends Emittery.Typed<
...
-            const sortedKeys = keys.sort((a, b) => Buffer.compare(a, b));
-
-            // find starting point in array of sorted keys
-            const startKeyBuffer = Data.from(startKey).toBuffer();
-            keys = sortedKeys.filter(key => {
-              if (Buffer.compare(startKeyBuffer, key) <= 0) {
-                return key;
-              }
-            });
+      type RangedStorageKeys = { keys: Buffer[]; nextKey: Data };
```
eshaben: should this type be moved to utils similarly to how `StorageRangeResult` type was handled?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1274,42 +1266,50 @@ export default class Blockchain extends Emittery.Typed<
...
+        const compare = (a: Buffer, b: Buffer) => a.compare(b) < 0;
+
+        const keys: Buffer[] = [];
+        const handleData = ({ key }) => {
+          // ignore anything that comes before our starting point
+          if (startKeyBuffer.compare(key) > 0) return;
+
+          // #4 - sort and filter keys
+          // insert the key exactly where it needs to go in the array
+          const position = findInsertPosition(keys, key, compare);
```
eshaben, line 1281:
```suggestion
          const position: number = findInsertPosition(keys, key, compare);
```

src/packages/utils/src/utils/find-insert-position.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @returns The index to the upper bound position for `val` in the range. If no
+ * element in the range compares greater than `val`, the function returns
+ * `array.length`.
+ */
+export function findInsertPosition<T>(
+  array: T[],
+  val: T,
+  comp: (a: T, b: T) => boolean
+) {
```
eshaben:
```suggestion
): number {
```

eshaben (commented):
just a couple small things... looks good though! thanks for cleaning it up... `keyStart`  😆

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: I'm okay with not adding the type annotation here as I've I added your other suggestion.

eshaben (approved):
LGTM

src/packages/utils/src/utils/find-insert-position.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * AKA `upper_bound`
+ *
+ * The elements are compared `comp`. The elements in the range shall already be
+ * sorted according to this same criterion (`comp`), or at least partitioned
+ * with respect to val.
```
gnidan: Written more actively, this paragraph's purpose might be easier to spot at-a-glance: _"as a precondition, elements in `array` must be..."_ (e.g.) instead of  _"shall already be"_ (e.g.)

src/packages/utils/src/utils/find-insert-position.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * The function optimizes the number of comparisons performed by comparing
+ * non-consecutive elements of the sorted range.
+ *
+ * The index into the `array` returned by this function cannot be equivalent to
+ * `val`, only greater.
```
gnidan:
```suggestion
 * The index into the `array` returned by this function will always be greater than
 * the index of the last-occurrence of `val`.
```

Is this right?

src/packages/utils/src/utils/find-insert-position.ts
```diff
@@ -0,0 +1,49 @@
...
+    offset += step;
+    if (!comp(val, array[offset])) {
+      first = ++offset;
+      count -= step + 1;
+    } else {
+      count = step;
+      offset = first;
+    }
+  }
+  return first;
```
gnidan: Might be worth commenting this a bit throughout the body here, just to describe the process of what's happening

src/chains/ethereum/ethereum/src/blockchain.ts
gnidan: Shouldn't need to trace the stack, no

gnidan (approved):
Here's a couple style suggestions. This looks good!

src/packages/utils/src/utils/find-insert-position.ts
davidmurdoch: The original comment was "the value pointed by the iterator returned by this function cannot be equivalent to val, only greater." I think yours says the same thing, but better.

src/packages/utils/src/utils/find-insert-position.ts
davidmurdoch: Good point, this comment was ported from the C++ docs, which is actually applied to vectors, which have some behaviors we don't replicate with JS arrays. Fix incoming.

PR #816, opened 2021-02-26
fix: make random seed actually random

davidmurdoch:
The `alea` RNG requires a starting seed (seedrandom doesn't).

Also, the types are wrong, so I removed them and opened an issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/51456

### DIFF
--- a/src/chains/ethereum/ethereum/package.json
+++ b/src/chains/ethereum/ethereum/package.json
@@ -81,7 +81,6 @@
     "@types/fs-extra": "9.0.2",
     "@types/lodash.clonedeep": "4.5.6",
     "@types/secp256k1": "4.0.1",
-    "@types/seedrandom": "2.4.28",
     "@types/ws": "7.2.7",
     "cheerio": "1.0.0-rc.3",
     "fs-extra": "9.0.1",
--- a/src/chains/ethereum/ethereum/package.json
+++ b/src/chains/ethereum/ethereum/package.json
@@ -81,6 +81,7 @@
     "@types/fs-extra": "9.0.2",
     "@types/lodash.clonedeep": "4.5.6",
     "@types/secp256k1": "4.0.1",
+    "@types/seedrandom": "3.0.0",
     "@types/ws": "7.2.7",
     "cheerio": "1.0.0-rc.3",
     "fs-extra": "9.0.1",
--- a/src/chains/ethereum/options/src/wallet-options.ts
+++ b/src/chains/ethereum/options/src/wallet-options.ts
@@ -1,9 +1,9 @@
 import { normalize } from "./helpers";
-import Seedrandom from "seedrandom";
+import seedrandom from "seedrandom";
 import { entropyToMnemonic } from "bip39";
 import { Definitions, DeterministicSeedPhrase } from "@ganache/options";
 
-const unseededRng = Seedrandom();
+const unseededRng = seedrandom();
 
 /**
  * WARNING: to maintain compatibility with ganache v2 this RNG only generates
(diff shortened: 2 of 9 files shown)


src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -244,7 +243,7 @@ export const WalletOptions: Definitions<WalletConfig> = {
     default: config =>
       config.deterministic
         ? DeterministicSeedPhrase
-        : randomAlphaNumericString(10, alea()),
+        : randomAlphaNumericString(10, new Alea(unseededRng)),
```
mikeseese: I don't believe this is correct.

Here shows the prng of calling `(new Seedrandom())()`:

image

But using the code you provided I get a predictable prng from alea:

image

Perhaps this should be:

```suggestion
        : randomAlphaNumericString(10, unseededRng),
```

I don't see a need to use Alea here, otherwise it would need to be `new Alea(unseededRng())`

mikeseese (changes requested):

src/chains/ethereum/options/src/wallet-options.ts
davidmurdoch: Ah, I good catch. I think I just missed _calling_ the seedrandom instance.

src/chains/ethereum/ethereum/package.json
```diff
@@ -81,7 +81,6 @@
     "@types/fs-extra": "9.0.2",
     "@types/lodash.clonedeep": "4.5.6",
     "@types/secp256k1": "4.0.1",
-    "@types/seedrandom": "2.4.28",
```
mikeseese, line 84: BTW, seems like this was fixed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51609

and IIRC how DefinitelyTypesd works, this should be upgraded to `@types/seedrandom@3.0.0`

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -1,10 +1,9 @@
 import { normalize } from "./helpers";
-import seedrandom from "seedrandom";
+import Seedrandom, { alea as Alea } from "seedrandom";
 import { entropyToMnemonic } from "bip39";
-
 import { Definitions, DeterministicSeedPhrase } from "@ganache/options";
 
-const { alea } = seedrandom;
+const unseededRng = new Seedrandom();
```
mikeseese: Per the [documentation](https://www.npmjs.com/package/seedrandom#cjs--nodejs-usage), I believe this should be (it fails to compile when using the latest types)

```suggestion
const unseededRng = Seedrandom();
```

src/chains/ethereum/options/src/wallet-options.ts
davidmurdoch:
```suggestion
        : randomAlphaNumericString(10, unseededRng),
```

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -244,7 +243,7 @@ export const WalletOptions: Definitions<WalletConfig> = {
     default: config =>
       config.deterministic
```
davidmurdoch:
```suggestion
      config.deterministic === true
```

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -1,10 +1,9 @@
 import { normalize } from "./helpers";
-import seedrandom from "seedrandom";
+import Seedrandom, { alea as Alea } from "seedrandom";
```
davidmurdoch:
```suggestion
import Seedrandom from "seedrandom";
```

PR #817, opened 2021-02-26
feat: allow starting up with 0 accounts

### LINKED ISSUE #60: Ability to specify zero accounts created/unlocked
## Expected Behavior
I would like to be able to specify zero accounts created/unlocked when starting ganache.

## Current Behavior
Setting` options = { total_accounts: 0 }` does not work as it will default to 10 accounts if set to `0` due to this line: https://github.com/trufflesuite/ganache-core/blob/develop/lib/statemanager.js#L42.

## Context
I am trying to test situations where a user signs out of MetaMask. If a user goes from a logged-in state to a logged-out state in MetaMask, the web3 object persists in the browser but the number of unlocked accounts available goes to zero. I would like to be able to test this behaviour in my code.

davidmurdoch:
Fixes https://github.com/trufflesuite/ganache-core/issues/60


src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -246,13 +246,13 @@ export default class Wallet {
         }
       }
     } else {
-      const numerOfAccounts = options.totalAccounts;
-      if (numerOfAccounts) {
-        accounts = Array(numerOfAccounts);
+      const numberOfAccounts = options.totalAccounts;
+      if (numberOfAccounts != null) {
```
eggplantzzz, line 255: If you made this a guard instead, like using `numberOfAccounts == null`, then you could have less indentation. Just an idea, not important.

src/packages/cli/src/initialize/ethereum.ts
```diff
@@ -6,41 +6,43 @@ export default function (provider: Provider, cliSettings: CliSettings) {
...
-    const strBalance = balance / ethInWei;
-    const about = balance % ethInWei === 0n ? "" : "~";
-    let line = `(${index}) ${toChecksumAddress(
-      address
-    )} (${about}${strBalance} ETH)`;
-
-    if (!accounts[address].unlocked) {
-      line += " 🔒";
-    }
+  if (addresses.length > 0) {
```
eggplantzzz, line 14: Would it be better to still print this information? Like tell them they have no accounts unlocked? It could be something like
```
Available Accounts
==================
(no accounts unlocked)
```

What do you think?

davidmurdoch: I only did it this way because we already omit sections in a similar fashion. But doing what you are suggesting seems more helpful. :+1:

PR #836, opened 2021-03-04
implement Filecoin database persistence

mikeseese:
This PR adds database persistence to the Filecoin branch. It moves most of the data objects to using database `Manager` classes to mimic how the `ethereum` flavor does it; see https://github.com/trufflesuite/ganache-core/tree/develop/src/chains/ethereum/ethereum/src/data-managers and https://github.com/trufflesuite/ganache-core/blob/develop/src/chains/ethereum/ethereum/src/database.ts if you would like to see the original classes that were used for this work.


tcoulter (approved):
Looks good!

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -63,56 +68,90 @@ export default class Blockchain extends Emittery.Typed<
...
+        const genesisTipset = new Tipset({
+          blocks: [genesisBlock],
+          height: 0
+        });
+
+        this.tipsetManager.earliest = genesisTipset; // initialize earliest
+        await this.tipsetManager.putTipset(genesisTipset); // sets latest
+        await this.#database.db!.put(
+          "latest-tipset",
+          Quantity.from(0).toBuffer()
```
davidmurdoch: This is the equivalent of `Buffer.from([])` (no data), aka `utils.BUFFER_EMPTY`, where `utils` is a named export of `@ganache/utils` (sorry about the sloppy export here :sweat_smile:). If you want `Buffer.from([])` you can use `BUFFER_EMPTY`, and for `Buffer.from([0])` there is a `BUFFER_ZERO`.

Quantity/Data are _intended_ to be used to convert to/from json-rpc QUANTITY/DATA.

Note: these instances aren't frozen, so someone could theoretically do `BUFFER_ZERO[0] = {not zero};`, so these instances should never be used outside ganache internals.

Not requesting a change here, just making sure it is returning what you expect.

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -63,56 +68,90 @@ export default class Blockchain extends Emittery.Typed<
...
-    setTimeout(async () => {
       await this.ipfsServer.start();
 
+      // Fire up the miner if necessary
+      if (this.options.miner.blockTime > 0) {
+        const intervalMine = () => {
+          this.mineTipset();
+        };
+
+        this.miningTimeout = setInterval(
```
davidmurdoch, line 137: You should probably use a recursive `setTimeout` pattern here, it's what ganache has always done (and still does in v7). I think of this setting more like we are _targeting_ `blockTime` blocks, not to average 1 block every `blockTime`. in Ethereum-flavored ganache, if block `1` takes a long time to mine, block `2` won't be mined until `blockTime` seconds _after_ the previous block completed mining.

src/chains/filecoin/filecoin/src/database.ts
```diff
@@ -1,7 +1,173 @@
...
+
+  #initialize = async () => {
+    const levelupOptions: any = { valueEncoding: "binary" };
+    const store = this.#options.db;
+    let db: LevelUp;
+    if (store) {
+      this.#rootStore = encode(store, levelupOptions);
+      // @ts-ignore - I get an error in ts-node that I don't get running ttsc:
+      // Argument of type 'AbstractLevelDOWN<any, any>' is not assignable to parameter of type 'string'
+      // But this error just doesn't make any sense. I'm just going to ignore the error
```
davidmurdoch, line 55: I just started getting this error after updating some dependencies. Some type def in some transitive package must have erroneously changed things on us :expressionless:

src/chains/filecoin/filecoin/src/database.ts
```diff
@@ -1,7 +1,173 @@
...
+      const ret = fn();
+      // PSA: don't let vscode (or yourself) rewrite this to `await` the
+      // `batch.write` call. The `finally` block needs to run _before_ the
+      // write promise has resolved.
+      prom = batch.write().then(() => ret);
+    } finally {
+      rootDb.put = originalPut;
+      rootDb.del = originalDel;
+    }
+    return prom;
```
davidmurdoch, line 131: I'm curious if you reviewed this batch implementation of mine at all. It let's us do nice things and all, but I do worry that I missed something.

src/chains/filecoin/types/src/data-managers/manager.d.ts
```diff
@@ -0,0 +1,23 @@
...
+import { LevelUp } from "levelup";
+import { SerializableObject, BaseConfig } from "../things/serializable-object";
+export declare type Instantiable<T> = {
+  new (...args: any[]): T;
+};
+export default class Manager<
+  T extends SerializableObject<C>,
+  C extends BaseConfig
+> {
+  #private;
```
davidmurdoch, line 11: Do these do anything?

src/packages/utils/src/utils/random-number-generator.ts
```diff
@@ -1,13 +1,24 @@
...
+  // here as it still applies and is valid code.
+  // https://github.com/trufflesuite/ganache-core/issues/756
+  constructor(seed?: string | null, state?: seedrandom.State) {
+    if (typeof seed === "string" && typeof state === "undefined") {
+      this.rng = seedrandom.alea(seed, { state: true });
+    } else if (typeof state === "object") {
+      // We can ignore seed even if it was provided.
+      // The user is reseeding the rng from a prior state,
+      // so let's initialize accordingly
+      this.rng = seedrandom.alea("", { state });
```
davidmurdoch, line 18: neat!

davidmurdoch (changes requested):
some comments/suggestions

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: Since `Buffer.from([...numbers])` expects `numbers` to each individually only be 8 bits, I think using some sort of serializable format to create our buffer array is a better approach. And I want it the initialization to be consistent with the further setting.

FWIW, I did check, but `Quantity.from(0).toBuffer()` is the equivalent of `Buffer.from([0])`:

```
> utils.Quantity.from(0)
0
> utils.Quantity.from(0).toBuffer()
<Buffer 00>
> Buffer.from([])
<Buffer >
> Buffer.from([0])
<Buffer 00>
```

I believe this is exactly what I want :) Thanks for making me double check though!

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: I'm fine with doing it with `setTimeout`; makes more sense to be consistent with the Ethereum flavor. It also is likely more accurate how chains work anyway. This implementation was actually from  Tim and just got up for reviewal due to code moving.

mikeseese: Since the mining logic was moved to a separate function as part of #842, can I implement the `setTimeout` logic change in that PR instead of this one?

src/chains/filecoin/filecoin/src/database.ts
mikeseese: I didn't, but I can! I kind of took the database stuff for granted

src/chains/filecoin/filecoin/src/blockchain.ts
davidmurdoch: Ah, well.... that's a bug. I guess the fix for that is only in my forking branch right now. Sorry!. It should return `BUFFER_EMPTY`, and it will in `develop` in the future (as ethereum's rlp encoding expects `0 == Buffer.from([])`.

Again, sorry I made you double check for no good reason.

mikeseese: We talked about this, and I'm going to change this to use `utils.uintToBuffer`

src/chains/filecoin/types/src/data-managers/manager.d.ts
mikeseese: Not that I know of, but everything under `src/chains/filecoin/types/src` is autogenerated by `ttsc` with the `npm run tsc.filecoin.declarations` command

src/chains/filecoin/filecoin/src/database.ts
mikeseese: 👌 This implementation looks sane to me. I think the only thing I'd add was a test to ensure the atomicity. I see that this is used in `blockchain.ts::#saveNewBlock` and works (aka tests work), but I do think a targeted set of tests would be good to verify this logic works as it looks. However, I don't see why it wouldn't work!

src/chains/filecoin/filecoin/src/database.ts
mikeseese: Did you want me to get to the bottom of it or not bother?

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: Deferred to #842: https://github.com/trufflesuite/ganache-core/pull/842#event-4466881006

src/chains/filecoin/filecoin/src/database.ts
davidmurdoch: eh, not bother.

src/chains/filecoin/filecoin/src/database.ts
davidmurdoch: Issue opened for that. Thanks! https://github.com/trufflesuite/ganache-core/issues/873

davidmurdoch (approved):

PR #837, opened 2021-03-04
implement multiple accounts through the db

mikeseese:
This PR adds the support necessary to generate a configurable amount of accounts, instead of only having one account. The implementation allows for recovery of all of those accounts on subsequent runs using the same DB without needing to specify the same wallet options (seed, total accounts, etc). Users only need to run `ganache filecoin --database.dbPath </path/to/db>` to resume where they left off.

### DIFF
--- a/src/chains/filecoin/filecoin/src/data-managers/account-manager.ts
+++ b/src/chains/filecoin/filecoin/src/data-managers/account-manager.ts
@@ -2,6 +2,7 @@ import Manager from "./manager";
 import { LevelUp } from "levelup";
 import { Account, AccountConfig } from "../things/account";
 import PrivateKeyManager from "./private-key-manager";
+import { Address } from "../things/address";
 
 export default class AccountManager extends Manager<Account, AccountConfig> {
   #privateKeyManager: PrivateKeyManager;
--- a/src/chains/filecoin/filecoin/src/data-managers/account-manager.ts
+++ b/src/chains/filecoin/filecoin/src/data-managers/account-manager.ts
@@ -33,7 +33,9 @@ export default class AccountManager extends Manager<Account, AccountConfig> {
   async getAccount(address: string): Promise<Account> {
     let account = await super.get(address);
     if (!account) {
-      account = new Account();
+      account = new Account({
+        address: new Address(address)
+      });
       await this.putAccount(account);
     }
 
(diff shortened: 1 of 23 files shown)


tcoulter (approved):
Cool cool!

src/chains/filecoin/filecoin/src/data-managers/account-manager.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  async putAccount(account: Account) {
+    await super.set(account.address.value, account);
+
+    if (account.address.privateKey) {
+      await this.#privateKeyManager.putPrivateKey(
+        account.address.value,
+        account.address.privateKey
+      );
+    }
```
davidmurdoch: I worry that a a `getAccount` call could occur between super.set and `putPrivateKey`, which may be able to get the account, but fail to get the private key.

Could you atomically write these two values?

Alternatively, if you write the `privateKey` first this _specific_ race condition goes away, though the inverse race condition now exists (though it may not currently be exploitable)

src/chains/filecoin/filecoin/src/data-managers/private-key-manager.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  async putPrivateKey(address: string, privateKey: string) {
+    await this.base.put(Buffer.from(address), Buffer.from(privateKey));
+    const addresses = await this.getAddressesWithPrivateKeys();
+    if (!addresses.includes(address)) {
+      addresses.push(address);
+      await this.base.put(
+        PrivateKeyManager.AccountsWithPrivateKeysKey,
+        Buffer.from(JSON.stringify(addresses))
+      );
```
davidmurdoch: I seeeee yoouuuuu... but choose to look away because it's highly unlikely for the number of addresses to grow large enough to cause problems.

Maybe a little `// TODO(perf)` comment and a new Issue labeled with a low "priority" would be enough for now?

src/packages/cli/src/cli.ts
```diff
@@ -123,12 +123,18 @@ async function startGanache(err: Error) {
...
+      await initializeFilecoin(
+        server.provider as FilecoinProvider,
+        cliSettings
+      );
       break;
     }
     case EthereumFlavorName:
     default: {
-      initializeEthereum(server.provider as EthereumProvider, cliSettings);
+      await initializeEthereum(
```
davidmurdoch: I don't see the need to make `initializeEthereum` async here and don't mind if the flavor-specific CLI initialization function interfaces differ, as the divergence has to happen at some point. I prefer keeping it synchronous until it becomes necessary.

davidmurdoch (changes requested):

davidmurdoch (approved):

PR #839, opened 2021-03-04
implement ID protocol addresses

mikeseese:
This PR adds support for the `ID` address protocol, and switches all of our usages of it (previously as `string`'s) to use the `Address` class with `AddressProtocol.ID`.


tcoulter (approved):
Fine by me!

mikeseese:
Woops; didn't meant to rerequest from @tcoulter. He still approves of this one.

davidmurdoch (approved):

PR #841, opened 2021-03-04
implement wallet modification/auxiliary functions

mikeseese:
This PR implements the remaining wallet functions:
- `Filecoin.WalletSetDefault` moves the specified address to the front of the array to become the new default address
- `Filecoin.WalletNew` generates a new account to the wallet
- `Filecoin.WalletList` lists the current addresses in the wallet
- `Filecoin.WalletHas` checks if the wallet has (and therefore can sign for) the specified address
- `Filecoin.WalletDelete` removes the specified address from our wallet; this action is irrevocable sans importing the private key again
- `Filecoin.WalletExport` exports the private key for the specified address if it is in the wallet
- `Filecoin.WalletImport` adds the private key/address to the wallet to be able to sign it
- `Filecoin.WalletSign` signs a buffer with the specified address's private key in the wallet
- `Filecoin.WalletSignMessage` signs a `Message` with the specified address's private key in the wallet
- `Filecoin.WalletVerify` checks if the provided signature is valid for the provided address (doesn't require the address to be in the wallet)
- `Filecoin.WalletValidateAddress` checks if the address is a valid address

### DIFF
--- a/src/chains/filecoin/filecoin/src/things/address.ts
+++ b/src/chains/filecoin/filecoin/src/things/address.ts
@@ -133,6 +133,32 @@ class Address extends SerializableLiteral<AddressConfig> {
     }
   }
 
+  async verifySignature(
+    buffer: Buffer,
+    signature: Signature
+  ): Promise<boolean> {
+    switch (this.protocol) {
+      case AddressProtocol.BLS: {
+        return await bls.verify(
+          signature.data,
+          buffer,
+          Address.recoverBLSPublicKey(this.value)
+        );
+      }
+      case AddressProtocol.SECP256K1: {
+        const hash = blake.blake2b(buffer, null, 32);
+        return secp256K1.ecdsaVerify(
+          signature.data.slice(0, 64), // remove the recid suffix (should be the last/65th byte)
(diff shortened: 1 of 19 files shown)


tcoulter (approved):
All good!

src/chains/filecoin/filecoin/src/things/address.ts
```diff
@@ -296,8 +311,90 @@ class Address extends SerializableLiteral<AddressConfig> {
...
+        throw new Error("invalid address length");
+      }
+      const id = parseInt(raw, 10);
+      if (isNaN(id) || id.toString(10) !== raw) {
+        throw new Error("invalid address payload");
+      }
+      return address;
+    }
+
+    const customBase32Alphabet = "abcdefghijklmnopqrstuvwxyz234567";
```
davidmurdoch: I know `customBase32Alphabet` is used in at least 1 other place. Could you make it a constant, or maybe just have a common custom base32.parse function?

davidmurdoch (changes requested):

davidmurdoch (approved):

PR #842, opened 2021-03-04
add miner enabling options and methods

mikeseese:
This PR adds the `miner.mine` option to be able to pause/start the miner; see more details from the [comment thread here](https://github.com/trufflesuite/ganache-core/issues/248#issuecomment-780777681). The provided `miner.mine` option is in addition to `miner.blockTime`. The latter dictates how frequently to mine (instantly or on an interval), and `miner.mine` specifies whether or not to actually use `miner.blockTime`.

### DIFF
--- a/src/chains/filecoin/filecoin/src/blockchain.ts
+++ b/src/chains/filecoin/filecoin/src/blockchain.ts
@@ -74,6 +74,7 @@ export default class Blockchain extends Emittery.Typed<
 
   private ipfsServer: IPFSServer;
   private miningTimeout: NodeJS.Timeout | null;
+  readonly #miningTimeoutLock: Sema;
   private rng: utils.RandomNumberGenerator;
 
   readonly #database: Database;
@@ -97,6 +98,7 @@ export default class Blockchain extends Emittery.Typed<
     this.ipfsServer = new IPFSServer(this.options.chain);
 
     this.miningTimeout = null;
+    this.#miningTimeoutLock = new Sema(1);
     // to prevent us from stopping while mining or mining
     // multiple times simultaneously
     this.#miningLock = new Sema(1);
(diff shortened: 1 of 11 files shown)


src/chains/filecoin/filecoin/tests/api/filecoin/ganache.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        assert.strictEqual(isEnabled, true);
+        assert.strictEqual(provider.blockchain.minerEnabled, true);
+      });
+
+      it("Ganache.MinerEnabledNotify", async () => {
+        assert.strictEqual(enabledChanges.length, 2);
+        assert.strictEqual(enabledChanges[0], false);
+        assert.strictEqual(enabledChanges[1], true);
+      });
```
tcoulter, line 186: Can we add a test that checks the effects of turning the miner off? e.g., when you start up a blockchain while mining on a timer, then disable the miner, can we ensure it doesn't continue to create blocks? Another good thing might be to start up the miner again afterward and ensure it continues mining after being disabled, then reenabled.

tcoulter (changes requested):
Looks good. One to two new tests would be awesome. See comment!

tcoulter (approved):
Nice work Seese!

davidmurdoch (dismissed):
Looks good. I should probably mimic your miner start/stop behavior on the Ethereum side!

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -238,6 +235,34 @@ export default class Blockchain extends Emittery.Typed<
...
+  enableMiner() {
+    this.#minerEnabled = true;
+    this.emit("minerEnabled", true);
+
+    if (this.options.miner.blockTime > 0) {
+      const intervalMine = () => {
+        this.mineTipset();
+      };
+
+      this.miningTimeout = setInterval(
```
mikeseese: We should use a recursive `setTimeout` implementation; see https://github.com/trufflesuite/ganache-core/pull/836#discussion_r595207331

mikeseese:
alright @davidmurdoch, I added the recursive `setTimeout` stuff in this PR

davidmurdoch (approved):

PR #843, opened 2021-03-04
implement different chain retrieval methods

mikeseese:
This PR adds various methods for retrieving indexed information from the chain to support Ganache UI usage:
- `Filecoin.ChainGetTipSet`
- `Filecoin.ChainGetTipSetByHeight`
- `Filecoin.ChainGetBlock`
- `Filecoin.ChainGetBlockMessages`
- `Filecoin.ChainGetMessage`

### DIFF
--- a/src/chains/filecoin/filecoin/src/things/block-header.ts
+++ b/src/chains/filecoin/filecoin/src/things/block-header.ts
@@ -172,7 +172,9 @@ class BlockHeader
         deserializedName: "timestamp",
         serializedName: "Timestamp",
         defaultValue: literal => {
-          return literal || new Date().getTime() / 1000;
+          return typeof literal !== "undefined"
+            ? literal
+            : new Date().getTime() / 1000;
         }
       },
       blockSignature: {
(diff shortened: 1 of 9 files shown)


tcoulter (approved):
All good!

src/chains/filecoin/filecoin/src/things/block-header.ts
```diff
@@ -171,8 +171,8 @@ class BlockHeader
       timestamp: {
         deserializedName: "timestamp",
         serializedName: "Timestamp",
-        defaultValue: () => {
-          return new Date().getTime() / 1000;
+        defaultValue: literal => {
+          return literal || new Date().getTime() / 1000;
```
davidmurdoch: how's this work? If `literal` is a `0` somehow will this behave appropriately?

mikeseese: This should probably be

```suggestion
          return typeof literal !== "undefined" ? literal : (new Date().getTime() / 1000);
```

mikeseese: I applied the above change, but FWIW, this field is unused by Ganache

davidmurdoch (approved):

PR #844, opened 2021-03-04
implement `Ganache.GetDealById`

mikeseese:
This PR adds the `Ganache.GetDealById` method to query deals using their `deal.DealID` rather than using `Filecoin.ClientGetDealInfo` (implemented in a subsequent branch) which uses `deal.ProposalCid` to query for the deal. This is used on the **Deals** screen in Ganache UI.

### DIFF
--- a/src/chains/filecoin/filecoin/src/blockchain.ts
+++ b/src/chains/filecoin/filecoin/src/blockchain.ts
@@ -67,7 +67,8 @@ export default class Blockchain extends Emittery.Typed<
   readonly #messagePoolLock: Sema;
 
   readonly deals: Array<DealInfo> = [];
-  readonly dealsByCid: Record<string, DealInfo> = {};
+  readonly dealsByCid: Map<string, DealInfo> = new Map<string, DealInfo>();
+  readonly dealsById: Map<number, DealInfo> = new Map<number, DealInfo>();
   readonly inProcessDeals: Array<DealInfo> = [];
 
   readonly options: FilecoinInternalOptions;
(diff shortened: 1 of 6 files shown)


src/chains/filecoin/filecoin/tests/api/filecoin/deals.test.ts
```diff
@@ -89,6 +89,20 @@ describe("api", () => {
...
+        assert.strictEqual(deals.length, 1);
+
+        const deal = await provider.send({
+          jsonrpc: "2.0",
+          id: "0",
+          method: "Ganache.GetDealById",
+          params: [deals[0].DealID]
+        });
+
+        assert.deepStrictEqual(deal, deals[0]);
```
tcoulter, line 104: Another good test would be to test the behavior when we pass in an unknown ID.

tcoulter (changes requested):
See comment!

tcoulter (approved):
Perfect!

src/chains/filecoin/filecoin/src/blockchain.ts
```diff
@@ -772,7 +773,10 @@ export default class Blockchain extends Emittery.Typed<
 
     // Because we're not cryptographically valid, let's
     // register the deal with the newly created CID
-    this.dealsByCid[proposalCid.value] = deal;
+    this.dealsByCid.set(proposalCid.value, deal);
+
+    // Lets keep deals by id for easy paginated lookup from Ganache UI
+    this.dealsById.set(deal.dealId, deal);
```
davidmurdoch, line 779: Are these ever cleaned up? I think these will need to be backed by a DB. :-/

davidmurdoch (changes requested):

src/chains/filecoin/filecoin/src/blockchain.ts
mikeseese: No they are not, but #849 replaces them with the appropriate logic when it all gets backed by the DB in that PR

davidmurdoch (approved):

PR #848, opened 2021-03-04
remove default pins from ipfs and persist ipfs database

mikeseese:
This PR removes default pinned items from IPFS (i.e. IPFS documentation comes pre-pinned) so the **Files** tab on Ganache UI only shows the files that were added post-startup.

This PR also moves the IPFS data directory to be in the database folder so that we can recover the IPFS files on subsequent boots.


tcoulter (approved):
A+!

src/chains/filecoin/filecoin/src/ipfs-server.ts
```diff
@@ -29,16 +28,11 @@ class IPFSServer {
...
-        if (err) {
-          return reject(err);
-        }
-        resolve(folder);
-      });
-    });
+  async start(parentDirectory: string) {
+    const folder = path.join(parentDirectory, "ganache-ipfs");
+    if (!fs.existsSync(folder)) {
+      await fs.promises.mkdir(folder);
```
davidmurdoch, line 34: I haven't followed how `parentDirectory` gets here - does this need the `{recursive: true}` in options?

mikeseese: It comes from https://github.com/trufflesuite/ganache-core/pull/848/files#diff-a8d7ab65cce14da36b4292d6f60e99a3d13c4eb75c33cdd6f62d88cefbcb7e9aR194 (`blockchain.ts:194`). `parentDirectory` is the `Blockchain.#database.directory` and is set after initialization. So the directory should exist (unless the user has deleted it after initialization (which would cause several other issue, if even possible)

So `{ recursive: true }` is not necessary, but can be added if you would prefer it. There shouldn't be any valid scenario (currently) that would need it.

davidmurdoch (approved):

PR #849, opened 2021-03-04
implement persistence for deals

mikeseese:
This PR moves deals from being a variable in Blockchain to the persisted database so that deals can be recovered on subsequent boots.

### DIFF
--- a/src/chains/filecoin/filecoin/src/data-managers/deal-info-manager.ts
+++ b/src/chains/filecoin/filecoin/src/data-managers/deal-info-manager.ts
@@ -1,0 +1,81 @@
+import Manager from "./manager";
+import { LevelUp } from "levelup";
+import { DealInfo, DealInfoConfig } from "../things/deal-info";
+import { SerializedRootCID } from "../things/root-cid";
+
+const NOTFOUND = 404;
+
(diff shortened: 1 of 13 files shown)


tcoulter (approved):
LGTM!

src/chains/filecoin/filecoin/src/data-managers/deal-info-manager.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (dealCid) {
+      return await this.get(dealCid["/"]);
+    } else {
+      return null;
+    }
+  }
+
+  private async putDealCids(cids: Array<SerializedRootCID>): Promise<void> {
+    await this.base.put(DealInfoManager.Deals, JSON.stringify(cids));
+  }
```
davidmurdoch, line 91: Since this is a big ask, can you make an issue for the below request, and then link the issue in a `// TODO` comment in this file?

---

This looks like it loads all Deal cids, and then all the deals themselves into memory, and then filters them.

We'll need to rework this in the future. leveldb has a `createReadStream` method that could help with some of this; but I think we'll also need to add another sublevel that acts as an index for deal states.

davidmurdoch (changes requested):

davidmurdoch (approved):

PR #850, opened 2021-03-04
some minor code cleanup

mikeseese:
This PR does some minor code cleanup (mainly changing `let` to `const` if the variable isn't reassigned and removing unused imports).


tcoulter (approved):
Way better!

davidmurdoch (approved):

PR #853, opened 2021-03-08
Enable webpack building for `@ganache/filecoin-options`

mikeseese:
This PR makes the necessary changes to enable webpack bundling for `@ganache/filecoin-options` so Ganache UI doens't need to run the provider to get default options

### DIFF
--- a/src/chains/filecoin/options/package.json
+++ b/src/chains/filecoin/options/package.json
@@ -23,7 +23,7 @@
     "directory": "src/chains/filecoin/options"
   },
   "scripts": {
-    "build": "npx webpack",
+    "build": "webpack",
     "tsc": "ttsc --build",
     "test": "nyc npm run mocha",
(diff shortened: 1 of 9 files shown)


tcoulter (approved):
✔️

src/chains/filecoin/options/package.json
```diff
@@ -1,26 +1,29 @@
...
+    "lib/src/index.d.ts",
+    "dist"
   ],
   "repository": {
     "type": "git",
     "url": "https://github.com/trufflesuite/ganache-core.git",
     "directory": "src/chains/filecoin/options"
   },
   "scripts": {
+    "build": "npx webpack",
```
davidmurdoch: i don't think you need `npx` here, since it is installed as a devDependency.

davidmurdoch (changes requested):

davidmurdoch (approved):

PR #857, opened 2021-03-09
fix: hex encode hash in  eth_getTransactionReceipt notice

### LINKED ISSUE #856: Ganache 7 prints out garbled transaction hashes in certain logging messages
When attempting to get the transaction receipt for a transaction that hasn't been mined yet, Ganache 7 will log a message that looks something like this:

```
 > Ganache `eth_getTransactionReceipt` notice: the transaction with hash
 > `���5
!�9�����k���t��0/u��Z��` has not
 > yet been mined. See https://trfl.co/v7-instamine for additional information.
```

As you can see, the transaction hash is terribly garbled.  As if it were printing out the bytes raw or something rather than converting them to hex.  This happens with every transaction hash I've seen.

I'm having this issue since installing / using `ganache@internal`.

davidmurdoch:
fixes #856

### DIFF
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -1216,7 +1216,8 @@ export default class EthereumApi implements types.Api {
   @assertArgLength(1)
   async eth_getTransactionReceipt(transactionHash: string) {
     const { transactions, transactionReceipts, blocks } = this.#blockchain;
-    const txHash = Data.from(transactionHash).toBuffer();
+    const dataHash = Data.from(transactionHash);
+    const txHash = dataHash.toBuffer();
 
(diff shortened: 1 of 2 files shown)


src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1244,7 +1245,7 @@ export default class EthereumApi implements types.Api {
       if (tx != null) {
         options.logging.logger.log(
           " > Ganache `eth_getTransactionReceipt` notice: the transaction with hash\n" +
-            ` > \`${txHash.toString()}\` has not\n` +
+            ` > \`${dataHash.toString()}\` has not\n` +
```
eggplantzzz, line 1248: Thanks for cleaning up after me!

eggplantzzz (approved):
I like it!

PR #860, opened 2021-03-10
fix filecoin stop logic and make it more explicit

mikeseese:
This PR resolves an issue that I discovered in Ganache UI where the database lock did not get unlocked properly.

The reproduction issue was:
1. Start ganache cli
2. Start ganache ui with same ports
3. UI would fail to start
4. Try to restart ganache ui
5. Receive error saying something along the lines of the `path/to/database/LOCK was already held by another process`

I tracked this down to the `packages/core/src/server.ts` not closing the `connector` because `#status === Status.closed`. This logic is invalid since `#status` is updated asynchronously, but `connector` is loaded on construction

Theoretically this bug would exist for the ethereum flavor too

In addition to solving this bug, this PR adds a `stopped` boolean to the filecoin blockchain to prevent multiple stops from being processed (which would cause us to constantly try to acquire a lock that purposely isn't released from a prior stop)

### DIFF
--- a/src/chains/filecoin/filecoin/src/blockchain.ts
+++ b/src/chains/filecoin/filecoin/src/blockchain.ts
@@ -87,6 +87,7 @@ export default class Blockchain extends Emittery.Typed<
   }
 
   private ready: boolean;
+  private stopped: boolean;
 
   constructor(options: FilecoinInternalOptions) {
     super();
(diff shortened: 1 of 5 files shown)


davidmurdoch (approved):

mikeseese:
I got approval from @tcoulter on Slack to merge this without his review (I also got @davidmurdoch's approval to merge without Tim's approval).

Tests fail in this PR due to not having fixes from #877 yet, but I'm going to apply fixes for that in a new PR and merge this for now to be done with it

PR #877, opened 2021-03-18
fix: rearchitect server status and move async functions from constructors to initialize functions

mikeseese:
This PR was driven by having an issue incorporating Ganache v7 into Ganache UI. If there was an error before the `Server.status` was set to `opening`, but the constructors started async operations (i.e. opening the database), then things would get locked without the ability to be closed properly during `Server.close()`

So this PR does a couple of things:
- Provide more/rearchitect the server `Status` enum to be more accurate and appropriate
- Move all asynchronous code from constructors to a dedicated asynchronous `initialize` functions so that we can guarantee that we can appropriately close things when things fail
- Coincidentally, this has caused `Ganache.provider()` to become asynchronous by nature. It always "was" and the work around was to use an event emitter to say when things were ready. This may get modified to go back to the old way (while still keeping the other benefits added in this PR) to maintain back-compatibility. However that decision is still waiting on feedback from the rest of the team. This PR won't get merged until the final direction has been decided and implemented in this PR. In the meantime, the other parts can be reviewed.

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -180,7 +180,6 @@ export default class Blockchain extends Emittery.Typed<
   constructor(
     options: EthereumInternalOptions,
     common: Common,
-    initialAccounts: Account[],
     coinbaseAddress: Address
   ) {
     super();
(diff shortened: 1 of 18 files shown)


src/packages/core/src/server.ts
```diff
@@ -59,28 +77,35 @@ export default class Server {
     return this.#status;
   }
 
-  constructor(serverOptions: ServerOptions = { flavor: DefaultFlavor }) {
-    const opts = (this.#options = serverOptionsConfig.normalize(serverOptions));
+  constructor(providerAndServerOptions: any = { flavor: DefaultFlavor }) {
```
davidmurdoch: Why the change to `any` here?

src/packages/core/src/server.ts
```diff
@@ -59,28 +77,35 @@ export default class Server {
...
+    this.#status = Status.ready;
+  }
+
+  private async initialize() {
     const connector = (this.#connector = ConnectorLoader.initialize(
-      serverOptions
+      this.#providerOptions
     ));
 
+    await connector.once("ready");
```
davidmurdoch: Is there a reason the server can't start its initialization until the connector is ready?

src/packages/core/src/server.ts
```diff
@@ -59,28 +77,35 @@ export default class Server {
...
     }
     this.#httpServer = new HttpServer(_app, connector);
   }
 
-  listen(port: number): Promise<void>;
-  listen(port: number, host: string): Promise<void>;
   listen(port: number, callback: Callback): void;
   listen(port: number, host: string, callback: Callback): void;
+  listen(port: number): Promise<void>;
+  listen(port: number, host: string): Promise<void>;
```
davidmurdoch: What's the reason for re-ordering these?

src/packages/core/src/server.ts
```diff
@@ -149,25 +182,36 @@ export default class Server {
...
+      // if closing or closed
+      throw new Error(`Server is already closing or closed.`);
     }
 
-    const _listenSocket = this.#listenSocket;
     this.#status = Status.closing;
+
+    // clean up the websocket objects
+    const _listenSocket = this.#listenSocket;
     this.#listenSocket = void 0;
```
davidmurdoch: I think this should be set to `null` here, to reflect your type change.

```suggestion
    this.#listenSocket = null;
```

src/packages/core/src/server.ts
```diff
@@ -149,25 +182,36 @@ export default class Server {
...
+    if (this.#httpServer !== null) {
+      this.#httpServer.close();
+    }
+
+    // cleanup the connector, provider, etc.
+    if (this.#connector !== null) {
+      await this.#connector.close();
+    }
+
+
```
davidmurdoch, line 236: Correct me if I'm wrong, but I think it's safe to remove these two guards, as well as the `_listenSocket` guard above, as these conditions should always evaluate to `true` here.

If the code changes in a way that causes these branches to evaluate to `false`, I'd like for it to throw an exception (implicit is fine), and not be guarded away, as when the server is `open`, these values must absolutely be non-null, or things are likely broken in some other subtle way.

src/packages/core/tests/connector.test.ts
```diff
@@ -4,12 +4,16 @@ import { Provider as EthereumProvider } from "@ganache/ethereum";
 
 describe("connector", () => {
   it("works without passing options", async () => {
-    assert.doesNotThrow(() => Ganache.provider());
+    assert.doesNotThrow(async () => {
+      const provider = Ganache.provider();
+      await provider.once("connect");
```
davidmurdoch: Is it required to `await provider.once("connect");` for these tests now?

davidmurdoch (changes requested):

src/packages/core/src/server.ts
mikeseese: It's possible we can move this to the bottom of the `initialize` function; I was try to prevent unnecessary dirty state if the initialize couldn't complete, but I can understand your desire here for performance rather than logical progression.

src/packages/core/src/server.ts
mikeseese: I'm not sure; I think I didn't fully understand that the `ServerOptions` type includes the provider options although `ServerOptions` const is defined as just the server related options in `src/packages/core/src/options/server-options.ts`. Feels like a confusing naming conflict.

I'll revert it back to `ServerOptions`

src/packages/core/src/server.ts
mikeseese: Seems to be a remnant from old code structure from another attempt that got copied and I didn't put it back in the same way

src/packages/core/src/server.ts
mikeseese: > as these conditions should always evaluate to `true` here.

I disagree with this. I don't see any evidence in the code that ensures they'll always be true.

For example, you could call `Server.listen()` and `uWS.App()` in `Server.initialize()` could throw, causing you to catch in the error in `listen()` which will call this function; `this.#httpSever = new HttpServer(...)` would never have been called.

If you tried to run this through typescript without the check, it would succeed because you don't have strict null checks turned on. If you had these turned on, typescript would tell you it could not ensure it was properly set. I definitely believe these guard if statements are necessary. Typescript would enforce them or a bang/optional-chain operator had you have strict null checks enabled.

> If the code changes in a way that causes these branches to evaluate to false, I'd like for it to throw an exception (implicit is fine), and not be guarded away, as when the server is open, these values must absolutely be non-null, or things are likely broken in some other subtle way.

This would potentially circumvent the error that is being caught in `listen()` and you'd be throwing a "could not initialize properly" error instead of the actual error that caused initialize to fail. This goes back to your assumption that these values are always non-null at this context, which is not the case. Not to be purposely repetitive, they could be not null in the event initialization fails, in which case would be handled and thrown in the `.catch(...)` logic at line 173.

src/packages/core/tests/connector.test.ts
mikeseese: No it's not required; I can remove them if you want. I figured it'd be good practice to use it, but if you'd prefer to use the request coordinators queuing logic in all the tests ok. I personally believe that should be a fallback rather than the main use case and we should be telling users to explicitly wait for Ganache to initialize in default use cases.

Regardless, I can also add a pair of tests that ensure the async logic here works properly.

davidmurdoch: I'd prefer to reduce boilerplate, rather than add to it. I'd prefer they awaits were removed.

And you know I always appreciate more tests! :-D

src/packages/core/src/server.ts
davidmurdoch: Ah, I missed the new call to `this.close()` in the catch. I think we need to move the `if (callbackIsFunction) callback!(err);` check on line `166` to the new catch block.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -214,113 +213,120 @@ export default class Blockchain extends Emittery.Typed<
...
 
-      this.coinbase = coinbaseAddress;
+  async initialize(initialAccounts: Account[]) {
+    await this.#database.initialize();
 
-      // create VM and listen to step events
-      this.vm = this.createVmFromStateTrie(
-        this.trie,
-        options.chain.allowUnlimitedContractSize
+    const database = this.#database;
```
davidmurdoch: Since we're assigning to `this.#database` to `database`, let's use the `database` variable in the `await this.#database.initialize();` line above (line `222`).

src/packages/core/src/server.ts
```diff
@@ -16,39 +16,57 @@ type Callback = (err: Error | null) => void;
...
 export enum Status {
   /**
-   * The connection is open and ready to communicate.
+   * The server is in an unknown state; perhaps construction didn't succeed
    */
-  open = 1,
+  unknown = 0,
   /**
-   * The connection is not yet open.
+   * The server has been constructed and is ready to be opened is open and ready to communicate.
```
davidmurdoch:
```suggestion
   * The Server has been constructed and is ready to be opened.
```

src/packages/core/src/server.ts
```diff
@@ -99,45 +124,53 @@ export default class Server {
...
-          `listen EADDRINUSE: address already in use ${
-            hostname || DEFAULT_HOST
-          }:${port}.`
-        );
-        if (callbackIsFunction) callback!(err);
-        else throw err;
-      }
+    const initializePromise = this.initialize();
+
+    const promise = initializePromise.then(() => {
```
davidmurdoch: I think we should still parallelize chain initialization and server start up here. I think it's still possible? So, instead of `initializePromise.then(...)`, perhaps we can `const promise = Promise.allSettled([initializePromise, new Promise(...)).catch(...);` (note: for this approach you'll probably need to use the `promise.allsettled` polyfill because of our Node 10 support)

src/packages/core/package.json
```diff
@@ -51,9 +51,11 @@
     "@ganache/options": "^0.1.0",
     "@ganache/tezos": "^0.1.0",
     "@ganache/utils": "^0.1.0",
+    "promise.allsettled": "^1.0.4",
```
davidmurdoch: can you pin this?

src/packages/core/package.json
```diff
@@ -51,9 +51,11 @@
...
     "uWebSockets.js": "github:uNetworking/uWebSockets.js#v18.4.0"
   },
   "devDependencies": {
+    "@types/promise.allsettled": "^1.0.3",
```
davidmurdoch: can you pin this?

src/packages/core/src/server.ts
```diff
@@ -136,6 +137,10 @@ export default class Server {
 
     const initializePromise = this.initialize();
 
+    // necessary for `Promise.allSettled` to be shimmed
+    // in `node@10`
+    allSettled.shim();
```
davidmurdoch: I don't think we should `shim` it everywhere; let's just use `allSettled([...])` directly.

davidmurdoch (changes requested):

src/packages/core/src/server.ts
mikeseese: Using `allSettled([...])` directly with the typescript version you're using gives this error in several tests: https://github.com/es-shims/Promise.allSettled/issues/5 (i have not tested an upgrade, but supposedly that fixes it)

davidmurdoch: Can we just ts-ignore it away for now? Upgrading breaks many other types.

mikeseese: No, I believe it's an error with the compiled JS output that typescript generates, not a typescript compiler error. I tested it to be sure:

image

src/packages/core/tests/connector.test.ts
mikeseese: If the extra test are not required, then I will not do them to try to save as much time as we can

src/packages/core/src/server.ts
mikeseese: I did add a more verbose comment with a todo in the event typescript gets upgraded

davidmurdoch (approved):

PR #882, opened 2021-03-24
Dockerfile for Docker Hub container image

kevinbluer:
As discussed, when we're ready to update the image at https://hub.docker.com/repository/docker/trufflesuite/ganache-cli, we'll need to update the "Source Repository" (under Build configurations) to point at this repo. In addition, given the `Dockerfile` isn't in the repo root we'll need to set the "Dockerfile location" under "Build Rules".

For reference, this was tested (and can be tested further as needed) via the following Docker repo / image: https://hub.docker.com/repository/docker/trufflesuite/ganache-core


trajan0x:
+1 on this, https://hub.docker.com/r/trufflesuite/ganache-cli is super out of date

PR #886, opened 2021-03-30
implement #877 async init changes for the filecoin flavor

mikeseese:
This PR implements the necessary changes from #877 (which was rebased into `filecoin`) for the filecoin flavor so that initialization can happen in an async method rather than the constructor

### DIFF
--- a/src/packages/core/src/server.ts
+++ b/src/packages/core/src/server.ts
@@ -183,9 +183,7 @@ export class Server extends Emittery<{ open: undefined; close: undefined }> {
           else throw err;
         }
       })
-    ]).then(() => {
-      return this.emit("open");
-    }).catch(async error => {
+    ]).catch(async error => {
       this.#status = Status.unknown;
       if (callbackIsFunction) callback!(error);
       await this.close();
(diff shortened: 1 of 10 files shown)


src/packages/core/src/server.ts
```diff
@@ -183,9 +183,7 @@ export class Server extends Emittery<{ open: undefined; close: undefined }> {
           else throw err;
         }
       })
-    ]).then(() => {
-      return this.emit("open");
```
mikeseese, line 187: I put this in the wrong spot during rebase; this change isn't one that is in `develop`. The `"open"` event was added in https://github.com/trufflesuite/ganache-core/pull/847/files#diff-b4ab586220123bb3565eb84b69a47164e6012bd51caea50e24bbe8b89879adfcR184

davidmurdoch (approved):

mikeseese:
I'm going to merge this one, but there are a handful of tests failing due to unrelated issues. I thought this PR would resolve them, but there are other issues. I will open a new PR with fixes :'(

PR #887, opened 2021-03-31
Fix Filecoin tests

mikeseese:
This PR fixes tests 2 root issues that were causing multiple tests from failing:
- Change how the result of `allSettled` is handled by using the root thenable rather than wrapping it in another promise. Without this, tests for node<12 were failing
- Use `ubuntu-18.04` to build the Linux images for https://github.com/trufflesuite/uws-js-unofficial (this is the version that the latest `uWebSockets.js` was used to build despite Alex's workflow uses `ubuntu-latest` [so it'll break if/when his `ubuntu-latest` moves to 20.04]). The error was seen in ubuntu 16.04 and 18.04 where GLIBC 2.28 could not be found

### DIFF
--- a/src/packages/core/src/server.ts
+++ b/src/packages/core/src/server.ts
@@ -183,7 +183,16 @@ export class Server extends Emittery<{ open: undefined; close: undefined }> {
           else throw err;
         }
       })
-    ]).catch(async error => {
+    ]).then((promiseResults) => {
+      if (promiseResults[0].status === "fulfilled" && promiseResults[1].status === "fulfilled") {
+        this.emit("open");
+      } else {
+        let reason = "";
+        reason += promiseResults[0].status === "rejected" ? `${promiseResults[0].reason}\n\n` : "";
+        reason += promiseResults[1].status === "rejected" ? promiseResults[1].reason : "";
+        throw reason;
+      }
+    }).catch(async error => {
       this.#status = Status.unknown;
       if (callbackIsFunction) callback!(error);
       await this.close();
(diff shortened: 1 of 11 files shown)


src/packages/core/src/server.ts
```diff
@@ -183,27 +183,24 @@ export class Server extends Emittery<{ open: undefined; close: undefined }> {
...
-    ]).catch(async error => {
+    ]).then((promiseResults) => {
+      if (promiseResults[0].status === "fulfilled" && promiseResults[1].status === "fulfilled") {
+        this.emit("open");
+      } else {
+        let reason = "";
+        reason += promiseResults[0].status === "rejected" ? `${promiseResults[0].reason}\n\n` : "";
+        reason += promiseResults[1].status === "rejected" ? promiseResults[1].reason : "";
+        throw reason;
+      }
```
davidmurdoch: Sorry I missed this before.

You shouldn't throw a string (or anything other than an Error), as callers likely expect it to be an Error with at least a `message` preoperty.

Even if this us updated to throw `new Error(reason)`, our stack trace is throw away, which makes me sad.

An `AggregateError` ([aggregate-error](https://www.npmjs.com/package/aggregate-error)) is probably appropriate here, and is already a transitive dependency of ours.

src/packages/core/src/server.ts
```diff
@@ -183,27 +183,24 @@ export class Server extends Emittery<{ open: undefined; close: undefined }> {
...
+    }).catch(async error => {
       this.#status = Status.unknown;
       if (callbackIsFunction) callback!(error);
       await this.close();
       throw error;
     });
```
davidmurdoch: Should this be something like?

```suggestion
    }).catch(error => {
      this.#status = Status.unknown;
	  await this.close();
	  throw error;
    }).catch(error => {
      if (callbackIsFunction) callback!(error);
      else throw error;
    });
```

Alternatively, since `this.close()` could also cause an error, we may want to implement `AggregateError` here, as well as above, but I don't know how AggregateError + AggregateError would behave. OOORRRRRR we ignore `this.close()` errors (`await this.close().catch(()=>{});` since this isn't the error the user would really care about anyway.

Really, though, it feels like this section needs to be refactored a bit to handle these error cases more eloquently. Any thoughts?

davidmurdoch (changes requested):

src/packages/core/src/server.ts
davidmurdoch: Another possibility here is to only send the _first_ error we get back to the user. It'll be clearer to the user what the problem is (since `AggregateError`s aren't something developers read every day). Doing it this way (first error only) may solves the complexity of handling these errors.

davidmurdoch (approved):

PR #892, opened 2021-04-02
refactor: increase the default ether balance for created accounts to 1000

eggplantzzz:
This PR addresses https://github.com/trufflesuite/ganache-core/issues/781.

### DIFF
--- a/src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
@@ -82,7 +82,7 @@ describe("api", () => {
 
         // Assert the starting balance is where we think it is, including tx costs.
         assert(
-          balance > 98900000000000000000 && balance < 99000000000000000000
+          balance > 990000000000000000000n && balance < 999000000000000000000n
         );
         startingBalance = balance;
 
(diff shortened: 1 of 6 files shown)


eggplantzzz:
Do we want to include tests for this? I suppose we could test it alongside tests verifying that manually setting balances works. Offhand I'm not sure if we have tests for that or where they belong.

davidmurdoch:
> Do we want to include tests for this? I suppose we could test it alongside tests verifying that manually setting balances works. Offhand I'm not sure if we have tests for that or where they belong.

There is a test, "eth_getBalance", "should return initial balance". You'll need to update it, and likely others as well.

src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
```diff
@@ -82,7 +82,7 @@ describe("api", () => {
 
         // Assert the starting balance is where we think it is, including tx costs.
         assert(
-          balance > 98900000000000000000 && balance < 99000000000000000000
+          balance > 990000000000000000000n && balance < 999000000000000000000n
```
eggplantzzz, line 85: Why do we test against a range here? Is it not totally deterministic?

davidmurdoch: This is one of the tests that was ported from the old tests, not sure why it was written this way.

eggplantzzz: Should I go ahead and make a specific assertion here?

davidmurdoch (approved):

PR #943, opened 2021-08-13
Create Transaction Factory and Introduce EIP2930 Access List Transactions

MicaiahReid:
Introduces typed transactions and the type 1, EIP-2930 transaction introduced in the Berlin hardfork.
`TransactionFactory` created to parse RPC tx data into appropriate types.


src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -163,10 +166,14 @@ export class RuntimeBlock {
       BUFFER_8_ZERO // nonce
     ];
     const { totalDifficulty } = header;
-    const txs: EthereumRawTx[] = [];
+    const txs: BlockDatabaseTransaction[] = []; // we don't want
     const extraTxs: GanacheRawBlockTransactionMetaData[] = [];
     transactions.forEach(tx => {
-      txs.push(tx.raw);
+      if (tx.raw[0][0] === undefined) {
```
MicaiahReid: Use `TransactionFactory.typeOfRaw(tx.raw) === LegacyTransaction` instead. There are some edge cases here where a LegacyTransaction could be mistaken for a typed one.

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -163,10 +166,14 @@ export class RuntimeBlock {
       BUFFER_8_ZERO // nonce
     ];
     const { totalDifficulty } = header;
-    const txs: EthereumRawTx[] = [];
+    const txs: BlockDatabaseTransaction[] = []; // we don't want
```
MicaiahReid: Remove silly comment

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -10,11 +10,12 @@ import { EthereumRawBlockHeader, serialize } from "./serialize";
...
 import { Block } from "./block";
 import {
-  EthereumRawTx,
+  BlockDatabaseTransaction,
   GanacheRawBlockTransactionMetaData,
-  RuntimeTransaction
+  LegacyDatabasePayload
 } from "@ganache/ethereum-transaction";
 import { StorageKeys } from "@ganache/ethereum-utils";
+import { TypedTransaction } from "@ganache/ethereum-transaction";
```
MicaiahReid: Move up with other `@ganache/ethereum-transaction` imports.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -64,6 +63,8 @@ import { ForkTrie } from "./forking/trie";
 import { LevelUp } from "levelup";
 import { activatePrecompiles } from "./helpers/precompiles";
 import TransactionReceiptManager from "./data-managers/transaction-receipt-manager";
+import { TypedTransaction } from "@ganache/ethereum-transaction";
```
MicaiahReid: Move up with other `@ganache/ethereum-transaction` imports.

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -20,12 +20,13 @@ import VM from "@ethereumjs/vm";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
 import replaceFromHeap from "./replace-from-heap";
 import { EVMResult } from "@ethereumjs/vm/dist/evm/evm";
-import { Params, RuntimeTransaction } from "@ganache/ethereum-transaction";
+import { Params } from "@ganache/ethereum-transaction";
 import { Executables } from "./executables";
 import { Block, RuntimeBlock } from "@ganache/ethereum-block";
+import { TypedTransaction } from "@ganache/ethereum-transaction";
```
MicaiahReid: Move up with other imports from `@ganache/ethereum-transaction`

src/chains/ethereum/transaction/src/access-lists.ts
```diff
@@ -0,0 +1,126 @@
...
+      AccessListJSON = json;
+    }
+
+    return {
+      AccessListJSON,
+      accessList: bufferAccessList
+    };
+  }
+
+  public static verifyAccessList(accessList: AccessListBuffer) {
```
MicaiahReid: Function is not used; remove it.

src/chains/ethereum/transaction/src/access-lists.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (storageSlots[storageSlot].length != 32) {
+          throw new Error(
+            "Invalid EIP-2930 transaction: storage slot length should be 32 bytes"
+          );
+        }
+      }
+    }
+  }
+
+  public static getAccessListJSON(accessList: AccessListBuffer) {
```
MicaiahReid: Function is not used; remove it.

src/chains/ethereum/transaction/src/access-lists.ts
```diff
@@ -0,0 +1,126 @@
...
+        JSONItem.storageKeys.push(
+          "0x" + setLengthLeft(storageSlot, 32).toString("hex")
+        );
+      }
+      accessListJSON.push(JSONItem);
+    }
+    return accessListJSON;
+  }
+
+  public static getDataFeeEIP2930(
```
MicaiahReid: Function is not used; consider removing. It could be a helpful endpoint to open up for users in the future.

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
```diff
@@ -0,0 +1,228 @@
...
+  Quantity,
+  keccak,
+  BUFFER_EMPTY,
+  BUFFER_32_ZERO,
+  RPCQUANTITY_EMPTY
+} from "@ganache/utils";
+import { Address } from "@ganache/ethereum-address";
+import type Common from "@ethereumjs/common";
+import { BN } from "ethereumjs-util";
+import { ecsign } from "ethereumjs-util";
```
MicaiahReid: Join imports from `ethereumjs-util`

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
```diff
@@ -0,0 +1,228 @@
...
+      this.chainId = Quantity.from(data.chainId);
+      this.gasPrice = Quantity.from(data.gasPrice);
+      const accessListData = AccessLists.getAccessListData(data.accessList);
+      this.accessList = accessListData.accessList;
+      this.accessListJSON = accessListData.AccessListJSON;
+    }
+  }
+
+  public toJSON = () => {
+    return {
```
MicaiahReid, line 91: The LegacyTransaction returns the type here, don't we need to? Test to make sure then add it.

src/chains/ethereum/transaction/src/fake-transaction.ts
```diff
@@ -1,19 +1,23 @@
...
 /**
  * A FakeTransaction spoofs the from address and signature.
  */
 
-export class FakeTransaction extends RuntimeTransaction {
-  constructor(data: RpcTransaction, common: Common) {
-    super(data, common);
+export class FakeTransaction {
+  constructor(data: TypedRpcTransaction, common: Common) {
+    //super(data, common);
```
MicaiahReid: Why did I add these comments??? I have no recollection of doing this.

src/chains/ethereum/transaction/src/fake-transaction.ts
```diff
@@ -1,19 +1,23 @@
-import { RpcTransaction } from "./rpc-transaction";
+import { TypedRpcTransaction } from "./rpc-transaction";
 import type Common from "@ethereumjs/common";
 import { RuntimeTransaction } from "./runtime-transaction";
```
MicaiahReid: Remove unused reference.

src/chains/ethereum/transaction/src/frozen-transaction.ts
```diff
@@ -92,5 +152,53 @@ export class FrozenTransaction extends BaseTransaction {
       r: this.r,
       s: this.s
     };
+    if (this.type) {
```
MicaiahReid: Will this ever not be true? Need to consider.

src/chains/ethereum/transaction/src/frozen-transaction.ts
```diff
@@ -92,5 +152,53 @@ export class FrozenTransaction extends BaseTransaction {
...
+    if (this.chainId) {
+      json.chainId = this.chainId;
+    }
+    if (this.accessList) {
+      json.accessList = this.accessListJSON;
+    }
+    return json;
   };
+
+  public toVmTransaction() {
```
MicaiahReid: Unused function. Remove it.

src/chains/ethereum/transaction/src/legacy-transaction.ts
```diff
@@ -0,0 +1,220 @@
...
+      to: this.to.isNull() ? null : this.to,
+      value: this.value,
+      gas: this.gas,
+      gasPrice: this.gasPrice,
+      input: this.data,
+      v: this.v,
+      r: this.r,
+      s: this.s
+    };
+    if (this.type) {
```
MicaiahReid: Will this ever not be true? Can possibly simplify this.

src/chains/ethereum/transaction/src/frozen-transaction.ts
MicaiahReid: Nvm, it is used. Keeping.

src/chains/ethereum/transaction/src/fake-transaction.ts
```diff
@@ -1,19 +1,23 @@
-import { RpcTransaction } from "./rpc-transaction";
```
MicaiahReid, line 1: Whole class is unused, can we delete?

MicaiahReid (commented):
Incorporated my own comments already.

src/chains/ethereum/transaction/src/transaction-receipt.ts
```diff
@@ -110,8 +149,7 @@ export class TransactionReceipt {
     blockLog.blockNumber = blockNumber;
     raw[3].forEach(l => blockLog.append(transactionIndex, transactionHash, l));
     const logs = [...blockLog.toJSON()];
-
-    return {
+    let json: TransactionReceiptJSON = {
```
davidmurdoch:
```suggestion
    const json: TransactionReceiptJSON = {
```

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -129,7 +135,6 @@ export class RuntimeBlock {
    * @param receiptTrie
    * @param bloom
    * @param stateRoot
-   * @param gasUsed
```
davidmurdoch, line 132: I don't think this was meant to be removed. Though these JSDOC params aren't actually adding any value, so they could just all be removed.

src/chains/ethereum/block/src/block.ts
```diff
@@ -108,20 +111,20 @@ export class Block {
...
-        Quantity.from(tx.r).toBuffer(),
-        Quantity.from(tx.s).toBuffer()
-      ]);
+      const typedTx = TransactionFactory.fromRpc(tx, common);
+      const raw = typedTx.toEthRawTransaction(
+        typedTx.v.toBuffer(),
+        typedTx.r.toBuffer(),
+        typedTx.s.toBuffer()
+      );
+      if (typedTx.type.toBuffer() === BUFFER_EMPTY) {
```
davidmurdoch: If we normalize the type for a legacy transaction to nothing (`null`), or even to `0`, everywhere we process this data we could then just compare it to a constant when needed in places like this as well as in places that use `typeOfRaw` (`typeOfRaw` would return the constant value (0/null , 1, 2, etc)).

So we could end up with something like `typedTx.type.toBuffer()[0] === TransactionType.Legacy` here.

davidmurdoch: I just realized this might complicate `toEthRawTransaction` in that we need to store the type for legacy transactions as `null`, IIRC.

src/chains/ethereum/block/src/block.ts
```diff
@@ -1,13 +1,16 @@
-import { Data, Quantity } from "@ganache/utils";
+import { BUFFER_EMPTY, Data, Quantity } from "@ganache/utils";
 import {
+  BlockDatabaseTransaction,
   BlockTransaction,
-  EthereumRawTx,
-  GanacheRawBlockTransactionMetaData
+  GanacheRawBlockTransactionMetaData,
+  LegacyTransaction,
```
davidmurdoch:
```suggestion
```

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -94,6 +97,7 @@ export class RuntimeBlock {
     coinbase: { buf: Buffer; toBuffer: () => Buffer };
     number: BnExtra;
     gasLimit: BnExtra;
+    gasUsed: BnExtra;
```
davidmurdoch, line 97: I'm not sure I can see the need to add this field. Maybe this is something the VM requires now?

src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
```diff
@@ -8,8 +8,10 @@ import type Common from "@ethereumjs/common";
 import { Data, Quantity } from "@ganache/utils";
 import {
   FrozenTransaction,
-  RpcTransaction,
-  RuntimeTransaction
+  RuntimeTransaction,
```
davidmurdoch:
```suggestion
```

src/chains/ethereum/transaction/src/fake-transaction.ts
davidmurdoch: Yes, it looks safe to delete this class.

src/chains/ethereum/transaction/src/raw.ts
```diff
@@ -1,18 +1,54 @@
...
+export type TypedDatabasePayload =
+  | LegacyDatabasePayload
+  | EIP2930AccessListDatabasePayload;
+export type BlockDatabaseTransaction =
+  | LegacyDatabasePayload
+  | EIP2930AccessListDatabaseTx;
+
 /**
  * Extra data Ganache stores as part of a transaction in order to support
  * account mascarading and quick lookups for transactions, blocks, and receipts.
```
davidmurdoch: let's sneak this typo fix in this PR :-)

```suggestion
 * account masquerading and quick lookups for transactions, blocks, and receipts.
```

src/chains/ethereum/transaction/src/signing.ts
```diff
@@ -140,31 +141,76 @@ const SHARED_BUFFER = Buffer.allocUnsafe(65);
...
 };
 
-export const computeIntrinsics = (
+export const computeInstrinsicsLegacyTx = (
   v: Quantity,
-  raw: EthereumRawTx,
+  raw: LegacyDatabaseTx,
   chainId: number
 ) => {
+  raw.shift();
```
davidmurdoch: This will mutate the `raw` array, which may be unexpected. I think you can update the `encodeRange` calls to use the appropriate indexes as well as update the use of `raw` in the `computeAddress` below.

src/chains/ethereum/transaction/src/legacy-transaction.ts
```diff
@@ -0,0 +1,206 @@
...
+  public gasPrice: Quantity;
+  public type: Quantity = Quantity.from("0x0");
+
+  public constructor(
+    data: LegacyDatabasePayload | TypedRpcTransaction,
+    common: Common
+  ) {
+    super(data, common);
+    if (Array.isArray(data)) {
+      this.nonce = Quantity.from(data[0], true);
```
davidmurdoch: I think this should be:

```suggestion
      this.nonce = Quantity.from(data[0]);
```

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
```diff
@@ -0,0 +1,229 @@
...
+  public type: Quantity = Quantity.from("0x1");
+
+  public constructor(
+    data: EIP2930AccessListDatabasePayload | TypedRpcTransaction,
+    common: Common
+  ) {
+    super(data, common);
+    if (Array.isArray(data)) {
+      this.chainId = Quantity.from(data[0]);
+      this.nonce = Quantity.from(data[1], true);
```
davidmurdoch: I think this should be:

```suggestion
      this.nonce = Quantity.from(data[1]);
```

src/chains/ethereum/transaction/src/legacy-transaction.ts
```diff
@@ -0,0 +1,206 @@
...
+      this.gasPrice = Quantity.from(data[1]);
+      this.gas = Quantity.from(data[2]);
+      this.to = data[3].length == 0 ? RPCQUANTITY_EMPTY : Address.from(data[3]);
+      this.value = Quantity.from(data[4]);
+      this.data = Data.from(data[5]);
+      this.v = Quantity.from(data[6]);
+      this.r = Quantity.from(data[7]);
+      this.s = Quantity.from(data[8]);
+      this.raw = [this.type.toBuffer(), ...data];
```
davidmurdoch: I have a hunch that we actually don't ever really need to store `this.type` in a legacy transaction's `raw`, as it gets stripped (as a side effect to calling `computeIntrinsics` a few lines below -- because it calls `raw.shift()`). My hunch is also telling me this would simplify our code in many places; we could just detect this transaction type by checking if the first byte of raw transaction's buffer is `> 0x7f`.

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
```diff
@@ -0,0 +1,229 @@
...
+import {
+  EIP2930AccessListDatabasePayload,
+  EIP2930AccessListDatabaseTx,
+  TypedDatabaseTransaction
+} from "./raw";
+import { AccessList, AccessListBuffer } from "@ethereumjs/tx";
+import { AccessLists } from "./access-lists";
+import { computeInstrinsicsAccessListTx } from "./signing";
+
+const CAPABILITIES: any[] = [2718, 2930];
```
davidmurdoch: Thoughts on using the following type in all the transaction types instead (haven't tested it)?

The `supports` function signatures would need to be updated to `(capability: number): boolean`, of course.

```suggestion
const CAPABILITIES = [2718, 2930] as const;
```

src/chains/ethereum/transaction/src/frozen-transaction.ts
davidmurdoch: I'm thinking that usage of the Frozen Transaction class should be replaced with the right transaction type class, as the correct types are needed in calls like `debug_traceTransaction`. There is some extra information FrozenTransactions have that the others do _not_, so we'll have to figure out how to handle that. I don't want to add a `LegacyFrozenTransaction`, `EIP1559FrozenTransaction`, etc.

davidmurdoch (changes requested):
This is looking great! So many complex changes so this is a partial review. Take a look at the comment and suggestions I've left. We can sync on some of the complex ones whenever.

Nice work!

src/chains/ethereum/block/src/runtime-block.ts
MicaiahReid: Yes, the new version of ethereumjs-vm needs this field.

src/chains/ethereum/transaction/src/access-lists.ts
davidmurdoch: The `getBaseFee` function of the EIP 2930 transaction needs this included in its computation, I believe.

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
```diff
@@ -0,0 +1,228 @@
...
+      AccessListJSON: this.accessListJSON,
+      getSenderAddress: () => ({
+        buf: sender,
+        equals: (a: { buf: Buffer }) => sender.equals(a.buf)
+      }),
+      /**
+       * the minimum amount of gas the tx must have (DataFee + TxFee + Creation Fee)
+       */
+      getBaseFee: () => {
+        const fee = this.calculateIntrinsicGas();
```
davidmurdoch, line 143: This should include the access list fee (`AccessLists.getDataFeeEIP2930(...)`).

src/chains/ethereum/transaction/src/access-lists.ts
```diff
@@ -0,0 +1,126 @@
...
+    common: Common
+  ): number {
+    const accessListStorageKeyCost = common.param(
+      "gasPrices",
+      "accessListStorageKeyCost"
+    );
+    const accessListAddressCost = common.param(
+      "gasPrices",
+      "accessListAddressCost"
+    );
```
davidmurdoch: Instead of using `common.param`, let's use our own params.ts file here. It'd end up being something like:

```typescript
export const Params = {
  /* ... */
  /**
   * Gas cost per storage key in an EIP-2930 Access List transaction
   */
  ACCESS_LIST_ADDRESS_GAS: 2400,

  /**
   * Gas cost per storage key in an EIP-2930 Access List transaction
   */
  ACCESS_LIST_STORAGE_KEY_GAS: 1900
};
  ```
  
I tend to stay away from `common.params` because it can do some heavy lifting in order to find the values it needs.

Now the question is: do we return a BigInt or a number here?

I don't think we need to _compute_ everything via BigInts (which are slower than numbers) because in order for an Access List's cost to exceed `Number.MAX_SAFE_INTEGER` you'd need a whole heck of a lot of entries, more than would ever make any sense for the real world or even theoretic testing.

I _think_ whenever we call this function it's going to be added to a BigInt, so maybe we should just convert the return value to a big int while returning, i.e., `return BigInt(address * accessListAddressCost ...);`.

Comment (davidmurdoch) [reply src/chains/ethereum/transaction/src/access-lists.ts]: Oh, and now that I think about it more, the `getAccessListData` function already loops over all the data. I wonder if it could make sense to just compute the total cost there. Would could just make it return a `cost` param in addition to the `AccessListJSON` and `accessList` props it already returns.

File: src/chains/ethereum/transaction/src/access-lists.ts [hunk 31]
```diff
@@ -0,0 +1,77 @@
+import {
+  AccessList,
+  AccessListBuffer,
+  AccessListItem,
+  isAccessList
+} from "@ethereumjs/tx";
+import Common from "@ethereumjs/common";
+import { bufferToHex, setLengthLeft, toBuffer } from "ethereumjs-util";
```
davidmurdoch:
```suggestion
import { bufferToHex, toBuffer } from "ethereumjs-util";
```

src/chains/ethereum/transaction/src/access-lists.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (accessList && isAccessList(accessList)) {
+      AccessListJSON = accessList;
+      const newAccessList: AccessListBuffer = [];
+
+      for (let i = 0; i < accessList.length; i++) {
+        const item: AccessListItem = accessList[i];
+        const addressBuffer = toBuffer(item.address);
+        const storageItems: Buffer[] = [];
+        for (let index = 0; index < item.storageKeys.length; index++) {
+          storageItems.push(toBuffer(item.storageKeys[index]));
```
davidmurdoch: Use `Data.from(..., 32).toBuffer()` here (and a few lines up) instead of `toBuffer`. We're working on optimizing these back-and-forth conversions of data types, and while we're not there yes, using our internal library for this should help in the future.

src/chains/ethereum/transaction/src/access-lists.ts
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+      bufferAccessList = accessList ? accessList : [];
+      // build the JSON
+      const json: AccessList = [];
+      for (let i = 0; i < bufferAccessList.length; i++) {
+        const data = bufferAccessList[i];
+        const address = bufferToHex(data[0]);
+        const storageKeys: string[] = [];
+        for (let item = 0; item < data[1].length; item++) {
+          storageKeys.push(bufferToHex(data[1][item]));
```
davidmurdoch: Let's use `Data.from(buffer, 32).toString()` here (same reason as above)

src/chains/ethereum/block/src/runtime-block.ts
davidmurdoch: Okay. We set it to the parent's gasUsed in some cases, which seems odd to me.

src/chains/ethereum/ethereum/tests/api/eth/sendTransaction.test.ts
```diff
@@ -167,6 +168,68 @@ describe("api", () => {
           });
         });
       });
+
+      describe("transaction types", () => {
+        let provider: EthereumProvider;
+        let from, to, accessListAcc: string;
```
davidmurdoch: Can you add the `:string` type to `from` and `to`, too?

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
davidmurdoch: From an offline convo: we agree this type makes no sense like this :-)

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -1,6 +1,317 @@
...
+    {
+      name: "ganache",
+      chainId: 1337,
+      comment: "Local test network",
+      bootstrapNodes: []
+    },
+    "berlin"
+  );
+  const options = EthereumOptionsConfig.normalize({});
+  const wallet = new Wallet(options.wallet);
```
davidmurdoch, line 46: You can get the actual private keys, instead of a generating a fake one via `wallet.initialAccounts`.

You can also initialize the options with `wallet: {accounts: [{secretKey: "0x" + "46".repeat(32), balance: 100n}]`. This could be a decent way testing deterministically, allowing us to hardcode results for things test that check the hash and signature fields on the transactions.

* the `"46".repeat(32)` secret key is an example key used in an EIP-155 (https://eips.ethereum.org/EIPS/eip-155, which is the only reason I suggested it.

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -1,6 +1,317 @@
...
+      let txFromRpc: LegacyTransaction;
+      it("generates legacy transactions from rpc data", async () => {
+        const rpc: TypedRpcTransaction = {
+          from: from,
+          to: to,
+          type: "0x0",
+          gasPrice: "0xffff"
+        };
+        txFromRpc = <LegacyTransaction>TransactionFactory.fromRpc(rpc, common);
+        txFromRpc.signAndHash(fakePrivateKey);
```
davidmurdoch: do we need to `signAndHash` here?

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -1,6 +1,317 @@
...
+  describe("LegacyTransaction Type", () => {
+    const rpc: TypedRpcTransaction = {
+      from: from,
+      to: to,
+      type: "0x0",
+      gasPrice: "0xffff"
+    };
+    const tx = <LegacyTransaction>TransactionFactory.fromRpc(rpc, common);
+    it("can be signed and hashed", () => {
+      // TODO: can we know the hash ahead of time if we supply the right data?
```
davidmurdoch: see above comment about deterministic accounts.

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -1,6 +1,317 @@
...
+      });
+      it("has a function to get sender address", () => {
+        assert.notDeepStrictEqual(vmTx.getSenderAddress(), undefined);
+      });
+      it("has a function to get base fee", () => {
+        assert.notDeepStrictEqual(vmTx.getBaseFee(), undefined);
+      });
+      it("has a function to get base upfront cost", () => {
+        assert.notDeepStrictEqual(vmTx.getUpfrontCost(), undefined);
+      });
```
davidmurdoch: I'd like exact values tested here, instead of `undefined`

davidmurdoch (changes requested):
some more comments. i've done an initial review of everything! yay!

src/chains/ethereum/block/src/runtime-block.ts
MicaiahReid: Removed all JSDOC params.

src/chains/ethereum/block/src/runtime-block.ts
MicaiahReid: From what I've seen, we do that whenever we're trying to simulate a transaction or estimate gas. We can discuss, but I'm pretty sure the cases where it's set to the parent's gasUsed makes sense.

src/chains/ethereum/transaction/src/legacy-transaction.ts
MicaiahReid: Yes, you are right. I've removed the type from the legacy transaction's `raw` data.

src/chains/ethereum/transaction/tests/index.test.ts
MicaiahReid: Signing and hashing fills out the v/r/s so that we get a legitimate raw transaction in the next two tests. That being said, this can definitely be cleaner. If we can make the accounts that we use the same for every run of the tests, couldn't we just hard-code the tx data to test?

src/chains/ethereum/transaction/src/access-lists.ts
MicaiahReid: Added to EIP-2930 Transaction's implementation of `getBaseFee`.

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
MicaiahReid: Added.

src/chains/ethereum/transaction/src/access-lists.ts
MicaiahReid: I've incorporated these suggestions for calculating the EIP-2930 base fee.

src/chains/ethereum/transaction/tests/index.test.ts
MicaiahReid: Updated tests to set private key.

src/chains/ethereum/transaction/tests/index.test.ts
MicaiahReid: Updated to use exact values.

src/chains/ethereum/transaction/tests/index.test.ts
MicaiahReid: Updated to have deterministic accounts

src/chains/ethereum/transaction/tests/index.test.ts
MicaiahReid: Since I've updated to use deterministic accounts, we can also have hard-coded tx data, so this is no longer needed.

PR #1065, opened 2021-08-23
fix: increase max payload size to match geth

davidmurdoch:
The default max payload size for websocket connections was quite small, and @cds-amal found a bug that was hitting our limits, except when run in node 16!

The reason it worked in node 16 is that our uWebsockets.js fork doesn't work with node 16, so instead, it uses a pure-JS fallback. This fallback didn't have any upper max payload size, so it happily accepts large payloads.

The limit introduced in this PR matches that of geth. This PR also updates our uws fork to a version that supports the max payload size option in the JS fallback.

### DIFF
--- a/src/chains/ethereum/ethereum/package-lock.json
+++ b/src/chains/ethereum/ethereum/package-lock.json
@@ -535,9 +535,9 @@
       }
     },
     "@trufflesuite/uws-js-unofficial": {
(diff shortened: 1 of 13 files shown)


cds-amal (approved):
Looks good! Thanks @davidmurdoch !

PR #1115, opened 2021-09-03
add `ganache:vm:tx:*` events to provider

davidmurdoch:
In addition to [EIP-1193's](https://eips.ethereum.org/EIPS/eip-1193) `"message"` event and the legacy `"data"` event, we're currently planning on adding 3 more events for the launch of Ganache 7: `"ganache:vm:tx:before"`, `"ganache:vm:tx:step"`, and `"ganache:vm:tx:after"`.

These events can be used to observe the lifecycle of any transaction executed via `*sendTransaction`, `eth_call`, `debug_traceTransaction`, or `debug_storageRangeAt`.

These share the [event paradigm that Truffle uses](https://www.trufflesuite.com/docs/truffle/advanced/event-system#how-to-define-your-event-handlers), but without any of the wildcard handling, i.e., no `"vm:\*"` (for now).

Each of these events will emit a `context` object which is a unique object that can be used to identify a transaction over the course of its lifecycle. For example:

```typescript
interface StepEvent {
  account: {
    nonce: bigint;
    balance: bigint;
    stateRoot: Buffer;
    codeHash: Buffer;
  };
  address: Buffer;
  codeAddress: Buffer;
  depth: number;
  gasLeft: bigint;
  gasRefund: bigint;
  memory: Buffer;
  memoryWordCount: bigint;
  opcode: {
    name: string;
    fee: number;
  };
  pc: number;
  returnStack: Buffer[];
  stack: Buffer[];
}

const contexts = new Map();
provider.on("ganache:vm:tx:before", (event: { context: {} }) => {
  contexts.set(event.context, []);
});
provider.on("ganache:vm:tx:step", (event: StepEvent) => {
  contexts.get(event.context).push(event.data);
});
provider.on("ganache:vm:tx:after", (event: { context: {} }) => {
  doAThingWithThisTransactionsSteps(contexts.get(event.context));
  contexts.delete(event.context);
});
```

The reason this `context` is necessary is that Ganache may run multiple transactions simultaneously, so `"ganache:vm:tx:step"` events from different transactions could be intermingled.

The above events will be emitted for `eth_call`, `*sendTransaction`, `debug_traceTransaction`, and `debug_storageRangeAt`.

Currently, we do not await the event listener's return value, however, we'll likely enable this in the future.


src/chains/ethereum/ethereum/src/provider-events.ts
```diff
@@ -0,0 +1,88 @@
...
+  readonly data: VmStepData;
+};
+
+function normalizeEvent(event: InterpreterStep) {
+  const { account, memory: originalMemory, opcode } = event;
+  const memoryLength = originalMemory.length;
+
+  // We need to copy some buffers so the user can't mutate them on us:
+  // Instead of making a bunch of individual buffers, we just make 1 and then
+  // it up when needed.
```
MicaiahReid: > we just make 1 and then
[REDACTED]
> it up when needed.

Why so secretive?

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -416,15 +431,19 @@ export default class Miner extends Emittery.Typed<
     origin: string,
     pending: Map<string, Heap<TypedTransaction>>
   ) => {
+    const context = {};
+    const vm = this.#vm;
+    this.emit("ganache:vm:tx:before", { context });
+    const stepListener = event => {
```
MicaiahReid, line 473: Could we not do:
```TypeScript
const stepListener = event => {
   this.emit("ganache:vm:tx:step", makeStepEvent(context, event));
};
if(this.#emitStepEvent) {
  vm.on("step", stepListener);
}
```

So we don't re-enter that listener with every step? Just disable it for the entire tx that's being run?
Maybe there are side effects I'm missing here.

src/packages/ganache/README.md
```diff
@@ -192,6 +192,59 @@ const provider = new ethers.providers.Web3Provider(ganache.provider());
 
 New RPC documentation coming soon! See https://github.com/trufflesuite/ganache/tree/master#options for Ganache v2 documentation.
 
+### Ganache Provider Events
```
MicaiahReid, line 195: Copying from slack so we have history here:

Alternate event system ideas:
1. Instead of `context`, name it `token` or `eventToken`.
Because the whole "unique token" thing above is kinda of weird and confusing: Instead of "lifecycle events" (`ganache:vm:tx:before`, `ganache:vm:tx:step`, `ganache:vm:tx:after`) we could emit a single `ganache:vm:tx` event that contains it's own subscription methods:
```TypeScript
interface TransactionEvent {
   onBefore(): (listener: () => void)
   onStep: (listener: (event: StepEvent) => void)
   onAfter(): (listener: () => void)
}
```
alternatively, it could just be a normal event emitter (`.on`), though this necessitates some additional performance considerations (as previously described).
If we did this the user would be in charge of keeping track of each transaction's context.

MicaiahReid (approved):
Approved with a few comments

src/chains/ethereum/ethereum/src/provider-events.ts
davidmurdoch: hahahaa. I somehow accidentally a word there!

davidmurdoch:
```suggestion
  // fill it in as needed.
```

src/chains/ethereum/ethereum/src/miner/miner.ts
davidmurdoch:
```suggestion
    // we always listen to the step event even if `#emitStepEvent` is false in
    // case the user starts listening in the middle of the transaction.
    const stepListener = event => {
```

src/packages/ganache/README.md
davidmurdoch: I've made this an issue for review for milestone 7.0.0 and ship the next alpha how things are now: https://github.com/trufflesuite/ganache/issues/1177

PR #1135, opened 2021-09-07
feat: london

### LINKED ISSUE #939: Support london hardfork
This is basically the same as  trufflesuite/ganache-cli#821 just for London, which is already deployed on Ropsten and Goerli and coming to Ethereum mainnet in early August.

## Expected Behavior

`ganache-cli --hardfork london` should work and behave like other Ethereum nodes running on London hardfork.

## Current Behavior

`Error: Hardfork with name london not supported`

## Possible Solution

I guess the solution is obvious, the EIPs in https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/mainnet-upgrades/london.md need to be implemented and followed when that hardfork setting is there. Most importantly, that means support of EIP-1559 transactions.

## Context

As with trufflesuite/ganache-cli#821, running ganache should be good in finding out gas requirements on real networks, which due to more changes in this hardfork compared to Berlin, is even more wrong for networks that are already on London.

That said, as a testing environment, it would be even more important to be able to run tests with EIP-1559 transaction semantics to get them right in various code that should run against a real network. Right now, that's now possible with using ganache.

## Your Environment
* Version used: `Ganache CLI v6.12.2 (ganache-core: 2.13.2)`
* NodeJS Version: `v10.24.0`
* Operating System and version (include distro if Linux): openSUSE Linux

davidmurdoch:
- Addition of EIP-1559 transactions.
- Added `baseFeePerGas` to a `block`
- When London is activated, use `baseFeePerGas` as the block's `gasPrice`
- When London is activated, re-prioritize the txPool based off of the transaction's new `effectiveGasPrice`
- Write tests for miner and txPool reordering
- Fix some instamine behavior
  - Added a `DESIGN-DECISIONS.md` to explain the change
 - Set london as the default hard fork
 - Fixed bug in how typed transactions are signed (no longer including the tx type)

### DIFF
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -10,13 +10,9 @@ import type Common from "@ethereumjs/common";
 import { encode, decode } from "@ganache/rlp";
 import { BaseFeeHeader, BlockHeader, makeHeader } from "./runtime-block";
 import { keccak } from "@ganache/utils";
-import {
-  EthereumRawBlockHeader,
-  GanacheRawBlock,
-  serialize
-} from "./serialize";
-import { Address } from "@ganache/ethereum-address";
+import { EthereumRawBlockHeader, GanacheRawBlock } from "./serialize";
 import { BlockParams } from "./block-params";
+
 export class Block {
   /**
    *  Base fee per gas for blocks without a parent containing a base fee per gas.
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -64,12 +64,11 @@ export class Block {
     });
   }
 
-  toJSON(includeFullTransactions = false) {
+  toJSON(includeFullTransactions = false, common: Common) {
     const hash = this.hash();
     const txFn = this.getTxFn(includeFullTransactions);
     const hashBuffer = hash.toBuffer();
     const number = this.header.number.toBuffer();
-    const common = this._common;
     const jsonTxs = this._rawTransactions.map((raw, index) => {
       const [from, hash] = this._rawTransactionMetaData[index];
       const extra: GanacheRawExtraTx = [
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -168,7 +168,7 @@ export class Block {
     // each block's base fee must be appropriately "floored" (Math.floor) before
     // the following block's base fee is calculated. If we don't do this we'll
     // end up with compounding rounding errors.
-    // FYI: the more performance but rounding error-prone would be:
+    // FYI: the more performant, but rounding error-prone, way is:
     // return lastMaxBlockBaseFee + (lastMaxBlockBaseFee * ((BASE_FEE_MAX_CHANGE_DENOMINATOR-1)**(blocks-1)) / ((BASE_FEE_MAX_CHANGE_DENOMINATOR)**(blocks-1)))
     while (--blocks) {
       lastMaxBlockBaseFee +=
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -194,6 +194,6 @@ export class Block {
         nextBaseFee = baseFeePerGas - adjustedFeeDelta;
       }
     }
-    return Quantity.from(nextBaseFee).toBuffer(); // TODO there must be a better way
+    return Quantity.from(nextBaseFee).toBuffer();
   }
 }
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -56,6 +56,7 @@ export class Block {
   }
 
   getTransactions() {
+    const common = this._common;
     return this._rawTransactions.map((raw, index) => {
       const [from, hash] = this._rawTransactionMetaData[index];
       const extra: GanacheRawExtraTx = [
@@ -65,7 +66,7 @@ export class Block {
         this.header.number.toBuffer(),
         Quantity.from(index).toBuffer()
       ];
-      return TransactionFactory.fromDatabaseTx(raw, this._common, extra);
+      return TransactionFactory.fromDatabaseTx(raw, common, extra);
     });
   }
 
@@ -74,6 +75,7 @@ export class Block {
     const txFn = this.getTxFn(includeFullTransactions);
     const hashBuffer = hash.toBuffer();
     const number = this.header.number.toBuffer();
+    const common = this._common;
     const jsonTxs = this._rawTransactions.map((raw, index) => {
       const [from, hash] = this._rawTransactionMetaData[index];
       const extra: GanacheRawExtraTx = [
@@ -83,7 +85,7 @@ export class Block {
         number,
         Quantity.from(index).toBuffer()
       ];
-      const tx = TransactionFactory.fromDatabaseTx(raw, this._common, extra);
+      const tx = TransactionFactory.fromDatabaseTx(raw, common, extra);
       return txFn(tx);
     });
 
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -159,7 +159,7 @@ export class Block {
     return nextBaseFee;
   }
 
-  static calcMaxNBlocksBaseFee(blocks: number, parentHeader: BaseFeeHeader) {
+  static calcNBlocksMaxBaseFee(blocks: number, parentHeader: BaseFeeHeader) {
     const { BASE_FEE_MAX_CHANGE_DENOMINATOR } = BlockParams;
 
     let maxPossibleBaseFee = this.calcNextBaseFeeBigInt(parentHeader);
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -162,7 +162,7 @@ export class Block {
   static calcMaxNBlocksBaseFee(blocks: number, parentHeader: BaseFeeHeader) {
     const { BASE_FEE_MAX_CHANGE_DENOMINATOR } = BlockParams;
 
-    let lastMaxBlockBaseFee = this.calcNextBaseFeeBigInt(parentHeader);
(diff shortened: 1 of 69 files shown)


src/chains/ethereum/block/src/block.ts
```diff
@@ -64,12 +64,11 @@ export class Block {
     });
   }
 
-  toJSON(includeFullTransactions = false) {
+  toJSON(includeFullTransactions = false, common: Common) {
```
MicaiahReid: `this` already has access to `common`, but sometimes when we initialize a block, we don't set common to save some extra work that would be done when it's set. (see comment on lines 184-187 of `runtime-block.ts` for more info on that.)

So instead, we pass in common to this function. Any thoughts on always passing common into the `block` constructor, allowing us to use `this._common` here? @davidmurdoch

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
```diff
@@ -242,4 +240,6 @@ export class EIP2930AccessListTransaction extends RuntimeTransaction {
       chainId
     );
   }
+
+  public updateEffectiveGasPrice() {}
```
MicaiahReid, line 252: Definitely open to suggestions on making this cleaner or omitting this method. From what I understand about the abstract method that calls this, we have to have this method defined.

MicaiahReid (approved):
Approved with a few questions for other reviewers' opinions!

src/chains/ethereum/block/src/block.ts
davidmurdoch: Probably fine to leave it as is and fix when we re-organize the transactions classes

src/packages/utils/src/utils/heap.ts
```diff
@@ -25,6 +27,18 @@ export class Heap<T> {
     }
   }
 
+  public refresh(context: U) {
```
davidmurdoch, line 40:
```suggestion
  /**
   * Updates all entries by calling the Heap's `refresher` function for each
   * item in the heap and then re-sorting.
   * @param context
   */
  public refresh(context: U) {
```

src/chains/ethereum/transaction/src/transaction-receipt.ts
```diff
@@ -163,17 +170,12 @@ export class TransactionReceipt {
...
     };
     if (transaction.type && common.isActivatedEIP(2718)) {
       json.type = transaction.type;
     }
-    if ("chainId" in transaction) {
-      json.chainId = transaction.chainId;
-    }
-    if ("chainId" in transaction) {
-      json.accessList = transaction.accessListJSON;
-    }
```
davidmurdoch, line 176: :+1:

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -48,6 +48,9 @@ export type BlockHeader = {
   baseFeePerGas?: Quantity;
 };
 
+export type BaseFeeHeader = BlockHeader &
```
MicaiahReid: Would it make more sense to define `BaseFeeHeader` in `block.ts`? We only ever reference it in `block.ts` and `api.ts`.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -483,8 +483,8 @@ export default class Blockchain extends Emittery.Typed<
       str += `  Contract created: ${Address.from(contractAddress)}${EOL}`;
     }
 
-    str += `  Gas usage: ${Quantity.from(receipt.raw[1]).toNumber()}${EOL}
```
MicaiahReid, line 487: Does this have anything to do with my change?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1086,8 +1092,13 @@ export default class EthereumApi implements Api {
     number: QUANTITY | Tag,
     index: QUANTITY
   ) {
-    const block = await this.eth_getBlockByNumber(number, true);
-    return block.transactions[parseInt(Quantity.from(index).toString(), 10)];
+    const blockchain = this.#blockchain;
```
MicaiahReid, line 1095: Does this have anything to do with my change?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1040,12 +1043,15 @@ export default class EthereumApi implements Api {
    */
   @assertArgLength(2)
   async eth_getTransactionByBlockHashAndIndex(hash: DATA, index: QUANTITY) {
-    const block = await this.eth_getBlockByHash(hash, true);
-    if (block) {
-      const tx = block.transactions[Quantity.from(index).toNumber()];
-      if (tx) return tx;
-    }
-    return null;
+    const blockchain = this.#blockchain;
```
MicaiahReid, line 1046: Does this have anything to do with my change?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1126,12 +1140,12 @@ export default class Blockchain extends Emittery.Typed<
         }
       }
     };
-
+    let mypromise;
```
MicaiahReid: Is this still necessary? I don't remember the details, but I thought the issues that were causing us to add this were only due to a malformed tx that was failing in `miner.runTx`.

src/chains/ethereum/transaction/src/eip1559-fee-market-transaction.ts
```diff
@@ -83,7 +106,7 @@ export class EIP1559FeeMarketTransaction extends RuntimeTransaction {
     this.updateEffectiveGasPrice();
   }
 
-  public toJSON(): EIP1559FeeMarketTransactionJSON {
+  public toJSON(_common?: Common): EIP1559FeeMarketTransactionJSON {
```
MicaiahReid, line 109: Do we need this? Doesn't seem like it's used.

src/chains/ethereum/transaction/src/rpc-transaction.ts
```diff
@@ -26,29 +26,29 @@ export type TypedRpcTransaction =
   | EIP2930AccessListRpcTransaction
   | EIP1559FeeMarketRpcTransaction;
 
-export type LegacyRpcTransaction = RpcTransaction & {
-  gasPrice?: string;
-  chainId?: never;
-  accessList?: never;
-  maxPriorityFeePerGas?: never;
-  maxFeePerGas?: never;
+export type LegacyRpcTransaction = Readonly<RpcTransaction> & {
```
MicaiahReid, line 29: Curious about the reasoning for this!

src/chains/ethereum/transaction/src/signing.ts
```diff
@@ -232,14 +245,24 @@ export const computeInstrinsicsFeeMarketTx = (
       encodedData.length + encodedSignature.length
     )
   ]);
+
+  const data = Buffer.concat([
+    typeBuf,
+    digest([encodedData.output], encodedData.length)
+  ]);
+  const senderPubKey = _ecdsaRecover(
```
MicaiahReid, line 254: Are we not bypassing our checks on the chainId portion of the signature for these?

src/chains/ethereum/transaction/src/transaction-factory.ts
```diff
@@ -168,59 +169,91 @@ export class TransactionFactory {
...
+    }
   }
   /**
    * Create a transaction from a `txData` object
    *
+   * When transaction types are activated (EIP 2718) the txData will be checked
+   * for a transaction envelope (first byte < 192) before determining the
+   * decoding strategy, otherwise it will be decoded as a Legacy Transaction. If
+   * the transaction contains a transaction envelop, but EIP 2718 is not active
+   * decoding will fail and an exception will be thrown.
```
MicaiahReid, line 205: It seems worthwhile to add an actual error for the case  where a transaction envelope is sent but EIP 2718 is not activated. Otherwise it's a random buffer related error from `decode`, which would be a bit confusing.

src/chains/ethereum/block/src/block.ts
```diff
@@ -156,44 +121,68 @@ export class Block {
...
+        (baseFeePerGas * gasUsedDelta) /
+        parentGasTarget /
+        BlockParams.BASE_FEE_MAX_CHANGE_DENOMINATOR;
+      nextBaseFee = baseFeePerGas - adjustedFeeDelta;
+    }
+
+    return nextBaseFee;
+  }
+
+  static calcMaxNBlocksBaseFee(blocks: number, parentHeader: BaseFeeHeader) {
```
MicaiahReid: I think the name `calcNBlocksMaxBaseFee` or `calcMaxBaseFeeNBlocks` would be better names for this function. The current name doesn't really convey what it returns, the maximum possible base fee after N blocks.

src/chains/ethereum/block/src/block.ts
```diff
@@ -156,44 +121,68 @@ export class Block {
...
+    const { BASE_FEE_MAX_CHANGE_DENOMINATOR } = BlockParams;
+
+    let lastMaxBlockBaseFee = this.calcNextBaseFeeBigInt(parentHeader);
```
MicaiahReid: I propose a rename of this to `maxPossibleBaseFee`. Since we use it in the while loop below, it is not always `last` block.

MicaiahReid (approved):

src/chains/ethereum/transaction/src/eip1559-fee-market-transaction.ts
davidmurdoch: I think the `_common` is needed because the super class needs it because `LegacyTransaction` needs it. The `_` makes the linter stop complaining about an unused variable.

src/chains/ethereum/transaction/src/rpc-transaction.ts
davidmurdoch: We should never mutate the json data a user gives us (like in `provider.send("eth_sendTransaction", transaction)`) but in one place in our code we were. Marking the type as `Readonly` prevents that sort of bug from happening.

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: These were types issues I uncovered while trying to fix another issue.

PR #1227, opened 2021-09-22
fix: handle initialization errors on server.listen, handle fallback rejections

### LINKED ISSUE #938: When forking, if we make archive state requests to infura, but we aren't using an archive node, the errors aren't handled at all

davidmurdoch:
fixes https://github.com/trufflesuite/ganache/issues/938

This is a all weird because I want users to be able to do const provider = Ganache.provider()... notice there is no await here! :eyes:  Ganache.provider does approx 3,452,345 async things on start up though, and if any of these promises reject no one is there to listen to that rejection; so node complains (but doesn't crash in Node v10-14).

Unfortunately, this PR does not fix this problem for Ganache.provider(), but does fix it for Ganache.server() by having the server.listen function, which is async, handle promise rejections "eventually".

In order to get the provider to handle these errors we'd have to hook into the first async method the user calls after we have finished (failing) initialization... which would probably be something like await provider.request({method:"eth_accounts", params: []})... which is a weird place to throw an initialization error :confused:.

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -256,140 +256,149 @@ export default class Blockchain extends Emittery.Typed<
     const options = this.#options;
     const instamine = this.#instamine;
 
-    let common: Common;
-    if (this.fallback) {
-      await Promise.all([database.initialize(), this.fallback.initialize()]);
-      common = this.common = this.fallback.common;
-      options.fork.blockNumber = this.fallback.blockNumber.toNumber();
-      options.chain.networkId = common.networkId();
-      options.chain.chainId = common.chainId();
-    } else {
-      await database.initialize();
-      common = this.common = createCommon(
-        options.chain.chainId,
-        options.chain.networkId,
-        options.chain.hardfork
-      );
-    }
+    try {
+      let common: Common;
+      if (this.fallback) {
+        await this.fallback.initialize();
+        await database.initialize();
 
-    const blocks = (this.blocks = await BlockManager.initialize(
-      this,
-      common,
-      database.blockIndexes,
-      database.blocks
-    ));
-
-    this.blockLogs = new BlockLogManager(database.blockLogs, this);
-    this.transactions = new TransactionManager(
-      options.miner,
-      common,
-      this,
-      database.transactions
-    );
-    this.transactionReceipts = new TransactionReceiptManager(
-      database.transactionReceipts,
-      this
-    );
-    this.accounts = new AccountManager(this);
-    this.storageKeys = database.storageKeys;
-
-    // if we have a latest block, use it to set up the trie.
-    const { latest } = blocks;
-    {
-      let stateRoot: Data | null;
-      if (latest) {
-        this.#blockBeingSavedPromise = Promise.resolve({
-          block: latest,
-          blockLogs: null
-        });
-        ({ stateRoot } = latest.header);
+        common = this.common = this.fallback.common;
+        options.fork.blockNumber = this.fallback.blockNumber.toNumber();
+        options.chain.networkId = common.networkId();
+        options.chain.chainId = common.chainId();
       } else {
-        stateRoot = null;
-      }
-      this.trie = makeTrie(this, database.trie, stateRoot);
-    }
-
-    // create VM and listen to step events
-    this.vm = await this.createVmFromStateTrie(
-      this.trie,
-      options.chain.allowUnlimitedContractSize,
-      true
-    );
-
-    {
-      // create first block
-      let firstBlockTime: number;
-      if (options.chain.time != null) {
-        // If we were given a timestamp, use it instead of the `_currentTime`
-        const t = options.chain.time.getTime();
-        firstBlockTime = Math.floor(t / 1000);
-        this.setTime(t);
-      } else {
-        firstBlockTime = this.#currentTime();
-      }
-
-      // if we don't already have a latest block, create a genesis block!
-      if (!latest) {
-        if (initialAccounts.length > 0) {
-          await this.#commitAccounts(initialAccounts);
-        }
-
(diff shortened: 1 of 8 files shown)


src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -256,140 +256,149 @@ export default class Blockchain extends Emittery.Typed<
...
-    const { latest } = blocks;
-    {
-      let stateRoot: Data | null;
-      if (latest) {
-        this.#blockBeingSavedPromise = Promise.resolve({
-          block: latest,
-          blockLogs: null
-        });
-        ({ stateRoot } = latest.header);
+    try {
```
davidmurdoch, line 259: this is just wrapping things in a try/catch

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -256,140 +256,149 @@ export default class Blockchain extends Emittery.Typed<
...
+      let common: Common;
+      if (this.fallback) {
+        await this.fallback.initialize();
+        await database.initialize();
```
davidmurdoch, line 263: This was changed because previously, if the `fallback` throws while database is still starting it becomes very tricky to ensure we can wait for the database to finish before trying to shut the blockchain class down. We _can_  do it, and I did, but it was a lot of code involving handling the weird return values of `Promise.allSettled` combined with throwing an `AggregateError` in some cases and not in others.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1567,25 +1576,27 @@ export default class Blockchain extends Emittery.Typed<
...
     }
 
+    this.#state = Status.stopping;
+
     // stop the polling miner, if necessary
     clearTimeout(this.#timer);
 
     // clean up listeners
-    this.vm.removeAllListeners();
+    this.vm && this.vm.removeAllListeners();
```
davidmurdoch, line 1596: Since we now call `stop` internally on initialization exceptions we might not have everything ready, so we check for things' existence before trying to clean them up.

src/chains/ethereum/ethereum/src/forking/handlers/http-handler.ts
```diff
@@ -172,7 +172,7 @@ export class HttpHandler extends BaseHandler implements Handler {
       if ("result" in result) {
         return result.result;
       } else if ("error" in result) {
-        throw result.error;
+        throw new CodedError(result.error.message, result.error.code);
```
davidmurdoch, line 175: don't throw plain objects, kids.

src/packages/core/src/connector-loader.ts
```diff
@@ -36,14 +36,16 @@ const initialize = <T extends FlavorName = typeof DefaultFlavor>(
...
   // The request coordinator is initialized in a "paused" state; when the
   // provider is ready we unpause.. This lets us accept queue requests before
   // we've even fully initialized.
-  connectPromise.then(requestCoordinator.resume);
 
-  return connector;
+  return {
+    connector,
+    promise: connectPromise.then(requestCoordinator.resume)
+  };
```
davidmurdoch, line 48: give connector loader callers, like server.ts, the ability to handle start up promise rejections.

src/packages/core/src/server.ts
```diff
@@ -6,6 +6,15 @@ import {
...
 
 import allSettled from "promise.allsettled";
+
+// This `shim()` is necessary for `Promise.allSettled` to be shimmed
+// in `node@10`. We cannot use `allSettled([...])` directly due to
+// https://github.com/es-shims/Promise.allSettled/issues/5 without
+// upgrading Typescript. TODO: if Typescript is upgraded to 4.2.3+
+// then this line could be removed and `Promise.allSettled` below
+// could replaced with `allSettled`.
+allSettled.shim();
```
davidmurdoch, line 16: I just moved this block to the global scope.

src/packages/core/src/server.ts
```diff
@@ -125,10 +134,16 @@ export class Server<
...
-      this.#providerOptions
-    ));
-    this.#initializer = this.initialize(connector);
+    const loader = ConnectorLoader.initialize(this.#providerOptions);
+    const connector = (this.#connector = loader.connector);
+
+    // Since the ConnectorLoader starts an async promise that we intentionally
+    // don't await yet we keep the promise around for something else to handle
+    // later.
+    this.#initializer = Promise.all([
```
davidmurdoch, line 143: this promise is awaited in the `listen` function below, and that is where a Connector/Provider's initialization rejections will get handled now.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -256,140 +256,152 @@ export default class Blockchain extends Emittery.Typed<
...
-      }
-      //#endregion
+        miner.on("block", this.#handleNewBlockData);
 
-      miner.on("block", this.#handleNewBlockData);
+        this.once("stop").then(() => miner.clearListeners());
+      }
+    } catch (e) {
+      // we failed to start up :-( bail!
+      this.#state = Status.stopping;
```
davidmurdoch, line 399: This huge try/catch is all for this little line here. If we fail to start the database (and everything else) let's make sure we can `stop` (stop requires things aren't in a `starting` stage... because we don't want to stop whilst starting otherwise the database can get fussy)

MicaiahReid: > stop requires things are in a `starting` stage...

Did you mean 

> stop requires things are in a `stopping` stage

MicaiahReid (approved):
Added one clarification comment to your annotations.

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: Thanks, I meant "stop requires things aren't in a starting stage"

PR #1237, opened 2021-09-22
Fix/same nonce replacement txs

### LINKED ISSUE #1218: Replacement transactions are run twice
If a transaction replaced another transaction in the pool (because it is "better" i.e., has a higher gas price) the better transaction will end up getting executed twice :exploding_head:

### LINKED ISSUE #1219: Transactions that aren't immediately executable can't be replaced by "better" transactions
The transaction pool only looks at immediately executable transactions (transactions whose `nonce` matches the account's current nonce); it doesn't consider any future nonce transactions for replacement.

MicaiahReid:
Prevents a transaction that replaces a previously queued transaction from being re-added to the transaction pool and thus run twice.

Previously when a transaction was sent we checked:
- do we have executable transactions in the transaction pool from the same origin?
- does this new transaction have the same nonce as one of those?
- if so, is this new transaction's gas price better?
- if so, replace the previous transaction with this new one.

We now perform all of those same checks against the future-nonce transactions (transactions whose nonce is higher than the sender account's nonce) in the transaction pool, not just the executable transactions.

Fixes #1219 and #1218

### DIFF
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -17,7 +17,33 @@ function byNonce(values: TypedTransaction[], a: number, b: number) {
     (values[b].nonce.toBigInt() || 0n) > (values[a].nonce.toBigInt() || 0n)
   );
 }
-
+/**
+ * Used to track a transaction's placement in the transaction pool based off
+ * of the its nonce.
+ */
+export enum TriageOptions {
+  /**
+   * Default value. A tx will be added to the future queue if it is not yet
+   * executable based off of the transaction's nonce.
+   */
+  FutureQueue = 0,
+  /**
+   * The transaction is currently executable based off the transaction's nonce.
+   */
+  Executable = 1,
+  /**
+   * The transaction is currently executable, has the same nonce as a pending
+   * transaction of the same origin, and has a gas price that is high enough to
+   * replace the currently pending transaction.
+   */
+  ReplacesPendingExecutable = 2,
+  /**
+   * The transaction is not currently executable but has the same nonce as a
+   * future queued transaction of the same origin and has a gas price that is
+   * high enough to replace the future queued transaction.
+   */
+  ReplacesFutureTransaction = 3
+}
 export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
   #options: EthereumInternalOptions["miner"];
 
@@ -104,10 +130,12 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
     const origins = this.#origins;
     const queuedOriginTransactions = origins.get(origin);
 
-    let transactionIsAlreadyQueued = false;
-    let isExecutableTransaction = false;
+    let transactionPlacement = TriageOptions.FutureQueue;
     const executables = this.executables.pending;
     let executableOriginTransactions = executables.get(origin);
+
+    const priceBump = this.#priceBump;
+    const newGasPrice = transaction.effectiveGasPrice.toBigInt();
 
     let length: number;
     if (
@@ -118,8 +146,6 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
       // executables queue already. Replace the matching transaction or throw this
       // new transaction away as necessary.
       const pendingArray = executableOriginTransactions.array;
-      const priceBump = this.#priceBump;
-      const newGasPrice = transaction.effectiveGasPrice.toBigInt();
       // Notice: we're iterating over the raw heap array, which isn't
       // necessarily sorted
       for (let i = 0; i < length; i++) {
@@ -139,7 +165,7 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
             // we don't want to mark this transaction as "executable" and thus
             // have it added to the pool again. so use this flag to skip
             // a re-queue.
-            transactionIsAlreadyQueued = true;
+            transactionPlacement = TriageOptions.ReplacesPendingExecutable;
 
             currentPendingTx.finalize(
               "rejected",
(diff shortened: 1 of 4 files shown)


src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -185,15 +216,69 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
...
+    ) {
+      // check if a transaction with the same nonce is in the origin's
+      // future queue already. Replace the matching transaction or throw this
+      // new transaction away as necessary.
+      const queuedArray = queuedOriginTransactions.array;
+      const priceBump = this.#priceBump;
+      const newGasPrice = transaction.effectiveGasPrice.toBigInt();
+      // Notice: we're iterating over the raw heap array, which isn't
+      // necessarily sorted
+      for (let i = 0; i < length; i++) {
```
MicaiahReid, line 266: This is mostly the same as the for loop [above](https://github.com/trufflesuite/ganache/pull/1237/files#diff-ca13abb534f8ddc94c9f9ae1309a903ad18f3f06f07b94653412d29f006178b2R151-R187), the differences being:
- above we set `transactionPlacement` to `TriageOptions.ReplacesExecutableTransaction`; here we set it to `TriageOptions.ReplacesFutureTransaction`
- above we need to track the `highestNonce`; here we do not.

I couldn't think of a great way to DRY this up. Every time I tried I only made it uglier. But I'm definitely open to suggestions.

davidmurdoch: Yeah, I see how it's tricky.

For the first problem, maybe you could pull this out into its own function and pass in the `transactionPlacement` value to use (`ReplacesExecutableTransaction` or `ReplacesFutureTransaction`)

and then also create a function whose sole purpose is to loop of the queue and then execute functions that you give it.

```
function loopTxs(txsToLoop: Transactions[], length: number, fnsToCall: (tx: transaction, nonce: bigint) => boolean[]) {
   for (let i = 0; i < length; i++) {
      const currentTx = txsToLoop[i];
      const thisNonce = currentTx.nonce.toBigInt();
      for (let j = 0, l = fnsToCall.length; j < l; j++) {
        // if a function returns false we can exit early without continuing to loop
        if ( !fnsToCall[j](currentTx, thisNonce) ) return;
      }
   }
}
```

For the `executableOriginTransactions` you'd pass it two functions, the one that does the replacement, and one that keeps track of the `highestNonce`, and for the `queuedOriginTransactions` you'd only pass it the replacement function.

The above example allows for returning `false` from any of the `fnsToCall` in order to stop the loop. This would be only be beneficial in the case of the `queuedOriginTransactions` in the case that it replaces a transaction in the queue, since once that happens we don't need to keep looping. I can explain this idea further over zoom if you'd like, as there is some additional finagling that you'd need to do.

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -185,15 +216,69 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
...
+          // oldPrice, throw out the old now.
+          if (!currentQueuedTx.locked && newGasPrice > thisPricePremium) {
+            // do an in-place replace without triggering a re-sort because we
+            // already know where this transaction should go in this "byNonce"
+            // heap.
+            queuedArray[i] = transaction;
+            // we don't want to mark this transaction as "FutureQueue" and thus
+            // have it added to the pool again. so use this flag to skip
+            // a re-queue.
+            transactionPlacement = TriageOptions.ReplacesFutureTransaction;
```
davidmurdoch: we should exit the loop early (`break;`) if we found a replacement, since we aren't ever going to find another.

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -221,7 +306,17 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
...
+      // we've replaced the best transaction from this origin for this nonce,
+      // and it is executable
+      return true;
+    } else if (
+      transactionPlacement === TriageOptions.ReplacesFutureTransaction
+    ) {
+      // we've replaced the best transaction from this origin for a future
+      // nonce, so this one isn't executable
+      return false;
+    } else if (transactionPlacement === TriageOptions.Executable) {
```
davidmurdoch: Let's reorder this so the most likely case, `TriageOptions.Executable`, is checked first. Also, what do you think about changing these ifs and elses to a `switch`?

src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
```diff
@@ -437,7 +437,7 @@ describe("api", () => {
             { from, to, value: value++, nonce: accountNonce + 3 }
           ]),
           send("eth_sendTransaction", [
-            { from, to, value: value++, nonce: accountNonce + 3 }
+            { from, to, value: value++, nonce: accountNonce + 4 }
```
davidmurdoch, line 440: Nice find!

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -17,7 +17,33 @@ function byNonce(values: TypedTransaction[], a: number, b: number) {
     (values[b].nonce.toBigInt() || 0n) > (values[a].nonce.toBigInt() || 0n)
   );
 }
-
+/**
+ * Used to track a transaction's placement in the transaction pool based off
+ * of the its nonce.
+ */
+export enum TriageOptions {
```
davidmurdoch: I think this should be `TriageOption`.


It  seems to make more sense when you look at it as a type that holds a singular value:

```
function fn(triageOption: TriageOption) {}
```

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -17,7 +17,33 @@ function byNonce(values: TypedTransaction[], a: number, b: number) {
...
+   * transaction of the same origin, and has a gas price that is high enough to
+   * replace the currently pending transaction.
+   */
+  ReplacesPendingExecutable = 2,
+  /**
+   * The transaction is not currently executable but has the same nonce as a
+   * future queued transaction of the same origin and has a gas price that is
+   * high enough to replace the future queued transaction.
+   */
+  ReplacesFutureTransaction = 3
```
davidmurdoch, line 82: These are excellent descriptions!

davidmurdoch (changes requested):

davidmurdoch:
Looks like tests aren't completing right now. See the `removes transactions that are in pending transactions at the start of evm_revert` test. It passes, but then mocha just quits. This seems to happen when Node.js has determined that the application has hit a dead end and thus no more events will be occurring. Let me know if you want to sync on this one!

davidmurdoch (approved):

PR #1248, opened 2021-09-24
Add in-memory cache for forking requests

### LINKED ISSUE #1224: Implement in-memory cache for forking requests

davidmurdoch:
closes #1224


src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -115,13 +115,22 @@ export default class BlockManager extends Manager<Block> {
...
+    if (typeof tagOrBlockNumber === "string") {
+      blockNumber = tagOrBlockNumber;
+    } else if (tagOrBlockNumber.toBigInt() > fallback.blockNumber.toBigInt()) {
+      // don't get the block if the requested block is _after_ our fallback's
+      // blocknumber because it doesn't exist in our local chain.
+      return null;
+    } else {
+      blockNumber = tagOrBlockNumber.toString();
+    }
+
```
davidmurdoch, line 130: previously, requesting blocks that exist on the original chain, but not locally, would always just return our lcoal block's "latest" instead of `null`. this fixes it -- probably. :-)

src/chains/ethereum/ethereum/src/forking/fork.ts
```diff
@@ -10,31 +10,32 @@ import { Block } from "@ganache/ethereum-block";
...
   blockNumber: Quantity | Tag.LATEST
 ) {
-  return fork
-    .request<string>("eth_getTransactionCount", [address, blockNumber])
-    .then(nonce => Quantity.from(nonce));
+  const nonce = await fork.request<string>("eth_getTransactionCount", [
+    address,
+    blockNumber
+  ]);
+  return Quantity.from(nonce);
```
davidmurdoch, line 38: `vscode suggested it could convert all these to `async`, so I clicked the button, and this is what it made. I like it.

src/chains/ethereum/ethereum/src/forking/handlers/base-handler.ts
```diff
@@ -33,24 +36,31 @@ export class BaseHandler {
...
-    const headers: Headers = {
-      "user-agent": userAgent
-    };
-    if (origin) {
-      headers["origin"] = origin;
-    }
+    this.valueCache = new LRU({
+      max: 1_073_741_824, // 1 gigabyte
+      length: (value, key) => {
+        return value.length + key.length;
```
davidmurdoch, line 42: To measure the cache size we measure the string length of the key and string length of the value. it's technically not perfect, but should be close enough for these purposes.

src/chains/ethereum/ethereum/src/forking/handlers/base-handler.ts
```diff
@@ -122,4 +132,46 @@ export class BaseHandler {
...
+      ...args: unknown[]
+    ) => Promise<{
+      response: { result: any } | { error: { message: string; code: number } };
+      raw: string | Buffer;
+    }>
+  ): Promise<T> {
+    const cached = this.getFromCache<T>(key);
+    if (cached !== undefined) return cached;
+
+    const promise = this.limiter.handle(send).then(({ response, raw }) => {
```
davidmurdoch, line 158: The `send` function actually does the sending, the limiter just schedules when to call that funciton. The `send` function returns a `{response, raw}` object here, `raw` is just used for the cache, and is a `string` or `Buffer`.

I did this because we can't measure the size of a json object for the cache, but we can for a string and Buffer.

src/chains/ethereum/ethereum/src/forking/handlers/base-handler.ts
```diff
@@ -33,24 +36,33 @@ export class BaseHandler {
...
+      max: 1_073_741_824, // 1 gigabyte
+      length: (value, key) => {
+        return value.length + key.length;
+      }
+    });
 
-    // we set our own Authentication headers, so username and password must be
-    // removed from the url. (The values have already been copied to the options)
-    url.password = url.username = "";
+    // we don't need to header-related things if we are using a provider
```
MicaiahReid:
```suggestion
    // we don't need header-related things if we are using a provider
```

src/chains/ethereum/ethereum/src/forking/handlers/provider-handler.ts
```diff
@@ -0,0 +1,77 @@
...
+      throw new Error("Forking `provider` must be EIP-1193 compatible");
+    }
+  }
+  public async request<T>(method: string, params: unknown[]) {
+    // format params via JSON stringification because the params might
+    // be Quantity or Data, which aren't valid as `params` themselves,
+    // but when JSON stringified they are
+    const strParams = JSON.stringify(params);
+
+    return await this.queueRequest<T>(`${method}:${strParams}`, () =>
```
MicaiahReid, line 70: Above in `http-handler`, the `key` for `queueRequest` is `JSON.stringify({ method, params })`, which won't be equal to the `key` for this request, `'${method}:${strParams}'`. Does that matter?

src/chains/ethereum/ethereum/tests/forking/forking.test.ts
```diff
@@ -39,6 +41,276 @@ describe("forking", () => {
       console.log(e);
     }
   });
+  describe("invalid/exceptional responses", async () => {
+    // doesn'contents of FAKE_BLOCK don't matter, it just needs to be parsable
```
MicaiahReid:
```suggestion
    // contents of FAKE_BLOCK don't matter, it just needs to be parsable
```

MicaiahReid (approved):

src/chains/ethereum/ethereum/src/forking/handlers/provider-handler.ts
davidmurdoch: It is fine. Each Handler maintains its own cache (and only one Handler can ever be used at a time anyway). The key just needs to be formatted the same every time it is used in the same Handler.

src/chains/ethereum/ethereum/src/forking/handlers/base-handler.ts
cds-amal: Your comment should be part of the code

Issue #1214, opened 2021-09-20
`eth_call` needs additional eip-1559 gas logic


davidmurdoch:
I've prototyped the following, but it is incomplete, and I haven't tested yet. See https://github.com/ethereum/go-ethereum/blob/b1a5e4afdd80e45b570c80fa0319ed2816a5e30b/accounts/abi/bind/backends/simulated.go#L586-L611

```typescript
  async eth_call(transaction: any, blockNumber: QUANTITY | Tag = Tag.LATEST) {
    const blockchain = this.#blockchain;
    const blocks = blockchain.blocks;
    const parentBlock = await blocks.get(blockNumber);
    const parentHeader = parentBlock.header;
    const options = this.#options;

    let gas: Quantity;
    if (typeof transaction.gasLimit === "undefined") {
      if (typeof transaction.gas !== "undefined") {
        gas = Quantity.from(transaction.gas);
      } else {
        // eth_call isn't subject to regular transaction gas limits by default
        gas = options.miner.callGasLimit;
      }
    } else {
      gas = Quantity.from(transaction.gasLimit);
    }

    let data: Data;
    if (typeof transaction.data === "undefined") {
      if (typeof transaction.input !== "undefined") {
        data = Data.from(transaction.input);
      }
    } else {
      data = Data.from(transaction.data);
    }

    // eth_call doesn't validate that the transaction has a sufficient
    // "effectiveGasPrice" *unless* the transaction has specified *non-zero*
    // `maxPriorityFeePerGas` or `maxFeePerGas` values.
    let baseFeePerGasBigInt = 0n;

    let gasPrice: Quantity;
    if (typeof transaction.gasPrice !== "undefined") {
      gasPrice = Quantity.from(transaction.gasPrice);
    } else {
      // User specified 1559 gas fields (or none), use those
      let maxFeePerGas = 0n;
      let maxPriorityFeePerGas = 0n;
      if (typeof transaction.maxFeePerGas !== "undefined") {
        maxFeePerGas = BigInt(transaction.maxFeePerGas);
      }
      if (typeof transaction.maxPriorityFeePerGas !== "undefined") {
        maxPriorityFeePerGas = BigInt(maxPriorityFeePerGas);
      }
      if (maxPriorityFeePerGas > 0 || maxFeePerGas > 0) {
        baseFeePerGasBigInt = Block.calcNextBaseFee(parentBlock);
        const a = maxFeePerGas - baseFeePerGasBigInt;
[...]

PR #1307, opened 2021-10-05
EIP-1559 Transaction Gas Fixes

### LINKED ISSUE #1213: Type 2 gasPrice needs to be handled better
I believe 1559 TX's `toVmTransaction` should return a `gasPrice` instead of maxPriorityFeePerGas or maxFeePerGas.

We may need to make the transaction's `extra` data include its block's `baseFeePerGas` so we can always compute the `effectiveGasPrice`?

MicaiahReid:
- Adds `effectiveGasPrice` to `GanacheRawExtraTx` so that it is stored in the db.
  - Uses forked tx's `gasPrice` to store `effectiveGasPrice` when creating the forked tx.
  - Uses stored `effectiveGasPrice` to set `gasPrice` when getting a tx from the db.
  - This fixes a bug where certain database transaction's `gasPrice` when fetching via `eth_getTransactionByHash` would sometimes be wrong.
- Updates test for `eth_getTransactionByHash` to check more values than just `transactionIndex`.
- Updates forking transaction test to make sure `gasPrice` values line up for the very specific fail case.
- Fixes broken tests in the miner that were caused by the the incorrect `gasPrice` bug
- Uses eip-1559 gas fields in eth_call
- Adds very basic tests for eth_call
- Fixes bug in `RuntimeBlock` where a 0 value for `baseFeePerGas` would completely leave out the `baseFeePerGas` field from the block.
- Fixes bug in `TransactionFactory` where a tx that omits the type but has `maxFeePerGas`/`maxPriorityFeePerGas` specified would have those fields overwritten. (no `type` => assume type 0, no `gasPrice` => upgrade to type 2 => assume no type 2 fields were given. the last assumption is wrong)
- Adds field checks that were previously left out of eip-1559 tx tests

Closes #1213 & #1214

### DIFF
--- a/src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
+++ b/src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
@@ -7,6 +7,7 @@ import PromiseQueue from "@ganache/promise-queue";
 import type Common from "@ethereumjs/common";
 import { Data, Quantity } from "@ganache/utils";
 import {
+  GanacheRawExtraTx,
   TransactionFactory,
   TypedRpcTransaction,
   TypedTransaction
@@ -47,19 +48,20 @@ export default class TransactionManager extends Manager<NoOp> {
       [Data.from(transactionHash).toString()]
     );
     if (tx == null) return null;
-    const extra = [
+    const blockHash = Data.from((tx as any).blockHash, 32);
+    const blockNumber = Quantity.from((tx as any).blockNumber);
+    const index = Quantity.from((tx as any).transactionIndex);
+
+    const extra: GanacheRawExtraTx = [
       Data.from(tx.from, 20).toBuffer(),
       Data.from((tx as any).hash, 32).toBuffer(),
-      Data.from((tx as any).blockHash, 32).toBuffer(),
-      Quantity.from((tx as any).blockNumber).toBuffer(),
-      Quantity.from((tx as any).transactionIndex).toBuffer()
-    ] as any;
+      blockHash.toBuffer(),
+      blockNumber.toBuffer(),
+      index.toBuffer(),
+      Quantity.from(tx.gasPrice).toBuffer()
+    ];
     const runTx = TransactionFactory.fromRpc(tx, fallback.common, extra);
-    return runTx.serializeForDb(
-      Data.from((tx as any).blockHash, 32),
-      Quantity.from((tx as any).blockNumber),
-      Quantity.from((tx as any).transactionIndex)
-    );
+    return runTx.serializeForDb(blockHash, blockNumber, index);
   };
 
   public async getRaw(transactionHash: Buffer): Promise<Buffer> {
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -145,6 +145,19 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
     if (transactorNoncePromise) {
       await transactorNoncePromise;
     }
+    // if the user called sendTransaction or signTransaction, effectiveGasPrice
+    // hasn't been set yet on the tx. calculating the effectiveGasPrice requires
+    // the block context, so we need to set it outside of the transaction. this
+    // value is updated in the miner as we're more sure of what block the tx will
+    // actually go on, but we still need to set it here for to check for valid
+    // transaction replacements of same origin/nonce transactions
+    if (
+      !transaction.effectiveGasPrice &&
+      this.#blockchain.common.isActivatedEIP(1559)
+    ) {
+      const baseFeePerGas = this.#blockchain.blocks.latest.header.baseFeePerGas;
+      transaction.updateEffectiveGasPrice(baseFeePerGas);
+    }
 
     // we should _probably_ cache `highestNonce`, but it's actually a really hard thing to cache as the current highest
     // nonce might be invalidated (like if the sender doesn't have enough funds), so we'd have to go back to the previous
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -145,11 +145,11 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
     if (transactorNoncePromise) {
       await transactorNoncePromise;
     }
-    // if the user called sendTransaction or signTransaction, effectiveGasPrice
+    // if the user called sendTransaction or sendRawTransaction, effectiveGasPrice
     // hasn't been set yet on the tx. calculating the effectiveGasPrice requires
     // the block context, so we need to set it outside of the transaction. this
     // value is updated in the miner as we're more sure of what block the tx will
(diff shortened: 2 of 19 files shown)


src/chains/ethereum/ethereum/tests/forking/transaction.test.ts
```diff
@@ -35,7 +43,7 @@ describe("forking", () => {
           .then(req => JSON.parse(req.text).result),
         provider.send("eth_getTransactionByHash", [txHash])
       ]);
-
+      console.log(tx, originalTx);
```
MicaiahReid:
```suggestion
```

src/chains/ethereum/transaction/src/transaction-factory.ts
```diff
@@ -117,8 +117,21 @@ export class TransactionFactory {
...
+              txData.maxFeePerGas === undefined ||
+              txData.maxFeePerGas === ""
+            ) {
+              tx.maxFeePerGas = Quantity.from(null);
+            }
+            if (
+              txData.maxPriorityFeePerGas === undefined ||
+              txData.maxPriorityFeePerGas === ""
+            ) {
+              tx.maxPriorityFeePerGas = RPCQUANTITY_ZERO;
```
davidmurdoch: Let's just do a falsey check on these e because  the empty string case should be impossible anyway, as these two tx.* values are Quantities, which won't allow `""` anyway.

src/chains/ethereum/transaction/src/eip1559-fee-market-transaction.ts
```diff
@@ -122,7 +122,14 @@ export class EIP1559FeeMarketTransaction extends RuntimeTransaction {
...
     }
-    this.updateEffectiveGasPrice();
+    // there's a chance this value was already set from the `extra` passed to
+    // the base transaction, and we don't want to overwrite it
+    if (
+      this.effectiveGasPrice === undefined ||
+      this.effectiveGasPrice.valueOf() === undefined
+    ) {
+      this.updateEffectiveGasPrice();
+    }
```
davidmurdoch: Maybe we should call `this.updateEffectiveGasPrice()` at the end of `autofillDefaultTransactionValues` instead of here. It seems out of place check these values here, as `maxFeePerGas`, which `updateEffectiveGasPrice` uses internally, might not be set yet (or `Quantity.isNull() === true`). I could be wrong though :sweat_smile:

src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
```diff
@@ -47,19 +48,20 @@ export default class TransactionManager extends Manager<NoOp> {
...
       Data.from(tx.from, 20).toBuffer(),
       Data.from((tx as any).hash, 32).toBuffer(),
-      Data.from((tx as any).blockHash, 32).toBuffer(),
-      Quantity.from((tx as any).blockNumber).toBuffer(),
-      Quantity.from((tx as any).transactionIndex).toBuffer()
-    ] as any;
+      blockHash.toBuffer(),
+      blockNumber.toBuffer(),
+      index.toBuffer(),
+      Quantity.from(tx.gasPrice).toBuffer()
```
davidmurdoch, line 61: The `tx` _might_ have an `effectiveGasPrice` field (depending on the forked network's rules and behavior), should we use that instead of `gasPrice`, if it is available?

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -146,6 +145,19 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
     if (transactorNoncePromise) {
       await transactorNoncePromise;
     }
+    // if the user called sendTransaction or signTransaction, effectiveGasPrice
```
davidmurdoch: `eth_signTransaction` won't enter the tx pool, will it?

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -146,6 +145,19 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
...
+    // value is updated in the miner as we're more sure of what block the tx will
+    // actually go on, but we still need to set it here for to check for valid
+    // transaction replacements of same origin/nonce transactions
+    if (
+      !transaction.effectiveGasPrice &&
+      this.#blockchain.common.isActivatedEIP(1559)
+    ) {
+      const baseFeePerGas = this.#blockchain.blocks.latest.header.baseFeePerGas;
+      transaction.updateEffectiveGasPrice(baseFeePerGas);
+    }
```
davidmurdoch, line 160: The tx pool does seem like a better place for this!

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -146,6 +145,19 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
     if (transactorNoncePromise) {
       await transactorNoncePromise;
     }
+    // if the user called sendTransaction or signTransaction, effectiveGasPrice
+    // hasn't been set yet on the tx. calculating the effectiveGasPrice requires
+    // the block context, so we need to set it outside of the transaction. this
+    // value is updated in the miner as we're more sure of what block the tx will
+    // actually go on, but we still need to set it here for to check for valid
```
davidmurdoch:
```suggestion
    // actually go on, but we still need to set it here to check for valid
```

src/chains/ethereum/ethereum/tests/api/eth/contracts/Example.sol
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  function setValue(uint val) public {
+    value = val;
+    emit ValueSet(val);
+  }
+
+  function destruct() public {
+    selfdestruct(msg.sender);
+  }
```
davidmurdoch: we will likely just forget these other methods are here "for use later", and end up adding them to another test contract once we port things over. So I'm in favor of removing what we don't use in these tests.

We can even rename this contract (and file) to something like `EthCall`, and try to stick to using this one for additional `eth_call`-related tests in the future. Thoughts?

```suggestion
```

davidmurdoch (changes requested):
Really great work here! Just a few comments/suggestions!

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid:
```suggestion
    // if the user called sendTransaction or sendRawTransaction, effectiveGasPrice
```

src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
MicaiahReid: Since the data from the fallback comes from `eth_getTransactionByHash`, I think we can be confident that it will be `gasPrice` rather than `effectiveGasPrice`. Both geth and our internal functions for `eth_getTransactionByHash` return just `gasPrice` (though it technically represents `effectiveGasPrice`.

davidmurdoch: Cool. Happy to be wrong!

davidmurdoch (approved):

PR #1339, opened 2021-10-08
perf: add a disk cache for forking requests

### LINKED ISSUE #1225: Implement persistent cache for forking requests

davidmurdoch:
closes #1225

This PR also implements RPC response streaming capabilities, specifically for large `debug_traceTransaction` results.


src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -128,7 +129,7 @@ export class RuntimeBlock {
       timestamp: new BnExtra(ts),
       baseFeePerGas:
         baseFeePerGas === undefined
-          ? undefined
+          ? new BnExtra(BUFFER_ZERO)
```
davidmurdoch, line 132: The VM always needs this set, so I'm setting it to 0 here.

src/chains/ethereum/ethereum/package.json
```diff
@@ -30,7 +30,7 @@
     "docs.preview": "ws --open --port 3010 --directory ../../../../docs",
     "tsc": "ttsc --build",
     "test": "nyc --reporter lcov npm run mocha",
-    "mocha": "cross-env TS_NODE_COMPILER=ttypescript TS_NODE_FILES=true mocha -s 0 -t 0 --exit --check-leaks --throw-deprecation --trace-warnings --require ts-node/register 'tests/**/*.test.ts'"
+    "mocha": "cross-env TS_NODE_COMPILER=ttypescript TS_NODE_FILES=true mocha -s 0 --exit --check-leaks --throw-deprecation --trace-warnings --require ts-node/register 'tests/**/*.test.ts'"
```
davidmurdoch, line 33: Removing the unlimited timeout because it shouldn't have been in there in the first place

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -673,7 +673,7 @@ export default class Blockchain extends Emittery.Typed<
 
       // commit accounts, but for forking.
       const stateManager = <DefaultStateManager>this.vm.stateManager;
-      stateManager.checkpoint();
+      await stateManager.checkpoint();
```
davidmurdoch, line 675: This await wasn't "technically" needed here, `checkpoint` is Typed as async, so i'm await it just to future proof.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -798,8 +798,16 @@ export default class Blockchain extends Emittery.Typed<
     return (this.#timeAdjustment = timestamp - Date.now());
   }
 
-  #deleteBlockData = (blocksToDelete: Block[]) => {
-    return this.#database.batch(() => {
+  #deleteBlockData = async (blocksToDelete: Block[]) => {
+    // if we are forking we need to make sure we clean up the forking related
+    // metadata that isn't stored in the trie
+    if ("revertMetaData" in this.trie) {
```
davidmurdoch, line 803: This check is a bit weird. The trie class forking uses has the method `revertMetaData` which deletes metadata that was previously committed to disk, which we need to do here!

Note: we don't delete "regular" trie changes during a revert because those are part of the MPT, and must be referenced by their state root. In other words: the data can't be read unless we somehow set the state root to one of these (now deleted) blocks' state roots. Since this can't happen we just let the data live on. We really _should_ clean it up, at least in those cases where users are persisting their database -- as this data does take up disk space.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1173,7 +1182,7 @@ export default class Blockchain extends Emittery.Typed<
       }
 
       const structLog: StructLog = {
-        depth: event.depth,
+        depth: event.depth + 1,
```
davidmurdoch, line 1184: I discovered that the top `depth` should be `1`, not `0` (which is what the VM event returns).

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1309,7 +1318,12 @@ export default class Blockchain extends Emittery.Typed<
     // The previous implementation had specific error handling.
     // It's possible we've removed handling specific cases in this implementation.
     // e.g., the previous incantation of RuntimeError
-    await runTransactions(vm, newBlock.transactions, newBlock);
+    await vm.stateManager.checkpoint();
```
davidmurdoch, line 1273: We need to `checkpoint` (and `revert`) to ensure changes made vs `debug_traceTransaction` aren't ever written to the canonical database trie.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1391,8 +1405,8 @@ export default class Blockchain extends Emittery.Typed<
       throw new Error("Unknown transaction " + transactionHash);
     }
 
-    const targetBlock = await this.blocks.get(
-      transaction.blockNumber.toBuffer()
+    const targetBlock = await this.blocks.getByHash(
```
davidmurdoch, line 1395: Getting by `blockHash` is more accurate, as it is possible for there to have existed two blocks with the same number.

src/chains/ethereum/ethereum/src/helpers/run-transactions.ts
```diff
@@ -17,13 +17,9 @@ export async function runTransactions(
...
-        tx: transactions[i] as any,
-        block: block as any
-      })
-      // we ignore transactions that error because we just want to _run_ these,
-      // transactions just to update the blockchain's state
-      .catch(() => {});
+    await vm.runTx({
+      tx: transactions[i] as any,
+      block: block as any
+    });
```
davidmurdoch: The VM should NEVER throw here, since these should all be valid transactions. If it does throw it means ganache is likely broken.

src/chains/ethereum/ethereum/src/connector.ts
```diff
@@ -45,6 +46,8 @@ export class Connector<
   implements IConnector<EthereumApi, R | R[], JsonRpcResponse> {
   #provider: EthereumProvider;
 
+  static BUFFERIFY_THRESHOLD: number = 100000;
```
davidmurdoch, line 49: I just picked a number that seemed big. :-/

src/chains/ethereum/ethereum/src/forking/fork.ts
```diff
@@ -105,11 +104,24 @@ export class Fork {
   #setBlockDataFromChainAndOptions = async () => {
     const options = this.#options;
     if (options.blockNumber === Tag.LATEST) {
-      // if our block number option is "latest" override it with the original
-      // chain's current blockNumber
-      const block = await fetchBlock(this, Tag.LATEST);
+      const latestBlock = await fetchBlock(this, Tag.LATEST);
+      let blockNumber = parseInt(latestBlock.number, 16);
+      const currentTime = BigInt((Date.now() / 1000) | 0); // current time in seconds
```
davidmurdoch: This logic is actually broken because:

1) a user's system clock can't be relied on.
2) block timestamps are permitted to be up to 900 seconds after the previous block (so basically 900 seconds in the future)

We could use a public NTP server to to sync the time, which isn't hard... but it now requires an Internet connection :-(

We should probably drop this in favor of a `"latest" - n` solution, where `n` is maybe `5 ` or so (configurable via options, of course).

davidmurdoch (commented):
partial self review

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -178,49 +178,6 @@ describe("provider", () => {
           await provider.send("debug_traceTransaction", [hash]);
         }, controlEvents);
       });
-      it("emits vm:tx:* events for debug_storageRangeAt", async () => {
```
davidmurdoch, line 181: removed as this doesn't send these events any more, since the transaction passed in isn't actually run (and it shouldn't have been run either!)

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -17,68 +16,48 @@ import * as lexico from "./lexicographic-key-codec";
...
-  } else {
-    // dump everything into the current (higher level) cache
-    const currentKeyValueMap = this.checkpoints[this.checkpoints.length - 1]
-      .keyValueMap;
-    keyValueMap.forEach((value, key) => currentKeyValueMap.set(key, value));
-  }
+
+function isEqualKey(encodedKey: Buffer, address: Buffer, key: Buffer) {
+  const decodedKey = lexico.decode(encodedKey);
+  const [, keyAddress, deletedKey] = decodedKey;
```
MicaiahReid:
```suggestion
  const [_, keyAddress, deletedKey] = decodedKey;
```

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -245,12 +287,16 @@ export class ForkTrie extends GanacheTrie {
    * Returns a copy of the underlying trie with the interface of ForkTrie.
    * @param includeCheckpoints - If true and during a checkpoint, the copy will contain the checkpointing metadata and will use the same scratch as underlying db.
```
MicaiahReid:
```suggestion
   * @param includeCheckpoints - If true and during a checkpoint, the copy will 
   * contain the checkpointing metadata and will use the same scratch as
   * underlying db.
```

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -90,56 +69,116 @@ export class ForkTrie extends GanacheTrie {
...
   async del(key: Buffer) {
     await this.lock.wait();
 
-    const hash = keccak(key);
-    const delKey = this.createDelKey(key);
-    const metaDataPutPromise = this.metadata.put(delKey, BUFFER_ZERO);
+    // we only track if the key was deleted (locally) for state tries _after_
+    // the fork block because we can't possibly delete keys _before_ the fork
+    // block, since those happened before ganache was even started
+    // This little optimization can debug_traceTransaction time _in half_.
```
MicaiahReid:
```suggestion
    // This little optimization can cut debug_traceTransaction time _in half_.
```

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -90,56 +69,116 @@ export class ForkTrie extends GanacheTrie {
...
+    if (!this.preForkBlock) {
```
MicaiahReid: Rename to `isPreForkBlock`!

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -225,12 +264,15 @@ export class ForkTrie extends GanacheTrie {
...
-      return null;
-    }
+    if (value != null) return value;
+
+    // since we don't have this key in our local trie check if we've have
+    // deleted it (locally)
+    // we only check if the key was deleted (locally) for state tries _after_
+    // the fork block because we can't possibly delete keys _before_ the fork
+    // block, since those happened before ganache was even started
+    // This little optimization can debug_traceTransaction time _in half_.
```
MicaiahReid:
```suggestion
    // This little optimization can cut debug_traceTransaction time _in half_.
```

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -245,12 +287,27 @@ export class ForkTrie extends GanacheTrie {
...
+      // Our `metadata.checkpoints` needs to be the same reference to the
+      // parent's metadata.checkpoints so that we can continue to track these
+      // changes on this copy, otherwise deletions made to a contract's storage
+      // may not be tracked.
+      // Note: db.checkpoints don't need this same treatment because of the way
+      // the statemanager uses a contract's trie: it doesn't ever save to it.
+      // Instead, it saves to its own internal cache, which eventually gets
+      // reverted or committed (flushed). Our metadata doesn't utilize a central
+      // cache.
+      // secureTrie.metadata.checkpoints = this.metadata.checkpoints;
```
MicaiahReid:
```suggestion
```

src/chains/ethereum/options/src/fork-options.ts
```diff
@@ -365,5 +404,17 @@ Defaults to: \`["User-Agent: Ganache/VERSION (https://www.trufflesuite.com/ganac
       "Restrict the number of requests per second sent to the fork provider. `0` means no limit is applied.",
     cliType: "number"
     //implies: ["url"]
+  },
+  noCache: {
+    normalize,
+    default: () => false,
+    cliDescription: "Don't cache forking requests in a persistent db.",
```
MicaiahReid:
```suggestion
    cliDescription: "Disables caching of all forking requests.",
```

src/chains/ethereum/options/src/fork-options.ts
```diff
@@ -365,5 +404,17 @@ Defaults to: \`["User-Agent: Ganache/VERSION (https://www.trufflesuite.com/ganac
       "Restrict the number of requests per second sent to the fork provider. `0` means no limit is applied.",
     cliType: "number"
     //implies: ["url"]
+  },
+  noCache: {
```
MicaiahReid: Rename to `disableCache`

src/chains/ethereum/options/src/fork-options.ts
```diff
@@ -160,6 +172,26 @@ export type ForkConfig = {
       type: number;
       hasDefault: true;
     };
+
+    /**
+     * Don't cache forking requests in a persistent db.
```
MicaiahReid: Disables caching of all forking requests.
```suggestion
     * Disables caching of all forking requests.
```

src/chains/ethereum/options/src/fork-options.ts
```diff
@@ -71,6 +72,17 @@ export type ForkConfig = {
       };
     };
 
+    /**
+     * Minimum age in seconds of the "latest" block. If the "latest" block is
+     * younger than this amount the block immediately preceding the latest block
+     * will be used instead.
```
MicaiahReid:
```suggestion
     * When the `fork.blockNumber` is set to "latest" (default), the number 
     * of blocks before the remote node's "latest" block to fork from.
```

src/chains/ethereum/options/src/fork-options.ts
```diff
@@ -71,6 +72,17 @@ export type ForkConfig = {
...
+     */
+    blockAge: {
```
MicaiahReid: Rename to `preLatestConfirmations`

src/chains/ethereum/options/src/fork-options.ts
```diff
@@ -270,9 +302,16 @@ Alternatively, you can use the \`fork.username\` and \`fork.password\` options.`
         return;
       }
     },
-    defaultDescription: `"${Tag.LATEST}"`
+    defaultDescription: `Latest block number`
     //implies: ["url"]
   },
+  blockAge: {
+    normalize: rawInput => BigInt(rawInput),
+    cliDescription: `Minimum age in seconds of the "latest" block. If the "latest" block is younger than this amount the block immediately preceding the latest block will be used instead.`,
```
MicaiahReid:
```suggestion
    cliDescription: `When the `fork.blockNumber` is set to "latest" (default), the number of blocks before the remote node's "latest" block to fork from.`,
```

src/chains/ethereum/ethereum/src/forking/fork.ts
davidmurdoch: After talking it over with @MicaiahReid, we've decided that we'll switch to the `"latest" - n` idea, but only fork "known" chain ids (chain id's that ethereumjs-common knows about). For all other chain id's we will just use the `"latest"` number itself.

src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
```diff
@@ -48,10 +48,15 @@ export default class TransactionManager extends Manager<NoOp> {
       [Data.from(transactionHash).toString()]
     );
     if (tx == null) return null;
+
     const blockHash = Data.from((tx as any).blockHash, 32);
     const blockNumber = Quantity.from((tx as any).blockNumber);
     const index = Quantity.from((tx as any).transactionIndex);
 
+    // don't get the transaction if the requested transaction is _after_ our
```
MicaiahReid, line 56: Is it possible for this to return false? If we've found the transaction from the fallback, we must have the block, right?

src/chains/ethereum/ethereum/src/forking/persistent-cache/helpers.ts
```diff
@@ -0,0 +1,220 @@
...
+    case "eth_getBalance":
+      return params[1];
+    case "eth_getStorageAt":
+      return params[2];
+    default:
+      return null;
+  }
+}
+
+export async function setDbVersion(db: LevelUp, version: Buffer) {
```
MicaiahReid, line 59: This function makes a total of zero (0) sense.

src/chains/ethereum/ethereum/src/forking/handlers/base-handler.ts
```diff
@@ -135,41 +141,96 @@ export class BaseHandler {
...
-
-    const promise = this.limiter.handle(send).then(({ response, raw }) => {
-      if (this.abortSignal.aborted) return Promise.reject(new AbortError());
+    if (!options.noCache) {
+      const memCached = this.getFromMemCache<T>(key);
+      if (memCached !== undefined) return memCached;
+
+      const diskCached = await this.getFromSlowCache<T>(method, params, key);
+      if (diskCached !== undefined) {
+        this.valueCache.set(key, Buffer.from(diskCached.raw));
```
MicaiahReid, line 181: I think this line has a chance of throwing. `this.getFromSlowCache` can return `null` in certain cases. So I think line 180 should be `if(diskCached != undefined) {`

MicaiahReid (changes requested):

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -51,7 +51,8 @@ describe("api", () => {
               block.totalDifficulty,
               `0x${((numberOfBlocksToMine + 1) * DEFAULT_DIFFICULTY).toString(
                 16
-              )}`
+              )}`,
+              `Didn't correctly mine ${numberOfBlocksToMine} blocks`
```
MicaiahReid: Don't love this comment since it doesn't line up with the point of this `it` or really have to do with the values that were checked. Maybe something like `Mined total difficulty, ${block.totalDifficulty} differs from sum of preceding block's difficulties.`

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -81,9 +82,14 @@ describe("api", () => {
...
             ]);
+            assert(
+              block,
+              `\`block\` is \`null\`; didn't correctly mine ${numberOfBlocksToMine} blocks`
+            );
             assert.strictEqual(
               block.totalDifficulty,
-              `0x${((numberOfBlocksToMine + 1) * difficulty).toString(16)}`
+              `0x${((numberOfBlocksToMine + 1) * difficulty).toString(16)}`,
+              `Didn't correctly mine ${numberOfBlocksToMine} blocks`
```
MicaiahReid: Same as above.

src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
davidmurdoch: When not forking `null` is what is returned if we didn't find the Transaction in the database, so it should be the same when we are forking as well.

src/chains/ethereum/ethereum/src/forking/handlers/base-handler.ts
davidmurdoch: Hm. I can't find where `getFromSlowCache` can return `null`

src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
davidmurdoch: It is possible if the user asks for a transaction hash that exists on the remote chain, but it occured _after_ our fork block.

src/chains/ethereum/options/src/fork-options.ts
```diff
@@ -71,6 +72,16 @@ export type ForkConfig = {
       };
     };
 
+    /**
+     * Minimum age in seconds of the "latest" block. If the "latest" block is
```
MicaiahReid: Looks like some of the suggested commits didn't work? For this comment and for the comment on `disableCache`

src/chains/ethereum/ethereum/src/forking/persistent-cache/ancestry.ts
```diff
@@ -0,0 +1,61 @@
...
+        ? null
+        : parent.closestKnownAncestor;
+      this.knownAncestors = new Set([parent.hash.toString("hex")]);
+    }
+  }
+
+  private async loadNextAncestor(next: Buffer) {
+    const k = next.toString("hex");
+    if (this.lock.has(k)) {
+      throw new Error("could not obtain lock");
```
MicaiahReid: I feel like this error could be a bit more descriptive.

src/chains/ethereum/ethereum/src/forking/persistent-cache/tree.ts
```diff
@@ -0,0 +1,64 @@
...
+  public key: Buffer;
+  public hash: Buffer;
+  public closestKnownAncestor: Buffer;
+  public closestKnownDescendants: Buffer[];
+
+  constructor(
+    height: Quantity,
+    hash: Data,
+    closestKnownAncestor: Buffer = BUFFER_EMPTY,
+    closestKnownDescendants: Buffer[] = []
```
MicaiahReid: As far as I've seen, this fourth option in the constructor isn't ever used. Can we remove and give it a default value?

MicaiahReid: Also, is it really the closest known descendants? Isn't it all known descendants? Consider rename to `orderedKnownDescendants`.

MicaiahReid (changes requested):

MicaiahReid (approved):

PR #1414, opened 2021-10-19
use new @trufflesuite/bigint-buffer

### LINKED ISSUE #1080: bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
```
C:\Projects>ganache
bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
ganache v7.0.0-alpha.0 (@ganache/cli: 0.1.1-alpha.0, @ganache/core: 0.1.1-alpha.0)
Starting RPC server
```

How to fix this `bigint` error?

MicaiahReid:
Previously the bigint-buffer library used by Ganache would fail to build on Windows unless `windows-build-tools` was installed. This would cause the (slower) JS fallback to be used and would display an ugly warning in the console on startup of Ganache. This PR's main objective is to use trufflesuite's new [bigint-buffer](https://github.com/trufflesuite/bigint-buffer) library, which prebuilds all binaries during CI to prevent this warning.

Additionally, this PR:
- removes unnecessary CI installations when Github runs the PR action
- Adds very basic test to check if fallback for bigint-buffer is used.
  - Note: the original bigint-buffer library does NOT fail to build and use the JS fallback in our CI tests. This is because the windows version used by Github actions has the necessary dependencies (python) installed by default. So while the test should accurately tell us if the fallback is used, it shouldn't ever happen in CI so it's not particularly useful.

Closes #1080

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -61,7 +61,7 @@ jobs:
       #   if: startsWith(matrix.os, 'windows-')
       #   run: npm config set msvs_version 2019 --global
 
-      - run: npm ci
+      - run: npm install
       - run: npm run tsc
       - run: npm test
         env:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,7 +15,7 @@ jobs:
         # monorepo we want to use 10.7.0 (earliest version the Ethereum side
         # supports) or 10.13.0 (first node 10 LTS version).
         node: [10.18.0, 10.x, 11.x, 12.x, 13.x, 14.x, 15.x, 16.x]
-        os: [windows-latest, ubuntu-18.04, ubuntu-20.04, macos-11.0]
+        os: [windows-2019, ubuntu-18.04, ubuntu-20.04, macos-11.0]
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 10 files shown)


.github/workflows/pr.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         # monorepo we want to use 10.7.0 (earliest version the Ethereum side
         # supports) or 10.13.0 (first node 10 LTS version).
         node: [10.18.0, 10.x, 11.x, 12.x, 13.x, 14.x, 15.x, 16.x]
-        os: [windows-2019, ubuntu-18.04, ubuntu-20.04, macos-11.0]
+        os: [windows-latest, ubuntu-18.04, ubuntu-20.04, macos-11.0]
```
davidmurdoch, line 18: Why the change to `windows-latest`? Using the `latest` build makes for potentionally unreproducable CI runs, as github can change what it targets at anytime.

src/packages/utils/tests/utils.test.ts
```diff
@@ -1,7 +1,28 @@
-"use strict";
+import assert from "assert";
+import sinon from "sinon";
+const BIGINT_ERROR =
+  "bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)";
 
-const utils = require("..");
+describe.only("@ganache/utils", () => {
```
davidmurdoch:
```suggestion
describe("@ganache/utils", () => {
```

davidmurdoch (changes requested):

.github/workflows/pr.yml
MicaiahReid:
```suggestion
        os: [windows-2019, ubuntu-18.04, ubuntu-20.04, macos-11.0]
```

davidmurdoch (approved):

Issue #996, opened 2019-10-27
`ganache-cli` doesn't really lock accounts


NicsTr:
## Expected Behavior

When using `--secure`, `ganache-cli` should not allow transactions from locked accounts.

## Current Behavior

Transactions are done as if the accounts were not locked.

## Steps to Reproduce (for bugs)

1. `npx ganache-cli --secure --deterministic -p 8000 -a 11`
2. After right configuration of web3js' provider to ganache : `web3.eth.personal.sendTransaction({from: "0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0", to: "0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1", value: web3.utils.toWei("1") }, '')`

## Context

Unless I wrongly understood how locked accounts should work, this seems like a pretty serious bug.
Please not that I figured a not so elegant workaround by using `--db` option and by relaunching `ganache-cli` with a smaller `-a` argument, causing it to initialize the 11 eleven accounts as expected but forbidding every transactions from an address that is not present in the second execution log.

## Your Environment
* Version used: v6.7.0
* Version of Truffle/Remix/Other tools used: ganache-core 2.8.0, web3js 1.2.2
* NodeJS Version: 10.16.3

davidmurdoch:
For ganache v7, let's rename the `--secure` flag to `--lock` and `--wallet.lock` (while keeping the `--secure` flag for legacy reasons)

We'll also add a `--passphrase` flag. The `--passphrase {string}`  flag by itself (sans `--lock`/`--secure`) will _not_ automatically lock accounts and will not be an error. This enables using `personal_lockAccount(address)` to lock the account (with the user-specified default passphrase) after start up.

Using `--lock`/`--secure` without the `--passphrase` flag will continue using `""` (empty string) as the passphrase.

Implementation detail:

We store account private keys that were created by `personal_` methods in their encrypted form, but really don't need to do this for accounts created with the default "" passphrase, so let's take advantage of this and optimize for better start up time by storing the private key directly (unencrypted).

PR #1513, opened 2021-11-06
fix: improve account locking

### LINKED ISSUE #1490: `evm_(un)lockUnknownAccount` should be named better and work alongside the `personal_` namespace
`evm_unlockUnknownAccount` currently unlocks accounts for use with `eth_sendTransaction`, but doesn't make the account available for us with `personal_sendTransaction`. It probably should.

Also, the name itself is confusing, what is "unknown" here even mean?

Maybe it should be `evm_addAccount(address, passphrase)` and `evm_removeAccount(address, passphrase)`, and then the user could use the `personal_` namespace to manage the locked status?

/cc @MicaiahReid

MicaiahReid:
Closes #1490 and #996 

- Renames `--secure` to `--lock`, keeping `--secure` as legacy alias.
 - Adds `wallet.passphrase` option.
   - Passphrase will be used to encrypt all accounts on startup and when locking accounts.
 - Adds `evm_addAccount` 
   - This adds an unknown account as a known account, giving a fake private key to the account
 - Adds `evm_removeAccount`
   - Thie removes a known account from the personal namespace
 - Removes `evm_unlockUnknownAccount` and `evm_lockUnknownAccount`
 - Skips account encryption on startup if `--lock` and `--passphrase` are not specified.


src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2892,10 +2894,7 @@ export default class EthereumApi implements Api {
     const fromString = tx.from.toString();
 
     const wallet = this.#wallet;
-    const encryptedKeyFile = wallet.encryptedKeyFiles.get(fromString);
-    if (encryptedKeyFile === undefined) {
-      throw new Error("no key for given address or file");
-    }
+    const secretKey = await wallet.getFromKeyFile(fromString, passphrase);
```
MicaiahReid: Previously, we would try to get the `fromString` out of `encryptedKeyFiles`. If the result was `undefined`, we'd throw, but if the result was `null`, we'd queue the transaction without the secret key.

I couldn't find a code path that would lead to the result of `encryptedKeyFiles.get(fromString)` being `null`, only undefined. So, the `getFromKeyFile` function throws for both `null` and `undefined`.

If we can find a case where `null` is a possible/valid result, we might need to back out this change.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -508,52 +508,51 @@ export default class EthereumApi implements Api {
...
-   * account was already unlocked. Throws an error if the account could not be
-   * unlocked.
+   * Note: accounts already known to the `personal` namespace and accounts
+   * returned by `eth_accounts` cannot be re-added using this method.
+   * @param address The address of the account to add to the `personal`
+   * namespace.
+   * @param passphrase The passphrase used to encrypt the account's private key.
+   * NOTE: this passphrase will be needed for all `personal` namespace calls
+   * that require a password.
+   * @returns The added account's private key if the account was successfully
```
davidmurdoch: Should we provide the fake key back to them? I don't think they can do anything with it as they can't sign a raw transaction with it (since signing removes the `from` and relies on the node to derive the from from the signature... which is not accurate).

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -508,52 +508,51 @@ export default class EthereumApi implements Api {
...
-   * @param address The address of the account to lock.
-   * @returns `true` if the account was locked successfully, `false` if the
-   * account was already locked. Throws an error if the account could not be
-   * locked.
+   * Note: accounts not known to the `personal` namespace cannot be removed
+   * using this method.
+   * @param address The address of the account to remove from the `personal`
+   * namespace.
+   * @param passphrase The passphrase used to decrypt the account's private key.
+   * @returns `address` if the account was successfully removed. Throws an error
```
davidmurdoch: These methods should probably return `boolean` values. Otherwise callers will have to `try`/`catch` calls to these methods. Feel free to argue with me on this :-)

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -300,23 +301,9 @@ export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
...
-        // see: https://github.com/ethereumjs/ethereumjs-monorepo/issues/829#issue-674385636
-        fakePrivateKey = Buffer.allocUnsafe(32).fill(0, 0, 31);
-        fakePrivateKey[31] = 1;
-      } else {
-        fakePrivateKey = Buffer.concat([from, from.slice(0, 12)]);
-      }
-      transaction.signAndHash(fakePrivateKey);
+      // if we don't have a secret key, the transaction should be signed
+      // already.
+      throw new CodedError(NO_SECRET_KEY, JsonRpcErrorCode.INVALID_INPUT);
```
davidmurdoch: Is it possible to get in this function without a `secretKey`? The function signature for `prepareTransaction` should probably change from `secretKey?: Data` to `secretKey: Data`, as should all the other functions that lead to `prepareTransaction`.

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -239,6 +248,8 @@ export default class Wallet {
       WEI * significand + fractional * (WEI / magnitude);
     const etherInWei = Quantity.from(defaultBalanceInWei);
     let accounts: Account[];
+    const passphrase = options.passphrase;
+    const lockAccounts = options.lock;
```
```suggestion
    const { passphrase, lockAccounts } = options;
```

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -168,8 +176,9 @@ export default class Wallet {
             throw new Error(`Invalid value specified in unlocked_accounts`);
         }
         if (unlockedAccounts.has(address)) continue;
-        // if we don't have the secretKey for an account we use `null`
-        unlockedAccounts.set(address, null);
+        // if we don't have the secretKey for an account we can make a fake one.
+        const privateKey = this.createFakePrivateKey(address);
+        unlockedAccounts.set(address, privateKey);
```
davidmurdoch, line 190: Do these accounts need to be added to `this.keyFiles`?

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -373,6 +424,96 @@ export default class Wallet {
...
+   * or the user specified --lock option) or unencrypted private keys.
+   * @param address The address whose private key is being stored.
+   * @param privateKey The passphrase to store.
+   * @param passphrase The passphrase to use to encrypt the private key. If
+   * passphrase is empty, the private key will not be encrypted.
+   * @param lock Flag to specify that accounts should be encrypted regardless
+   * of if the passphrase is empty.
+   */
+  public addToKeyFileSync(
+    address: string,
```
davidmurdoch: Use `Address` here instead of `string`, and then `toString()` will be all lowercase and you won't need the `toLowerCase()` calls (same for the `addToKeyFile` method)

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -373,6 +424,96 @@ export default class Wallet {
...
+    }
+  }
+
+  /**
+   * Fetches the private key for a specific address. If the keyFile is encrypted
+   * for the address, the passphrase is used to decrypt.
+   * @param address The address whose private key is to be fetched.
+   * @param passphrase The passphrase used to decrypt the private key.
+   */
+  public async getFromKeyFile(address: string, passphrase: string) {
```
davidmurdoch: You might want to change `address: string` to `address: Address` here, too, to avoid the `toLowerCase()` requirement.

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -373,6 +424,96 @@ export default class Wallet {
...
+    const keyFile = this.keyFiles.get(address.toLowerCase());
+    if (keyFile === undefined || keyFile === null) {
```
davidmurdoch, line 515: @cds-amal would be proud! :laughing:

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -373,6 +424,96 @@ export default class Wallet {
...
+    }
+    if (keyFile.encrypted === true) {
+      return this.decrypt(keyFile.key, passphrase);
+    } else {
+      // if the keyFile is not marked as encrypted, they should provide no
+      // passphrase. so we'll make it look like they gave the "wrong" passphrase
+      // by throwing the same error that's thrown when decrypting
+      if (passphrase) {
+        throw new Error(
+          'could not decrypt key with given password (default password for accounts created at startup is "")'
```
davidmurdoch: :+1:  I like the additional information here!

Perhaps we should only put the parenthetical in the when `options.passphrase` is `=== ""`?

Oh, I think we should use "passphrase" instead of "password".

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -429,28 +566,58 @@ export default class Wallet {
...
+      throw new Error("cannot remove unknown account");
     }
 
-    // a duration <= 0 will remain unlocked
-    const durationMs = (duration * 1000) | 0;
-    if (durationMs > 0) {
-      const timeout = setTimeout(this.#lockAccount, durationMs, lowerAddress);
-      unref(timeout);
-      this.lockTimers.set(lowerAddress, timeout as any);
+    const privatKey = await this.getFromKeyFile(lowerAddress, passphrase);
```
davidmurdoch: `privatKey` -> `privateKey` :-)

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -274,15 +284,22 @@ export const WalletOptions: Definitions<WalletConfig> = {
...
-  secure: {
+  lock: {
     normalize,
     cliDescription:
       "Lock available accounts by default (good for third party transaction signing).",
     default: () => false,
-    legacyName: "secure",
-    cliAliases: ["n", "secure"],
+    legacyName: "lock",
+    cliAliases: ["n", "secure", "lock", "lock"],
```
davidmurdoch:
```suggestion
    cliAliases: ["n", "secure", "lock"],
```

src/chains/ethereum/ethereum/tests/transaction-pool.test.ts
```diff
@@ -437,4 +433,18 @@ describe("transaction pool", async () => {
       transaction.serialized.toString()
     );
   });
+
+  it("rejects unsigned transactions with no secret key pprovided", async () => {
```
davidmurdoch:
```suggestion
  it("rejects unsigned transactions with no secret key provided", async () => {
```

src/chains/ethereum/ethereum/tests/api/personal/personal.test.ts
```diff
@@ -520,5 +519,35 @@ describe("api", () => {
         });
       });
     });
+
+    describe("startup passphrase", () => {
```
davidmurdoch, line 523: :+1: !

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1686,12 +1685,8 @@ export default class EthereumApi implements Api {
       this.#options
     );
 
-    if (isUnlockedAccount) {
-      const secretKey = wallet.unlockedAccounts.get(fromString);
-      return blockchain.queueTransaction(tx, secretKey);
-    } else {
-      return blockchain.queueTransaction(tx);
-    }
+    const secretKey = wallet.unlockedAccounts.get(fromString);
```
davidmurdoch: Can we change the lines above: 
```
const isUnlockedAccount = wallet.unlockedAccounts.has(fromString);

if (!isUnlockedAccount) {
```
to 

```
const secretKey = wallet.unlockedAccounts.get(fromString);

if ( secretKey === undefined ) {
```

?

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -183,11 +192,11 @@ export default class Wallet {
       accountsCache[i] = strAddress;
       knownAccounts.add(strAddress);
 
-      // if the `secure` option has been set do NOT add these accounts to the
+      // if the `lock` option has been set do NOT add these accounts to the
       // unlockedAccounts, unless the account was already added to
       // unlockedAccounts, in which case we need to add the account's private
       // key.
-      if (opts.secure && !unlockedAccounts.has(strAddress)) continue;
+      if (opts.lock && !unlockedAccounts.has(strAddress)) continue;
```
davidmurdoch: Do these accounts need to be added to `this.keyFiles`?

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/wallet.ts
cds-amal: 🥲

src/chains/ethereum/options/src/wallet-options.ts
gnidan: you sure you don't mean
```suggestion
    cliAliases: ["n", "secure", "lock", "lock", "lock"],
```

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: I did this originally, but it broke some tests. This is because `eth_sendRawTransaction` does not have a secret key, but rather already has a signature. So there are other valid paths where a secret key isn't necessary. However, if there's no secret key and the transaction hasn't been signed, it's now invalid and we give an error. I don't know if this code path is actually possible, though.

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -248,8 +248,7 @@ export default class Wallet {
       WEI * significand + fractional * (WEI / magnitude);
     const etherInWei = Quantity.from(defaultBalanceInWei);
     let accounts: Account[];
-    const passphrase = options.passphrase;
-    const lockAccounts = options.lock;
+    const { passphrase, lockAccounts } = options;
```
davidmurdoch: opps! sorry!

```
const { passphrase, lock: lockAccounts } = options;
```

src/chains/ethereum/ethereum/src/api.ts
cds-amal: 🥲

src/chains/ethereum/ethereum/src/wallet.ts
MicaiahReid: There are times when `addToKeyFile` is called from the api where we only receive a `string` from the user.

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: Upon further discussion, we realized that if a raw transaction is sent and makes it this far into the pool, the transaction has already been constructed and verification of a proper signature (i.e. can be used to recover from address) was sent. So we should be safe to remove this check.

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -411,7 +413,7 @@ export default class Wallet {
     }
 
     if (!localMac || !mac.toBuffer().equals(localMac)) {
-      throw new Error("could not decrypt key with given password");
+      throw new Error("could not decrypt key with given passphrase");
```
cds-amal, line 432: Just an observation, this error message seems similar to another one later in the file which appends `(default passphrase for accounts created at startup is "")`.

MicaiahReid: Yes, this is intentional! The one stating `(default passphrase for accounts created at startup is "")` is just extra specific because we know in that case that they didn't provide a passphrase at startup, so they need to send `""` to decrypt.

src/chains/ethereum/ethereum/src/wallet.ts
MicaiahReid: Done as suggested!

src/chains/ethereum/ethereum/src/transaction-pool.ts
davidmurdoch: :+1:

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2830,15 +2820,12 @@ export default class EthereumApi implements Api {
    */
   @assertArgLength(2, 3)
   async personal_unlockAccount(
-    address: DATA,
+    strAddress: DATA,
```
davidmurdoch: The name will be displayed in the interactive docs, so I think it should remain as `address`.

MicaiahReid: Good find!

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -567,27 +568,29 @@ export default class Wallet {
     return true;
   }
 
-  public async addUnknownAccount(lowerAddress: string, passphrase: string) {
+  public async addUnknownAccount(address: Address, passphrase: string) {
+    const lowerAddress = address.toString();
     // if we "know" about this account, it cannot be added this way
     if (this.knownAccounts.has(lowerAddress)) {
-      throw new Error("cannot add known/personal account");
+      false;
```
davidmurdoch: `false;` - Dwight Schrute 🤣

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2889,13 +2871,10 @@ export default class EthereumApi implements Api {
       throw new Error("from not found; is required");
     }
 
-    const fromString = tx.from.toString();
+    const fromAddress = new Address(tx.from.toString());
```
davidmurdoch: I bet we can change `BaseTransaction`'s `from` type to `Address` and then we don't need to do the `toString()` dance here.

src/chains/ethereum/ethereum/tests/api/evm/evm.test.ts
```diff
@@ -154,97 +155,87 @@ describe("api", () => {
...
-      it("should return `false` upon lock if account isn't locked (unknown account)", async () => {
-        const address = "0x942d35Cc6634C0532925a3b844Bc454e4438f450";
-        const result = await provider.send("evm_lockUnknownAccount", [address]);
+      it("should not remove an account that isn't known to the personal namespace", async () => {
+        const address = "0x742d35Cc6634C0532925a3b844Bc454e4438f44e";
+        const result = await provider.send("evm_removeAccount", [
+          address,
+          passphrase
+        ]);
+        // cannot add account
```
davidmurdoch:
```suggestion
        // cannot remove account
```

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -130,17 +130,27 @@ export type WalletConfig = {
...
+    lock: {
       type: boolean;
       hasDefault: true;
       legacy: {
         /**
-         * @deprecated Use wallet.secure instead
+         * @deprecated Use wallet.lock instead
          */
-        secure: boolean;
+        lock: boolean;
```
davidmurdoch: the legacy name is still `secure`.

```suggestion
        secure: boolean;
```

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -275,15 +285,22 @@ export const WalletOptions: Definitions<WalletConfig> = {
...
   },
-  secure: {
+  lock: {
     normalize,
     cliDescription:
       "Lock available accounts by default (good for third party transaction signing).",
     default: () => false,
-    legacyName: "secure",
-    cliAliases: ["n", "secure"],
+    legacyName: "lock",
```
davidmurdoch:
```suggestion
    legacyName: "secure",
```

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -275,15 +285,22 @@ export const WalletOptions: Definitions<WalletConfig> = {
...
+    cliAliases: ["n", "secure", "lock"],
     cliType: "boolean"
   },
+  passphrase: {
+    normalize,
+    cliDescription: "Passphrase to use when locking accounts.",
+    default: () => "",
+    cliAliases: ["passphrase", "password"],
```
davidmurdoch: I not convinced we need to alias this to `password`.

`cliAliases` create aliases at the top level (`ganache --passphrase`) instead of namespace (`ganache --wallet.passphrase`). Do you think we should add passphrase to the top level?


```suggestion
    cliAliases: ["passphrase"],
```

davidmurdoch (changes requested):
Left some standalone comments and suggestions. 

I wonder if we should revert the changes to `docs`, since the release process creates them automatically. I'm thinking `docs` should represent the released version of ganache only? At least until we get stable out, and then we may want to host and track multiple docs versions at a time (7.0.0, 7.0.1, 7.0.2-alpha.0, 7.0.2-nightly.4, etc). Thoughts?

Also can you revert the changes to the `package-lock.json` and `npm-shrinkwrap.json` files, as I have a PR out that reverts the change that causes them to be wrong in the first place, and I should probably add these lock file changes to that PR.

MicaiahReid:
> Left some standalone comments and suggestions.
> 
> I wonder if we should revert the changes to `docs`, since the release process creates them automatically. I'm thinking `docs` should represent the released version of ganache only? At least until we get stable out, and then we may want to host and track multiple docs versions at a time (7.0.0, 7.0.1, 7.0.2-alpha.0, 7.0.2-nightly.4, etc). Thoughts?
> 
> Also can you revert the changes to the `package-lock.json` and `npm-shrinkwrap.json` files, as I have a PR out that reverts the change that causes them to be wrong in the first place, and I should probably add these lock file changes to that PR.

I've reverted `docs`, `package-lock.json`, and `npm-shrinkwrap.json` changes!

davidmurdoch (approved):
I really appreciate how you are always a gentleman. You are proof that chivalry is not dead.

PR #1537, opened 2021-11-11
fix: make forking chainId aware

### LINKED ISSUE #628: investigate if forking should use the chainId of the original chain
Perhaps it should use the original chainId only for `eth_call`/`debug_traceTransaction` calls before the fork block number, while still using the configured/default id for RPCs after the fork block number.

/cc @haltman-at, @gnidan

davidmurdoch:
fixes #628

This change causes ganache to use its default chainId instead of the fork's chainId. This prevents replay attacks, but breaks `debug_traceTransaction` and `eth_call`. So this PR also makes `debug_traceTransaction` and `eth_call` "aware" of the chainId of the original chain so that they can use the correct chainId in block contexts that require it (on or before the fork block number).

### DIFF
--- a/src/chains/ethereum/ethereum/src/forking/fork.ts
+++ b/src/chains/ethereum/ethereum/src/forking/fork.ts
@@ -91,6 +91,8 @@ export class Fork {
       fetchNetworkId(this)
     ]);
 
+    this.chainId = chainId;
+
     this.common = Common.forCustomChain(
       KNOWN_CHAINIDS.has(chainId) ? chainId : 1,
       {
@@ -237,8 +239,8 @@ export class Fork {
    */
   public getCommonForBlockNumber(common: Common, blockNumber: BN) {
     const bigIntBlockNumber = Quantity.from(blockNumber.toBuffer()).toBigInt();
-    if (bigIntBlockNumber < this.blockNumber.toBigInt()) {
-      // we are before our fork block
+    if (bigIntBlockNumber <= this.blockNumber.toBigInt()) {
+      // we are at or before our fork block
 
       if (KNOWN_CHAINIDS.has(this.chainId)) {
         // we support this chain id, so let's use its rules
--- a/src/chains/ethereum/ethereum/src/forking/fork.ts
+++ b/src/chains/ethereum/ethereum/src/forking/fork.ts
@@ -12,6 +12,7 @@ import { Account } from "@ganache/ethereum-utils";
 import BlockManager from "../data-managers/block-manager";
 import { ProviderHandler } from "./handlers/provider-handler";
 import { PersistentCache } from "./persistent-cache/persistent-cache";
+import { BN } from "ethereumjs-util";
 
 async function fetchChainId(fork: Fork) {
   const chainIdHex = await fork.request<string>("eth_chainId", []);
@@ -41,16 +42,18 @@ export class Fork {
   public common: Common;
   #abortController = new AbortController();
   #handler: Handler;
-  #options: EthereumInternalOptions["fork"];
+  #options: EthereumInternalOptions;
   #accounts: Account[];
   #hardfork: string;
 
   public blockNumber: Quantity;
   public stateRoot: Data;
   public block: Block;
+  public chainId: number;
 
   constructor(options: EthereumInternalOptions, accounts: Account[]) {
-    const forkingOptions = (this.#options = options.fork);
+    this.#options = options;
+    const forkingOptions = options.fork;
     this.#hardfork = options.chain.hardfork;
     this.#accounts = accounts;
 
@@ -94,7 +97,7 @@ export class Fork {
         name: "ganache-fork",
         defaultHardfork: this.#hardfork,
         networkId,
-        chainId,
+        chainId: this.#options.chain.chainId,
         comment: "Local test network fork"
       }
     );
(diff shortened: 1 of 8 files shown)


src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
```diff
@@ -1,23 +1,26 @@
...
       let from, to: string;
       let contractAddress: string;
       let tx: object;
 
+      before("compile", () => {
+        contract = compile(join(__dirname, "./contracts/EthCall.sol"), {
+          contractName: "EthCall"
+        });
+      });
+
```
davidmurdoch, line 23: When debugging fork tests this contract complication keep popping up. It should never have been outside a `before` call anyway.

src/chains/ethereum/ethereum/tests/api/eth/contracts/EthCall.sol
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: MIT
```
davidmurdoch, line 1: silencing a warning from the solc compiler

src/chains/ethereum/ethereum/tests/forking/contracts/Forking.sol
```diff
@@ -32,7 +32,7 @@ contract Forking {
   // initial value is 0, set to 1 in constructor
   uint public value4 = 0;
 
-  constructor() public payable {
+  constructor() payable {
```
davidmurdoch, line 35: The solc compiler was warning me about `public`, so I removed it.

src/chains/ethereum/ethereum/tests/forking/contracts/Forking.sol
```diff
@@ -61,4 +61,10 @@ contract Forking {
   function destruct() public {
     selfdestruct(payable(address(msg.sender)));
   }
+
+  function getChainId() public pure returns (uint256 chainId) {
+    assembly {
+      chainId := chainid()
```
davidmurdoch, line 67: This is the test change you should actually care about. :-)

src/chains/ethereum/ethereum/tests/forking/forking.test.ts
```diff
@@ -15,6 +16,42 @@ import compile from "../helpers/compile";
 import path from "path";
 import { CodedError } from "@ganache/ethereum-utils";
 
+async function deployContract(
```
davidmurdoch, line 19: I just moved this to a more global scope so I could reuse it.

src/chains/ethereum/transaction/src/eip1559-fee-market-transaction.ts
```diff
@@ -105,12 +105,6 @@ export class EIP1559FeeMarketTransaction extends RuntimeTransaction {
     } else {
       if (data.chainId) {
         this.chainId = Quantity.from(data.chainId);
-        if (this.common.chainId() !== this.chainId.toNumber()) {
-          throw new CodedError(
-            `Invalid chain id (${this.chainId.toNumber()}) for chain with id ${common.chainId()}.`,
-            JsonRpcErrorCode.INVALID_INPUT
-          );
-        }
       } else {
```
davidmurdoch, line 108: Deleted because I need this to happen. Maaayyybbbe we could work around it, but I'm not sure it'll be a problem since `data.chainId` should only be coming from the database? :thinking:  hmmmm

src/chains/ethereum/ethereum/src/forking/fork.ts
```diff
@@ -222,4 +227,36 @@ export class Fork {
...
+
+  /**
+   * If the blockNumber is before our fork.blockNumber return a Common instance
+   * applying the rules from the remote chain's via it's original chainId. If
+   * the remote chain's chain id is now "known", return a common with our local
+   * common's rules applied, but with the remote chain's chainId. If the block
+   * is greater than or equal to our fork.blockNumber returns `common`.
+   * @param common
+   * @param blockNumber
+   */
```
MicaiahReid:
```suggestion
 /**
   * If the `blockNumber` is before our `fork.blockNumber`, return a `Common`
   * instance, applying the rules from the remote chain's `common` via its
   * original `chainId`. If the remote chain's `chainId` is now "known", return
   * a `Common` with our local `common`'s rules applied, but with the remote
   * chain's `chainId`. If the block is greater than or equal to our
   * `fork.blockNumber` return `common`.
   * @param common
   * @param blockNumber
   */
```

MicaiahReid: I don't really have much of a preference of "chain id" vs `chainId` or common vs `common`, I just care that we're consistent. So no hard feelings if you want to switch it back. Also there's a few grammar fixes here.

src/chains/ethereum/ethereum/tests/forking/forking.test.ts
```diff
@@ -959,3 +976,147 @@ describe("forking", function () {
...
+      ({
+        contractBlockNum,
+        contractAddress,
+        contractBlockNum,
+        methods
+      } = await deployContract(remoteProvider, remoteAccounts));
+    });
+
+    before("fork from mainnet at contractBlockNum + 1", async () => {
+      // progress the remote provider forward 1 additional block so we can call
```
MicaiahReid:
```suggestion
      // progress the remote provider forward 2 additional blocks so we can call
```

MicaiahReid (approved):

PR #1589, opened 2021-11-16
fix: return error for `eth_call` when call errors

### LINKED ISSUE #1496: eth_call should return JSON-RPC error on revert
When `eth_call` causes a transaction revert, the JSON-RPC method should return an error rather than a result that just contains the revert data.  Currently (using Ganache from git) `eth_call` returns data like this for a call that reverts:

    0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000008408c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000324552433732313a207472616e7366657220746f206e6f6e20455243373231526563656976657220696d706c656d656e746572000000000000000000000000000000000000000000000000000000000000000000000000000000000000

This seems to be basically the same issue that Geth had (and resolved) in https://github.com/ethereum/go-ethereum/issues/19027.

davidmurdoch:
fixes #1496

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -1107,13 +1107,9 @@ export default class Blockchain extends Emittery.Typed<
       context: transactionContext
     });
     if (result.execResult.exceptionError) {
-      if (this.#options.chain.vmErrorsOnRPCResponse) {
-        // eth_call transactions don't really have a transaction hash
-        const hash = RPCQUANTITY_EMPTY;
-        throw new RuntimeError(hash, result, RETURN_TYPES.RETURN_VALUE);
-      } else {
-        return Data.from(result.execResult.returnValue || "0x");
-      }
+      // eth_call transactions don't really have a transaction hash
+      const hash = RPCQUANTITY_EMPTY;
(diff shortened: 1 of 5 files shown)


MicaiahReid (approved):
I've tested the change and it all looks good.

I do think this change gives further support for the idea of reversing `--vmErrorsOnRPCResponse` to something more like `--suppressVmErrors`, where the vm errors are enabled by default and can optionally be ignored.

It seems strange that the default use case would be that most vm errors are disabled, except for those from `eth_call` which are always enabled.

Then again, if that's what geth does that's what geth does.

davidmurdoch:
Allowing users to do `vmErrorsOnRPCResponse = false` was so they can make ganache more like a real node, since when running a transaction real nodes don't throw on errors. This makes sense, because transactions aren't run immediately, but they are prioritized and only run when the miner chooses to put the transaction in a block -- and sometimes that can take days (or never).

`eth_call` is often used by developers to learn why a previously run transaction reverted. This PR makes that possible _by default_ again.

Still +1 on picking a new name for this flag. Changing the default mode would also require changing the default for `legacyInstamine` to `true`, which doesn't make it very legacy any more. haha

PR #1591, opened 2021-11-16
fix: remove bigint's `n` from cli output

### LINKED ISSUE #1341: Startup display on Ganache 7 includes "n" after some numbers
E.g.:

```
Default Gas Price
==================
2000000000n

BlockGas Limit
==================
12000000n

Call Gas Limit
==================
9007199254740991n
```

(Although the real thing is colorized, which is important, because presumably you don't want to fix this in a way that loses the color.)

Presumably, this is happening because this is how `util.inspect` formats bigints.  However, the fact that these numbers are represented internally as bigints is a purely internal matter; no "n" should appear in printouts like this.

davidmurdoch:
fixes #1341

Was:
image

Now:
image

### DIFF
--- a/src/packages/cli/src/initialize/ethereum.ts
+++ b/src/packages/cli/src/initialize/ethereum.ts
@@ -11,9 +11,6 @@ function capitalizeFirstLetter(string: string) {
 function color(str: string) {
   return chalk`{hex("${TruffleColors.porsche}") ${str}}`;
 }
-function k(str: string) {
-  return chalk`{hex("${TruffleColors.porsche}").bgHex("#3C2C30") ${str}}`;
-}
 
 export default function (provider: Provider, cliSettings: CliSettings) {
   const liveOptions = provider.getOptions();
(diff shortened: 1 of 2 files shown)


MicaiahReid (approved):
I tell you with this code, You have ascended.

src/packages/cli/src/initialize/ethereum.ts
```diff
@@ -1,16 +1,26 @@
...
 import type { Provider } from "@ganache/ethereum";
 import { toChecksumAddress } from "ethereumjs-util";
 import { CliSettings } from "../types";
 
+function color(str: string) {
+  return chalk`{hex("${TruffleColors.porsche}") ${str}}`;
+}
+function k(str: string) {
+  return chalk`{hex("${TruffleColors.porsche}").bgHex("#3C2C30") ${str}}`;
+}
```
davidmurdoch: remove unused function
```suggestion
```

PR #1650, opened 2021-11-24
fix: update equality checks for replacement transactions to prevent erroneous underpriced errors

MicaiahReid:
Fixes bug where replacement transactions with `effectiveGasPrice = previousTxGasPrice + (previousTxGasPrice * priceBumpPercent` are not viable replacements (essentially a `<` to a `<=`). This edge case is actually pretty common because metamasks "Cancel transaction" functionality sets to this exact value.

Updates logic to verify replacement transactions so that `maxFeePerGas` and `maxPriorityFeePerGas` are both always checked against the previous transaction if the transactions have those fields. If not, those fields are defaulted to `gasPrice`. This is in line with geth's implementation.

Adds `--miner.priceBump` option to make the priceBump requirement configurable.

### DIFF
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -111,6 +111,9 @@ export enum TriageOption {
 export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
   #options: EthereumInternalOptions["miner"];
 
+  /**
+   * Minimum price bump percentage needed to replace a transaction that already exists in the transaction pool.
+   */
   #priceBump: bigint;
 
   #blockchain: Blockchain;
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -36,6 +36,16 @@ function shouldReplace(
   if (replaceeNonce !== replacerNonce) {
     return false;
   }
+
+  // if the transaction being replaced is in the middle of being mined, we can't
+  // replpace it so let's back out early
+  if (replacee.locked) {
+    throw new CodedError(
+      TRANSACTION_LOCKED,
+      JsonRpcErrorCode.TRANSACTION_REJECTED
+    );
+  }
+
   const replacerTip =
     "maxPriorityFeePerGas" in replacer
       ? replacer.maxPriorityFeePerGas.toBigInt()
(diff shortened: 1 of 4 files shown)


src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -27,24 +28,45 @@ import { TypedTransaction } from "@ganache/ethereum-transaction";
...
+      : replacee.effectiveGasPrice.toBigInt();
+  const replaceeMaxFee =
+    "maxFeePerGas" in replacee
+      ? replacee.maxFeePerGas.toBigInt()
+      : replacee.effectiveGasPrice.toBigInt();
+
+  const tipPremium = replaceeTip + (replaceeTip * priceBump) / 100n;
+  const maxFeePremium = replaceeMaxFee + (replaceeMaxFee * priceBump) / 100n;
 
   // if our replacer's price is `gasPrice * priceBumpPercent` better than our
```
davidmurdoch: can you update this comment?

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -83,10 +105,7 @@ export enum TriageOption {
 export default class TransactionPool extends Emittery.Typed<{}, "drain"> {
   #options: EthereumInternalOptions["miner"];
 
-  /**
-   * Minimum price bump percentage to replace an already existing transaction (nonce)
-   */
-  #priceBump: bigint = 10n;
+  #priceBump: bigint;
```
davidmurdoch, line 117: I think the old comment may still be good to keep here.

src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -158,6 +158,18 @@ export type MinerConfig = {
...
     };
+
+    /**
+     * Minimum price bump percentage needed to replace a transaction that already exists in the transaction pool.
+     *
+     * @defaultValue ""
+     */
+    priceBump: {
+      type: bigint;
+      rawType: number | bigint;
```
davidmurdoch: Since the `cliType` is string and the type is `bigint` (normalized by calling `BigInt(rawType)`) we will accept things like `0o10` (octal for 8n) and `0b1111` (binary for 15n), which I don't think we accept anywhere else.

Additionally, the `rawType` should probably include `string` (since `cliType` is string), so that we can accept the same hex strings on the command line and in typescript (`ganache --miner.priceBump 0xff` and `ganache.provider({miner:{priceBump: "0xff"}})`.

src/chains/ethereum/utils/src/errors/errors.ts
```diff
@@ -38,3 +38,9 @@ export const VM_EXCEPTION = "VM Exception while processing transaction: ";
  */
 export const VM_EXCEPTIONS =
   "Multiple VM Exceptions while processing transactions: : \n\n";
+
+/**
+ * Returned if a replacement transaction is sent while the potentially replaced transaction is being mined.
+ */
+export const TRANSACTION_LOCKED =
+  "transaction can't be replaced, mining has already started. (we don't really know how you got here, couldy you open an issue with reproduction steps? https://github.com/trufflesuite/ganache/issues/new)";
```
davidmurdoch:
```suggestion
  "transaction can't be replaced, mining has already started. (please open an issue with reproduction steps: https://github.com/trufflesuite/ganache/issues/new)";
```

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: I moved this comment over to the cli description for the `miner.priceBump` option. But I'm fine with having it here too, I've added it back with a minor adjustment to match what's in the `miner-options`.

src/chains/ethereum/options/src/miner-options.ts
MicaiahReid: I added `string` to `rawType`. I don't think any changes need to be made to the normalization, this will accept  `0xf`, `0b11`, and `20` values, as do other options (i.e. `--miner.defaultGasPrice`)

davidmurdoch (approved):

PR #1661, opened 2021-11-24
fix: ensure `eth_sendTransaction` returns the correct error when the account has insufficient funds

davidmurdoch:


src/chains/ethereum/ethereum/tests/api/eth/sendTransaction.test.ts
```diff
@@ -46,6 +46,30 @@ describe("api", () => {
...
+          const [from, to] = await p.send("eth_accounts");
+          const balance = await p.send("eth_getBalance", [from]);
+          const types = ["0x0", "0x1", "0x2"] as const;
+          for (let i = 0; i < types.length; i++) {
+            await assert.rejects(
+              p.send("eth_sendTransaction", [
+                { type: types[i], from, to, value: balance }
+              ]),
+              {
+                message: `VM Exception while processing transaction: sender doesn't have enough funds to send tx. The upfront cost is: 1000000099667968750000 and the sender's account (${from}) only has: ${BigInt(
```
MicaiahReid: I think we'll need to get each block's `baseFeePerGas` to calculate the upfront cost rather than hardcoding it. In instamine mode, after each block is sent a block is still mined even if the transaction is rejected. After each block, the `baseFeePerGas` is updated. Maybe a workaround could be to set `miner: { blockTime: 10000 }` for the provider so that all three transactions are rejected on the same block.

MicaiahReid (changes requested):

MicaiahReid (approved):

PR #1740, opened 2021-12-04
fix: prevent DEBUG env var from crashing ganache

davidmurdoch:
@ethereumjs/vm uses the `DEBUG` env var and the `debug` npm module to output debug information during runtime. We don't currently use the @ethereum/tx library for our transactions, so our transaction are "missing" some fields that are used only when DEBUG is true. This causes ganache to error when it shouldn't.

Other potential solutions:
 * implement all of the debug-related properties everywhere
 * use @ethereumjs/tx (and probably `/account`, and a `/block, too)
 
A benefit of just removing the `debug` module and replacing `process.env.DEBUG` with `false` is that the resulting build is smaller (and slightly faster due to the minifier removing dead code). Downside: no one can use `debug` for ganache's internals (I doubt anyone was wanting this anyway?) and if we want to use the debug module in the future we can't... because it's removed by the build process.

### DIFF
--- a/src/packages/ganache/webpack/polyfills/debug.ts
+++ b/src/packages/ganache/webpack/polyfills/debug.ts
@@ -1,0 +1,1 @@
+export const debug = () => () => {};
--- a/src/packages/ganache/webpack/webpack.browser.config.ts
+++ b/src/packages/ganache/webpack/webpack.browser.config.ts
@@ -35,7 +35,10 @@ const config: webpack.Configuration = merge({}, base, {
       // mcl-wasm may be needed when creating a new @ethereumjs/vm and requires a browser version for browsers
       "mcl-wasm": require.resolve("mcl-wasm/browser"),
       // ws doesn't work in the browser, isomorphic-ws does
-      ws: require.resolve("isomorphic-ws/")
+      ws: require.resolve("isomorphic-ws/"),
+      // we don't use the debug module internally, so let's just not include it
+      // in any package.
+      debug: require.resolve("./polyfills/debug")
     }
   },
   output: {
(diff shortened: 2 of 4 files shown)


MicaiahReid (approved):
I'm curious, do you have any repro steps for times when Ganache would error when it shouldn't before this fix?

davidmurdoch:
> I'm curious, do you have any repro steps for times when Ganache would error when it shouldn't before this fix?

Woops. I should have provided this!


Using ganache 7.0.0-beta.1:

```bash
ganache -d
```

Then from a separate terminal run:

```bash
curl -H 'Content-Type: application/json'   --data '{"jsonrpc":"2.0","method": "eth_estimateGas", "params": [ {"from": "0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1", "to": "0x016c5aa3Fd1e8cF85b31F0F69F9BE82f2f52e764"} ],"id":1}' http://localhost:8545
```

 and you'll get:
 
 ```json
{"id":1,"jsonrpc":"2.0","error":{"message":"e.tx.isSigned is not a function","stack":"TypeError: e.tx.isSigned is not a function\n    at VM._runTx (/home/david/.nvm/versions/node/v12.0.0/lib/node_modules/ganache/dist/node/0.js:2:1115879)\n    at processTicksAndRejections (internal/process/task_queues.js:88:5)\n    at async VM.runTx (/home/david/.nvm/versions/node/v12.0.0/lib/node_modules/ganache/dist/node/0.js:2:1122709)\n    at async exactimate (/home/david/.nvm/versions/node/v12.0.0/lib/node_modules/ganache/dist/node/1.js:2:104682)","code":-32700}}
 ```

PR #1780, opened 2021-12-08
fix: version disparity between startup/`web3_clientVersion`

MicaiahReid:
Fixes issue with startup version being different from version received when using `web3_clientVersion` by using webpack to set version environment variables on build. Uses "DEV" version as fallback when environment variables aren't availble.

Positive side affect is we were able to remove ts-transformer-inline-file lib.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "semver": "7.3.5",
     "shx": "0.3.3",
     "ts-node": "9.1.1",
-    "ts-transformer-inline-file": "0.1.1",
     "ttypescript": "1.5.12",
     "typescript": "4.1.3",
     "validate-npm-package-name": "3.0.0",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "postpublish": "git restore \"$(lerna list --parseable --scope ganache)\"/npm-shrinkwrap.json",
     "start": "lerna exec --loglevel=silent --scope ganache -- npm run start --silent -- ",
     "test": "lerna exec --concurrency 1 -- npm run test",
-    "tsc": "ttsc --build src",
+    "tsc": "tsc --build src",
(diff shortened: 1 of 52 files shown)


src/tsconfig-base.json
```diff
@@ -15,18 +15,13 @@
...
-      "ES2020"
-    ],
+    "lib": ["ES2020"],
     "experimentalDecorators": true,
     "plugins": [
-      {
-        "transform": "ts-transformer-inline-file/transformer"
-      },
       {
         "transform": "@zerollup/ts-transform-paths",
```
davidmurdoch: Can you try removing this one too (I don't think it was even used)

package.json
```diff
@@ -49,7 +49,6 @@
     "semver": "7.3.5",
     "shx": "0.3.3",
     "ts-node": "9.1.1",
-    "ts-transformer-inline-file": "0.1.1",
```
davidmurdoch, line 52: You should commit the changed package-lock.json file as well.

davidmurdoch (changes requested):
YAY! (changes requested)

package.json
MicaiahReid: For some reason there weren't any available to commit. `npm run reinstall` seemed to do the trick.

src/packages/ganache/npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "ganache",
-	"version": "7.0.0-alpha.2",
```
davidmurdoch, line 3: I think you need to update and commit this file, as it still contains a `ttypescript` reference, but I don't think it should. Can you check?

davidmurdoch (approved):

PR #1861, opened 2021-12-19
fix: a typo in `evm_setTime` docs

gas1cent:
`This is will result in an invalid state` -> `This will result in an invalid state`


davidmurdoch:
Good catch! Can you only include the change to api.ts? We'll automatically generate the docs as part of the release.

gas1cent:
@davidmurdoch yeah, sure. Thanks for the tip 😉

davidmurdoch (approved):

PR #2055, opened 2022-01-13
chore: update licenses for 2022

davidmurdoch:
This just updates everything to 2022, but maybe we should make it say a year range instead?


MicaiahReid (approved):
Incredible work.

davidmurdoch:
> Incredible work.

But should they be year ranges? :thinking:

PR #2094, opened 2022-01-18
ci: add master branch to release action

davidmurdoch:


.github/workflows/release.yml
```diff
@@ -76,7 +95,7 @@ jobs:
       # this needs `--no-verify-access` until https://github.com/lerna/lerna/issues/2788 is fixed
       - name: Release to npm
         run: |
-          $(npm bin)/lerna publish from-package --yes --dist-tag ${GITHUB_REF##*/} --pre-dist-tag ${GITHUB_REF##*/} --no-verify-access
+          $(npm bin)/lerna publish from-package --yes --dist-tag ${TAG} --pre-dist-tag ${TAG} --no-verify-access
```
davidmurdoch, line 96: @cds-amal how does this line look to you?

cds-amal: IT looks ok to me as long as lerna is installed :) how did the action run?

.github/workflows/release.yml
```diff
@@ -33,11 +33,26 @@ jobs:
...
-          git add .
+          echo "TAG=latest" >> $GITHUB_ENV
+
+      - name: Set TAG for non-master to the branch name
+        if: ${{ github.ref != 'refs/heads/master' }}
+        run: |
+          echo "TAG=${GITHUB_REF##*/}" >> $GITHUB_ENV
+
+      - name: Update package versions for latest release (master)
+        if: ${{ env.TAG == 'latest' }}
```
MicaiahReid, line 47: since after `if: ${{ github.ref == 'refs/heads/master' }}` we set the tag to `'latest'`, can't we just remove this if statement and include the following `run: $(npm bin)/lerna version patch --no-git-tag-version --no-push --yes --exact` inside of it?

Not particularly important, but could be a slight improvement.

MicaiahReid (approved):

.github/workflows/release.yml
davidmurdoch: Yeah, we could do that, and avoid these weird `if: `s would be kinda nice. The "step" would just be doing two things now. It'd be named something like "Set TAG for master to latest and update versions for master".

.github/workflows/release.yml
davidmurdoch: [It ran](https://github.com/trufflesuite/ganache/runs/4859668937?check_suite_focus=true)! This step wasn't included in the test action because it has side effects.

cds-amal (approved):
Looks good, pending tests outcome

Issue #2133, opened 2022-01-21
Ganache v7.0.0: Javascript traces inside revert message


Uxio0:
We are using ganache to test Gnosis Safe, and we are trying to update from v6 to v7, and we detected a weird behaviour. Even if `--chain.vmErrorsOnRPCResponse false`, when having a `revert` message there are Javascript traces in there that prevent the parsing of the revert error:

```
{"id":1,"jsonrpc":"2.0","error":{"message":"VM Exception while processing transaction: revert \\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\xef\xbf\xbdu","stack":"RuntimeError: VM Exception while processing transaction: revert \\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\xef\xbf\xbdu\\n    at Blockchain.simulateTransaction (/home/uxio/.config/yarn/global/node_modules/ganache/dist/node/1.js:2:48034)","code":-32000,"name":"RuntimeError","data":{"hash":null,"programCounter":112,"result":"0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000008775","reason":"\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\xef\xbf\xbdu","message":"revert"}}}
```

It's not easily reproducible as it involves deploying the Safe contracts https://github.com/gnosis/gnosis-py/runs/4897023415?check_suite_focus=true#step:10:370

[...]

PR #2186, opened 2022-01-25
fix: gethify eth call errors (#2133)

MicaiahReid:
Before this change, errors from `eth_call` were formatted in our "non-standard" `vmErrorsOnRPCResponse` format, which uses the error's `data` property to store some extra helpful information (program counter, hash, etc.):
```
{
   "error": {
     "message": "...",
     "code": ...
     "data": {
          ... bunch of props, like `hash` and `programCounter`...
         "result": "<raw revert hex string>" <---- this moves (See below)
     }
   } 
}
```

 Now, the error's `data` property only contains the raw revert hex string, which should more closely match real node's error handling:
```
{
   "error": {
     "message": "...", // <- we'll change the message
     "code": ...
     "data": "<raw revert hex string>"  <---- new home
   } 
}
```

Our hope is that this will allow users to remove any conditionals handling errors differently between Ganache and real Ethereum nodes.

### DIFF
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -114,6 +114,10 @@ type TypedData = Exclude<
 //#endregion
 
 //#region helpers
+/**
+ * Combines RuntimeErrors for a list of rejected or reverted transactions.
+ * @param transactions Array of transactions with errors to assert.
+ */
 function assertExceptionalTransactions(transactions: TypedTransaction[]) {
   let baseError: string = null;
   let errors: string[];
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -9,7 +9,7 @@ import {
   TraceDataFactory,
   TraceStorageMap,
   RuntimeError,
-  RETURN_TYPES,
+  CallError,
   StorageKeys,
   StorageRangeResult,
   StorageRecords,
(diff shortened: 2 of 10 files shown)


davidmurdoch (approved):

fabianorodrigo:
Guys, 
I have a doubt about this changes implemented on v7.0.1.

I'm using the package @openzeppelin/test-helpers in my unit tests, where I have oracles like this:

`
await expectRevert(
      gameContract.bet(score, betTokenAmount, {
        from: bettor,
      }),
      "ERC20: transfer amount exceeds allowance"
    );
`
Won't Ganache give me that message I passed to `revert` inside my contract?

Sorry if I misunderstand. It's because when I run run the above test on Ganache v2.5.4 it works/passes, but after upgrading to v7.0.1  it's failing:


`
      Wrong kind of exception received
      + expected - actual

      -Transaction: 0x3dfa34902aaceb3fd6e7176180d65bc9c0b7ec664ddad54c5700ab2f42d6214c exited with an error (status 0). 
      -     Please check that the transaction:
      -     - satisfies all conditions set by Solidity `require` statements.
      -     - does not trigger a Solidity `revert` statement.
      +ERC20: transfer amount exceeds allowance`

davidmurdoch:
@fabianorodrigo  Have you updated your truffle version to the latest version?

cds-amal:
hi @fabianorodrigo, you'll need to update to the latest version of truffle `npm -g uninstall truffle && npm -g install truffle` and your project tests should work. Note, I pushed a PR to align your tests with openzeppelin-test-helper documentation that you should review.

fabianorodrigo:
@cds-amal , 
Thank you very much! I've already updated to Truffle v5.4.31 but after reading your answer I realized that it was at another workstation.
Now I updated Truffle and everything is fine.
Thanks!

PR #2268, opened 2022-02-02
fix: `eth_subscribe` logs

MicaiahReid:
Previously, logs returned via `eth_subscribe("logs")` returned all logs in an array. [Geth's logging](https://geth.ethereum.org/docs/rpc/pubsub) and Ganache v6 emit separately for each event. This change makes it so we emit separately for each log once again.

Also, I've added some tests for filtering `eth_subscribe("logs")` events.

### DIFF
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -2078,8 +2078,6 @@ export default class EthereumApi implements Api {
         const unsubscribe = this.#blockchain.on(
           "blockLogs",
           (blockLogs: BlockLogs) => {
-            // TODO: move the JSON stringification closer to where the message
-            // is actually sent to the listener
             for (const log of blockLogs.filter(addresses, topics)) {
               promiEvent.emit("message", {
                 type: "eth_subscription",
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -2080,16 +2080,15 @@ export default class EthereumApi implements Api {
           (blockLogs: BlockLogs) => {
             // TODO: move the JSON stringification closer to where the message
             // is actually sent to the listener
-            const result = JSON.parse(
-              JSON.stringify([...blockLogs.filter(addresses, topics)])
-            );
-            promiEvent.emit("message", {
-              type: "eth_subscription",
-              data: {
-                result,
-                subscription: subscription.toString()
-              }
-            });
+            for (const log of blockLogs.filter(addresses, topics)) {
+              promiEvent.emit("message", {
+                type: "eth_subscription",
+                data: {
+                  result: log,
+                  subscription: subscription.toString()
+                }
+              });
+            }
           }
         );
         subscriptions.set(subscription.toString(), unsubscribe);
--- a/src/chains/ethereum/ethereum/tests/api/eth/logs.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/logs.test.ts
@@ -117,6 +117,7 @@ describe("api", () => {
           });
 
           it("filters subscription by address", async () => {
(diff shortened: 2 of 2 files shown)


src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2080,16 +2080,15 @@ export default class EthereumApi implements Api {
           (blockLogs: BlockLogs) => {
             // TODO: move the JSON stringification closer to where the message
             // is actually sent to the listener
```
davidmurdoch: looks like we don't need the comment anymore?

src/chains/ethereum/ethereum/tests/api/eth/logs.test.ts
```diff
@@ -33,20 +29,37 @@ describe("api", () => {
...
                 if (message.data.subscription === subId) {
+                  messages.push(message);
+                  if (expectedMessageCount > messages.length + 1) {
+                    getMessagesForSub(subId, expectedMessageCount, messages);
+                  }
                   resolve(message);
                 }
               });
             });
+            return messages;
```
davidmurdoch: The recursion seems strange here. Something like the following should do the same job and I think it seems a bit simpler. Thoughts?

```typescript
          const getMessagesForSub = async (
            subId,
            expectedMessageCount
          ) => {
            const messages = [];
            return await new Promise<any[]>(resolve => {
              provider.on("message", message => {
                if (message.data.subscription === subId) {
                  messages.push(message);
                  if (expectedMessageCount === messages.length) {
                    resolve(messages);
                  }
                }
              });
            });
```

src/chains/ethereum/ethereum/tests/api/eth/logs.test.ts
```diff
@@ -62,39 +75,145 @@ describe("api", () => {
...
+            const filtered = getMessagesForSub(
+              subscriptionId,
+              numberOfLogs,
+              []
+            );
+            await provider.send("eth_sendTransaction", [{ ...filteredTx }]);
+
+            const noMessages = await Promise.race([
+              filtered,
+              new Promise(resolve => setTimeout(resolve, 3000, "timeout"))
```
davidmurdoch: Interesting approach here. I wonder if there is a better way? Maybe listening for a `newHeads` event would work? Though i'm not sure that would guarantee what we are trying to guarantee here. hmmmmmm :thinking:

davidmurdoch (changes requested):
This looks really great. Especially those new tests!

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: Yep, good find!

src/chains/ethereum/ethereum/tests/api/eth/logs.test.ts
MicaiahReid: Yours is waaaaaay better, fixing.

src/chains/ethereum/ethereum/tests/api/eth/logs.test.ts
MicaiahReid: It looks like `.emit("blockLogs", blockLogs)` is always emitted immediately before `.emit("block", block)`, so presumably `logs` subscriptions would beat `newHeads` in a race, unless the logs were successfully filtered. Probably is a bit frail for a test, though.

davidmurdoch: Yeah, it sounds like we'd also need to write a test here that ensures the order of messages here :-/ hmmm

MicaiahReid: See updated code, we've managed to work around the timeout.

davidmurdoch (approved):

PR #2331, opened 2022-02-09
fix: serialize eth_subscribe log data

MicaiahReid:


src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -244,4 +245,40 @@ describe("provider", () => {
...
+    let web3: Web3;
+    let accounts: string[];
+
+    beforeEach(async () => {
+      provider = await getProvider({
+        chain: { networkId },
+        wallet: { deterministic: true }
+      });
+      web3 = new Web3();
+      // TODO: remove "as any" once we've fixed our typing issues with web3
```
davidmurdoch: Can you add a reference to the PRs that will fix this?

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -244,4 +245,40 @@ describe("provider", () => {
...
+      accounts = await web3.eth.getAccounts();
+    });
+
+    it("returns things via legacy", async () => {
+      const subscription = web3.eth.subscribe("newBlockHeaders");
+      assert(subscription != null);
+      assert.notStrictEqual(subscription.id, false);
+
+      // if the data isn't properly serialized before emitting, web3 won't
+      // ever catch this
```
davidmurdoch: What do you mean by "catch" here? Would "handle" be a better word?

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -244,4 +245,40 @@ describe("provider", () => {
...
+
+  describe("web3 compatibility", () => {
+    const networkId = 1234;
+    let provider: EthereumProvider;
+    let web3: Web3;
+    let accounts: string[];
+
+    beforeEach(async () => {
+      provider = await getProvider({
+        chain: { networkId },
```
davidmurdoch: Do we need to set a `networkId` here? If not I think it should be removed.

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -244,4 +245,40 @@ describe("provider", () => {
...
+
+    it("returns things via legacy", async () => {
+      const subscription = web3.eth.subscribe("newBlockHeaders");
+      assert(subscription != null);
+      assert.notStrictEqual(subscription.id, false);
+
+      // if the data isn't properly serialized before emitting, web3 won't
+      // ever catch this
+      subscription.on("data", data => {
+        assert.strictEqual(
```
davidmurdoch: I think this test will still pass even if `data` handler isn't handled.

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -359,13 +359,14 @@ export default class EthereumProvider
     >;
     if (promise instanceof PromiEvent) {
       promise.on("message", data => {
+        const serialized = JSON.parse(JSON.stringify(data));
```
davidmurdoch: I don't think `serialized` is the right name here. Maybe `normalizedData` or `clonedData` would be better?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2054,13 +2054,10 @@ export default class EthereumApi implements Api {
...
             (result as any).baseFeePerGas = header.baseFeePerGas;
           }
-
-          // TODO: move the JSON stringification closer to where the message
-          // is actually sent to the listener
           promiEvent.emit("message", {
             type: "eth_subscription",
             data: {
-              result: JSON.parse(JSON.stringify(result)),
+              result: result,
```
```suggestion
              result,
```

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -244,4 +245,40 @@ describe("provider", () => {
...
+        wallet: { deterministic: true }
+      });
+      web3 = new Web3();
+      // TODO: remove "as any" once we've fixed our typing issues with web3
+      web3.setProvider(provider as any);
+      accounts = await web3.eth.getAccounts();
+    });
+
+    it("returns things via legacy", async () => {
+      const subscription = web3.eth.subscribe("newBlockHeaders");
```
davidmurdoch: How does this work without awaiting it first? :thinking:

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/provider.ts
MicaiahReid: I like `normalizedData`. Updated!

src/chains/ethereum/ethereum/tests/provider.test.ts
MicaiahReid: Removed.

src/chains/ethereum/ethereum/tests/provider.test.ts
MicaiahReid:
```suggestion
      // TODO: remove "as any" once we've fixed our typing issues
      // with web3 (https://github.com/ChainSafe/web3.js/pull/4761)
```

src/chains/ethereum/ethereum/tests/provider.test.ts
MicaiahReid: Yes, I definitely wasn't thinking of "catch" as an actual technical term here. Handle is better.
```suggestion
      // ever handle this
```

src/chains/ethereum/ethereum/tests/provider.test.ts
MicaiahReid: Fixed test!

davidmurdoch (approved):

Issue #1245, opened 2021-09-23
get forking working in the browser


davidmurdoch:
We at least: need to polyfill `http` and `https`, use isomporphic-ws in the browser, and fix the `new URL` issue.

davidmurdoch:
closed by #3130

PR #2345, opened 2022-02-11
Add polyfill for http and https

rmeissner:
Contributes to https://github.com/trufflesuite/ganache/issues/1245

Changes:
- sets `http` polyfill to `stream-http`
- sets `https` polyfill to `https-browserify`

Notes:
- `isomporphic-ws` was already set


rmeissner:
I tested this by using ganache in a react app. I worked well with the dev server, but if I perform a production optimized build I get a compiler error `TypeError: /ganache/src/packages/ganache/dist/web/ganache.min.js: Duplicate declaration "a"` ... not sure why.

rmeissner:
I couldn't get the URL error fixed easily. Any idea how to go at this? I thought I can polyfill this, but it doesn't work and now I was thinking about using a wrapper library such as `native-url`, but then the used type definitions don't work in all places out of the box :(

davidmurdoch:
The browser has a native `URL` support, so I'm not sure what's going on with that. I'll take a closer look at this today.

rmeissner:
Yes. I found an article that it should be possible just to remove the import. But this doesn't work in all places (it does in some)

davidmurdoch:
We're finally getting this working. Thanks for starting it! I've moved the work from this PR over to #3130 so I can finish up the work on a branch based on this repository. See: #3130!

PR #2397, opened 2022-02-17
fix: use `host` param when passed in `server.listen`

### LINKED ISSUE #2358: Ganache binding to all network interfaces even with `-h localhost`
When I run
```ganache --fork -n -h localhost```
I see the following with `lsof`:
```node    488738 root   20u  IPv6 28301584      0t0  TCP *:8545 (LISTEN)```
Because I'm running it on a popular cloud VM without a firewall, I also get a flood of people trying to steal the forked ETH.

### PR DESCRIPTION
**Pull request by davidmurdoch**
fixes #2358

### DIFF
--- a/src/packages/core/src/server.ts
+++ b/src/packages/core/src/server.ts
@@ -109,7 +109,7 @@ export const _DefaultServerOptions = serverDefaults;
  */
 export class Server<
   T extends FlavorName = typeof DefaultFlavor
-> extends Emittery<{ open: undefined; close: undefined }> {
+  > extends Emittery<{ open: undefined; close: undefined }> {
   #options: InternalOptions;
   #providerOptions: Options<T>;
   #status: number = ServerStatus.unknown;
@@ -180,10 +180,9 @@ export class Server<
     host?: string | Callback,
     callback?: Callback
   ): void | Promise<void> {
-    let hostname: string = null;
     if (typeof host === "function") {
       callback = host;
-      hostname = null;
+      host = null;
     }
     const callbackIsFunction = typeof callback === "function";
     const status = this.#status;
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Review by MicaiahReid — APPROVED** [review]
Cool tests! But it looks like they're only passing on Ubuntu :(

File: src/packages/core/tests/server.test.ts [hunk 1]
```diff
@@ -59,6 +63,152 @@ describe("server", () => {
...
+    }
+
+    function getHost(info: NetworkInterfaceInfo, interfaceName: string) {
+      // a link-local ipv6 address starts with fe80:: and _must_ include a "zone_id"
+      if (info.family === "IPv6" && info.address.startsWith("fe80::")) {
+        if (process.platform == "win32") {
+          // on windows the zone_id is the scopeid
+          return `${info.address}%${info.scopeid}`;
+        } else {
+          // on *nix the zone_id is the `interfaceName`
```
davidmurdoch, line 120: this is why we can't have nice things

src/packages/core/src/server.ts
```diff
@@ -184,6 +184,8 @@ export class Server<
     if (typeof host === "function") {
       callback = host;
       hostname = null;
+    } else {
```
cds-amal: Small tiny, itsy bitsy, style note: If you don't set `hostname = null` in the truthy consequence block, this `hostname` can be assigned to the result of a ternary

src/packages/core/tests/server.test.ts
```diff
@@ -59,6 +63,149 @@ describe("server", () => {
     }
   }
 
+  // skip this test unless in GitHub Actions, as this test iterates over
+  // all available network interfacesnand network interfaces on user
+  // machines are unpredictible and may behave in ways that we don't care
+  // about.
+  (process.env.GITHUB_ACTION ? describe : describe.skip)("listen", function () {
```
cds-amal, line 70: TIL!

cds-amal (commented):
Interesting! I think these tests are more sound than the previous ones. Nice work on OSX interface filtering. I wonder what's going on with Windows networking 🤔

src/packages/core/src/server.ts
davidmurdoch: I could probably just rewrite this to get rid of `hostname` and reuse `host`. Is that what you're thinking?

cds-amal: Oh, that would be better than what I was thinking.

davidmurdoch:
Waiting for merge this one until after https://github.com/trufflesuite/ganache/pull/2503 so I can stop playing the Restart All Tests game.

PR #2405, opened 2022-02-18
fix: allow hex string for `coinbase` option in cli

davidmurdoch:


MicaiahReid (changes requested):
Any chance we can get tests for this? (I'd be happy to write them if you don't have time, but it would not be until next week)

davidmurdoch:
Only reason I didn't add tests is because we don't yet have a testing framework set up for the cli itself.

MicaiahReid (approved):

PR #2528, opened 2022-03-04
chore: update @ethereumjs/vm to v5.9.0, add support for sepolia

davidmurdoch:
We've updated `@ethereumjs/vm` to v5.9.0 and added support for forking the new "sepolia" test network.


src/chains/ethereum/ethereum/package.json
```diff
@@ -70,14 +70,15 @@
     "env-paths": "2.2.1",
     "eth-sig-util": "2.5.3",
     "ethereumjs-abi": "0.6.8",
-    "ethereumjs-util": "7.1.3",
+    "ethereumjs-util": "7.1.4",
     "keccak": "3.0.1",
+    "encoding-down": "7.1.0",
```
davidmurdoch, line 75: We were already using encoding-down directly, but were able to  import it only because it was present within a transitive dependency.

src/chains/ethereum/ethereum/package.json
```diff
@@ -98,7 +101,6 @@
     "@types/subleveldown": "4.1.1",
     "@types/superagent": "4.1.10",
     "@types/ws": "^7.4.7",
-    "abstract-leveldown": "7.2.0",
```
davidmurdoch, line 101: We were only using the types, not the package itself.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -61,7 +61,7 @@ import {
 } from "@ethereumjs/vm/dist/state/index";
 import { GanacheTrie } from "./helpers/trie";
 import { ForkTrie } from "./forking/trie";
-import { LevelUp } from "levelup";
+import type { LevelUp } from "levelup";
```
davidmurdoch, line 68: The types got weird at one point, so I was attempting to figure it out and one of the steps was to ensure that we were using only the type, and not the runtime. Since many of these imports really are type-only imports I've left these changes in.

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -1,14 +1,13 @@
...
 import Blockchain from "./blockchain";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
-import levelup from "levelup";
-import type { LevelUp } from "levelup";
-import leveldown from "leveldown";
 import sub from "subleveldown";
 import encode from "encoding-down";
+import leveldown from "leveldown";
+import type { LevelUp } from "levelup";
+const levelup = require("levelup");
```
davidmurdoch, line 10: The latest LevelUp types are exported at the wrong location so we can't use the runtime and the types in the same import. See: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/59192

src/chains/ethereum/ethereum/src/forking/cache.ts
```diff
@@ -1,19 +1,18 @@
...
-    super(trie);
+    /**
+     * Looks up address in underlying trie.
+     * @param address - Address of account
+     */
+    const lookupAccount = async (address: Address) => {
+      const rlp = await (trie as ForkTrie).get(address.buf);
+      return rlp ? Account.fromRlpSerializedAccount(rlp) : new Account();
+    }
+    super({ getCb: lookupAccount, putCb: trie.put.bind(trie), deleteCb: trie.del.bind(trie) });
```
davidmurdoch, line 16: This is the real change. The @ethereumjs/vm `Cache` no longer accepts a Merkle Patricia Trie instance, but instead takes the options listed here. It also replaces the `_lookupAccount` function internally with the `getCb` option, which is why it was removed below.

I don't know what these have a "Cb" prefix and don't know what `Cb` means in this context. I can tell you that these are _not_ "callbacks" though, they actually fetch the data and return Promises.
.

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -291,7 +291,7 @@ export class ForkTrie extends GanacheTrie {
    */
   copy(includeCheckpoints: boolean = true) {
     const db = this.db.copy() as CheckpointDB;
-    const secureTrie = new ForkTrie(db._leveldb, this.root, this.blockchain);
+    const secureTrie = new ForkTrie(db._leveldb as LevelUp, this.root, this.blockchain);
```
davidmurdoch, line 294: `merkle-patricia-tree` uses a different version of the `CheckpointDB` types so we have to cast it to our own.

src/chains/ethereum/ethereum/src/helpers/trie.ts
```diff
@@ -21,7 +21,7 @@ export class GanacheTrie extends SecureTrie {
    */
   copy(includeCheckpoints = true) {
     const db = this.db.copy();
-    const secureTrie = new GanacheTrie(db._leveldb, this.root, this.blockchain);
+    const secureTrie = new GanacheTrie(db._leveldb as LevelUp, this.root, this.blockchain);
```
davidmurdoch, line 24: `merkle-patricia-tree` uses a different version of the `CheckpointDB` types so we have to cast it to our own.

src/chains/filecoin/filecoin/package.json
```diff
@@ -60,16 +60,17 @@
     "@ganache/filecoin-options": "0.1.4",
     "@ganache/utils": "0.1.4",
     "@trufflesuite/uws-js-unofficial": "20.4.0-unofficial.3",
+    "@types/abstract-leveldown": "7.2.0",
```
davidmurdoch, line 63: Again, we use the types directly, not the library itself.

src/chains/filecoin/filecoin/package.json
```diff
@@ -60,16 +60,17 @@
...
     "@types/bn.js": "5.1.0",
     "@types/deep-equal": "1.0.1",
-    "@types/levelup": "4.3.0",
+    "@types/encoding-down": "5.0.0",
```
davidmurdoch, line 66: We now use these types internally, we were just using it as `any` before.

src/chains/filecoin/filecoin/src/database.ts
```diff
@@ -54,10 +55,7 @@ export default class Database extends Emittery {
     const store = this.#options.db;
     let db: LevelUp;
     if (store) {
-      this.#rootStore = encode(store, levelupOptions);
-      // @ts-ignore - I get an error in ts-node that I don't get running ttsc:
-      // Argument of type 'AbstractLevelDOWN<any, any>' is not assignable to parameter of type 'string'
-      // But this error just doesn't make any sense. I'm just going to ignore the error
+      this.#rootStore = encode(store as AbstractLevelDOWN, levelupOptions);
       db = levelup(this.#rootStore!, {});
```
davidmurdoch, line 59: It works now, I guess.

src/packages/utils/src/utils/constants.ts
```diff
@@ -15,4 +15,4 @@ export const RPCQUANTITY_ONE = Quantity.from(1n);
 export const RPCQUANTITY_GWEI = Quantity.from(1000000000);
 export const WEI = 1000000000000000000n as const;
 
-export const KNOWN_CHAINIDS = new Set([1, 3, 4, 5, 42]);
+export const KNOWN_CHAINIDS = new Set([1, 3, 4, 5, 42, 11155111]);
```
davidmurdoch, line 18: `11155111` is the `sepolia` chain's `ChainID`.

Might be worth noting that we are NOT adding this to our list of Infura known networks because Infura doesn't know about it right now.

MicaiahReid (approved):
It just needs a description that can be added to the release and looks good to go!

PR #2529, opened 2022-03-04
feat: add __experimental_info export to core

davidmurdoch:
This is an internal and private feature for [Truffle](https://github.com/trufflesuite/truffle). Truffle needed a way to programmatically get the list of chains Ganache fully supports in order to enable the dry-run feature for those chains.

This introduces a new experimental and private (this will likely change in a future release!) `__experimental_info` export:

```typescript
Readonly<{
  version: string,
  fork: Readonly<{
    /**
     * Chains Ganache is known to be compatible with. Operations performed
     * locally at historic block numbers will use the Ethereum Virtual Machine
     * OPCODEs, gas prices, and EIPs that were active at the time the historic
     * block originally took place.
     */
    knownChainIds: number[],
  }>
}>
```

### DIFF
--- a/src/packages/core/index.ts
+++ b/src/packages/core/index.ts
@@ -1,11 +1,23 @@
 import { ConnectorsByName, DefaultFlavor, FlavorName } from "@ganache/flavors";
+import { KNOWN_CHAINIDS } from "@ganache/utils";
 import ConnectorLoader from "./src/connector-loader";
 import { ProviderOptions, ServerOptions } from "./src/options";
 import Server from "./src/server";
 export { Server, ServerStatus, _DefaultServerOptions } from "./src/server";
-
 export type { Provider, EthereumProvider, FilecoinProvider } from "@ganache/flavors";
 export type { ProviderOptions, ServerOptions } from "./src/options";
+export type _ExperimentalInfo = Readonly<{
+  version: string,
+  fork: Readonly<{
+    /**
+     * Chains fully supported by Ganache forking.
+     */
+    supportedChainIds: number[],
+  }>
+}>
+
+
+const version = process.env.VERSION || "DEV";
 
 /**
  * @public
@@ -46,6 +58,18 @@ const Ganache = {
   ): ConnectorsByName[T]["provider"] => {
     const loader = ConnectorLoader.initialize<T>(options);
     return loader.connector.provider;
+  },
+  /**
+   * 
+   * @experimental
+   */
+  __experimental_info(): _ExperimentalInfo {
+    return {
+      version,
+      fork: {
+        supportedChainIds: Array.from(KNOWN_CHAINIDS)
+      }
+    };
   }
 };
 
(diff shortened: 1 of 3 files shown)


src/packages/core/tests/interface.test.ts
```diff
@@ -1,14 +1,18 @@
 import Ganache from "..";
 import * as assert from "assert";
+import { KNOWN_CHAINIDS } from "@ganache/utils";
 
 describe("interface", () => {
-  it("has an interface", () => {
+  it.only("has an interface", () => {
```
davidmurdoch:
```suggestion
  it("has an interface", () => {
```

src/packages/core/index.ts
```diff
@@ -1,11 +1,23 @@
...
 import Server from "./src/server";
 export { Server, ServerStatus, _DefaultServerOptions } from "./src/server";
-
 export type { Provider, EthereumProvider, FilecoinProvider } from "@ganache/flavors";
 export type { ProviderOptions, ServerOptions } from "./src/options";
+export type _ExperimentalInfo = Readonly<{
+  version: string,
+  fork: Readonly<{
+    /**
+     * Chains fully supported by Ganache forking.
```
MicaiahReid: I think it would be good to be explicit about what we mean by "fully supported" here.

I could imagine a world where we also have "partially supported" chain IDs, where we know up to what point Ganache works, and where you might start to find differences.

davidmurdoch: By "Fully supported" I really mean "ethereumjs-vm claims to know the rules" and I just assume they get it right. We do nothing with the Chain ID other than pass it to ethereumjs-vm. :thinking:

MicaiahReid: I see, that makes sense. In the future it would be really cool to have a test suite that "proves" "full support", or points to the differences between Ganache + some chain ID and the same transactions on that chain.

Maybe it spins up a test Ganache instance at block `n` of the chain we want to test, and has a reference instance of Ganache at block `n+1`. Then it runs the same transactions that were on the `+1` block onto the `n` block and finds the differences in the resulting block/transactions.

davidmurdoch: Updated the code and comment and create a new issue: https://github.com/trufflesuite/ganache/issues/2560

src/packages/core/index.ts
```diff
@@ -46,6 +61,18 @@ const Ganache = {
...
+  },
+  /**
+   * 
+   * @experimental
+   */
+  __experimental_info(): _ExperimentalInfo {
+    return {
+      version,
+      fork: {
+        knownChainIds: Array.from(KNOWN_CHAINIDS)
```
haltman-at, line 73: The method seems good enough for me!  Earlier you made it sound like there might be a more expansive thing with more info about the various chains, which might be nice for later, but if that's a bunch more work, well, `knownChainIds` does the job! :)

davidmurdoch: Oh yeah, we could potentially stuff a whole bunch of interesting stuff in here in the future.

haltman-at (approved):
Seems good to me!  Maybe more can go here in the future, but this should be good enough for now.

MicaiahReid (approved):

PR #905, opened 2021-05-18
feat: Allow contract code and balance overrides

dekz:
Wanted to gather some early feedback on this WIP PR. https://github.com/trufflesuite/ganache-core/issues/554

We'd love to have override support, similar to geth, in ganache. https://geth.ethereum.org/docs/rpc/ns-eth

The feature in geth is fantastic as it allows us to query a contract without requiring it to be deployed onto the chain. It also allows us to use this query contract on an archive node (i.e before it could conceivably be deployed). It also allows us to replace token contracts with dummy contracts and override balances/allowances for simulation and testing. 

Would love to hear your thoughts on whether this will be a PR you'd accept and any hints on the approach taken.

PR #2565, opened 2022-03-09
feat: allow balance, code, nonce, and state overrides in `eth_call`

### LINKED ISSUE #554: Add support for eth_call state-overrides
Geth has support for [performing state-overrides](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set) when doing an `eth_call`. It would be great if Ganache could support this same behavior.

domob1812:
This PR adds support for overriding the contract code and storage and the account balance and nonce for simulations with `eth_call`.  This feature is similarly [supported by geth](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set).

The `eth_call` RPC method now accepts a third override parameter which is a map from an address to the desired state of that address. The state object has the following keys, all of which are optional:

>`balance`: `QUANTITY` - The balance to set for the account before executing the call.
`nonce`: `QUANTITY` - The nonce to set for the account before executing the call.
`code`: `DATA` - The EVM bytecode to set for the account before executing the call.
`state`\*: `OBJECT` - Key-value mapping of storage slot to value that will *clear all* slots and then *override individual* slots in the account storage before executing the call.
`stateDiff`\*: `OBJECT` - Key-value mapping of storage slot to value that will override *individual* slots in the account storage before executing the call.
_*Note - `state` and `stateDiff` fields are mutually exclusive._

So, an example override object using all fields overriding multiple addresses could look like the following:
```JSON
{
  "0xd9c9cd5f6779558b6e0ed4e6acf6b1947e7fa1f3": { 
    "nonce": "0xa", 
    "balance": "0xffff", 
    "state": { 
      "0000000000000000000000000000000000000000000000000000000000000001": "0xbaddad42baddad42baddad42baddad42baddad42baddad42baddad42baddad42"
    } 
  },
  "0xebe8efa441b9302a0d7eaecc277c09d20d684540": { 
    "code": "0x123456", 
    "stateDiff": {
      "0000000000000000000000000000000000000000000000000000000000000002": "0xbaddad42baddad42baddad42baddad42baddad42baddad42baddad42baddad42"
    } 
  }
}
```

Like all state changes made with `eth_call`, the changes are ephemeral and only last for that specific run of `eth_call`. We'd like to give a huge thank you to @domob1812  and @dekz for their initial contributions to this one, you both were a huge help.

The PR introducing this feature uses @dekz's #905 as a starting point, and fixes #554 and #2934.


domob1812:
Note that geth also supports overriding storage slots, which this PR does not yet implement (because I personally don't need it and also #905 did not implement it).

I think this can be added as a follow-up when needed.  But if you prefer to include support with this PR, let me know and I can try to add it.

MicaiahReid:
@domob1812 Thank you for the PR! We are reviewing and adding support for overriding storage slots while we're at it. This should hopefully be available in the next release!

domob1812:
Ok cool, thanks a lot!  If you need any help from my side, let me know.  I can also work on the storage slots, although I guess you know the involved code much better than me (of course).

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -73,6 +73,7 @@ import {
 } from "./provider-events";
 
 import mcl from "mcl-wasm";
+import { keccak256 } from "ethereumjs-util";
```
davidmurdoch: Use keccak from `import { keccak } from "@ganache/utils";` as it's slightly more efficient than keccak256. Let's also move the import up near the others, the `mcl` import as some additional loading code that I'd like to keep near by (see the `mclPromise` code below).

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,57 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+          } as any;
+        }
+        if (code) {
+          const codeBuffer = Data.from(code).toBuffer();
+          // The ethereumjs-vm StateManager does not allow to set empty code,
+          // therefore we will manually set the code hash when "clearing" the contract code
+          const codeHash =
+            codeBuffer.length > 0 ? keccak256(codeBuffer) : KECCAK256_NULL;
+          account.codeHash = codeHash;
+          this.trie.db.put(codeHash, codeBuffer);
```
davidmurdoch: should this be awaited?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,57 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     return vm;
   };
 
+  applySimulationOverrides = async (vm: VM, overrides: CallOverrides) => {
+    const stateManager = vm.stateManager;
+    for (const [
+      address,
+      { balance, nonce, code, state, stateDiff }
+    ] of Object.entries(overrides)) {
```
davidmurdoch: `Object.entries` iterates over the entire collection, so we are effectively iterating twice and doing some extra work in the process.

Here is a hacked together pseudo-benchmark demonstrating the difference in speed:

```javascript
(function() {
  const object = {
    key1: 123,
    key2: 123,
    key3: 123,
    key4: 123,
    key5: 123,
    key6: 123,
    key7: 123,
    key8: 123,
    key9: 123,
    key10: 123,
    key11: 123,
    key12: 123,
  }

  console.time("entries");
  let lengtha = 0;
  for (let i = 0; i < 1e3; i++) {
    const resulta = [];
    for (const [key, value] of Object.entries(object)) {
      resulta.push(key + value); // do some work
    }
    lengtha += resulta.length;
  }
  console.timeEnd("entries");
  console.log("lengtha", lengtha);

  console.time("for in");
  let lengthb = 0;
  for (let i = 0; i < 1e3; i++) {
    const resultb = [];
    for (const key in object) {
      if (object.hasOwnProperty(key)) {
        const value = object[key];
        resultb.push(key + value); // do some work
      }
    }
    lengthb += resultb.length;
  }
  console.timeEnd("for in");
  console.log("lengthb", lengthb);
}());
```

Let's investigate if it's be worth changing this to use `for ... in` instead of `for ... of Object.entries`.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2633,19 +2642,30 @@ export default class EthereumApi implements Api {
...
     const blocks = blockchain.blocks;
     const parentBlock = await blocks.get(blockNumber);
     const parentHeader = parentBlock.header;
     const options = this.#options;
 
+    // override state and stateDiff fields are mutually exclusive
+    if (overrides && overrides.state && overrides.stateDiff) {
+      throw new Error("both state and stateDiff overrides specified");
+    }
+
```
davidmurdoch: I'm thinking it may make sense to move this check to the `applySimulationOverrides` function, which might allow us to condense some logic. We are checking if `overrides.state` and `overrides.stateDiff` exist in both locations already. I'm okay leaving it here if anyone else feels strongly about it.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,57 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+      }
+      // group together overrides that update storage
+      if (state || stateDiff) {
+        let storageEntries = [];
+        // override.state clears all storage and sets just the specified slots
+        if (state) {
+          await stateManager.clearContractStorage(vmAddr);
+          storageEntries = Object.entries(state);
+        } else {
+          storageEntries = Object.entries(stateDiff);
```
davidmurdoch: Same notes about `Object.entries` from above applies here as well.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1050,6 +1123,10 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         if (to) stateManager.addWarmedAddress(to.buf);
       }
 
+      // If there are any overrides requested for eth_call, apply
+      // them now before running the simulation.
+      await this.applySimulationOverrides(vm, overrides);
```
davidmurdoch: I think we need to guard this with an `if (overrides) {` check, as `null` is a possible value here, as default parameters (like `overrides= {}`) only kick in for `undefined`. Or, a better solution may be to check that `typeof overrides === "object"`, and throw if overrides in given but is not a valid `overrides` object. I'm not sure what to do with `null` in this case. Note: I _don't_ think we need to check for only expected _properties_ on the object, as we already allow "rogue" properties almost everywhere else (like: `eth_sendTransaction`). 🤔

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: It should be noted that we have a decently typed `hasOwn` helper function as part of `@ganache/utils` that does the `object.hasOwnProperty(key)` check see above in a "more correct" way; if we switch to `for in` we should use it :-)

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: Another note: should we add a test passing in `null`, and maybe some other incorrect overrides `values`?

davidmurdoch (changes requested):
Looking really great! Exciting to finally see this feature land!

I've requested some changes. Let me know if either of you (@domob1812 @MicaiahReid ) have questions or disagree with any of my comments!

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Done!

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Ya

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Cool stuff! Thanks for the lesson. I'll implement `for ... in`

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Fixed

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: I've added tests that pass junk values for the overrides.

Adding an `if (overrides) {` around this line seems superfluous, since `null`, `undefined`, and `{}` all won't make it inside the first `for(const address in overrides) {` loop of the `applySimulationOverrides` function. Do you see any potential problems with that?

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: My only reason for wanting it here was that a lot of work is done to get to `applySimulationOverrides`, but we already know we're going to throw early on. Also can we really condense that logic much? Since this is checking if `state && stateDiff` exist while the other checks if `state || stateDiff` exist.

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: I think it a previous version `Object.keys(overrides)` was called, so it would have thrown without the additional check.

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: Oh wait, I don't think checking `overrides.state` and `overrides.stateDiff` makes sense here at all, because Overrides has address keys (`"0x12...34"`), not properties like `state`, `stateDiff`, etc.

This check should probably be done within the `for (const address in overrides) {`, right?

In any case I was thinking we could something like:

```javascript
function applySimulationOverrides(){
  for (const address in overrides) {
    // ...
    if (state) {
      if (stateDiff) throw ...; // <-----------
      // ...
      applyNewState(...);
    } else if (stateDiff) {
      if (state) throw ...; // <-----------
      // ...
      applyNewState(...);
    }
  }
}
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,73 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+          // and slot, but not a value we can actually set in the storage. if
+          // so, we don't want to set the storage, and we also don't want to
+          // clear it out
+          let hasValidValue = false;
+          for (const slot in state) {
+            const value = state[slot];
+            if (!hasOwn(state, slot) || value == null || value === "") continue;
+            hasValidValue = true;
+          }
+          if (!hasValidValue) return;
```
davidmurdoch: is `return` correct here, or do we want to `continue`?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,73 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+          // override.state clears all storage and sets just the specified slots
+          await stateManager.clearContractStorage(vmAddr);
+          newState = state;
+        } else {
+          newState = stateDiff;
+        }
+        for (const slot in newState) {
+          const value = newState[slot];
+          // don't set storage for invalid values
+          if (!hasOwn(newState, slot) || value == null || value === "")
```
davidmurdoch: I think we only need to do this check for `stateDiff`, as the `state` branch already does this. Should be just remove the check from the `state` section? Alternatively, since we're already looping over all data in `state` in order to create the `newState` should we `putContractStorage` then, instead of here?

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: You're right, I'm wrong. Fixed.

MicaiahReid:
I've done a lot of testing in Geth to see how their `eth_call` overrides work. Here are the results:
<google-sheets-html-origin>

Override Type | Tested Value | Geth Response
-- | -- | --
Balance | null | Balance remains
Balance | "" | Balance 0
Balance | "0x", undefined, "123", {}, 123, -9 | Rejected
Code | null | Code remains
Code | "" | Clears code
Code | "0x" | Clears code
Code | undefined, "123", {}, 123, -9 | Rejected
StateDiff | setting value to:null, "0xbaddad42", "", "0x", undefined, "123", {}, 123, -9 | Rejected
StateDiff | setting slot to:null, "0xbaddad42", "", "0x", undefined, "123", {}, 123, -9 | Rejected
State | setting value to:null, "0xbaddad42", "", "0x", undefined, "123", {}, 123, -9 | Rejected
State | setting slot to:null, "0xbaddad42", "", "0x", undefined, "123", {}, 123, -9 | Rejected
Nonce | null | Accepted - makes same contract hash as not overriding nonce
Nonce | "" | Accepted - sets nonce to "0x0"
Nonce | "0x", undefined, "123", {}, 123, -9 | Rejected

I've made updates so that Ganache behaves the same way, but there's some very specific data validation that we're now doing in the `applySimulationOverrides` that makes me wonder if we should have some more general rpc data types that align with Geth's.

Also, we should probably consider matching our `evm_setAccountNonce`, `evm_setAccountBalance`, `evm_setAccountCode`, and `evm_setAccountStorageAt` implementations to have the same data validation as this.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: I've reworked this section to be more efficient.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2633,12 +2642,18 @@ export default class EthereumApi implements Api {
...
-  @assertArgLength(1, 2)
-  async eth_call(transaction: any, blockNumber: QUANTITY | Tag = Tag.latest) {
+  @assertArgLength(1, 3)
+  async eth_call(
+    transaction: any,
+    blockNumber: QUANTITY | Tag = Tag.latest,
+    overrides: CallOverrides = {}
+  ) {
     const blockchain = this.#blockchain;
     const common = this.#blockchain.common;
```
```suggestion
    const common = blockchain.common;
```

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2613,9 +2613,18 @@ export default class EthereumApi implements Api {
    * * `value`: `QUANTITY` (optional) - Integer of the value in wei.
    * * `data`: `DATA` (optional) - Hash of the method signature and the ABI encoded parameters.
    *
+   * State Override Object - An address to state mapping used to override the vm state for a
+   * given address. Each address mps to:
```
```suggestion
   * State Override object - An address-to-state mapping, where each entry specifies some
   * state to be ephemerally overridden prior to executing the call. Each address maps to an
   * object containing:
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,115 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     return vm;
   };
 
+  validateStorageOverride = (slot: string, value: string) => {
+    // assume Quantity will handle other types, these are just special string cases
+    if (typeof slot === "string" && slot !== "" && slot.indexOf("0x") === 0) {
+      // assume we're starting with 0x cause Quantity will verify if not
+      const noPrefix = slot.slice(2);
+      if (noPrefix.length != 64) {
```
davidmurdoch: Can we can skip the `slice(2)` and check if it is `slot.length !== 66` instead?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,115 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+      );
+    }
+    // assume Quantity will handle other types, these are just special string cases
+    if (
+      typeof value === "string" &&
+      value !== "" &&
+      value.indexOf("0x") === 0
+    ) {
+      const noPrefix = value.slice(2);
+      if (noPrefix.length != 64) {
```
davidmurdoch: Can we can skip the slice(2) and check if it is slot.length !== 66 instead?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,115 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     return vm;
   };
 
+  validateStorageOverride = (slot: string, value: string) => {
```
davidmurdoch: `validateStorageOverride(slot: string, value: string, fieldName: string) => {` would let us be more specific in our error message, as we wouldn't need to say `"State/StateDiff override ..."`, but could instead do ``` `${fieldName} override ...` ```.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,115 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+      const vmAddr = { buf: Address.from(address).toBuffer() } as any;
+      // group together overrides that update the account
+      if (nonce != null || balance != null || code != null) {
+        const account = await stateManager.getAccount(vmAddr);
+
+        if (nonce != null) {
+          account.nonce = {
+            toArrayLike: () =>
+              // geth treats empty strings as "0x0" nonce for overrides
+              Quantity.from(nonce === "" ? "0x0" : nonce).toBuffer()
```
```suggestion
              nonce === "" ? BUFFER_ZERO : Quantity.from(nonce).toBuffer()
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,115 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+          } as any;
+        }
+        if (balance != null) {
+          account.balance = {
+            toArrayLike: () =>
+              // geth treats empty strings as "0x0" balance for overrides
+              Quantity.from(balance === "" ? "0x0" : balance).toBuffer()
```
```suggestion
              nonce === "" ? BUFFER_ZERO : Quantity.from(balance).toBuffer()
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,115 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+          } as any;
+        }
+        if (code != null) {
+          // geth treats empty strings as "0x" code for overrides
+          const codeBuffer = Data.from(code === "" ? "0x" : code).toBuffer();
```
```suggestion
          const codeBuffer = code === "" ? BUFFER_EMPTY : Data.from(code).toBuffer();
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,115 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+          // The ethereumjs-vm StateManager does not allow to set empty code,
+          // therefore we will manually set the code hash when "clearing" the contract code
+          const codeHash =
+            codeBuffer.length > 0 ? keccak(codeBuffer) : KECCAK256_NULL;
+          account.codeHash = codeHash;
+          await this.trie.db.put(codeHash, codeBuffer);
```
davidmurdoch: This puts the `codeHash: codeBuffer` into the global trie. We need to set it only for cloned (and checkpointed) trie, which is available in the `stateTrie` variable in the `simulateTransaction` function that calls this method.

If i'm not \wrong about the above then we don't yet have exhaustive tests that check if `eth_call` has side effects and we should add those.

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: I think i'd like to see `validateStorageOverride` and `applySimulationOverrides` moved to a new file (call.s, or `simulation.ts`? I dunno), as blockchain.ts is getting big (we should probably move other things to other files one of these days as well).

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -647,6 +668,115 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+              await stateManager.clearContractStorage(vmAddr);
+              clearedState = true;
+            }
+            const slotBuf = Quantity.from(slot).toBuffer();
+            const valueBuf = Quantity.from(value).toBuffer();
+            await stateManager.putContractStorage(vmAddr, slotBuf, valueBuf);
+          }
+        } else {
+          for (const slot in stateDiff) {
+            // don't set storage for invalid values
```
davidmurdoch: I think this comment outdated. remove?

```suggestion
```

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2633,12 +2642,18 @@ export default class EthereumApi implements Api {
    * const simpleSol = "0x6080604052600560008190555060858060196000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c80633fa4f24514602d575b600080fd5b60336049565b6040518082815260200191505060405180910390f35b6000548156fea26469706673582212200897f7766689bf7a145227297912838b19bcad29039258a293be78e3bf58e20264736f6c63430007040033";
    * const [from] = await provider.request({ method: "eth_accounts", params: [] });
    * const txObj = { from, gas: "0x5b8d80", gasPrice: "0x1dfd14000", value:"0x0", data: simpleSol };
-   * const result = await provider.request({ method: "eth_call", params: [txObj, "latest"] });
+   * const slot = "0x0000000000000000000000000000000000000000000000000000000000000005"
+   * const override = { [from]: { balance: "0x3e8", "nonce: "0x5", code: "0xbaddad42", stateDiff: { [slot]: "0xbaddad42"}}}
+   * const result = await provider.request({ method: "eth_call", params: [txObj, "latest", override] });
```
```suggestion
   * const overrides = { [from]: { balance: "0x3e8", "nonce: "0x5", code: "0xbaddad42", stateDiff: { [slot]: "0xbaddad42"}}}
   * const result = await provider.request({ method: "eth_call", params: [txObj, "latest", overrides] });
```

davidmurdoch (approved):

davidmurdoch (changes requested):
meant to request changes on the last review. so i'm reeviewing again.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Fixed.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Fixed.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: I've moved the `validateStorageOverride` and `applySimulationOverrides` functions to the preexisting `/helpers/run-call.ts` file. I've also added your suggestion to specify the `fieldName` in the storage override validation and updated the tests accordingly.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: I've fixed this to use the `stateTrie`.

I've also created issue #2934 to capture the need for `eth_call` tests.

MicaiahReid (approved):

davidmurdoch (approved):
<3

PR #2572, opened 2022-03-10
Fix not working example of get "eth_accounts"

MatthieuScarset:
Cannot use `await` inside the server function when using Ganache as a JSON-RPC web server.

### DIFF
--- a/src/packages/ganache/README.md
+++ b/src/packages/ganache/README.md
@@ -163,10 +163,11 @@ server.listen(PORT, err => {
 
   console.log(`ganache listening on port ${PORT}...`);
   const provider = server.provider;
-  const accounts = await provider.request({
+  provider.request({
     method: "eth_accounts",
     params: []
-  });
+  }).then(accounts => console.log(accounts));
+
 });
 ```
 


src/packages/ganache/README.md
```diff
@@ -163,10 +163,11 @@ server.listen(PORT, err => {
 
   console.log(`ganache listening on port ${PORT}...`);
   const provider = server.provider;
-  const accounts = await provider.request({
+  provider.request({
     method: "eth_accounts",
     params: []
-  });
+  }).then(accounts => console.log(accounts));
+
```
davidmurdoch, line 170: This should probably be:

```javascript
server.listen(PORT, async (err) => {
  if (err) throw err;
  console.log(`ganache listening on port ${PORT}...`);
  const provider = server.provider;
  const accounts = await provider.request({
    method: "eth_accounts",
    params: []
  });
```

davidmurdoch (changes requested):

davidmurdoch:
Thanks for bringing this to our attention. I'm closing in favor of: https://github.com/trufflesuite/ganache/pull/2789

PR #2602, opened 2022-03-14
fix: ensure impersonated "large" accounts use compliant (fake) private keys

### LINKED ISSUE #2586: Transaction from unlocked account fails with error: "The nonce generation function failed, or the private key was invalid":
Hi, I'm unlocking this address: `0xffffFFFfFFffffffffffffffFfFFFfffFFFfFFfE` on a mumbai fork to try to emulate polygon system account.
When I send the following transaction I get in the ganache v7.0.0 container log I get the error `"The nonce generation function failed, or the private key was invalid"`:
```
mumbai_1 | > eth_sendTransaction: [ 
mumbai_1 | > { 
mumbai_1 | > "gas": "0xcd3f", 
mumbai_1 | > "from": "0xfffffffffffffffffffffffffffffffffffffffe", 
mumbai_1 | > "to": "0x0000000000000000000000000000000000001001", 
mumbai_1 | > "data": "...."
```
Even a transfer tx fails with the same error: 
```
mumbai_1   |    >  eth_sendTransaction: [
mumbai_1   |    >   {
mumbai_1   |    >     "from": "0xfffffffffffffffffffffffffffffffffffffffe",
mumbai_1   |    >     "to": "0x5f890c9522dce5670d741d4277bfcc2d9ca8af02",
mumbai_1   |    >     "value": "0x12",
mumbai_1   |    >     "gasPrice": "0x77359400"
mumbai_1   |    >   }
```
Is seems this error happens when trying to ecsign with the fake private key.

Any idea ?

davidmurdoch:
Transactions from "impersonated" accounts are still "signed" by Ganache, but the private key is faked. It was previously possible for the fake private key we generate for some accounts to cause an error when a transaction is eventually signed.

Fixes https://github.com/trufflesuite/ganache/issues/2586

### DIFF
--- a/src/chains/ethereum/ethereum/src/wallet.ts
+++ b/src/chains/ethereum/ethereum/src/wallet.ts
@@ -23,6 +23,8 @@ import { writeFileSync } from "fs";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
 import { Address } from "@ganache/ethereum-address";
 
+const TWELVE_255s = Buffer.allocUnsafe(12).fill(255);
+
 //#region Constants
 const SCRYPT_PARAMS = {
   dklen: 32,
@@ -626,7 +628,7 @@ export default class Wallet {
 
   public createFakePrivateKey(address: string) {
     let fakePrivateKey: Buffer;
-    const addressBuf = Data.from(address).toBuffer();
+    const addressBuf = Address.from(address).toBuffer();
     if (addressBuf.equals(ACCOUNT_ZERO)) {
       // allow signing with the 0x0 address...
       // always sign with the same fake key, a 31 `0`s followed by a single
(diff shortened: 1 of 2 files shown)


davidmurdoch:
WIP! I meant to make this a draft. I need to separate these two fixes out into separate PR :-)

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -636,7 +638,40 @@ export default class Wallet {
...
+      // then there is a chance that the fakePrivateKey will be too large.
+      // `Buffer.compare` is a native method in V8, so it should be pretty fast.
+      // example: the address `0xffffffffffff{anything can go here}` generates a
+      // bad fakePrivateKey but, `0xfffffffffffe{anything can go here}` never
+      // will. There are obviously many chances for a false positive, but the
+      // next condition in the `while` loop will catch those.
+      if (first12.compare(TWELVE_255s) === 0) {
+        // keccak returns a 32 byte hash of the input data, which is the exact
+        // length we need for a private key.
+        // note: if keccak can return it's own input as it's output, then this
```
```suggestion
        // note: if keccak can return its own input as its output, then this
```

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -636,7 +638,40 @@ export default class Wallet {
...
+      // next condition in the `while` loop will catch those.
+      if (first12.compare(TWELVE_255s) === 0) {
+        // keccak returns a 32 byte hash of the input data, which is the exact
+        // length we need for a private key.
+        // note: if keccak can return its own input as its output, then this
+        // loops forever. The chances of this happening are impossibly low, so
+        // it's not worth the effort to check, but it would be interesting if
+        // someone reported an issue that can cause this for a specific
+        // address!
+        fakePrivateKey = keccak(fakePrivateKey);
```
```suggestion
        while (
          BigInt(`0x${fakePrivateKey.toString("hex")}`) >=
          SECP256K1_MAX_PRIVATE_KEY
        ) {
          // keccak returns a 32 byte hash of the input data, which is the exact
          // length we need for a private key.
          // note: if keccak can return its own input as its output, then this
          // loops forever. The chances of this happening are impossibly low, so
          // it's not worth the effort to check, but it would be interesting if
          // someone reported an issue that can cause this for a specific
          // address!
          fakePrivateKey = keccak(fakePrivateKey);
        }
```

davidmurdoch:
closing in favor of #2944

PR #2610, opened 2022-03-16
Fixes #2580 - parse port to number type if provided in a different type

### LINKED ISSUE #2580: Incoherent "Address already in use 127.0.0.1:XXXX" issue when using Ganache programmatically (JS)
I'm having an issue with Ganache when trying to use it _programmatically_ with a JavaScript script.

Basically **it doesn't want to start** saying **port is already in use** - whereas it is not :shrug:  

It **does start** with the same port if I run the CLI directly.

How to reproduce: 
- create a script for Ganache (see code below)
- run `node ganache.js` => does not work, results in  "Address already in use 127.0.0.1:XXXX", no matter the port
- run `./node_modules/.bin/ganache -p XXXX` => it works 

Here's the content of the script `ganache.js` - based on the example provided in README:
```JavaScript
require("dotenv").config();
const ganache = require("ganache");
const Web3 = require('web3');

const options = {
  account_keys_path: "ganache.json",
  wallet: {
    accounts: [
      {
        secretKey: `${process.env.PRIVATE_KEY}`,
        balance: Web3.utils.toHex(`${process.env.INITIAL_BALANCE}`)
      }
    ],
  }
};

const port = process.env.PORT || 8545;
const server = ganache.server(options);
server.listen(port, async err => {
  if (err) throw err;
  console.log(`ganache listening on port ${port}...`);
  const provider = server.provider;

  await provider.request({
    method: "eth_accounts",
    params: []
  }).then(accounts => console.log(accounts));

});
```

jeffsmale90:
I didn't end up changing the listen() signature, I agree that having a specific signature makes sense.

I did check the http.server.listen() behaviour, AFAIK our implementation now matches it wrt providing port number as string.

### DIFF
--- a/src/packages/core/src/server.ts
+++ b/src/packages/core/src/server.ts
@@ -185,6 +185,17 @@ export class Server<
       host = null;
     }
     const callbackIsFunction = typeof callback === "function";
+
+    // Method signature specifies port: number, but we parse a non-number if provided
+    // Matches the behaviour of http.server.listen https://www.w3schools.com/nodejs/met_server_listen.asp
+    if (isNaN(port)) {
+      const err = new Error(`Provided port is not a valid value: ${port}.`);
+      return callbackIsFunction
+        ? process.nextTick(callback!, err)
+        : Promise.reject(err);
+    }
+    const portNumber = typeof port === "number" ? port : parseInt(port);
+
     const status = this.#status;
     if (status === ServerStatus.closing) {
       // if closing
@@ -195,7 +206,7 @@ export class Server<
     } else if ((status & ServerStatus.openingOrOpen) !== 0) {
       // if opening or open
       const err = new Error(
-        `Server is already open, or is opening, on port: ${port}.`
+        `Server is already open, or is opening, on port: ${portNumber}.`
       );
       return callbackIsFunction
         ? process.nextTick(callback!, err)
@@ -214,11 +225,11 @@ export class Server<
           host
             ? (this.#app as any).listen(
               host,
-              port,
+              portNumber,
               LIBUS_LISTEN_EXCLUSIVE_PORT,
               resolve
             )
-            : this.#app.listen(port, LIBUS_LISTEN_EXCLUSIVE_PORT, resolve);
+            : this.#app.listen(portNumber, LIBUS_LISTEN_EXCLUSIVE_PORT, resolve);
         }
       ).then(listenSocket => {
         if (listenSocket) {
@@ -228,7 +239,7 @@ export class Server<
           this.#status = ServerStatus.closed;
           const err = new Error(
             `listen EADDRINUSE: address already in use ${host || DEFAULT_HOST
-            }:${port}.`
+            }:${portNumber}.`
           );
           throw err;
         }
--- a/src/packages/core/src/server.ts
+++ b/src/packages/core/src/server.ts
@@ -186,15 +186,21 @@ export class Server<
     }
     const callbackIsFunction = typeof callback === "function";
 
-    // Method signature specifies port: number, but we parse a non-number if provided
-    // Matches the behaviour of http.server.listen https://www.w3schools.com/nodejs/met_server_listen.asp
-    if (isNaN(port)) {
-      const err = new Error(`Provided port is not a valid value: ${port}.`);
+    // Method signature specifies port: number, but we parse a string if provided
+    // inspiration taken from nodejs internal port validator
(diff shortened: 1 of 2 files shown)


src/packages/core/src/server.ts
```diff
@@ -185,6 +185,17 @@ export class Server<
       host = null;
     }
     const callbackIsFunction = typeof callback === "function";
+
+    // Method signature specifies port: number, but we parse a non-number if provided
+    // Matches the behaviour of http.server.listen https://www.w3schools.com/nodejs/met_server_listen.asp
+    if (isNaN(port)) {
```
davidmurdoch: So `isNaN` is really weird... things like `isNaN(true)` return `false` 😱 (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#examples for more examples)

We may want to take inspiration from what Node.js does internally:

https://github.com/nodejs/node/blob/8c4b8b201ada6b76d5306c9c7f352e45087fb4a9/lib/internal/validators.js#L208-L219

The reason we don't get this error organically is because we don't use the Node.js `server` module; we use https://github.com/uNetworking/uWebSockets.js.

Something to note about node.js's validation code is that the `allowZero` flag is interesting, as a port of `0` can be used to instruct Node.js to pick the port automatically. `uWebSockets` [supports this](https://github.com/uNetworking/uWebSockets.js/blob/9b3efbdfa315f4ec559fc8a6afe78083007bb7df/examples/AutomaticPortSelection.js), but Ganache doesn't currently provide a way of returning the automatically selected port to the user.

Oh! Something to note, but that is not really directly relevant to this PR: uWebsockets.js is a "native" library written in C++ and is currently pre-compiled for Node.js v14-17, for Node.js 12 (and some niche/rare CPU architectures) we fallback to a pure JavaScript version: https://github.com/trufflesuite/uws-js-unofficial/tree/master/src/fallback.

Anyway, fully supporting automatic port selection by specifying port `0` is definitely out of scope of this PR, but we should probably make a new "enhancement" GitHub Issue for this so we can follow up on it later.

src/packages/core/tests/server.test.ts
```diff
@@ -299,6 +299,47 @@ describe("server", () => {
...
+
+      try {
+        const req = request.post("http://localhost:5432");
+        await req.send(jsonRpcJson);
+      } finally {
+        await teardown();
+      }
+    });
+
+    it("accepts port as string type (hexidecimal)", async () => {
```
davidmurdoch: hexidecimal  -> hexadecimal

src/packages/core/tests/server.test.ts
```diff
@@ -299,6 +299,47 @@ describe("server", () => {
...
+      s = Ganache.server(defaultOptions);
+      try {
+        await assert.rejects(s.listen(<number><unknown>portString), {
+          message: `Provided port is not a valid value: ${portString}.`
+        });
+      } finally {
+        await teardown();
+      }
+    });
+
```
davidmurdoch, line 342: Can you add some tests for other "weird" values? Like `null`, `""`, `true`/`false`, negative values, `Infinity`, etc?

davidmurdoch (changes requested):
Looks great!

I've left some comments requesting some changes. Let me know if you need additional explanations on anything I've suggested here.

src/packages/core/src/server.ts
```diff
@@ -185,6 +185,23 @@ export class Server<
       host = null;
     }
     const callbackIsFunction = typeof callback === "function";
+
+    // Method signature specifies port: number, but we parse a string if provided
+    // inspiration taken from nodejs internal port validator
+    // https://github.com/nodejs/node/blob/8c4b8b201ada6b76d5306c9c7f352e45087fb4a9/lib/internal/validators.js#L208-L219
+    if ((typeof port !== 'number' && typeof port !== 'string') ||
+      (typeof port === 'string' && String.prototype.trim.apply(port).length === 0) ||
```
davidmurdoch: We don't need to do `String.prototype.trim.apply(port)` here. Node.js references their internal version just in case userland code has modified `trim`.  We're OK using "tampered" built-ins because it's close to impossible for us to avoid it (and sometimes user's may want to changes these built-ins in order to affect the code they are running).

jeffsmale90: Fair enough - I was thinking something about avoiding calling .trim() on a non-string type, but we check the type right before this call.

davidmurdoch (approved):

PR #2634, opened 2022-03-18
perf: speed up large transaction traces by buffering fragmented send

robmcl4:
`debug_traceTransaction` is _extremely_ slow when using websocket fragmented send. This is largely because the generator returned by `bufferify` emits extremely small size buffers, which are each sent as a websocket fragment. This adds a large amount of unnecessary overhead.

This change introduces a buffer to reduce the number of websocket fragments.

For a quick performance measurement I trace a transaction with 16,600 structLogs and measure the total round-trip time as observed by a web3py client. To do this measurement I also:

1. Decreased the [`BUFFERIFY_THRESHOLD`](https://github.com/trufflesuite/ganache/blob/386771d84a9985f6d4b61b262f2be3cda896162e/src/chains/ethereum/ethereum/src/connector.ts#L49) to 10,000 (otherwise I receive an exception because of string too long)
2. Increased [`idleTimeout`](https://github.com/trufflesuite/ganache/blob/386771d84a9985f6d4b61b262f2be3cda896162e/src/packages/core/src/servers/ws-server.ts#L51) to 10 minutes.
3. Increased [`MAX_PAYLOAD_SIZE`](https://github.com/trufflesuite/ganache/blob/386771d84a9985f6d4b61b262f2be3cda896162e/src/packages/core/src/servers/ws-server.ts#L35) to `1 * 1024 * 1024 * 1024`.

Without buffer: timeout at 10 minutes
With buffer: 8.283 seconds


davidmurdoch:
Oh! So awesome that you took this on! I'm excited to review this when I get back in the office. In the future we plan on making debug_traceTransaction output directly to the socket (avoiding the JSON step entirely), but that will be a while and will be fairly complex. So I definitely like this solution in the in interim!

robmcl4:
After testing a bit with some heavy traces I noticed that payloads were being corrupted. Some investigation showed that this was due to a sort of race-condition with the buffer: after flushing, we may write the next data portion into the buffer before the prior content was written to the network. This is addressed in commit [b25d327](https://github.com/trufflesuite/ganache/pull/2634/commits/b25d327d0d29eb8ae22e5ddde53d777a75866344), which allocates an entirely new buffer each time we flush content to the websocket.

davidmurdoch:
Awesome. Can you run `npx prettier --write ./src/packages/core/src/servers/ws-server.ts` and commit the changes? Also, you mentioned you had to change `BUFFERIFY_THRESHOLD` and `idleTimeout` but I don't see those changes in this PR; were those changes not necessary after all?

davidmurdoch:
@robmcl4 Would you like us to go ahead and finish this PR up ourselves? No problem either way!

robmcl4:
Ah apologies for dropping off .. yes if you could finish this up that would be great

davidmurdoch:
@robmcl4 you mentioned you had to change BUFFERIFY_THRESHOLD and idleTimeout but I don't see those changes in this PR; were those changes not necessary after all?

robmcl4:
Ah, so wrt. idleTimeout I had to increase that in order to have a 'true' understanding of how long it takes the non-buffered fragmented send to complete.

wrt. BUFFERIFY_THRESHOLD I did end up needing to lower it because sometimes the JSON-ified string exceeded max string length

Which is to say, I believe lowering BUFFERIFY_THRESHOLD is appropriate

src/packages/core/src/connector-loader.ts
```diff
@@ -17,9 +17,9 @@ const initialize = <T extends FlavorName = typeof DefaultFlavor>(
   let asyncRequestProcessing: boolean;
 
   if ("chain" in options && "asyncRequestProcessing" in options["chain"]) {
-    asyncRequestProcessing = options["chain"].asyncRequestProcessing;
+    asyncRequestProcessing = options.chain.asyncRequestProcessing;
```
davidmurdoch, line 20: Once upon a time typescript couldn't figure out the types here, now it can. This is a weird change to have in this PR and I can remove it if anyone objects.

src/packages/core/src/options/server-options.ts
```diff
@@ -1,5 +1,6 @@
 import { DefaultFlavor, FilecoinFlavorName } from "@ganache/flavors";
 import { Definitions } from "@ganache/options";
+import { kMaxLength } from "buffer";
```
davidmurdoch, line 3: A Buffer's maximum length allowed by node may differ between versions, architectures, and OSes, so we can use this value to get the maximum length allowed for single buffer instance.

src/packages/core/src/options/server-options.ts
```diff
@@ -42,6 +43,17 @@ export type ServerConfig = {
       type: string;
       hasDefault: true;
     };
+
+    /**
+     * For memory and performance reasons ganache may respond with chunked
+     * transfer-encoding over HTTP and fragmented send over WebSockets.
+     * This option allows you to control the approximate size of each chunk.
```
davidmurdoch, line 50: The reason it is the "approximate size":.

If a piece a data we are processing is too big to fit in the chunk we're been building we go ahead and send the chunk, even though it isn't yet full. If the piece of data we are processing is too big to fit in out chunk we send the whole piece of data as is. So the actual chunk size sent is rarely ever _exactly_ `chunkSize`.

src/packages/core/src/servers/http-server.ts
```diff
@@ -93,10 +94,57 @@ function sendResponse(
   });
 }
 
-export type HttpServerOptions = Pick<InternalOptions["server"], "rpcEndpoint">;
+function sendChunkedResponse(
```
davidmurdoch, line 97: I think the code comments are pretty decent here, but an explanation and walk-through over zoom may help in the review, because I still find it easy to get lost here. :-)

src/packages/core/tests/server.test.ts
```diff
@@ -1230,58 +1261,287 @@ describe("server", () => {
       });
     });
 
-    it("responds with transfer-encoding: chunked responses when bufferification is triggered", async () => {
-      // this test needs to set BUFFERIFY_THRESHOLD before starting the server
-      await teardown();
+    describe("bufferification", () => {
+      it("responds over websockets when bufferification is triggered", async () => {
```
davidmurdoch, line 1296: This is an existing test and I didn't change anything; the git diff is just going a bit crazy here.

davidmurdoch (commented):
I've reviewed my changes myself to add additional context for other reviewers.

src/packages/core/src/servers/utils/fragment-generator.ts
```diff
@@ -0,0 +1,42 @@
...
+  value: Buffer,
+  bufferSize: number
+) {
+  // Use a large buffer to reduce round-trips
+  let buf = Buffer.allocUnsafe(bufferSize);
+  let offset = 0;
+  let done = false;
+  do {
+    const length = value.byteLength;
+    if (offset > 0 && length + offset > bufferSize) {
```
davidmurdoch: Probably should add a clarifying comment here

src/packages/core/src/servers/http-server.ts
```diff
@@ -93,10 +94,57 @@ function sendResponse(
...
+  response: HttpResponse,
+  closeConnection: boolean,
+  statusCode: HttpResponseCodes,
+  contentType: RecognizedString | null,
+  data: Generator<Buffer, any, unknown>,
+  chunkSize: number,
+  writeHeaders: (response: HttpResponse) => void = noop
+) {
+  const { value: first } = data.next();
```
davidmurdoch: I think this line can be moved to `getFragmentGenerator` itself

src/packages/core/src/servers/http-server.ts
```diff
@@ -93,10 +94,57 @@ function sendResponse(
...
+  const fragments = getFragmentGenerator(data, first, chunkSize);
+  // get our first fragment
+  const { value: firstFragment } = fragments.next();
+  // check if there is any more fragments after this one
+  let { value: prevFragment, done } = fragments.next();
+  // if there are no more fragments send the "firstFragment" via `sendResponse`,
+  // as we don't need to chunk it.
+  if (done) {
+    sendResponse(
+      response,
```
davidmurdoch, line 115: what happens if `response` is `undefined` here?

src/packages/core/src/servers/http-server.ts
```diff
@@ -93,10 +94,57 @@ function sendResponse(
...
+  data: Generator<Buffer, any, unknown>,
+  chunkSize: number,
+  writeHeaders: (response: HttpResponse) => void = noop
+) {
+  const { value: first } = data.next();
+  const fragments = getFragmentGenerator(data, first, chunkSize);
+  // get our first fragment
+  const { value: firstFragment } = fragments.next();
+  // check if there is any more fragments after this one
+  let { value: prevFragment, done } = fragments.next();
```
davidmurdoch: might want to add a `.hasMore` method here that doesn't actually create the next fragment, but does determine if it _will_ has more. (maybe just make an issue for this)

src/packages/core/src/servers/ws-server.ts
```diff
@@ -20,6 +20,46 @@ type WebSocketCapableFlavorMap = {
...
+      // since we have at least two fragments send the first one now that it
+      // is "full"
+      ws.sendFirstFragment(firstFragment as RecognizedString, useBinary);
+      // at this point `lastFragment` is the next fragment that should be sent
+      // but it might also be our last fragment. If it is, we MUST use
+      // `sendLastFragment` to send it. So we iterate over all fragments,
+      // sending the _previous_ fragment while caching the current (next)
+      // fragment to be send in the next iteration, or via `sendLastFragment`
+      // when `nextFragment` is also the last one.
+      let nextFragment: Buffer;
```
davidmurdoch: this doesn't need to be in this scope. just put it in the for loop.

src/packages/core/src/connector-loader.ts
jeffsmale90: I approve - it's nice to tidy things up

src/packages/core/src/servers/utils/fragment-generator.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (offset > 0 && byteLength + offset > bufferSize) {
+      yield buf.subarray(0, offset);
+      // Reset the buffer. Since `uws` sends packets asynchronously,
+      // it is important that we allocate a new buffer for the next
+      // frame. This avoids overwriting data before it is sent. The
+      // reason we need to do this is likely because we do not yet
+      // handle backpressure. Part of handling backpressure will
+      // involve the drain event and only sending while
+      // `ws.getBufferedAmount() < ACCEPTABLE_BACKPRESSURE`.
+      // See https://github.com/trufflesuite/ganache/issues/FIX THIS
```
jeffsmale90:
```suggestion
      // See https://github.com/trufflesuite/ganache/issues/2790
```

src/packages/core/src/servers/utils/fragment-generator.ts
```diff
@@ -0,0 +1,43 @@
...
+    // Store value in buffer if it fits (but don't store it if it is the exact
+    // same size as bufferSize)
+    if (byteLength < bufferSize) {
+      // copy from value into buffer
+      if (buf === null) buf = Buffer.allocUnsafe(bufferSize);
+      value.copy(buf, offset, 0, byteLength);
+      offset += byteLength;
+    } else {
+      // Cannot fit this value in buffer, send it directly.
+      // Buffer has just been flushed so we do not need to worry about
```
jeffsmale90: It's not immediately obvious _why_ we can depend on the buffer having just been flushed.

It might be worthwhile linking back to line 11 in the comment and explaining that either `offset === 0` (buffer has not yet been written to) _or_ `byteLength + offset > bufferSize` causing the buffer to be flushed.

Alternatively, that might just make this comment unwieldy 🤷

src/packages/core/tests/server.test.ts
```diff
@@ -1257,58 +1261,287 @@ describe("server", () => {
...
+          sendFirstFragment: () =>
+            assert.fail("should not have used sendFirstFragment"),
+          sendLastFragment: () =>
+            assert.fail("should not have used sendLastFragment")
+        };
+        const dataGenerator = function* () {
+          for (const chunk of chunks) yield chunk;
+        };
+        sendFragmented(mockWebsocket, dataGenerator(), false, chunkSize);
+        assert.strictEqual(sendCallCount, 1, "send called too many times!");
```
jeffsmale90:
```suggestion
        assert.strictEqual(sendCallCount, 1, "send called incorrect number of times!");
```

Or not enough times?

src/packages/core/tests/server.test.ts
```diff
@@ -1257,58 +1261,287 @@ describe("server", () => {
...
+          // fits in a fragment
+          Buffer.from("world", "utf8")
+        ];
+        const dataGenerator = function* () {
+          for (const response of expectedResponses) yield response;
+        };
+        const receivedFragments = [];
+        const mockWebsocket: any = {
+          cork: (callback: () => void) => callback(),
+          sendFragment: (value: string) => receivedFragments.push(value),
```
jeffsmale90: Should we distinguish between `sendFragment` / `sendFirstFragment` / `sendLastFragment`?

ie if `sendFragment` were just called 3 times, then this test would incorrectly pass.

jeffsmale90 (changes requested):

src/packages/core/src/servers/ws-server.ts
```diff
@@ -19,6 +20,48 @@ type WebSocketCapableFlavorMap = {
...
+    let { value: maybeLastFragment, done } = fragments.next();
+    // if there are no more fragments send the "firstFragment" via `send`, as
+    // we don't need to chunk it.
+    if (done) {
+      ws.send(firstFragment as RecognizedString, useBinary);
+    } else {
+      // since we have at least two fragments send the first one now that it
+      // is "full"
+      ws.sendFirstFragment(firstFragment as RecognizedString, useBinary);
+      // at this point `maybeLastFragment` is the next fragment that should be
```
MicaiahReid, line 46: I could try to suggest corrections on this comment block but I think it would be easier if you went over it :laughing: I get the gist of what you're saying, but there definitely are some typos

davidmurdoch: Huh, I don't recall having a stroke whilst writing that comment. ¯\\_(ツ)_/¯

I've degibberished the comment so hopefully it makes sense now!

src/packages/core/src/servers/utils/fragment-generator.ts
davidmurdoch: Would this be better?

```
// Cannot fit this value in buffer, send it directly.
// Buffer has just been flushed (since the condition `byteLength + offset > bufferSize`
// will always be true here, which will have triggered the flush above) so
// we do not need to worry about out-of-order send.
```

src/packages/core/tests/server.test.ts
davidmurdoch: good catch!

src/packages/core/src/servers/utils/fragment-generator.ts
davidmurdoch: I added this here: https://github.com/trufflesuite/ganache/pull/2634/files#diff-0a3bee479fd46bab9ab4aa0beb7492951a25bce519c7196daf63a587e94294e8R35-R38

Let me know if it works or if it can be improved!

jeffsmale90: Love it ❤️ !

src/packages/core/tests/server.test.ts
```diff
@@ -1257,58 +1261,301 @@ describe("server", () => {
...
+
+        // send the data!
+        sendFragmented(mockWebsocket, dataGenerator(), false, chunkSize);
+
+        assertMockSocketResults(receivedFragments, expectedResponses);
+      });
+
+      it("handles small+small+large bufferification edge-case", () => {
+        // we should receive 2 fragments, the first of which is very large and
+        // must be sent by itself, the second of which is very tiny.
```
```suggestion
        // we should receive 2 fragments, the first of which is small and
        // the second of which is large and must be sent by itself.
```

MicaiahReid (changes requested):

jeffsmale90 (approved):

davidmurdoch (approved):

jeffsmale90 (approved):

MicaiahReid (approved):

jeffsmale90 (approved):

PR #2667, opened 2022-03-21
fix: Close http connections after http-server.close() has been called

### LINKED ISSUE #2185: Shutting down the Ganache Server v7.0.0
Platform: macOS Big Sur 11.6
Ganache Version: v7.0.0
Node: v14.17.0 
npm: v6.14.13

Description:

Server does not shut down with `Ctrl` + `C`

Steps to reproduce:

1. Run a test build of [MetaMask](https://github.com/MetaMask/metamask-extension#building-locally) by running `yarn build:test`
2. In a separate terminal start Ganache by running `ganache`
3. Press `Ctrl` + `C` to shutdown the server

Expected result:
Ganache Server is shut down

Actual result:
Ganache Server continues running after receiving the shutdown signal

Notes:
[Video attachment](https://drive.google.com/file/d/1uHM4SR7pvEF0xnKFLCe1GCJTlhWboGpW/view?usp=sharing)

Tested the following versions in the screen capture above.

Ganache CLI v6.12.2 (ganache-core: 2.13.2)

- Server shuts down immediately 

ganache v7.0.0 (@ganache/cli: 0.1.1, @ganache/core: 0.1.1)

- Server continues running

jeffsmale90:
fixes #2185


src/packages/core/src/server.ts
```diff
@@ -185,6 +185,17 @@ export class Server<
...
+    // Method signature specifies port: number, but we parse a non-number if provided
+    // Matches the behaviour of http.server.listen https://www.w3schools.com/nodejs/met_server_listen.asp
+    if (isNaN(port)) {
+      const err = new Error(`Provided port is not a valid value: ${port}.`);
+      return callbackIsFunction
+        ? process.nextTick(callback!, err)
+        : Promise.reject(err);
+    }
+    const portNumber = typeof(port) === "number" ? port : parseInt(port);
+
```
davidmurdoch: You probably want to do an interactive rebase in order to drop this commit. See https://stackoverflow.com/a/40926981 for a way to do this.

src/packages/core/tests/server.test.ts
```diff
@@ -350,6 +392,68 @@ describe("server", () => {
...
+    it("closes even if a persistent http connection is open", async () => {
+      const agent = new http.Agent({
+        keepAlive: true
+      });
+
+      await setup();
+      const jsonRpcJson: any = {
+        jsonrpc: "2.0",
+        id: "1",
+        method: "eth_block_number",
```
davidmurdoch: I think this should be `"eth_blockNumber"`

jeffsmale90: You are absolutely right! Fixed.

src/packages/core/src/servers/http-server.ts
```diff
@@ -237,13 +243,18 @@ export default class HttpServer {
...
+
   public close() {
-    // currently a no op.
+    // flags the server as closing, indicating the connection should be closed with subsequent responses
+    // as there is no way presently to close existing connections outside of the request/response context
+    // see discussion: https://github.com/uNetworking/uWebSockets.js/issues/663#issuecomment-1026283415
+    this.#isClosing = true;
   }
-}
+}
```
davidmurdoch: Let's keep the newline at the end of files. Some devs like to use `cat` to view files, and the newline makes that process easier (so I've been told). :-)

```suggestion
}

```

src/packages/core/tests/server.test.ts
```diff
@@ -62,6 +62,49 @@ describe("server", () => {
...
+        .on("end", () => resolve({ status: 200, body: JSON.parse(data) }));
+    });
+    req.on("error", (err) => reject(err));
+    req.write(data);
+    req.end();
+    return deferred;
+  }
 
   // skip this test unless in GitHub Actions, as this test iterates over
   // all available network interfacesnand network interfaces on user
```
```suggestion
  // all available network interfaces and network interfaces on user
```

src/packages/core/src/servers/http-server.ts
```diff
@@ -71,6 +71,7 @@ function prepareCORSResponseHeaders(method: HttpMethods, request: HttpRequest) {
 
 function sendResponse(
   response: HttpResponse,
+  closeConnection: boolean,
```
MicaiahReid, line 74: Instead of increasing the number of parameters passed to this function, can we just use `this.#isClosing` within `sendResponse`?

MicaiahReid (approved):
Looks good! I have one spelling fix (from legacy code that I just happened to see) and on potential thing to make things a little cleaner.

src/packages/core/src/servers/http-server.ts
davidmurdoch: We can't; it is a global function.

src/packages/core/tests/server.test.ts
```diff
@@ -62,6 +62,49 @@ describe("server", () => {
...
   }
+  /**
+   * Sends a post request to the server and returns the response.
+   * @param address
+   * @param port
+   * @param json
+   * @param agent
+   * @returns
+   */
+  function post(host: string, port: number, json: any, agent?: any) {
```
davidmurdoch, line 74: Is this `post` and the other `post` below still similar enough that we can combine them into one function?

jeffsmale90: Yeah my bad - I have removed the duplicate.

davidmurdoch (approved):

PR #2716, opened 2022-03-28
fix: Add tests for json-rpc-data and address to describe expected behaviour 

### LINKED ISSUE #1594: Fix `toString` Padding for `Address` types
The original intent for `Address` was to have the following functionality:
```TypeScript
const address = new Address("0x1");
console.log(address.toString()) // 0x0000000000000000000000000000000000000001
```
where the original value, "0x1" is stored, but the `toString()` representation pads the values with zeros to be a proper address. This isn't actually working, and likely points to overall issues in `json-rpc-data` and `json-rpc-base-types` that could have side affects to consider.

As an additional note, this also leads to `evm_addAccount` adding invalid account addresses to the `personal` namespace.

jeffsmale90:
fixes #1594

### DIFF
--- a/src/packages/utils/tests/json-rpc-data.test.ts
+++ b/src/packages/utils/tests/json-rpc-data.test.ts
@@ -2,106 +2,186 @@ import assert from "assert";
 import {Data} from "..";
 
 describe("json-rpc-data", () => {
-    it("should stringify without specifying byteLength", () => {
-      const cases = [
-        {input: "0x1", expected: "0x1"}, // I'm not sure whether this is the correct behaviour - should it return a single byte length string - ie 0x01?
-        {input: "0x01", expected: "0x01"},
-        {input: "0x0123456789abcdef", expected: "0x0123456789abcdef"},
-        {input: Buffer.from([0x01]), expected: "0x01"}
-      ];
+  const validValues = [ "0x", "0x1", "0x1234", Buffer.from([]), Buffer.from([0x12,0x34]) ];
+  const invalidValues: any[] = [ "1234", 1234n, NaN/*, undefined, null, 1234, [], {}, "0x-1234"*/ ]; // todo: this should be addressed in rewrite of json-rpc-data
+                                                                                                     // See related https://github.com/trufflesuite/ganache/issues/2728
+  const validBytelengths = (() => { let i = 0; return [...new Array(100)].map(_ => i++); })(); // [0...99]
+  const invalidBytelengths: any[] = [ -1, "1", {}, [], null, NaN ];
+  const inputOf32Bytes = "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef";
 
-      cases.forEach(c => {
-        const d = new Data(c.input);
-        const stringified = d.toString();
+  function getExpectedString(value: Buffer|string, bytelength?: number) {
+    let expected: string;
 
-        assert.equal(stringified, c.expected);
+    if (typeof value === "string") {
+      expected = value.slice(2);
+    } else if (Buffer.isBuffer(value)) {
+      expected = (<Buffer>value).toString("hex");
+    } else {
+      throw new Error(`Type not supported ${typeof value}`)
+    }
+    if (bytelength !== undefined) {
+      const padCharCount = (bytelength - expected.length / 2) * 2; // (desired byte count - actual byte count) * 2 characters per byte
+      if (padCharCount > 0) {
+        expected = "0".repeat(padCharCount) + expected;
+      } else {
+        expected = expected.slice(Math.abs(padCharCount));
+      }
+    }
+    return "0x" + expected;
+  }
+
+  describe("ctor", () => {
+    it("should accept different values", () => {
+      validValues.forEach(value => {
+        const d = new Data(value);
       });
     });
 
-    it("should stringify to fixed byteLength", () => {
-      const cases = [
-        {input: "0x1", length: 1,  expected: "0x01"},
-        {input: "0x01", length: 2, expected: "0x0001"},
-        {input: "0x01", length: 10, expected: "0x00000000000000000001"},
-        {input: Buffer.from([0x01]), length: 2, expected: "0x0001"}
-      ];
-
-      cases.forEach(c => {
-        const d = new Data(c.input);
-        const stringified = d.toString(c.length);
-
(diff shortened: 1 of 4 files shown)


jeffsmale90:
I spent some time looking into this today, and have a reasonable grasp on what's going on.

I just wanted to check whether my expectations around the behaviour are correct, so this PR only contains a couple of tests describing that behaviour. Don't expect anything to work, and don't take this as a full and complete set of tests for this behaviour!

Thanks.

jeffsmale90:
If you have a look at the build, pretty much all of these tests currently fail, except for the first set where no byteLength parameter is passed, as in the current implementation, the byteLength parameter has no impact on the toString() result.

MicaiahReid (commented):
This looks good! If you implemented these changes I think it will give us a lot more confidence in our `Data` types. I think the biggest concern is all of the possible side affects this fix could cause. Let's hope our tests are robust enough to catch them!

One more thing to note that I realize as I was digging into the code for the `Data` constructor. It only allows a `value` of type `string | Buffer`, but we only validate against `bigint`. It might be worth adding extra validation? ¯\_(ツ)_/¯

jeffsmale90:
There's plenty of opportunity to improve this - ie. I tested the performance impact of removing the weakmaps, and it's quite significant, but I wanted to fix the issue without changing too much, with the intention of addressing remaining problems when we rework this.

src/packages/utils/tests/json-rpc-data.test.ts
```diff
@@ -0,0 +1,187 @@
+import assert from "assert";
+import {Data} from "..";
+
+describe("json-rpc-data", () => {
+  const validValues = [ "0x", "0x1", "0x1234", Buffer.from([]), Buffer.from([0x12,0x34]) ];
+  const invalidValues: any[] = [ "1234", 1234n, NaN/*, undefined, null, 1234, [], {}, "0x-1234"*/ ]; // todo: this should be addressed in rewrite of json-rpc-data
+                                                                                                     // See related https://github.com/trufflesuite/ganache/issues/2728
```
MicaiahReid:
```suggestion
                                                                                                     // See related https://github.com/trufflesuite/ganache/labels/json-rpc%20refactor
```
Since it's more a collection of issues rather than a specific issue, I made a label in github and added that label to the relevant issues.

src/packages/utils/tests/json-rpc-data.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      if (padCharCount > 0) {
+        expected = "0".repeat(padCharCount) + expected;
+      } else {
+        expected = expected.slice(Math.abs(padCharCount));
+      }
+    }
+    return "0x" + expected;
+  }
+
+  describe("ctor", () => {
```
MicaiahReid: I think I'm a bit slow on this one, what's ctor?

MicaiahReid: Ahhh, constructor! Let's spell it out for people like me, haha
```suggestion
  describe("constructor", () => {
```

src/packages/utils/tests/json-rpc-data.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+        assert.equal(s, expected);
+      });
+    });
+
+    it("should truncate the value with a smaller bytelength", () => {
+      const d = new Data("0x1234567890abcdef");
+      const s = d.toString(2);
+
+      assert.equal(s, "0x1234");
```
MicaiahReid: Can we use your `getExpectedString` function here too just for consistency? That function can truncate too, right?

src/packages/utils/tests/json-rpc-data.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  describe("toBuffer()", () => {
+    it("should create a buffer", () => {
+      const expected = Buffer.from([0x12, 0x34]);
+      const d = new Data("0x1234");
+      const b = d.toBuffer();
+
+      assert.deepEqual(b, expected);
+    });
+
+    // todo: this should be addressed in rewrite of json-rpc-data
```
MicaiahReid:
```suggestion
    // todo: this should be addressed in rewrite of json-rpc-data https://github.com/trufflesuite/ganache/labels/json-rpc%20refactor
```

src/packages/utils/tests/json-rpc-data.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    it.skip("should create a buffer with a smaller bytelength", () => {
+      const expected = Buffer.from([0x12, 0x34]);
+      const d = new Data("0x123456789abcdef", 2);
+      const b = d.toBuffer();
+
+      assert.deepEqual(b, expected);
+    });
+
+    // todo: this should be addressed in rewrite of json-rpc-data
```
MicaiahReid:
```suggestion
    // todo: this should be addressed in rewrite of json-rpc-data https://github.com/trufflesuite/ganache/labels/json-rpc%20refactor
```

MicaiahReid (changes requested):
Looks good! Tiny nit-picks in the tests.

jeffsmale90:
Lucky you suggested the change to use getExpectedString - because it didn't truncate properly!

The reason I had a couple of special cases explicitly defined in the tests was to explain the behaviour around truncating / padding. Using getExpectedString instead and adding the 32byte value to the validValues array, I found that the special cases were duplicates, so I removed them and added a comment to the getExpectedString function to explain.

MicaiahReid (approved):

src/packages/utils/tests/json-rpc-data.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (typeof value === "string") {
+      expected = value.slice(2);
+    } else if (Buffer.isBuffer(value)) {
+      expected = (<Buffer>value).toString("hex");
+    } else {
+      throw new Error(`Type not supported ${typeof value}`)
+    }
+    if (bytelength !== undefined) {
+      /*
+      If the value is longer than the specified bytelength, then the result my be left padded.
```
MicaiahReid:
```suggestion
      If the value is longer than the specified bytelength, then the result must be left padded.
```

MicaiahReid:
@jeffsmale90 I added one typo fix to that new comment, but I've approved.

PR #2811, opened 2022-04-04
perf: improve performance of the persistent cache used in forking

jeffsmale90:
This PR makes the level get stream's 'lt' argument more specific in order to retrieve fewer entries and convert block hash to Buffer type immediately on initialization (rather than every time we get / put an entry).


jeffsmale90:
Running the convex-shutdown-simulation, I'm seeing a 3.5% performance improvement (~870ms) with this change.

davidmurdoch (approved):

PR #2840, opened 2022-04-06
fix: reject transaction if maxFeePerGas is less than blockBaseFeePerGas of the pending block

### LINKED ISSUE #2176: In Geth chain-mode, logic to accept/reject transactions based on gas price/limit should match Geth
If a transaction gets into the miner's `runTx` function, and it fails in the VM, that `catch` block should probably add the transaction back into the pool.

Check against geth if:
1. a transaction like this: `{"from" : "0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1", "gas": "0x5b8d80", "maxFeePerGas": "0x1" }` (too low `maxFeePerGas`) immediately returns an error like this `VM Exception while processing transaction: Transaction's maxFeePerGas (1) is less than the block's baseFeePerGas`, and if
2. the transaction is then available in `txpool_content`

### LINKED ISSUE #2831: Don't remove a transaction from the txpool if it's maxFeePerGas < baseFeePerGas
If a transaction's `maxFeePerGas` is too low (< `baseFeePerGas` for the pending block) after the transaction is removed from the "pending" queue, an error will be thrown within the vm:

https://github.com/trufflesuite/ganache/blob/8f8feb4473a5cc770c4ea92daeb5469dc932a433/src/chains/ethereum/ethereum/src/miner/miner.ts#L461

During analysis of the behaviour in Geth, we discovered that the transaction will stay "pending" until the `baseFeePerGas` reduces sufficiently for the transaction to be included in a block.

We should update Ganache to ensure that a pending transaction with insufficient gas fee remains in the "pending" queue.

In order to reproduce this behaviour in Geth:
- Create a "future nonce" transaction with relatively low `maxFeePerGas` (ie. the `baseBlockFee` of the pending block)
- Create a large number of high-`maxFeePerGas` transactions to fill the first block (from a different address to the "future nonce" transaction above)
- Start the miner in order to create the block 1 (this will drive up the `baseBlockFee`)
- Stop the miner immediately
- Create a "filler" transaction in order to make the "future nonce" transaction valid
- Create a large number of high-`maxFeePerGas` transactions to fill up the block and ensure that the `baseBlockFee increases further
- Start the miner in order to create the block 2 (which will contain the "filler" transaction, but _not_ the "future nonce" transaction (as per https://github.com/trufflesuite/ganache/issues/2526#issuecomment-1089579021)
- Wait for the "future nonce" transaction to eventually be mined

There should be a number of empty blocks created after block 2, before the "future nonce" transaction is included (ie the `baseBlockFee` reduces sufficiently)

jeffsmale90:
Fixes #2176

### DIFF
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -476,7 +476,7 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
       );
     }
 
-    // transaction maxGasPerFee must be greater or equal to baseFeePerGas _of the pending block_
+    // transaction maxGasPerFee must be greater or equal to baseFeePerGas _of the next block_
     if ("maxFeePerGas" in transaction && !transaction.maxFeePerGas.isNull()) {
       const nextBaseFee = Block.calcNextBaseFee(this.#blockchain.blocks.latest);
       if (transaction.maxFeePerGas.toBigInt() < nextBaseFee) {
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -478,7 +478,6 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
 
     // transaction maxGasPerFee must be greater or equal to baseFeePerGas _of the pending block_
     if ("maxFeePerGas" in transaction && !transaction.maxFeePerGas.isNull()) {
-      console.log(this.#blockchain.blocks.latest);
       const nextBaseFee = Block.calcNextBaseFee(this.#blockchain.blocks.latest);
       if (transaction.maxFeePerGas.toBigInt() < nextBaseFee) {
         return new CodedError(
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -9,11 +9,13 @@ import {
   UNDERPRICED,
   REPLACED,
   TRANSACTION_LOCKED,
-  INSUFFICIENT_FUNDS
+  INSUFFICIENT_FUNDS,
+  ERROR_FEE_CAP_TOO_LOW
 } from "@ganache/ethereum-utils";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
 import { Executables } from "./miner/executables";
 import { TypedTransaction } from "@ganache/ethereum-transaction";
+import {Block} from "@ganache/ethereum-block";
 
 /**
  * Checks if the `replacer` is eligible to replace the `replacee` transaction
@@ -472,6 +474,18 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
         INTRINSIC_GAS_TOO_LOW,
         JsonRpcErrorCode.INVALID_INPUT
       );
+    }
+
+    // transaction maxGasPerFee must be greater or equal to baseFeePerGas _of the pending block_
+    if ("maxFeePerGas" in transaction && !transaction.maxFeePerGas.isNull()) {
+      console.log(this.#blockchain.blocks.latest);
+      const nextBaseFee = Block.calcNextBaseFee(this.#blockchain.blocks.latest);
+      if (transaction.maxFeePerGas.toBigInt() < nextBaseFee) {
+        return new CodedError(
+          ERROR_FEE_CAP_TOO_LOW,
+          JsonRpcErrorCode.TRANSACTION_REJECTED
+        );
+      }
     }
 
     return null;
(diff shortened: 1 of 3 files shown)


davidmurdoch:
Can you rebase this on current `develop`?

MicaiahReid:
So with further testing I discovered that transactions with too small gas price were accepted if the gas limit is specified. I opened an [issue with Geth](https://github.com/ethereum/go-ethereum/issues/24661), and they pretty much said it's expected behavior. Here is a summary of how they handle these cases and their reasoning:

| Gas Limit | Gas Price | Result | Notes |
| -------- | -------- | -------- | ----- |
| Unspecified     | Unspecified     | No error     | Geth chooses a good gas price, then uses that price to estimate gas limit. |
| Specified     | `< baseFeePerGas`     | No error     | Geth blindly accepts the gas price because the transaction has a gas limit. |
| Unspecified     | `< baseFeePerGas`     | `Error: maxFeePerGas is less than block base fee.`     | Because no gas limit was specified, Geth needs to estimate it. To estimate it, it runs the transaction against the next block. Because the gas price is too low, it's rejected. |

I think the key difference between us and Geth in this case is that our `eth_estimateGas` function does not enforce a `baseFeePerGas`. This can be seen when we make the `RuntimeBlock` in `eth_estimateGas`:
```JavaScript
  parentHeader.parentHash,
  parentHeader.miner,
  tx.gas.toBuffer(),
  parentHeader.gasUsed.toBuffer(),
  parentHeader.timestamp,
  options.miner.difficulty,
  parentHeader.totalDifficulty,
  0n // no baseFeePerGas for estimates
);
```

So I think rather than this current change, we may just want to pass in the _next_ block's `baseFeePerGas` rather than `0n`. Thoughts? @davidmurdoch @jeffsmale90

MicaiahReid (changes requested):
So with further testing I discovered that transactions with too small gas price were accepted if the gas limit is specified. I opened an [issue with Geth](https://github.com/ethereum/go-ethereum/issues/24661), and they pretty much said it's expected behavior. Here is a summary of how they handle these cases and their reasoning:

| Gas Limit | Gas Price | Result | Notes |
| -------- | -------- | -------- | ----- |
| Unspecified     | Unspecified     | No error     | Geth chooses a good gas price, then uses that price to estimate gas limit. |
| Specified     | `< baseFeePerGas`     | No error     | Geth blindly accepts the gas price because the transaction has a gas limit. |
| Unspecified     | `< baseFeePerGas`     | `Error: maxFeePerGas is less than block base fee.`     | Because no gas limit was specified, Geth needs to estimate it. To estimate it, it runs the transaction against the next block. Because the gas price is too low, it's rejected. |

I think the key difference between us and Geth in this case is that our `eth_estimateGas` function does not enforce a `baseFeePerGas`. This can be seen when we make the `RuntimeBlock` in `eth_estimateGas`:
```JavaScript
  parentHeader.parentHash,
  parentHeader.miner,
  tx.gas.toBuffer(),
  parentHeader.gasUsed.toBuffer(),
  parentHeader.timestamp,
  options.miner.difficulty,
  parentHeader.totalDifficulty,
  0n // no baseFeePerGas for estimates
);
```

So I think rather than this current change, we may just want to pass in the _next_ block's `baseFeePerGas` rather than `0n`. Thoughts? @davidmurdoch @jeffsmale90

MicaiahReid:
Upon further inspection, it doesn't look like we call `eth_estimateGas` as liberally as Geth does. We only estimate gas for transactions that omit `gasLimit` if their `--miner.defaultTransactionGasLimit="estimate"`, so my suggested fix wouldn't work in all cases.

Regardless, the current implementation now rejects the transaction in _more_ cases than Geth. In my eyes, we could:
1. Match Geth exactly by setting the `baseFeePerGas` in the `eth_estimateGas` function as mentioned above, and checking that the gas price is greater than the `baseFeePerGas` in the `autofillDefaultTransactionValues` branch that uses the default gas limit.
2. Add a `--chain.mode="geth"` option that allows a user to use ganache in a more strict mode (following geth's rules) and a more lenient mode (we don't ever error for gas being too low).
3. Leave ganache as is to be more lenient and, in our eyes, intuitive.

jeffsmale90:
@MicaiahReid @davidmurdoch I think the best way forward is to discard this change, and consider adding a strict "Geth" mode as @MicaiahReid recommended where we pass the _actual_ next block baseFeePerGas into vm.runTx (instead of 0).

What do you guys think?

MicaiahReid:
@jeffsmale90 I'm on board for that! 
I think in all the actions would be:
 - make an issue for a general Geth compatibility mode feature
 - update #2176 with details on exactly how Geth does it
 - mark #2176 as blocked by the issue you make for the Geth compatibility mode feature
 - (optionally) make a label that we can apply to those two issues and any future geth compatibility mode features

@davidmurdoch thoughts?

davidmurdoch:
@MicaiahReid That sounds like a great plan to me!

jeffsmale90:
Closing this PR, after following the actions @MicaiahReid listed above.

PR #2859, opened 2022-04-08
Feat/version check

davidmurdoch:


tenthirtyone:
### Description
`version-check` (VC) alerts the user when it detects a newer version is available. It does so in two ways:

1. When the user starts `ganache` from the command line through a banner message.
2. When the `ganache --version` command switch is run as a single string.

VC does two things. 
1. On startup, VC will report based on an existing `latestVersion` stored on disk, comparing it to the currently running software version (`currentVersion`). 
2. It makes a request to a configured url and expects back a semver response representing the `latestVersion` that is then persisted to disk.

#### Requirements
VC does not block the currently running application when it makes a request to check on the `latestVersion`. It is specced to be eventually consistent, assuming users will start/stop the tool implementing VC frequently. At each startup, VC checks against the  value of the previously fetched `latestVersion`. 

VC tracks the version that has been reported to the user to reduce its chattiness to the user. Upon start, VC will persist a `latestVersionLogged` representing the semver of the last version it displayed to the user upon starting ganache. VC will not display the banner message more than once for a given version. When the `ganache --version` command is run, VC will always display the single string message (if a newer version is known). 

On install, VC defaults the `latestVersion` and `latestVersionLogged` to `0.0.0`. This will always be below the first-run of the latest install. 

VC is specced to fail silently. If a request fails, returns invalid semver or any combination of current/latest version is not valid semver it will quit without informing the user. It is self-healing if the API incorrectly reports a version and is later corrected it will inform the user of the latest, correct, version. 

VC's relies on the same `conf` package as truffle to manage the data persistence across environments (Linux, Mac, Windows).

VC relies on the `semver` package to perform `diff` and validation of semver. 

#### Usage
VC is flexible in how it may be implemented

``` javascript
const vc = new VersionCheck(currentVersionSemVer, [config], [logger]);
```

The constructor requires a `currentVersionSemVer` and accepts an optional config and logger function. 

``` javascript
vc
  .init();
  .log();
```

`init` will send the request to fetch and set the `latestVersion` on disk. 
`log` will perform the required notification checks and, if they pass, log the banner message to the user based on the current `latestVersion` on disk. 

`init` will not await the fetch, it fires the request and does not concern itself with the outcome or results. If the desired behavior is to wait for a response before logging to the user (forgoing the eventual consistency spec), VC can be used as such:

``` javascript
await vc.getLatestVersion();
vc.log();
```

This usage will perform the the same steps as using `init`, however the `latestVersion` stored on disk will be the version returned from the fetch. This could be desirable by applications that are not concerned with latency/server timeout/ or that have configured a low `ttl` value in the VC config. 

``` javascript
vc.getVersionMessage();
```

`getVersionMessage` returns a single string noting the `currentVersion` and `latestVersion` if a semver change is detected. This string is appended to the `detailedVersion` string in `@ganache/cli` for to support the `ganache --version` command switch.

#### Test

From a develop environment, start ganache using:
``` bash
$ export VERSION="1.2.3"; npm run start
```

##### Expected Output
1.) No banner message is displayed (first run, assumes latest was installed)
``` bash
~/proj/ganache (feat/version-check) » export VERSION="1.2.3"; npm run start                                                                                                                                                                                                                                  hayek@rothbard

> root@ start /home/hayek/proj/ganache
> lerna exec --loglevel=silent --scope ganache -- npm run start --silent -- 

Debugger listening on ws://127.0.0.1:9229/b4dc2d61-8f90-4a2c-80ff-14db24160453
For help, see: https://nodejs.org/en/docs/inspector
ganache v1.2.3 (@ganache/cli: DEV, @ganache/core: DEV)
Starting RPC server
```
2.) A config file now exists for the user `less ~/.config/@ganache/version-check-nodejs/config.json` (on linux)
```
{
        "config": {
                "packageName": "ganache",
                "enabled": true,
                "url": "https://version.trufflesuite.com/",
                "ttl": 300,
                "latestVersion": "7.0.5",
                "latestVersionLogged": "0.0.0"
        }
}
```

Restart `ganache` (`export VERSION="1.2.3"; npm run start`)

##### Expected Output
1) Banner message displays, `latestVersion` > `latestVersionLogged`
``` bash
export VERSION="1.2.3"; npm run start                                                                                                                                                                                                                                 hayek@rothbard

> root@ start /home/hayek/proj/ganache
> lerna exec --loglevel=silent --scope ganache -- npm run start --silent -- 

Debugger listening on ws://127.0.0.1:9229/911542db-3daa-433c-a31f-2aa662a084e2
For help, see: https://nodejs.org/en/docs/inspector
ganache v1.2.3 (@ganache/cli: DEV, @ganache/core: DEV)
Starting RPC server

                        ╔══════════════════════════════════════════════════════════════════════╗                        
                        ║                                                                      ║                        
                        ║        New major version of ganache available! 1.2.3 ⇢ 7.0.5         ║                        
                        ║  Changelog: https://github.com/trufflesuite/ganache/releases/v7.0.5  ║                        
                        ║             Run npm install -g ganache@7.0.5 to update!              ║                        
                        ║                                                                      ║                        
                        ╚══════════════════════════════════════════════════════════════════════╝                        
```
2) Config updates `latestVersionLogged` 
``` bash
{
        "config": {
                "packageName": "ganache",
                "enabled": true,
                "url": "https://version.trufflesuite.com/",
                "ttl": 300,
                "latestVersion": "7.0.5",
                "latestVersionLogged": "7.0.5"
        }
}
```

Restart `ganache` (`export VERSION="1.2.3"; npm run start`)

##### Expected Results
1.) Banner message does not display, user has already seen this version.
``` bash
export VERSION="1.2.3"; npm run start                                                                                                                                                                                                                           130 ↵ hayek@rothbard

> root@ start /home/hayek/proj/ganache
> lerna exec --loglevel=silent --scope ganache -- npm run start --silent -- 

Debugger listening on ws://127.0.0.1:9229/27f051e4-1c84-4bdf-a41d-95601466ba2b
For help, see: https://nodejs.org/en/docs/inspector
ganache v1.2.3 (@ganache/cli: DEV, @ganache/core: DEV)
Starting RPC server
```

Run `npm start -- --version`                                                                                                                                                                

##### Expected Result
``` bash
ganache v1.2.3 (@ganache/cli: DEV, @ganache/core: DEV)
note: there is a new version available! 1.2.3 -> 7.0.5
```

package.json
```diff
@@ -17,7 +17,7 @@
     "prepublishOnly": "cd scripts && ts-node filter-shrinkwrap.ts \"$(lerna list --parseable --scope ganache)\"/npm-shrinkwrap.json",
     "postpublish": "git restore \"$(lerna list --parseable --scope ganache)\"/npm-shrinkwrap.json",
     "start": "lerna exec --loglevel=silent --scope ganache -- npm run start --silent -- ",
-    "test": "lerna exec --concurrency 1 -- npm run test",
+    "test": "lerna exec --concurrency 1 --scope @ganache/version-check -- npm run test",
```
MicaiahReid:
```suggestion
    "test": "lerna exec --concurrency 1 -- npm run test",
```
I assume this was just intended for while you were testing?

MicaiahReid: If so, one way you can set the scope in your CLI _before_ running `npm run test` (so that you don't actually have to alter the package.json like you did here is:
`$(npm bin)/lerna exec --scope @ganache/version-check -- npm t`

src/packages/version-check/README.md
```diff
@@ -0,0 +1,3 @@
+# `@ganache/version-check`
+
+> TODO: description
```
MicaiahReid: The comment you left on the PR would make an _excellent_ README, imo

src/packages/version-check/index.ts
```diff
@@ -0,0 +1,8 @@
+/*!
+ * @ganache/version-check
+ *
+ * @author David Murdoch
+ * @license MIT
+ */
+
+export * from "./src/";
```
MicaiahReid:
```suggestion
export * from "./src/";

```

MicaiahReid: Newline to appease the linter.

src/packages/version-check/src/index.ts
```diff
@@ -0,0 +1,278 @@
...
+        })
+        .setTimeout(ttl, () => {
+          req.close();
+          reject();
+        });
+    });
+  }
+
+  alreadyLoggedThisVersion() {
+    return this._config.latestVersionLogged >= this._config.latestVersion;
```
MicaiahReid: Is there a better way to do this check? It will start to fail eventually:
```
return "v7.1.0" >= "v7.10.0" // false, as it should
return "v7.9.0" >= "v7.10.0" // true, saying we've already logged this version when we haven't :/
````

Comment (MicaiahReid) [reply src/packages/version-check/src/index.ts]: It probably would be overkill for this, but we already have the `"lexicographic-key-codec"` library imported elsewhere in ganache. We could lexicographically encode the versions and compare those ¯\_(ツ)_/¯

File: src/packages/cli/src/cli.ts [hunk 5]
```diff
@@ -127,6 +133,8 @@ async function startGanache(err: Error) {
   }
   started = true;
 
+  versionCheck.init().log();
```
MicaiahReid: I kind of wonder if we should call this at the end of the `startGanache` function. I tested this and it's easy for the banner to get lost because we log so much data to the console on startup.

src/packages/version-check/tests/index.test.ts
```diff
@@ -0,0 +1,1209 @@
...
+    it("sets an optional config", () => {
+      vc = new VersionCheck(testVersion, testConfig);
+
+      assert.deepStrictEqual(
+        vc._config,
+        testConfig,
+        "testConfig is not set in the version checker."
+      );
+    });
+    it("sets an optional sparse config", () => {});
```
MicaiahReid: Either add a test or turn into a TODO comment with a link to an issue if it's awaiting a feature.

src/packages/version-check/tests/index.test.ts
```diff
@@ -0,0 +1,1209 @@
...
+      assert.equal(
+        latestVersionLogged,
+        vc._config.latestVersionLogged,
+        "setLatestVersionLogged incorrectly set"
+      );
+    });
+  });
+
+  describe("canNotifyUser", () => {
+    it("false if !currentVersion or currentVersion === falsy", () => {
```
MicaiahReid, line 222: Maybe add a new test for comparing weird values like `"v7.9.0"` vs `"v7.10.0"` once those cases are fixed.

MicaiahReid (changes requested):
Great work! And you did an awesome job documenting this stuff. 

In addition to my comments, there's one issue I noticed when testing. I'm getting this error in the console when testing in VS Code:
`Starting inspector on 127.0.0.1:9229 failed: address already in use`

It happens when I run ganache more than once in a row. It looks like the VS code inspector is being held open. Maybe by the request for the version number?

I don't know for sure what the cause is, but I switched to develop and rebuilt and the issue does not occur over there, so I think it was introduced with this change.

src/packages/version-check/src/index.ts
davidmurdoch: A previous implementation split the version strings apart to compare them:

https://github.com/trufflesuite/ganache/pull/2859/commits/082fb00c4875f12ad489181d747259b1bd3cd52d#diff-d3c52bfdae42da5ce88cd5db96d272177e44283f952b3545816b03f08a2c8b85R21-R29

```typescript
function getUpgradeType(current: string, update: string) {
  const [_, major, minor, patch] = current.match(semverRegex).slice(1, 4).map(Number);
  const [updateMajor, updateMinor, updatePatch] = update.match(semverRegex).slice(1, 4).map(Number);

  return updateMajor !== major ? 'major'
    : updateMinor !== minor ? 'minor'
      : updatePatch !== patch ? 'patch'
        : null;
}
```

src/packages/version-check/tests/index.test.ts
tenthirtyone: Added support for `v`, '-alpha' and '-beta' tagging.

src/packages/version-check/src/index.ts
tenthirtyone: The unit testing sussed out some issues with the regex/split strategy and my regex is weak so I added `semver` to hide my shame. On the plus side, support is in for `-alpha` and `-beta` in prepatch/prerelease capacity and 'v' version strings like 'v1.2.3'

tenthirtyone: The unit tests 

```
0.0.1 -> 0.1.1 (reports patch)
0.1.0 -> 0.1.1 (reports major)
1.0.0 -> 1.0.1 (reports major)
0.1.0 -> 1.1.0 (reports minor)
1.0.0 -> 1.1.0 (reports major)
```

tenthirtyone:
@MicaiahReid Super Sleuthing catching the multiple instances bug. I was able to reproduce and troubleshoot. Looks like it was where I had `req.end` in the `fetch...`

This should be ready for the next pass.

MicaiahReid (approved):
Awesome work @tenthirtyone! Tested and everything seems to be working great.

PR #2892, opened 2022-04-12
ci: automated release improvements

### LINKED ISSUE #2279: ci: make sure release makes a v7.*.* tag, not just ganache@7.*.*

### LINKED ISSUE #2882: `develop` branch doesn't currently allow robot@trufflesuite.com to merge into develop
Before we fully automate our publishing step we'll need to set up a github account for `robot@trufflesuite.com` and add it to the ganache branch settings as an account that can merge into develop.

MicaiahReid:
This change updates the release automation process in a few ways:
 1. Adds a `vX.x.x` git tag to each release, in addition to the `ganache@X.x.x` tag that already was automatically added - Fixes #2279.
 2. After a successful `"latest"` release (into master), merges master back into develop to keep their commit history in sync.
 3. Signs all automation commits from @TrufBot - Fixes #2882.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,3 +131,10 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
             INFURA_KEY=${{ secrets.INFURA_KEY }}
+
+      - name: Merge changes back into develop for latest release (master)
+        if: ${{ env.TAG == 'latest' }}
+        run: |
+          git checkout develop
+          git merge master
+          git push origin develop
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,10 +16,16 @@ jobs:
           # use node 14 until we can evaluate using npm 7+ and lock file version 2
           node-version: 14
 
-      - name: Set git identity
+      - name: Import Robot GPG key
+        uses: crazy-max/ghaction-import-gpg@v4
+        with:
+          gpg_private_key: ${{ secrets.TRUFBOT_GPG_SECRET_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -130,3 +136,10 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
             INFURA_KEY=${{ secrets.INFURA_KEY }}
+
+      - name: Merge changes back into develop for latest release (master)
+        if: ${{ env.TAG == 'latest' }}
+        run: |
+          git checkout develop
+          git merge master
+          git push origin develop
```
davidmurdoch, line 145: I worry what will happen if someone commits to `develop` while a release is running. Will the merge cause things to be out of order. Could we end up with a timeline like this:

```
chore(release): publish v7.0.4
feat: a fix to land in v7.1.0
fix: v7.0.3 was broken, this fixes it
```

Would it be best to back-date our `chore(release):` commit to the same timestamp as the commit (tip of `master`) that initiated the release? I just don't know enough about git :thinking:

cds-amal: I think it would fail when pushing to origin/develop @davidmurdoch. The commits are on local refs and resolved by either git fetch, git push, both of which would fail if the tips don't match.

MicaiahReid: @davidmurdoch I also intend to restrict pushing to `develop` by anyone other than TrufBot. For now we have few enough people that we'll just need to coordinate during releases.

davidmurdoch (approved):
i hope it works

PR #2944, opened 2022-04-20
fix: enforce eip-2 imposed limits and secp256k1 upper bound for private keys

### LINKED ISSUE #2586: Transaction from unlocked account fails with error: "The nonce generation function failed, or the private key was invalid":

### LINKED ISSUE #2600: All transaction signatures whose s-value is greater than secp256k1n/2 are now considered invalid, but ganache accepts them
According to EIP-2 "All transaction signatures whose s-value is greater than secp256k1n/2 are now considered invalid."
the signature generated by hardware wallet has s component greater than secp256k1n/2 .

Ganache currently permits this greater "s" value.

davidmurdoch:
Sending transactions from an impersonated account with a "large" account number, like `fffffffffffffffffffffffffffffffffffffffe`, would result in the error "The nonce generation function failed, or the private key was invalid" due to the way we fake transaction signing in ganache. Previously we would take the account number plus the first 12 bytes of the account number, `fffffffffffffffffffffffffffffffffffffffe` + `ffffffffffffffffffffffff`, and would use that as a fake private key. This results in an invalid key, as secp256k1, the elliptic curve used in Ethereum cryptography, has an effective maximum private key value of `0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140n` (AKA `secp256k1_n - 1`, or the total number of non-trivial points on the curve). This fixes #2586.

While implementing this fix it was discovered that we were not rejecting transactions with too-high s-values; i.e., s-values that are greater than `(secp256k1_n - 1) / 2`. This restriction was added way back in Ethereum's first hardfork, homestead, as part of [EIP-2](https://eips.ethereum.org/EIPS/eip-2) in order to remove the possibility of "malleable" transactions. While somewhat unrelated to the core reason for this fix, it has been added as part of this PR. This fixes #2600.

### DIFF
--- a/src/chains/ethereum/ethereum/src/wallet.ts
+++ b/src/chains/ethereum/ethereum/src/wallet.ts
@@ -14,7 +14,7 @@ import {
   WEI
 } from "@ganache/utils";
 import { privateToAddress } from "ethereumjs-util";
-import secp256k1 from "@ganache/secp256k1";
+import secp256k1, { SECP256K1_N } from "@ganache/secp256k1";
 import { mnemonicToSeedSync } from "bip39";
 import { alea } from "seedrandom";
 import crypto from "crypto";
@@ -22,6 +22,8 @@ import createKeccakHash from "keccak";
 import { writeFileSync } from "fs";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
 import { Address } from "@ganache/ethereum-address";
+
+const TWELVE_255s = Buffer.allocUnsafe(12).fill(255);
 
 //#region Constants
 const SCRYPT_PARAMS = {
@@ -626,7 +628,7 @@ export default class Wallet {
 
   public createFakePrivateKey(address: string) {
     let fakePrivateKey: Buffer;
-    const addressBuf = Data.from(address).toBuffer();
+    const addressBuf = Address.from(address).toBuffer();
     if (addressBuf.equals(ACCOUNT_ZERO)) {
       // allow signing with the 0x0 address...
       // always sign with the same fake key, a 31 `0`s followed by a single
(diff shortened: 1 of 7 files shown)


src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -636,7 +638,45 @@ export default class Wallet {
...
+      // will. There are obviously many chances for a false positive, but the
+      // next condition in the `while` loop will catch those.
+      if (first12.compare(TWELVE_255s) === 0) {
+        while (
+          BigInt(`0x${fakePrivateKey.toString("hex")}`) >=
+          SECP256K1_MAX_PRIVATE_KEY
+        ) {
+          // keccak returns a 32 byte hash of the input data, which is the exact
+          // length we need for a private key.
+          // note: if keccak can return it's own input as it's output, then this
```
```suggestion
          // note: if keccak can return its own input as its output, then this
```

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -23,6 +23,8 @@ import { writeFileSync } from "fs";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
 import { Address } from "@ganache/ethereum-address";
 
+const TWELVE_255s = Buffer.allocUnsafe(12).fill(255);
```
MicaiahReid, line 26: This is a great variable name. Bravo.

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -169,6 +170,65 @@ describe("@ganache/ethereum-transaction", async () => {
...
+          const genesis = Common.forCustomChain(
+            "mainnet",
+            {
+              name: "ganache",
+              chainId: 1,
+              comment: "Local test network",
+              bootstrapNodes: []
+            },
+            // EIP-2 was in homestead, the first hardfork, so we need to create
+            // a common at the genesis (eke chainstart) so we can test at a fork
```
MicaiahReid:
```suggestion
            // a common at the genesis (aka chainstart) so we can test at a fork
```
Unless `eke` means something I'm not aware of?

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -169,6 +170,65 @@ describe("@ganache/ethereum-transaction", async () => {
         const txFromRpc = TransactionFactory.fromRpc(tempAccessListTx, common);
         assert.strictEqual(txFromRpc.type.toString(), "0x0");
       });
+
+      describe("EIP-2", () => {
```
MicaiahReid, line 174: Cool test!

MicaiahReid (changes requested):

MicaiahReid:
Reviewed! This also needs a description that is release ready.

src/chains/ethereum/ethereum/src/wallet.ts
davidmurdoch: Thanks for noticing. This line took me two weeks to get right!

MicaiahReid (approved):

PR #2962, opened 2022-04-23
ci: use semantic commits to determine next release version and fix final merge step

davidmurdoch:
Our previous release mostly worked (it failed to auto-merge back into the develop branch) but had the wrong version number for a neat feature release. This PR should fix both of these issues. :crossed_fingers:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          # we need all commit history so we can merge master into develop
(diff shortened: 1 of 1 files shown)


MicaiahReid (dismissed):

MicaiahReid (approved):
I hope this works!

davidmurdoch:
We need to think this over a bit more, as this won't actually work once merged into master, because `git log ^origin/master origin/master` will be empty, as everything from develop is already in master by the time this runs. An alternative is to use the most recent semantic-commit git tag (like "v7.0.4") that exists in master to determine the "range" to compare.

davidmurdoch:
closing in favor of #2979

PR #2979, opened 2022-04-25
ci: use semantic commits to determine next release version and fix final merge step

davidmurdoch:
Our previous release mostly worked (it failed to auto-merge back into the develop branch) but had the wrong version number for the next feature release. The release process wasn't detecting feature commits, which would cause feature releases to be versioned incorrectly (patch update instead of a minor update). This PR should fix both of these issues.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          # we need all commit history so we can merge master into develop
+          fetch-depth: 0
       - uses: actions/setup-node@v1
         with:
           # use node 14 until we can evaluate using npm 7+ and lock file version 2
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -11,6 +11,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          # we need all commit history so we can merge master into develop
+          fetch-depth: 0
```
davidmurdoch, line 15: The last steps:

```
git checkout develop
git merge origin/master
git push origin develop
```

need the complete history for `master` and `develop`. `fetch-depth: 0` causes `actions/checkout@v2` to fetch all history.

.github/workflows/release.yml
```diff
@@ -141,5 +179,5 @@ jobs:
         if: ${{ env.TAG == 'latest' }}
         run: |
           git checkout develop
-          git merge master
+          git merge origin/master
```
davidmurdoch, line 182: I don't know why, but `origin/` was required in my testing.

.github/workflows/release.yml
```diff
@@ -49,9 +51,45 @@ jobs:
         run: |
           echo "TAG=${GITHUB_REF##*/}" >> $GITHUB_ENV
 
+      - name: Count features
+        if: ${{ env.TAG == 'latest'}}
+        # get all commits in develop that aren't in master, find any that start with feat: or feat(*):, then count them
+        # 1. `git tag --sort=-version:refname --no-contains HEAD`
+        #    - get a list of all tags that do NOT contain out current HEAD commit
+        #    - sort by version name
```
cds-amal, line 59: I didn't know about `--sort=-version:refname`.  N.B, the Sort order defaults to the value configured for the `tag.sort` variable from the applied `.gitconfig` or lexicographic order otherwise.  Mentioning this in case anyone on the team has that variable defined, they may see different results. 

>       --sort=<key>
           Sort based on the key given. Prefix - to sort in descending order of the value. You may use the --sort=<key> option multiple times, in which case the last key becomes the
           primary key. Also supports "version:refname" or "v:refname" (tag names are treated as versions). The "version:refname" sort order can also be affected by the
           "versionsort.suffix" configuration variable. The keys supported are the same as those in git for-each-ref. Sort order defaults to the value configured for the tag.sort
           variable if it exists, or lexicographic order otherwise. See git-config(1).

.github/workflows/release.yml
cds-amal: Did master and origin/master match in your test environment?

cds-amal (approved):
This is cool ...and looks good to me!

.github/workflows/release.yml
davidmurdoch: IIRC, locally, both work. When tested in GitHub Actions  origin/master (technically we tested the branch named origin/fix-ci) worked.

davidmurdoch: I'm thinking it has something to do with the way it's checked out (it's not cloned!): https://github.com/trufflesuite/ganache/runs/6165271216?check_suite_focus=true#step:2:44

```
/usr/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/*
```

PR #2983, opened 2022-04-26
perf: improve JSON-RPC data types; better, faster, stronger

### LINKED ISSUE #2725: Ensure that only hex values are used in hex strings
Currently `json-rpc-base-types` validates that a string is prefixed with "0x" but not that the actual data is hex data. For example, `"0x a string"` is valid according to the base type.

### LINKED ISSUE #2728: Don't allow negative numbers to be used as `Quantity`
Don't do it!

### LINKED ISSUE #2857: RPC Quantity should not allow `"0x"`
Geth errors with: `cannot unmarshal hex string \"0x\" into Go struct field`

jeffsmale90:
As part of an ongoing effort to make Ganache :zap: Lightning-Fast :zap:, the internal representations of JSON-RPC data (`Data`, `Quantity`, and `Address`) have been rewritten to be quicker, and does not change how you should use Ganache.

These classes also now validate input values correctly, following the guidelines  [established by the Ethereum Foundation](https://ethereum.org/en/developers/docs/apis/json-rpc/). This makes Ganache behave more correctly, but may cause some failures in existing code bases where invalid values are being passed into RPC functions.

This change has a significant impact on forking performance. [The Convex Shutdown benchmark](https://github.com/mds1/convex-shutdown-simulation) simulates a call to Convex's `systemShutdown` method. With a hot cache, Ganache's performance, with and without this change is below:

| Ganache Version | Cached     |
| -----------------|------------|
| 7.2.0 - before      | 0m28.015s |
| 7.3.0 - after         | 0m19.312s |

Which is an approximately 31% speed up!

_Note: Benchmarks were performed on macOS 12.2.1 with M1 Pro CPU and 16 GB LPDDR5._

### DIFF
--- a/src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
+++ b/src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
@@ -6,6 +6,7 @@ import Blockchain from "../blockchain";
 import PromiseQueue from "@ganache/promise-queue";
 import type Common from "@ethereumjs/common";
 import { Data, Quantity } from "@ganache/utils";
+import { Address } from "@ganache/ethereum-address";
 import {
   GanacheRawExtraTx,
   TransactionFactory,
@@ -57,8 +58,8 @@ export default class TransactionManager extends Manager<NoOp> {
     if (!fallback.isValidForkBlockNumber(blockNumber)) return null;
 
     const extra: GanacheRawExtraTx = [
-      Data.toBuffer(tx.from, 20),
-      Data.from((tx as any).hash, 32).toBuffer(),
+      Address.toBuffer(tx.from),
+      Data.toBuffer((tx as any).hash, 32),
       blockHash.toBuffer(),
       blockNumber.toBuffer(),
       index.toBuffer(),
--- a/src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
+++ b/src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
@@ -44,7 +44,7 @@ export default class TransactionManager extends Manager<NoOp> {
   fromFallback = async (transactionHash: Buffer) => {
     const { fallback } = this.#blockchain;
     const tx = await fallback.request<Transaction>("eth_getTransactionByHash", [
-      Data.from(transactionHash).toString()
+      Data.toString(transactionHash)
     ]);
     if (tx == null) return null;
 
@@ -57,12 +57,12 @@ export default class TransactionManager extends Manager<NoOp> {
     if (!fallback.isValidForkBlockNumber(blockNumber)) return null;
 
     const extra: GanacheRawExtraTx = [
-      Data.from(tx.from, 20).toBuffer(),
+      Data.toBuffer(tx.from, 20),
       Data.from((tx as any).hash, 32).toBuffer(),
       blockHash.toBuffer(),
       blockNumber.toBuffer(),
       index.toBuffer(),
-      Quantity.from(tx.gasPrice).toBuffer()
+      Quantity.toBuffer(tx.gasPrice)
     ];
     const common = fallback.getCommonForBlockNumber(
       fallback.common,
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -524,7 +524,7 @@ describe("api", () => {
                 },
                 {
                   junk: "0x",
-                  error: `Cannot wrap "0x" as a json-rpc Quantity type; Quantity must contain at least one digit`
+                  error: `Cannot wrap "0x" as a json-rpc Quantity type; strings must contain at least one hexadecimal symbol.`
                 },
               ],
               contractMethod: `0x${methods["getBalance(address)"]}${encodedAddr}`
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -508,7 +508,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `cannot convert string value "123" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: {},
@@ -545,7 +545,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `cannot convert string value "123" into type \`Data\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: {},
@@ -571,7 +571,7 @@ describe("api", () => {
                 },
                 {
                   junk: "",
-                  error: `cannot convert string value "" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: "0x",
@@ -583,7 +583,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `cannot convert string value "123" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: {},
@@ -611,7 +611,7 @@ describe("api", () => {
                 },
                 {
                   junk: "",
-                  error: `cannot convert string value "" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: "0x",
@@ -623,7 +623,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `cannot convert string value "123" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: {},
@@ -642,16 +642,16 @@ describe("api", () => {
               junks: [
                 {
                   junk: null,
-                  error: `cannot convert string value "null" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`,
+                  error: `Cannot wrap string value "null" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`,
                   expectedValue: null
                 },
                 {
                   junk: undefined,
-                  error: `cannot convert string value "undefined" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "undefined" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: "",
-                  error: `cannot convert string value "" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: "0x",
@@ -663,11 +663,11 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `cannot convert string value "123" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: {},
-                  error: `cannot convert string value "[object Object]" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "[object Object]" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 }
                 // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
                 // { junk: "0xa string", error: `` }
@@ -682,16 +682,16 @@ describe("api", () => {
               junks: [
                 {
                   junk: null,
-                  error: `cannot convert string value "null" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`,
+                  error: `Cannot wrap string value "null" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`,
                   expectedValue: null
                 },
                 {
                   junk: undefined,
-                  error: `cannot convert string value "undefined" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "undefined" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: "",
-                  error: `cannot convert string value "" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: "0x",
@@ -703,11 +703,11 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `cannot convert string value "123" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: {},
-                  error: `cannot convert string value "[object Object]" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "[object Object]" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 }
                 // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
                 // { junk: "0xa string", error: `` }
@@ -734,7 +734,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `cannot convert string value "123" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: {},
@@ -787,7 +787,7 @@ describe("api", () => {
                 await assert.rejects(
                   prom,
                   new Error(error),
-                  `Failed junk data validation for "${type}" override type with value "${junk}".`
+                  `Failed junk data validation for "${type}" override type with value "${junk}". Expected error: ${error}`
                 );
               } else {
                 assert.strictEqual(
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -524,7 +524,7 @@ describe("api", () => {
                 },
                 {
                   junk: "0x",
-                  error: `Cannot wrap "0x" as a json-rpc Quantity type; strings must contain at least one hexadecimal symbol.`
+                  error: `Cannot wrap "0x" as a json-rpc Quantity type; strings must contain at least one hexadecimal character.`
                 },
               ],
               contractMethod: `0x${methods["getBalance(address)"]}${encodedAddr}`
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -513,13 +513,19 @@ describe("api", () => {
                 {
                   junk: {},
                   error: `Cannot wrap a "object" as a json-rpc type`
-                }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
-                // { junk: "0xa string", error: `` }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2728 is closed
-                // { junk: -9, error: `` },
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2857 is closed
-                //{ junk: "0x", error: `` },
+                },
+                {
+                  junk: "0xa string",
+                  error: `Cannot wrap string value "0xa string" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: -9,
+                  error: "Cannot wrap a negative value as a json-rpc type"
+                },
+                {
+                  junk: "0x",
+                  error: `Cannot wrap "0x" as a json-rpc Quantity type; Quantity must contain at least one digit`
+                },
               ],
               contractMethod: `0x${methods["getBalance(address)"]}${encodedAddr}`
             },
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -20,7 +20,7 @@ import { GanacheTrie } from "../../../src/helpers/trie";
 import { Transaction } from "@ganache/ethereum-transaction";
 
 const encodeValue = (val: number | string) => {
-  return Quantity.from(val).toBuffer().toString("hex").padStart(64, "0");
+  return Quantity.toBuffer(val).toString("hex").padStart(64, "0");
 };
 
 async function deployContract(provider, from, code) {
@@ -77,7 +77,7 @@ describe("api", () => {
         it("executes a message call", async () => {
           const result = await provider.send("eth_call", [tx, "latest"]);
           // gets the contract's "value", which should be 5
-          assert.strictEqual(Quantity.from(result).toNumber(), 5);
+          assert.strictEqual(Quantity.toNumber(result), 5);
         });
 
         it("does not create a transaction on the chain", async () => {
@@ -100,7 +100,7 @@ describe("api", () => {
           };
           const result = await provider.send("eth_call", [tx, "latest"]);
           // we can still get the result when the gasPrice is set
-          assert.strictEqual(Quantity.from(result).toNumber(), 5);
+          assert.strictEqual(Quantity.toNumber(result), 5);
         });
 
         it("allows eip-1559 fee market transactions", async () => {
@@ -114,13 +114,13 @@ describe("api", () => {
 
           const result = await provider.send("eth_call", [tx, "latest"]);
           // we can still get the result when the maxFeePerGas/maxPriorityFeePerGas are set
-          assert.strictEqual(Quantity.from(result).toNumber(), 5);
+          assert.strictEqual(Quantity.toNumber(result), 5);
         });
 
         it("allows gas price to be omitted", async () => {
           const result = await provider.send("eth_call", [tx, "latest"]);
           // we can get the value if no gas info is given at all
-          assert.strictEqual(Quantity.from(result).toNumber(), 5);
+          assert.strictEqual(Quantity.toNumber(result), 5);
         });
 
         it("rejects transactions that specify both legacy and eip-1559 transaction fields", async () => {
@@ -850,10 +850,10 @@ describe("api", () => {
             block: block
           };
           ethereumJsFromAddress = new EthereumJsAddress(
-            Quantity.from(from).toBuffer()
+            Quantity.toBuffer(from)
           );
           ethereumJsToAddress = new EthereumJsAddress(
-            Quantity.from(to).toBuffer()
+            Quantity.toBuffer(to)
           );
           // set up a real transaction
           transaction = {
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -3,7 +3,7 @@ import { EthereumProvider } from "../../../src/provider";
 import getProvider from "../../helpers/getProvider";
 import compile, { CompileOutput } from "../../helpers/compile";
 import { join } from "path";
-import { BUFFER_EMPTY, Data, Quantity, RPCQUANTITY_ONE } from "@ganache/utils";
+import { BUFFER_EMPTY, Data, Quantity } from "@ganache/utils";
 import { CallError } from "@ganache/ethereum-utils";
 import Blockchain from "../../../src/blockchain";
 import Wallet from "../../../src/wallet";
@@ -836,7 +836,7 @@ describe("api", () => {
             gas.toBuffer(),
             parentHeader.gasUsed.toBuffer(),
             parentHeader.timestamp,
-            RPCQUANTITY_ONE, // difficulty
+            Quantity.One, // difficulty
             parentHeader.totalDifficulty,
             parentHeader.baseFeePerGas.toBigInt()
           );
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -525,7 +525,7 @@ describe("api", () => {
                 {
                   junk: "0x",
                   error: `Cannot wrap "0x" as a json-rpc Quantity type; strings must contain at least one hexadecimal character.`
-                },
+                }
               ],
               contractMethod: `0x${methods["getBalance(address)"]}${encodedAddr}`
             },
@@ -556,11 +556,15 @@ describe("api", () => {
                 {
                   junk: {},
                   error: `Cannot wrap a "object" as a json-rpc type`
+                },
+                {
+                  junk: "0xa string",
+                  error: `Cannot wrap string value "0xa string" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: -9,
+                  error: "Cannot wrap a negative value as a json-rpc type"
                 }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
-                // { junk: "0xa string", error: `` }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2728 is closed
-                // { junk: -9, error: `` },
               ],
               contractMethod: `0x${methods["getCode(address)"]}${encodedContractAddress}`
             },
@@ -594,13 +598,19 @@ describe("api", () => {
                 {
                   junk: {},
                   error: `Cannot wrap a "object" as a json-rpc type`
+                },
+                {
+                  junk: "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth",
+                  error: `Cannot wrap string value "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: -9,
+                  error: "Cannot wrap a negative value as a json-rpc type"
+                },
+                {
+                  junk: "0x",
+                  error: "StateDiff override data must be a 64 character hex string. Received 0 character string."
                 }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
-                // { junk: "0xa string", error: `` }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2728 is closed
-                // { junk: -9, error: `` },
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2857 is closed
-                //{ junk: "0x", error: `` },
               ],
               contractMethod: `0x${methods["getStorageAt(uint256)"]}${slot}`
             },
@@ -634,13 +644,19 @@ describe("api", () => {
                 {
                   junk: {},
                   error: `Cannot wrap a "object" as a json-rpc type`
+                },
+                { // State override data must be 64 characters long in order to hit this validation
+                  junk: "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth",
+                  error: `Cannot wrap string value "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: -9,
+                  error: "Cannot wrap a negative value as a json-rpc type"
+                },
+                {
+                  junk: "0x",
+                  error: "State override data must be a 64 character hex string. Received 0 character string."
                 }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
-                // { junk: "0xa string", error: `` }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2728 is closed
-                // { junk: -9, error: `` },
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2857 is closed
-                //{ junk: "0x", error: `` },
               ],
               contractMethod: `0x${methods["getStorageAt(uint256)"]}${slot}`
             },
@@ -674,13 +690,19 @@ describe("api", () => {
                 {
                   junk: {},
                   error: `Cannot wrap string value "[object Object]" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth",
+                  error: `Cannot wrap string value "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: -9,
+                  error: `Cannot wrap string value "-9" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: "0x",
+                  error: "StateDiff override slot must be a 64 character hex string. Received 0 character string."
                 }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
-                // { junk: "0xa string", error: `` }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2728 is closed
-                // { junk: -9, error: `` },
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2857 is closed
-                //{ junk: "0x", error: `` },
               ],
               contractMethod: `0x${methods["getStorageAt(uint256)"]}${slot}`
             },
@@ -714,13 +736,19 @@ describe("api", () => {
                 {
                   junk: {},
                   error: `Cannot wrap string value "[object Object]" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                { // State override must be 64 characters long in order to hit this validation
+                  junk: "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth",
+                  error: `Cannot wrap string value "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: -9,
+                  error: `Cannot wrap string value "-9" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: "0x",
+                  error: `State override slot must be a 64 character hex string. Received 0 character string.`
                 }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
-                // { junk: "0xa string", error: `` }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2728 is closed
-                // { junk: -9, error: `` },
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2857 is closed
-                //{ junk: "0x", error: `` },
               ],
               contractMethod: `0x${methods["getStorageAt(uint256)"]}${slot}`
             },
@@ -745,13 +773,19 @@ describe("api", () => {
                 {
                   junk: {},
                   error: `Cannot wrap a "object" as a json-rpc type`
+                },
+                {
+                  junk: "0xa string",
+                  error: `Cannot wrap string value "0xa string" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                },
+                {
+                  junk: -9,
+                  error: "Cannot wrap a negative value as a json-rpc type"
+                },
+                {
+                  junk: "0x",
+                  error: `Cannot wrap "0x" as a json-rpc Quantity type; strings must contain at least one hexadecimal character.`
                 }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
-                // { junk: "0xa string", error: `` }
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2728 is closed
-                // { junk: -9, error: `` },
-                // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2857 is closed
-                //{ junk: "0x", error: `` },
               ],
               contractMethod: `0x${methods["createContract(bytes)"]}${simpleSol}`
             }
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -609,7 +609,8 @@ describe("api", () => {
                 },
                 {
                   junk: "0x",
-                  error: "StateDiff override data must be a 64 character hex string. Received 0 character string."
+                  error:
+                    "StateDiff override data must be a 64 character hex string. Received 0 character string."
                 }
               ],
               contractMethod: `0x${methods["getStorageAt(uint256)"]}${slot}`
@@ -645,7 +646,8 @@ describe("api", () => {
                   junk: {},
                   error: `Cannot wrap a "object" as a json-rpc type`
                 },
-                { // State override data must be 64 characters long in order to hit this validation
+                {
+                  // State override data must be 64 characters long in order to hit this validation
                   junk: "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth",
                   error: `Cannot wrap string value "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth" as a json-rpc type; the input value contains an invalid hex character.`
                 },
@@ -655,7 +657,8 @@ describe("api", () => {
                 },
                 {
                   junk: "0x",
-                  error: "State override data must be a 64 character hex string. Received 0 character string."
+                  error:
+                    "State override data must be a 64 character hex string. Received 0 character string."
                 }
               ],
               contractMethod: `0x${methods["getStorageAt(uint256)"]}${slot}`
@@ -701,7 +704,8 @@ describe("api", () => {
                 },
                 {
                   junk: "0x",
-                  error: "StateDiff override slot must be a 64 character hex string. Received 0 character string."
+                  error:
+                    "StateDiff override slot must be a 64 character hex string. Received 0 character string."
                 }
               ],
               contractMethod: `0x${methods["getStorageAt(uint256)"]}${slot}`
@@ -737,7 +741,8 @@ describe("api", () => {
                   junk: {},
                   error: `Cannot wrap string value "[object Object]" as a json-rpc type; strings must be prefixed with "0x".`
                 },
-                { // State override must be 64 characters long in order to hit this validation
+                {
+                  // State override must be 64 characters long in order to hit this validation
                   junk: "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth",
                   error: `Cannot wrap string value "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth" as a json-rpc type; the input value contains an invalid hex character.`
                 },
@@ -892,9 +897,7 @@ describe("api", () => {
           ethereumJsFromAddress = new EthereumJsAddress(
             Quantity.toBuffer(from)
           );
-          ethereumJsToAddress = new EthereumJsAddress(
-            Quantity.toBuffer(to)
-          );
+          ethereumJsToAddress = new EthereumJsAddress(Quantity.toBuffer(to));
           // set up a real transaction
           transaction = {
             from,
--- a/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
@@ -508,7 +508,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be prefixed with "0x".`
                 },
                 {
                   junk: {},
@@ -516,7 +516,7 @@ describe("api", () => {
                 },
                 {
                   junk: "0xa string",
-                  error: `Cannot wrap string value "0xa string" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "0xa string" as a json-rpc type; the input value contains an invalid hex character.`
                 },
                 {
                   junk: -9,
@@ -551,7 +551,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be prefixed with "0x".`
                 },
                 {
                   junk: {},
@@ -559,7 +559,7 @@ describe("api", () => {
                 },
                 {
                   junk: "0xa string",
-                  error: `Cannot wrap string value "0xa string" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "0xa string" as a json-rpc type; the input value contains an invalid hex character.`
                 },
                 {
                   junk: -9,
@@ -581,7 +581,7 @@ describe("api", () => {
                 },
                 {
                   junk: "",
-                  error: `Cannot wrap string value "" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "" as a json-rpc type; strings must be prefixed with "0x".`
                 },
                 {
                   junk: "0x",
@@ -593,7 +593,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be prefixed with "0x".`
                 },
                 {
                   junk: {},
@@ -601,7 +601,7 @@ describe("api", () => {
                 },
                 {
                   junk: "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth",
-                  error: `Cannot wrap string value "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "0xnothexnothexnothexnothexnothexnothexnothexnothexnothexnothexnoth" as a json-rpc type; the input value contains an invalid hex character.`
                 },
                 {
                   junk: -9,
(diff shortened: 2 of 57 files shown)


jeffsmale90:
I've rewritten these classes with minimal changes to the external API. These classes represent their values internally as Buffer, so that the conversions (toString, toBuffer etc) each have a single input type, which decreases the permutations!

`Data.toBuffer` initially did not respect ByteLength, which I have fixed (this caused a couple of tests to fail - mostly where a fixed length Data type was being converted to Buffer before being hashed. This can be seen here: https://github.com/trufflesuite/ganache/pull/2983/files#diff-d29601cdcc68f88433762afa09d2953d76f16de73eede59fceab658d6558844aL411

I added static conversion methods to Data and Quantity, so that we can do things like `Quantity.toBuffer("0x1234")`, which we do quite a lot. I've extracted the changes to _use_ these methods into a separate commit. (Although I then made some subsequent commits which makes it a little harder.

This improves the overall performance significantly, but is not the fasted possible implementation of these classes. I intend to reduce the number of calls to these functions by removing them out of `persistent-cache.ts` and other places where they are used in tight loops, which will reduce the impact significantly.

There's a couple of TODOs attached to the related issues. I'll clean those up once this has passed.

jeffsmale90:
I rewrote this, with a focus on simplicity, which also has improved the performance of the convex shutdown benchmark.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,14 +1,47 @@
+import { BUFFER_EMPTY } from "../../utils";
 import { BaseJsonRpcType } from "./json-rpc-base-types";
-import { strCache, toStrings } from "./json-rpc-base-types";
 
 function validateByteLength(byteLength?: number) {
-  if (typeof byteLength !== "number" || !(byteLength >= 0)) {
+  if (byteLength !== undefined && (typeof byteLength !== "number" || byteLength < 0 || !isFinite(byteLength))) {
```
davidmurdoch: Since this function is only called internally we can probably get rid of the validation altogether here.

src/packages/utils/tests/json-rpc-data.test.ts
```diff
@@ -98,7 +106,13 @@ describe("json-rpc-data", () => {
   });
 
   describe("toString()", () => {
-    it("should stringify without arguments", () => {
+    it("should work with uneven length", () => {
+      const d= new Data("0x123");
```
davidmurdoch: `0x123` is invalid for the Data type; it _is_ valid for the Quantity type though.

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,73 @@
...
+  return input;
+}
+
+const TYPE_TO_PARSER_MAP = {
+  number: parseAndValidateNumberInput,
+  bigint: parseAndValidateBigIntInput,
+  string: parseAndValidateStringInput
+};
+
+export function getParseAndValidateFor(input: number | bigint | string | Buffer): any {
```
MicaiahReid: `input`'s type can be replaced with `JsonRpcInputArg`

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const VALIDATE_REGEX = /^0x[0-9a-f]*$/i;
+export function parseAndValidateStringInput(input: string): Buffer {
+  if (!VALIDATE_REGEX.test(input)) {
+    throw new Error(`Cannot wrap string value "${input}" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`);
+  }
+  //preserve length
+  let hexValue = (<string>input).slice(2);
+
+  if (hexValue.length & 1) {
```
MicaiahReid, line 63: From what I can tell this ` & 1` business is to see if it's and odd length? It's obvious enough if you give it much thought, but I think it's worth explaining they "why" in a comment.

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export function parseAndValidateBufferInput(input: Buffer): Buffer {
+  return input;
+}
+
+export function parseAndValidateNumberInput(input: number): Buffer {
+  if (input < 0) {
+    throw new Error("Cannot wrap a negative value as a json-rpc type");
+  }
+  if (<number>input % 1) {
```
MicaiahReid: Is casting to `number` necessary in this function?

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,73 @@
+import {bigIntToBuffer} from "../../utils/bigint-to-buffer";
```
MicaiahReid: I would love to see jsdoc comments on this whole file :)

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,107 @@
...
 export class Data extends BaseJsonRpcType {
-
-  constructor(value: string | Buffer, private _byteLength?: number) {
+  constructor(value: JsonRpcDataInputArg, private _byteLength?: number) {
     super(value);
+
     if (typeof value === "bigint") {
       throw new Error(`Cannot create a ${typeof value} as a Data`);
     }
     if (_byteLength !== undefined) {
```
MicaiahReid: I think this check is redundant because we check it in the `validateByteLength` function

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,107 @@
...
+  private static padString(value: string, byteLength: number | undefined) {
+    let paddedString;
+    const charLength = byteLength * 2
+  
+    if (byteLength === undefined || charLength === value.length) {
+      return value;
+    }
+  
+    // (desired byte count - actual byte count) * 2 characters per byte
+    const padCharCount = (byteLength - value.length / 2) * 2;
```
MicaiahReid:
```suggestion
    const padCharCount = charLength - value.length;
```
Check me on this, but I _think_ this is the same and saves a little math.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,107 @@
...
+    if (padCharCount === 0) {
```
MicaiahReid: I think this case can never be true because of the `charLength === value.length` check above. Probably can be removed. Or, if it helps readability we could remove it above and keep it here.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,107 @@
...
     return new Data(value, byteLength);
   }
+
+  private static validateByteLength(byteLength?: number) {
+    if (byteLength !== undefined && (typeof byteLength !== "number" || byteLength < 0 || !isFinite(byteLength))) {
+      throw new Error(`byteLength must be a number greater than or equal to 0, provided: ${byteLength}`);
+    }
+  }
+  
+  private static padString(value: string, byteLength: number | undefined) {
```
MicaiahReid: `padString` _feels_ like a weird name to me. I don't know if it's a formal definition, but to me padding is additive, but this function can go either way. Maybe `padOrSliceString`? Or `correctStringLength`? ¯\_(ツ)_/¯

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,107 @@
...
+      paddedString = value;
+    } else if (padCharCount > 0) {
+      paddedString = "0".repeat(padCharCount) + value;
+    } else {
+      paddedString = value.slice(0, charLength);
+    }
+    return paddedString;
+  }
+  
+  private static padBuffer(value: Buffer, byteLength: number | undefined) {
```
MicaiahReid: Same thing for this name as for the `padString` function

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,79 +1,103 @@
-import { bufferToBigInt, BUFFER_EMPTY } from "../../utils";
-import { BaseJsonRpcType } from "./json-rpc-base-types";
-// TODO(perf): rewrite this stuff since it isn't really caching anything
+import { bufferToBigInt } from "../../utils/buffer-to-bigint";
+import { BaseJsonRpcType, JsonRpcInputArg } from "./json-rpc-base-types";
+const BUFFER_EMPTY = Buffer.alloc(0);
```
MicaiahReid: Do we want to take this from `constants.ts`?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,79 +1,103 @@
...
+      if (this.bufferValue[firstNonZeroByte] !== 0) break;
+    }
+
+    let val = this.bufferValue.toString("hex", firstNonZeroByte);
+
+    if (val.length === 0 || val === "0") {
+      // RPC Quantities must represent `0` as `0x0`
+      return this._nullable ? null : Quantity.DEFAULT_STRING_VALUE;
+    }
+    if (val[0] === "0") {
```
MicaiahReid: Can you explain this line? It isn't clear to me why we're doing this

MicaiahReid (commented):
Your code is very readable and easy to follow, great work! Since this is still in a draft state, I've reviewed the base types, data, and quantity implementations and not all of the files that actually implement them yet.

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: Nope! I think earlier this was in a big ol' case statement, with union type.

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: Good call

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: Yep - I've added a comment to explain.

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: Done!

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: Yes!

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: Good call!

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: Yes, correct!

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: Agreed - I changed to "bufferToFixedByteLength" and "stringToFixedByteLength". (I originally has stringToFixedLength, but it was strange that we then passed in a _byte_length rather than char length. I made the string and buffer signatures match, even though it made sense to pass byte length to bufferToFixedLength).

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: I did, but then we get circular references :/

I thought about removing the static RPC-JSON types from constants, and moving them to static properties on the RPC-JSON classes themselves.

Keen to hear people's thoughts?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: Yeah, that was wrong. I've fixed it up.

`Quantity.from(Buffer.alloc(0), true).toString()` would return `null` :/ which is now covered in this test: https://github.com/trufflesuite/ganache/blob/74a17afe9d084e888d4b4895690ab0af859b9013/src/packages/utils/tests/json-rpc-quantity.test.ts#L54

src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
```diff
@@ -58,12 +58,12 @@ export default class TransactionManager extends Manager<NoOp> {
     if (!fallback.isValidForkBlockNumber(blockNumber)) return null;
 
     const extra: GanacheRawExtraTx = [
-      Data.from(tx.from, 20).toBuffer(),
+      Data.toBuffer(tx.from, 20),
```
davidmurdoch: Looks like this one should be `Address.toBuffer(tx.from)`

src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
```diff
@@ -235,7 +235,8 @@ describe("api", () => {
         const promises = ids.map(id =>
           send("evm_revert", [id]).then(result =>
             assert.strictEqual(result, false)
-          )
+          ).catch(() => {})
+          // either result is false, or it throws
```
davidmurdoch: Why are we catching the assertion error here? Doesn't this only ensure this test can never fail?

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -408,7 +408,7 @@ describe("@ganache/ethereum-transaction", async () => {
       tx.signAndHash(privKeyBuf);
       assert.strictEqual(
         tx.hash.toString(),
-        "0x078395f79508111c9061f9983d387c8b7bfed990dfa098497aa4d34b0e47b265"
+        "0x1d37ff36e8af71fca608ad0f92b52e741b381663ebfe02d0dd1117a4462c206f"
```
davidmurdoch: At this point in my review i'm unsure why the `tx.hash` would change. I suspect I will find my answer later, but just in case I don't... here is a comment to ask about it :-)

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: To avoid the circular refs you could create a new `buffer-constants.ts` file that the jsp-rpc classes and `constants.ts` could import. Only real benefit of doing it this way instead of moving those constants to the static classes is that there would be less refactoring required.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,132 @@
...
     if (value instanceof Quantity) return value;
-    const q = new Quantity(value);
-    q._nullable = nullable;
-    return q;
+    return new Quantity(value, nullable);
+  }
+
+  constructor(value: JsonRpcInputArg, nullable?: boolean) {
+    super(value);
+    if (typeof(value) === "string" && value === "0x") {
```
davidmurdoch: Is the following faster?

```suggestion
    if (value === "0x") {
```

davidmurdoch: Note: you can put this check before the `super` call if you so desired:

```
    if (typeof(value) === "string" && value === "0x") {
       //       ...
    }
    super(value);
```

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,97 @@
...
+    throw new Error("Cannot wrap a decimal as a json-rpc type");
+  }
+  if (!isFinite(input)) {
+    throw new Error(`Cannot wrap a ${input} as a json-rpc type`);
+  }
+  if (input === 0) {
+    return BUFFER_EMPTY;
+  } else {
+    return uintToBuffer(input as number);
+  }
```
davidmurdoch: I think if `input === 0` is likely to be the most common case it should be first, as it will short-circuit all these other checks.

`if (Object.is(input % 1, 0)) { ... }` will do the work of `isFinite`, `input % 1`, and `input < 0` all at once!

It works because a negative integer module 1 (`-5 % 1`) is `-0`, which doesn't usually mean much (since `-0 === 0`) BUT `Object.is` is even stricter than strict equals, and returns `false`! :exploding_head: 

It might be worth benchmarking? We'll want to keep the nice error messages specific to the value ("decimal" and "negative"), so the existing checks will need to remain, but they could be performed only in the failure case (`Object.is` returns value).

Note: `Object.is` might be super expensive, so there is still a chance this slows everything down!

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Parse and validate a {@link bigint} to {@link Buffer} as internal representation for a JSON-RPC data type.
+ * @param  {bigint} input - must be a positive integer, or null.
+ * @returns {Buffer}
+ */
+export function parseAndValidateBigIntInput(input: bigint): Buffer {
+  if (input < 0n) {
+    throw new Error("Cannot wrap a negative value as a json-rpc type");
+  }
+  return input === 0n ? BUFFER_EMPTY : bigIntToBuffer(input as bigint);
```
davidmurdoch: `input === 0n` should probably be the first check.

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Parse and validate a {@link string} to {@link Buffer} as internal representation for a JSON-RPC data type.
+ * @param  {string} input - must be a hex encoded integer prefixed with "0x".
+ * @returns Buffer
+ */
+export function parseAndValidateStringInput(input: string): Buffer {
+  if (!VALIDATE_REGEX.test(input)) {
+    throw new Error(`Cannot wrap string value "${input}" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`);
+  }
+
+  let hexValue = (<string>input).slice(2);
```
davidmurdoch: Do we need to cast input to a string here?

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  // hexValue must be an even number of hexidecimal characters in order to correctly decode in Buffer.from
+  // see: https://nodejs.org/api/buffer.html#buffers-and-character-encodings
+  if (hexValue.length & 1) {
+    hexValue = `0${hexValue}`;
+  }
+
+  return Buffer.from(hexValue, "hex");
```
davidmurdoch: Is this function a hot path? I bet the regex check makes it much slower than it needs to be; it's probably a bit more efficient (i'd guess about 10-20%) to loop over the input string "manually", checking and converting the hex pairs to a uint as we do, and then set that value on a preallocated Buffer instance.

Think it's worth pursuing?

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
```diff
@@ -1,116 +1,43 @@
...
+    this.bufferValue = parseAndValidate(value);
   }
 
   toString(): string | null {
-    let str = strCache.get(this);
-    if (str === void 0) {
-      str = "0x" + toStrings.get(this)();
-      strCache.set(this, str);
+    if (this.bufferValue == null) {
+      return <null>this.bufferValue;
```
davidmurdoch: Any reason not to do:

```suggestion
      return null;
```

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,132 @@
...
+    }
+
+    let result: number;
+    if (length > 6) {
+      const trimmedBuffer = firstNonZeroByte === 0 ? this.bufferValue : this.bufferValue.slice(this.bufferValue);
+      result = Number(bufferToBigInt(trimmedBuffer));
+    } else {
+      result = this.bufferValue.readUIntBE(firstNonZeroByte, length);
+    }
+    if (!Number.isSafeInteger(result)) {
```
davidmurdoch: I think this check only needs to occur in the `length > 6` case.


I also wonder if `Number.isSafeInteger` is overkill, as it checks the lower bound as well, and the `result` here can never be negative (maybe comparing to `Number.MAX_SAFE_INTEGER` would be marginally faster?). Up to you if want to test it or not, I'm just writing this since I thought of it. I doubt `Number.isSafeInteger` is super slow.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,132 @@
...
   }
+
   valueOf(): bigint {
-    const value = this.value;
-    if (value === null) {
-      return value as null;
-    } else if (value === undefined) {
-      return value as undefined;
+    if (this.bufferValue == null) {
+      return <null>this.bufferValue;
```
davidmurdoch: `return null` is probably fine here.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,132 @@
...
+    }
+    return firstNonZeroByte;
+  }
+
+  static toBuffer(value: JsonRpcInputArg, nullable?: boolean): Buffer {
+    return Quantity.from(value, nullable).toBuffer();
+  }
+
+  static toString(value: JsonRpcInputArg, nullable?: boolean): string {
+    return Quantity.from(value, nullable).toString();
```
davidmurdoch, line 139: I'd love to see if these static helpers can benefit by using a shared buffer instance, since the underlying Buffer gets thrown away right now.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,132 @@
...
-    } else if (this.value == null) {
-      return "0x";
-    } else {
-      return super.toString();
+    if (this.bufferValue == null) {
+      return this._nullable? this.bufferValue : Quantity.ZERO_VALUE_STRING;
     }
+
+    let firstNonZeroByte = 0;
+    for (firstNonZeroByte = 0; firstNonZeroByte < this.bufferValue.length; firstNonZeroByte++) {
```
davidmurdoch: Should we use the new `this.findFirstNonZeroByteIndex` function here?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,132 @@
...
+    const firstNonZeroByte = this.findFirstNonZeroByteIndex();
+
+    const length = this.bufferValue.length - firstNonZeroByte;
+    if (length === 0) {
+      return 0;
+    }
+
+    let result: number;
+    if (length > 6) {
+      const trimmedBuffer = firstNonZeroByte === 0 ? this.bufferValue : this.bufferValue.slice(this.bufferValue);
```
davidmurdoch: Buffer's `slice` is deprecated as of Node v18 (https://nodejs.org/api/buffer.html#bufslicestart-end), we need to transition to `subarray`.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,132 @@
...
-  public toBigInt(): bigint | null {
-    const value = this.value;
+    const firstNonZeroByte = this.findFirstNonZeroByteIndex();
 
-    // TODO(perf): memoize this stuff
-    if (Buffer.isBuffer(value)) {
-      const bigInt = bufferToBigInt(value);
-      return bigInt == null ? (this._nullable ? null : 0n) : bigInt;
+    if (firstNonZeroByte > 0) {
+      return this.bufferValue.slice(firstNonZeroByte);
```
davidmurdoch: Buffer's `slice` is deprecated as of Node v18 (https://nodejs.org/api/buffer.html#bufslicestart-end), we need to transition to `subarray`.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,103 @@
...
     return new Data(value, byteLength);
   }
+
+  private static validateByteLength(byteLength?: number) {
+    if (byteLength !== undefined && (typeof byteLength !== "number" || byteLength < 0 || !isFinite(byteLength))) {
+      throw new Error(`byteLength must be a number greater than or equal to 0, provided: ${byteLength}`);
+    }
+  }
+
+  private static stringToFixedByteLength(value: string, byteLength: number | undefined) {
```
davidmurdoch: Should we allow these `byteLength`s to be `undefined` in the `toFixed*` functions? Perhaps the caller should check if the length is undefined and then conditionally call the function? The name seems mismatched (since it doesn't always return a fixed length).

davidmurdoch (changes requested):
This is all so so much better!

I'm not 100% through everything, but since I've got so many comments piled up I thought I should send it along sooner rather than later!

src/chains/ethereum/transaction/tests/index.test.ts
jeffsmale90: Ah, probably should have added a comment on that - the hashes have changed because we now respect bytelength on toBuffer. There's some fixed length buffers fed into the hash function, so previously they were trimmed.

src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
```diff
@@ -621,7 +621,7 @@ describe("api", () => {
                 },
                 {
                   junk: "123",
-                  error: `cannot convert string value "123" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "123" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 },
                 {
                   junk: {},
```
MicaiahReid, line 646: @jeffsmale90 this file has a bunch of TODOs with linked issues. This PR fixes them, would like to uncomment these TODOs and put in the appropriate error, or would you like me to?

Also, can we mark this PR as fixing these linked issues?

MicaiahReid (changes requested):

src/chains/ethereum/transaction/tests/index.test.ts
davidmurdoch: Oooohhh. This might be a mistake. IIRC Buffers are supposed to be compact when hashing (leading 0s removed). I might be wrong, but we need to double check!

jeffsmale90: I looked further into this, the "offending" statement is here: https://github.com/trufflesuite/ganache/blob/d1b9bbda7bcc3ca12973c93313d78365e8b82c98/src/chains/ethereum/transaction/src/access-lists.ts#L30

If I remove the bytelength argument, making it (previously this was ignored anyways):
`const addressBuffer = Data.toBuffer(item.address);`

The hash is `0x078395f79508111c9061f9983d387c8b7bfed990dfa098497aa4d34b0e47b265` which is the previously expected value.

v, r, s all use `Quantity` type, so will explicitly strip leading 0 bytes however.

davidmurdoch: > The [EIP-2718](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2718.md) `TransactionPayload` for this transaction is `rlp([chainId, nonce, gasPrice, gasLimit, to, value, data, accessList, signatureYParity, signatureR, signatureS])`.
> ...
> For the transaction to be valid, `accessList` must be of type `[[{20 bytes}, [{32 bytes}...]]...]`, where `...` means "zero or more of the thing to the left".
Per https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2930.md

Looks like we _do_ need to ensure a correct `byteLength`, but the `byteLength` should be `20` (so we can use `Address.toBuffer` here?).

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: I'm a little mixed about this, as although it's only used internally, having nice, useful errors is helpful.

On balance, given we have typing on the byteLength parameter, I guess removing the validation is reasonable.

src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
jeffsmale90: Yeah, that was a bit of a half effort. 

The change here is that `Quantity` now validates inputs, so `evm_revert` doesn't return false, rather it throws.

Thinking further about this, I think we should just remove -1 (as that's an invalid Quantity value), and rather test the specific values that are disallowed by this function (ie 0, based on this comment: https://github.com/trufflesuite/ganache/blob/c4be4b7a278d5645c26b4499aecf18d9c2410492/src/chains/ethereum/ethereum/src/blockchain.ts#L875)

src/chains/ethereum/transaction/tests/index.test.ts
jeffsmale90: I've added a comment to this effect - as well as enforcing that the accesslist storagekey part is 32 bytes

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: Definitely seems unnecessary to check the type _and_ value 🤷 

Regarding the super call - nice! I thought that super() needed to be the first statement in the constructor, I've learned it's a little more nuanced than that:
>A 'super' call must be the first statement in the constructor to refer to 'super' or 'this' when a derived class contains initialized properties, parameter properties, or private identifiers.

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: So this is real interesting, and I had a bit of a read about -0. There's apparently only two ways to differentiate -0 / 0:

`Object.is(-0, 0)`, as you suggested and `1/0 === 1/-0` 🤷 

Both of these are _marginally_ slower than what I have above, so I'm going to leave the validation as is, but move the zero check up as you suggested. Sound good?

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: I did some investigation, and found that the fastest method is actually to check the string starts with 0x, and assert that the resulting buffer length is correct (Buffer.from will return an empty buffer if the string doesn't conform to hex encoding - I added a comment explaining this too)

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: Oh, this was changed in Typescript 4.6: https://github.com/microsoft/TypeScript/commit/b7fee7f1ee78d05c793409069ddd00105eddad26

We are on 4.5.2, so could consider a major bump. What's your thoughts?

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
jeffsmale90: The previous implementation explicitly distinguished between null and undefined in at least one case: https://github.com/trufflesuite/ganache/blob/c4be4b7a278d5645c26b4499aecf18d9c2410492/src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts#L75

There's a test that asserts that `nonce === undefined`:

https://github.com/trufflesuite/ganache/blob/c4be4b7a278d5645c26b4499aecf18d9c2410492/src/chains/ethereum/ethereum/tests/transaction-pool.test.ts#L460

Do you have thoughts on whether we _should_ distinguish?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: I did test this, and the two are basically the same in terms of performance. Good call moving the check into the length > 6 case though!

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: I started reading through nodejs Buffer pooling implementation, and I have some thoughts about how we can benefit from bypassing this and providing our own pooling, potentially in all cases, not just the statics.

I'd like to explore this outside of this PR, as it could take some time!

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: Agree - that makes a bunch of sense

src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
jeffsmale90: Yeah - awesome. I wanted to get at least vague buy in before I jumped into those. I'll pick those up in here.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: As above - I'm unsure of the implications, happy to defer to your expertise though.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: I've moved them to statics as a separate commit so that we can revert easily if desired - keen to hear your thoughts on this approach: https://github.com/trufflesuite/ganache/pull/2983/commits/93f6d565a7e617d5fbbaaaa331beb8ef24d82ea7

jeffsmale90:
Fixes #2725 #2728 #2857

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: I like it!

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,104 @@
...
+  // hexValue must be an even number of hexidecimal characters in order to correctly decode in Buffer.from
+  // see: https://nodejs.org/api/buffer.html#buffers-and-character-encodings
+  if (hexValue.length & 1) {
+    hexValue = `0${hexValue}`;
+  }
+  const byteLength = Math.ceil(input.length / 2 - 1);
+
+  const _buffer = Buffer.from(hexValue, "hex");
+  if (input.slice(0, 2).toLowerCase() !== "0x" || _buffer.length !== byteLength) {
+    // Buffer.from will return an empty buffer if the input does not conform to hexidecimal encoding
```
davidmurdoch:
```suggestion
    // Buffer.from will return an empty buffer if the input does not conform to hexadecimal encoding
```

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: :+1:

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: I just think it is more readable is all.

src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
```diff
@@ -229,15 +229,10 @@ describe("api", () => {
...
+      it("evm_revert returns false for out-of-range subscriptionId", async () => {
         const { send } = context;
-        const ids = [-1, Buffer.from([0])];
-        const promises = ids.map(id =>
-          send("evm_revert", [id]).then(result =>
-            assert.strictEqual(result, false)
-          )
-        );
-        await Promise.all(promises);
+        const result = await send("evm_revert", [0]);
```
davidmurdoch, line 235: Since we don't need to test `-1` here anymore, as `Quantity` can never store negative numbers (yay!) there is a check in blockchain.ts, `// snapshot ids can't be < 1, so we do a quick sanity check here` , that can now be removed,

src/chains/ethereum/transaction/src/runtime-transaction.ts
```diff
@@ -24,7 +22,7 @@ import { TransactionReceipt } from "./transaction-receipt";
 import { Address } from "@ganache/ethereum-address";
 
 export const toValidLengthAddress = (address: string, fieldName: string) => {
-  const buffer = Data.from(address).toBuffer();
+  const buffer = Address.toBuffer(address);
   if (buffer.byteLength !== Address.ByteLength) {
```
davidmurdoch: I don't think `Data.from(address).toBuffer()` should be changed to `Address.toBuffer(address)` in this case. I think this function is validating that the given string is already of the proper length (`Address.ByteLength`) and changing it to use `Address.toBuffer` will cause it to _always_ result in having a `byteLength` of `Address.ByteLength`.

The `toValidLengthAddress` function is used when validating the `from` and `to` address of a JSON transaction object. I believe we need to ensure that the `from` and `to` addresses are always 40 hex characters.

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
davidmurdoch: I think we should tighten up our types and enable `strictNullChecks` in our tsconfig.json some day Soon:tm:. But today is probably not that day. :-)

In this case maybe we should delete `valueOf`? Other than that test it looks like it is only _explicitly_ used in blockchain.ts's `revert`; and we can probably change that to not need it without much effort.

I originally intended on using `valueOf` to do arithmetic on Quantity types without having to call `toBigInt()` first: e.g.:

```javascript
// `valueOf` magic
Quantity.from("0x5") + Quantity.from(Buffer.from(2)) // result: 7n !
```

Typescript didn't like it though, but the `valueOf` function remains. I'm not sure if there are any implicit usages remaining.

I vote to remove `valueOf` if it's trivial to do so!

davidmurdoch (changes requested):

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
```diff
@@ -1,116 +1,43 @@
...
-    | number
-    | bigint
-    | string
-    | Buffer
-> {
+export class BaseJsonRpcType {
   public [Symbol.toStringTag]: string;
 
-  protected value: T;
+  protected bufferValue: any | null;
```
davidmurdoch: Should this be?

```suggestion
  protected bufferValue: Buffer | null;
```

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: I'll have to get that right one of these times...

src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
jeffsmale90: Nice! I like deleting code.

src/chains/ethereum/transaction/src/runtime-transaction.ts
jeffsmale90: Good call - I've changed this to `Data.toBuffer()` instead.

jeffsmale90:
I've addressed all of the comments on here (unless I missed something) and rebased onto develop.

I still need to address the Address length check, and remove valueOf. I ran out of time so will get to this in the morning 🤞

src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
jeffsmale90: This _does_ cover the case of snapshotId = 0 though, so that piece needs to stay :(

davidmurdoch: It looks like it was never necessary even before this change, as an number `<= 0` will return `false` as since a snapshot with that id can't exist anyway.

```
    // snapshot ids can't be < 1, so we do a quick sanity check here
    if (rawValue < 1n) {
      return false;
    }

    const snapshots = this.#snapshots;
    const snaps = snapshots.snaps;
    const snapshotIndex = Number(rawValue - 1n);
    const snapshot = snaps[snapshotIndex];

    if (!snapshot) {
      return false;
    }
```

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,105 @@
...
+    return noopParseAndValidate;
+  }
+
+  const type = typeof input;
+  const parser = TYPE_TO_PARSER_MAP[type];
+  if (parser === undefined) {
+    throw new Error(`Cannot wrap a "${type}" as a json-rpc type`);
+  }
+
+  return parser;
```
davidmurdoch: What the reason for returning a parse function instead of calling the parse function with the given `input` immediately?

src/packages/utils/src/things/json-rpc/input-parsers.ts
davidmurdoch: Missed this comment last time around. the latest updates sound good to me!

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,95 @@
...
+
+  private static stringToFixedByteLength(value: string, byteLength: number) {
+    const desiredCharLength = byteLength * 2
+
+    if (desiredCharLength === value.length) {
+      return value;
+    }
+
+    const padCharCount = desiredCharLength - value.length;
+    let fixedLengthValue;
```
davidmurdoch:
```suggestion
    let fixedLengthValue: string;
```

src/chains/ethereum/transaction/src/runtime-transaction.ts
davidmurdoch: FYI: latest code still shows this line as `Addres.toBuffer(address)`

jeffsmale90:
I've created https://github.com/trufflesuite/ganache/issues/3169 to cover removing the valueOf functions, because removing them is not trivial, but I think it would be worth doing.

src/chains/ethereum/transaction/src/runtime-transaction.ts
jeffsmale90: Yeah, I ran into an issue last night where this was causing test failures, so left it for today. That's why I converted to draft, but I had already clicked the "re-request" when I realised.

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
jeffsmale90: I created this case https://github.com/trufflesuite/ganache/issues/3169 to cover removing valueOf, as there's at least some implicit usage of the function that make it slightly less trivial to remove.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: Fixed this in a nother commit.

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: I was thinking to keep that separated to support testing, but in retrospect we are still able to test the parsing effectively.

I've moved the selection of the parser back into the constructor of the base rpc json object where I think it better fits.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: We're on 4.6.4 now :-)

src/packages/utils/src/things/json-rpc/input-parsers.ts
davidmurdoch: Created https://github.com/trufflesuite/ganache/issues/3173 so we can follow up on this at a later date.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,95 @@
...
+
+    if (this.bufferValue == null) {
+      return BUFFER_EMPTY;
+    }
+
+    if (length == undefined || length === this.bufferValue.length) {
+      return this.bufferValue;
+    }
+
+    return length === undefined? this.bufferValue: Data.bufferToFixedByteLength(this.bufferValue, length);
```
davidmurdoch: Can `length` ever be `undefined` at this point?

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,95 @@
...
-  public toString(byteLength?: number): string {
-    if (byteLength === undefined) {
-      byteLength = this._byteLength;
+
+  public toBuffer(byteLength?: number): Buffer {
+    const length = byteLength || this._byteLength;
+
+    if (this.bufferValue == null) {
+      return BUFFER_EMPTY;
+    }
```
davidmurdoch: The following is probably faster in case where `this.bufferValue` is null/undefined, and the same speed when it is not.

```suggestion
    if (this.bufferValue == null) {
      return BUFFER_EMPTY;
    }
    
    const length = byteLength || this._byteLength;
```

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,95 @@
...
+
+    if (length == undefined || length === this.bufferValue.length) {
+      return this.bufferValue;
+    }
+
+    return length === undefined? this.bufferValue: Data.bufferToFixedByteLength(this.bufferValue, length);
+  }
+
+  public static from(value: JsonRpcDataInputArg | Data, byteLength?: number) {
+    if (value instanceof Data) {
```
davidmurdoch: In theory this can never happen since `JsonRpcDataInputArg` doesn't allow it. I don't think we should allow passing Data/Quantity to Data/Quantity.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,95 @@
...
+      return value;
```
davidmurdoch: I know i just commented this, but It's now gone from my view, so sorry if you are seeing it twice:

`JsonRpcDataInputArg` doesn't allow Data (or Quantity), so this can't be true (in theory). I don't think we should allow for this case.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,135 @@
...
-    const q = new Quantity(value);
-    q._nullable = nullable;
-    return q;
+    return new Quantity(value, nullable);
+  }
+
+  constructor(value: JsonRpcInputArg, nullable?: boolean) {
+    super(value);
+    if (value === "0x") {
+      throw new Error('Cannot wrap "0x" as a json-rpc Quantity type; Quantity must contain at least one digit');
```
davidmurdoch: In theory this shouldn't ever happen right? In any case, I don't think "digit" is the right word here, since it can be non-digits (if we are referring to `value` itself)

jeffsmale90: It _shouldn't_ happen, but it _can_, this is a fix for https://github.com/trufflesuite/ganache/issues/2857

Agreed the message isn't perfect - what do you think of `Cannot wrap "0x" as a json-rpc Quantity type; strings must contain at least one hexadecimal symbol.`? It's consistent with other validation messages, if a little awkward.

jeffsmale90: I also added a test to target this specific validation case.

davidmurdoch: Ah, that makes sense!

Does the word "character" read better?

```
Cannot wrap "0x" as a json-rpc Quantity type; strings must contain at least one hexadecimal character.
````

Comment (jeffsmale90) [reply src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts]: Yeah I went back and forward between those two. Character / Symbol... 

I think both are a little awkward, so I'll defer to you :)

File: src/chains/ethereum/ethereum/tests/api/eth/call.test.ts [hunk 42]
```diff
@@ -663,11 +669,11 @@ describe("api", () => {
...
-                  error: `cannot convert string value "[object Object]" into type \`Quantity\`; strings must be hex-encoded and prefixed with "0x".`
+                  error: `Cannot wrap string value "[object Object]" as a json-rpc type; strings must be hex-encoded and prefixed with "0x".`
                 }
                 // TODO: add this back once https://github.com/trufflesuite/ganache/issues/2725 is closed
```
MicaiahReid: @jeffsmale90 there were 20 occurrences in all of this or similar TODO's I think. Sorry, I should have been more clear in the last comment. 

This code is a bit repetitive, I probably should have cleaned this up more, I was very done with this feature at the time, haha.

Some sections have 3 TODOs, some only 2. This should correctly reflect the expected behavior for that section.

jeffsmale90: Oops! Sorry for missing that! 😅

jeffsmale90:
@davidmurdoch I added a couple more tests around nullable `Quantities` to ensure that we cover those cases you were questioning.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,91 @@
...
-    throw new Error(`byteLength must be a number greater than or equal to 0, provided: ${byteLength}`);
-  }
-}
+export type JsonRpcDataInputArg = string | Buffer;
 
 export class Data extends BaseJsonRpcType {
+  public static Empty = Data.from(BUFFER_EMPTY);
 
-  constructor(value: string | Buffer, private _byteLength?: number) {
+  constructor(value: JsonRpcDataInputArg, private _byteLength?: number) {
```
tenthirtyone, line 9: I am totally fine with defining the properties from the constructor, but because the rest of the repo follows a pattern of defining properties separate from the constructor I would suggest something like this:

``` typescript
export class Data {
private _byteLength?: number | null;

constructor(value: JsonRpcDataInputArg, byteLength?: number) {

this._byteLength = byteLength
  }
}
```

It also makes it more explicit. It may not be obvious to someone who doesn't know the property will be set on the class from the constructor.

Check out https://github.com/trufflesuite/ganache/blob/develop/src/packages/core/src/server.ts#L110 as an example class to pattern/copy from.

I think this is also consistent with the way `json-rpc-base-types.ts` is written as well so it may be a good opportunity to refactor this

tenthirtyone, line 9: Would a unit test like this pass:
``` javascript
    it("does not handle strings that are not prefixed with 0x", () => {
      const input = "bad input";
      const data = new Data(input);
    });
```

but something like this would pass:
``` javascript
    it("does not handle strings that are not prefixed with 0x", () => {
      const input = "bad input";
      const data = new Data(Buffer.from(input));
    });
```


I'm not familiar with the implementation details but the constructor seems tuned for the 'happy path'. Would it make sense to bottleneck the constructor more to test for the input it is looking for and handle the fail cases explicitly?

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,91 @@
...
-    value: T,
-    byteLength?: number
-  ) {
+
+    const length = byteLength || this._byteLength;
+    if (length == undefined || length === this.bufferValue.length) {
+      return this.bufferValue;
+    }
+
+    return length === undefined? this.bufferValue: Data.bufferToFixedByteLength(this.bufferValue, length);
```
tenthirtyone: Is `this.buffervalue...` unreachable code because of the check on L38?

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,91 @@
...
-      return `0x${str}`;
-    }
-  }
-  public static from<T extends string | Buffer = string | Buffer>(
-    value: T,
-    byteLength?: number
-  ) {
+
+    const length = byteLength || this._byteLength;
+    if (length == undefined || length === this.bufferValue.length) {
```
tenthirtyone, line 38: Why are we coercing length to check for undefined?

Since this is a perf pr I want to share that `===` is about 1/3 faster than `==` _when coercion occurs_ otherwise they are about equivalent but we are still talking μseconds. 

If the types are the same, the algorithm is exactly the same, if not `==` will attempt to coerce the value. 

So, if `length` is `undefined` `===` and `==` are equivalent. If `length` is not `undefined` coercion occurs. So, to be more clear, we are using double equals here to shortcut explicitly checking `length === undefined || length === null` and for every instance where `length` is not `typeof undefined` we are attempting to coerce it. 

https://dorey.github.io/JavaScript-Equality-Table/

I saw a bunch of " == " that could be opti'd for perf if this is of benefit. 

Alternatively, how about 

``` javascript
!length
```
as
``` javascript
!undefined && !null === true
```

or would `0` mess with that?

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -0,0 +1,70 @@
...
+const BUFFER_EMPTY = Buffer.allocUnsafe(0);
+
+export type JsonRpcInputArg = number | bigint | string | Buffer;
+
+/**
+ * Parse and validate a {@link number} to {@link Buffer} as internal representation for a JSON-RPC data type.
+ * @param {number} input - must be a positive, finite integer, or null.
+ * @returns {Buffer}
+ */
+export function parseAndValidateNumberInput(input: number): Buffer {
```
tenthirtyone, line 13: This is maybe more a gap in my typescript understanding and ramp up on ganache and a chance for me to learn. 

Because this is handling data that comes from the user, even though we are defining "This must be a number" is it possible that the user-defined input passed down at runtime could be something other than a number? 

e.g. if I did something like sent `eth_getTransactionReceipt` through RPC and passed "I am not a number or valid tx hash" as the payload would that be handled?

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
```diff
@@ -1,57 +1,91 @@
...
+
+    const strValue = this.bufferValue.toString("hex");
+    return `0x${Data.stringToFixedByteLength(strValue, length)}`;
   }
-  public toString(byteLength?: number): string {
-    if (byteLength === undefined) {
-      byteLength = this._byteLength;
+
+  public toBuffer(byteLength?: number): Buffer {
+    if (this.bufferValue == null) {
```
tenthirtyone, line 33: Why are we coercing to check for null?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,135 @@
...
+      return 0;
+    }
+
+    let result: number;
+    if (length > 6) {
+      const trimmedBuffer = firstNonZeroByte === 0 ? this.bufferValue : this.bufferValue.subarray(firstNonZeroByte, length);
+      result = Number(bufferToBigInt(trimmedBuffer));
+
+      if (!Number.isSafeInteger(result)) {
+        console.warn(`0x${this.bufferValue.toString("hex")} is too large - the maximum safe integer value is 0${Number.MAX_SAFE_INTEGER.toString(16)}`);
```
tenthirtyone: I see the warning but I do not see where `result` is being set to `Number.MAX_SAFE_INTEGER.toString(16)` or that value is being returned instead.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -1,87 +1,135 @@
...
+
+    const firstNonZeroByte = this.findFirstNonZeroByteIndex();
+
+    const length = this.bufferValue.length - firstNonZeroByte;
+    if (length === 0) {
+      return 0;
+    }
+
+    let result: number;
+    if (length > 6) {
```
tenthirtyone, line 91: Why 6? 

Is it worth setting a var for the magic number?

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
```diff
@@ -1,116 +1,62 @@
...
-      toStrings.set(this, () => value.toString("hex"));
-      bufCache.set(this, value);
-      self[Symbol.toStringTag] = "Buffer";
+
+  constructor(value: JsonRpcInputArg) {
+    if (value == null) {
+      this.bufferValue = null;
+    } else if (Buffer.isBuffer(value)) {
+      // empty buffer should be treated as null
+      this.bufferValue = value.length === 0 ? null: value;
```
tenthirtyone: Could this be consolidated into something like

``` javascript
if (value === null || (Buffer.isBuffer(value) && value.length === 0) {
  // codepath for nulls
  this.bufferValue = null;
} else {
 // codepath for non-null
 this.bufferValue = value;
}

```

tenthirtyone (changes requested):

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
davidmurdoch: note: looks like we need to run prettier on these changes.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: @tenthirtyone the error message states that  the given value is too largebecausee the maximum _safe_ integer value is MAX_SAFE_INTEGER. It is not implying the maximum value is MAX_SAFE_INTEGER.

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: That second case _would_ pass, but it would be expected to, as `Buffer.from(string)` will use utf8 encoding by default - so we end up passing a byte array into the `Data` constructor.

jeffsmale90: That's a fair point - I've tried to adopt the coding style of the codebase, but am sure to miss some!

Updated to use a private property - also in Quantity#nullable

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: I think that the performance implications here are probably fairly negligible.

I've used == comparator in a couple of places (length == undefined and bufferValue == null). In both cases, I am comparing with the type that they _should_ be (ie when we parse the input, we explicitly set the bufferValue to null, and length should either be defined by the user, or undefined [although they could explicitly set to null]).

I think in the interests of readability, leaving this as is makes sense, especially given that this is negligible cost compared with the other work we are doing. Happy to look into this further, and get some data to validate one way or the other if you disagree.

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: Yes - as discussed, the following files handle type checking and input validation:

https://github.com/trufflesuite/ganache/blob/perf/json-rpc/src/packages/utils/src/things/json-rpc/input-parsers.ts
https://github.com/trufflesuite/ganache/blob/perf/json-rpc/src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts

There's a strict set of requirements for these input types, which adhere to the rules outlined here: https://eth.wiki/json-rpc/API (under "Hex value encoding")

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: as described above regarding coercion for undefined check

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: We return whatever value is parsed, with the warning indicating that the returned value might start to lose precision.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: I've added a comment, so that the reasoning is more clear (it's more clear than a constant IMO, but open to other POV).

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
jeffsmale90: It could be - but it makes the branching more complicated, which I'm not a huge fan of (also we'd need to either check Buffer.isBuffer twice, or store that value).

Open to other POV though!

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: Good find!

src/packages/utils/src/things/json-rpc/json-rpc-data.ts
jeffsmale90: Reverted this change - as the implications are more wide reaching. Still think it's a good change, but I'll defer it for later :)

MicaiahReid (approved):

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
tenthirtyone: Resolving with comment~

I strongly urge adding something like the following to https://github.com/trufflesuite/ganache/blob/develop/src/packages/utils/src/utils/constants.ts

``` javascript
export const BITS_IN_BYTE = 8;
export const MAX_BUFFER_BITS = 48;
export const MAX_BUFFER_BYTES = MAX_BUFFER_BITS / BITS_IN_BYTE; // 6
```

and changing
``` javascript
import { BUFFER_EMPTY, BUFFER_ZERO } from "../../utils/constants";
```

to
``` javascript
import { BUFFER_EMPTY, BUFFER_ZERO, MAX_BUFFER_BYTES } from "../../utils/constants";
```

tenthirtyone (approved):

davidmurdoch (approved):
Yay!

PR #2988, opened 2022-04-26
chore: support node 18

davidmurdoch:
Ganache now supports and is tested on Node versions 12.0.0, 12.x, 14.x, 16.x, and 18.x on operating systems Windows 2019, Ubuntu 18.04, Ubuntu 20.04, and macOS 11, with the exception of Node v18 on Ubuntu 18.04, as Ubuntu 18.04 is not supported by Node v18.

Happy upgrading!


src/packages/core/tests/server.test.ts
```diff
@@ -154,6 +166,7 @@ describe("server", () => {
 
           for (const info of interfaceInfo) {
             const host = getHost(info, interfaceName);
+            console.log(host);
```
```suggestion
```

MicaiahReid (changes requested):
Just one small change!

MicaiahReid (approved):

PR #2989, opened 2022-04-26
Release v7.1.0

davidmurdoch:
This release is intended to re-release v7.0.5, but as v7.1.0. v7.0.5 was supposed to be released with the v7.1.0 version number.

This PR requires that the repository does NOT have a v7.0.5 git tag. I've deleted this tag. The tag should be re-added at the appropriate commit AFTER 7.1.0 is automatically released once this PR is merged. If the tag is added back _before_ we release this PR will end up causing a release of version v7.0.6 :grimacing:.

The reason for this fragility is that there is a feature commit included in v7.0.5 that we need to be counted for v7.1.0 instead. Our automatic releases use the most recent semver git tag to determine the commit range (LATEST_TAG through current HEAD) to consider when determining the semver version kind (minor/patch).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          # we need all commit history so we can merge master into develop
+          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


MicaiahReid (approved):
I hope it works!

PR #3016, opened 2022-05-02
fix: save `evm_mine` blocks before returning

### LINKED ISSUE #3060: Race condition between assignment of "latest" tag and persistence of the latest raw block data causing intermittent test failures
If you wrap [the "should set storage slot and delete after" evm API test](https://github.com/trufflesuite/ganache/blob/917503b0ff9635241cc179bc2763ac94b4dfa859/src/chains/ethereum/ethereum/tests/api/evm/evm.test.ts#L207-L234) in a for loop such that it executes 100 times, you'll observe that it fails intermittently, due to the RPC method handler for `eth_getStorageAt` [throwing because it can't find a block](https://github.com/trufflesuite/ganache/blob/917503b0ff9635241cc179bc2763ac94b4dfa859/src/chains/ethereum/ethereum/src/api.ts#L1669).

This occurs even though the test isn't specifying the block number in its call to `eth_getStorageAt`, meaning it's implicitly using the "latest" tag. This implies that `blockchain.blocks.getEffectiveBlockNumber` is occasionally returning the number for a block that hasn't yet been persisted to the database.

If you refactor the API handler for `eth_getStorageAt` to avoid the use of `blockchain.blocks.getRawByBlockNumber` and instead prefer `blockchain.blocks.get`, the test in question succeeds 100% of the time. While this adequately works around the problem, I don't think that this is a proper fix for the issue, as it seems that the real problem is a lack of consistency between the in-memory and on-disk representations of the block's current head state.

Note that several other RPC method handlers appear to use the same pattern for reading the raw block data, and are therefore likely also affected by this issue.

MicaiahReid:
Though it was rare, sometimes calling `evm_mine` would return before the mined block was actually saved. If you polled for the block that Ganache had claimed to mine directly after calling `evm_mine`, that block could not yet exist.

This change ensures that the block is saved _before_ emitting the `"newHeads"` subscription message and before returning during an `evm_mine`. Fixes #3060.

#### Potential Side Effect
This change does have one potential side effect, though we really doubt anyone should be impacted, and we consider it a bug fix. Sometimes when a block is mined, Ganache waits until the next iteration of the event loop (using [`setImmediate`](https://nodejs.dev/learn/understanding-setimmediate)) to emit the mined block. This is to ensure that the user has time to listen for the new block's `"message"` before Ganache emits it. 

Previously, Ganache would delay emitting this event if  `--miner.instamine="eager"` AND `--miner.blockTime=0`. These settings mean that transactions are saved before returning the transaction (as apposed to returning a hash as the transaction enters the transaction pool in `"strict"` mode), and the miner immediately starts mining transactions as they enter the transaction pool (as opposed to mining a block every `blockTime` seconds).*

Now, Ganache delays emitting this event when `--miner.instamine="eager"` _regardless of how `blockTime` is set_. The instamine mode affects when the transaction is returned to the user, so it should be the driving factor behind delaying the event. If in `--miner.blockTime=0` mode (the default) you somehow relied on the event being emitted at a very specific event tick, this may cause issues for you. For most of us mere mortal developers, this won't make a difference.

_*Note: We'll have a blog post or discussion providing a deeper dive into all of our mining modes soon that should help further clarify `--miner.instamine`, `--miner.blockTime`, and `--chain.vmErrorsOnRPCResponse`._


src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -570,14 +572,15 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     timestamp?: number,
     onlyOneBlock: boolean = false
   ) => {
-    await this.#blockBeingSavedPromise;
```
MicaiahReid, line 573: The miner has its own internal mechanism for making sure only one block can be mined at a time, so I don't think this is necessary. It didn't affect the tests, but this is the scariest part of the change for me :sweat_smile:

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -570,14 +572,15 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
     const nextBlock = this.#readyNextBlock(this.blocks.latest, timestamp);
+    const transactions = await this.#miner.mine(
+      nextBlock,
+      maxTransactions,
+      onlyOneBlock
+    );
+    await this.#blockBeingSavedPromise;
```
MicaiahReid, line 582: Await _after_ mining so we don't return until mining is actually fully completed. I tested this against the convex shutdown repo and it did not show a performance decrease. But this would more have an impact if there is a test that sends many transactions rather than just one big one. I'm open to ways to further test for performance degradation.

MicaiahReid: Regardless, I hope to follow this PR with one that will allow the miner to manage the queue of mining requests and when the blocks being saved actually needs to be awaited. I just wanted to get this fix in sooner so I can feel like I'm making progress :smile:

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: :sweat_smile:

davidmurdoch:
Thoughts on adding tests for this new timing behavior?

MicaiahReid:
> Thoughts on adding tests for this new timing behavior?

I mean our `block is null; didn't correctly mine ${numberOfBlocksToMine} blocks` test _kind of_ tests this, right? :smile: 

Yeah, I'll get some tests added.

MicaiahReid:
@davidmurdoch I've added a test!

I'm instantiating a provider where I specify the db using the `{database: { db: ... }}` flag, and I'm passing in memdown. I'm using [patch-package](https://www.npmjs.com/package/patch-package) to update memdown's `_batch` function to delay 20ms before  saving. This `_batch` function is used by our blockchain to save blocks to the db.

I've created [a branch](https://github.com/trufflesuite/ganache/pull/3102) that has these tests _without_ the actual race condition fixes as a proof that the tests do what I say they should, and that my changes actually fix the bug.

src/chains/ethereum/ethereum/patches/memdown+6.1.1.patch
```diff
@@ -0,0 +1,53 @@
+diff --git a/node_modules/memdown/memdown.js b/node_modules/memdown/memdown.js
```
davidmurdoch: In the near future ganache is very likely to ship with memdown, and this patch will probably break things for us.

I think a better solution would be monkey-patch (or duck punch, if you'd rather :fist_oncoming: :duck:) a MemDOWN's `_batch` instance. We also need to ensure that the `_batch` instead we patch is actually called by the test in question, otherwise a future change might accidentally stop using the patched version, then later regress into a race condition again, but then it will still pass CI because the race condition doesn't happen very often.


Oh, I also thought of an alternative approach that doesn't require `setTimeout`, if you'd rather try this. We can detect when `_batch` is called form without our tests and use this to ensure everything happens in the right order. I haven't thought through it fully, so maybe this is worse than simulating a slow file system.

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/patches/memdown+6.1.1.patch
jeffsmale90: Would it be possible to wrap memdown with a shim?

We could either introduce a delay, or potentially block calls to `_batch` until a handle is called to release the call?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -535,9 +535,11 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     storageKeys: StorageKeys;
     transactions: TypedTransaction[];
   }) => {
-    this.#blockBeingSavedPromise = this.#blockBeingSavedPromise
-      .then(() => this.#saveNewBlock(blockData))
-      .then(this.#emitNewBlock);
+    this.#blockBeingSavedPromise = this.#blockBeingSavedPromise.then(() => {
+      const saveBlockProm = this.#saveNewBlock(blockData);
+      saveBlockProm.then(this.#emitNewBlock);
```
jeffsmale90, line 540: I guess this change is to ensure that the event emission and resolution of #blockBeingSavedPromise all happen in the correct order - would it be worthwhile adding a comment to make it more clear?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -535,9 +535,11 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+      return saveBlockProm;
+    });
 
     await this.#blockBeingSavedPromise;
     return;
```
jeffsmale90: (clearly not your change, but) Is this `return;` statement needed?

src/chains/ethereum/ethereum/patches/memdown+6.1.1.patch
MicaiahReid: I've changed the approach to monkey-patch the memdown instance rather than using patch-package. It seems to work consistently, so let's go with this for now!

src/chains/ethereum/ethereum/tests/api/evm/evm.test.ts
```diff
@@ -132,6 +133,31 @@ describe("api", () => {
...
+        // race condition where a block was mined and returned by evm_mine
+        // before it actually saved to the database. for history, the race
+        // condition issue is documented here:
+        // https://github.com/trufflesuite/ganache/issues/3060
+        memdown._batch = function (array, options, callback) {
+          setTimeout(() => {
+            memdown.prototype._batch(array, options, callback);
+          }, 20);
+        };
+        const options = { database: { db: memdown() } };
```
davidmurdoch: Does this code work? I was thinking it should be more like (totally not tested!):

```javascript
        const db = memdown();
        let customBatchCalled = false;
        db._batch = (...args) => {
            setTimeout(() => {
              customBatchCalled = true;
              Reflect.apply(memdown.prototype._batch, db, args);
            }, 20);
          });
        };
        const options = { database: { db } };
        
        // ... other tests
        
        // make sure monkey patching _batch_ was actually used (protect against future changes in `memdown`):
        assert(customBatchCalled);
```

Also, I wonder how we can determine if 20ms is enough time to wait? :thinking:

MicaiahReid: I've implemented your suggestion!

From my earlier comment: 
>I've created https://github.com/trufflesuite/ganache/pull/3102 that has these tests without the actual race condition fixes as a proof that the tests do what I say they should, and that my changes actually fix the bug.

So running this test with 20ms on a branch without the actual fix allowed us to consistently recreate this race condition. You'll recall that without the slowdown, the race condition causing an error in our tests was rare.

davidmurdoch: Oh cool. That was fast!

I'm just wondering how we can _know_ 20ms is enough? I haven't given it deep consideration, and it's probably okay at 20ms.

davidmurdoch (approved):

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: It sure isn't! Removed it.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Done, does that comment help?

jeffsmale90 (approved):

PR #3032, opened 2022-05-05
fix: only mine one block in interval mining

### LINKED ISSUE #3030: `blockTime > 0` mode mines more than one block in each call to the miner
When `blockTime > 0`, the miner should behave more like a real node in that a new block begins mining every `blockTime` seconds. It should mine a single block, then wait `blockTime` seconds again, regardless of how many transactions are in the pool.

Currently, ganache mines _all_ transactions in the pool, regardless of how many blocks it takes, every time `blockTime` elapses.

MicaiahReid:
In a real Ethereum node, blocks are committed [approximately every 15 seconds](https://ethereum.org/en/developers/docs/blocks/#why-blocks). Ganache allows users to configure how often this mining takes place in this "interval" mining mode with the `--miner.blockTime` flag, and it allows users to "instamine" transactions by setting `--miner.blockTime=0` (the default).

For each block that is made in a real Ethereum node, the transactions that are included on the block is based off of many factors including the number of transactions in the transaction pool, the gas price and gas limit of those transactions, and the block's gas limit. Regardless of the ordering, the number of transactions included on the block is limited by that block gas limit. Then, once a single block is mined, approximately 15 seconds elapses before another block is generated.

In the interval mining mode before this change, Ganache would begin mining after the `blockTime` elapses, but it would mine all transactions in the pool, regardless of how many blocks it would take. This change fixes Ganache's behavior to be more like a real Ethereum node. Now only one block will be mined whenever `blockTime` elapses, even if that means leaving some transactions in the transaction pool until the next block is generated. This change fixes #3030.

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -362,7 +362,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
             unref((this.#timer = setTimeout(next, minerOpts.blockTime * 1e3)));
           // when interval mining, only one block should be mined. the block
           // can, however, be filled
-          const next = () => mineAll(Capacity.FillBlock, true).then(wait);
+          const next = () => {
+            mineAll(Capacity.FillBlock, true).then(wait);
+          };
           wait();
         }
         //#endregion
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -359,7 +359,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
           // interval mining
           const wait = () =>
             // unref, so we don't hold the chain open if nothing can interact with it
-            unref((this.#timer = setTimeout(next, minerOpts.blockTime * 1e3)));
+            (this.#timer = setTimeout(next, minerOpts.blockTime * 1e3));
           // when interval mining, only one block should be mined. the block
           // can, however, be filled
           const next = () => {
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -358,7 +358,6 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         } else {
           // interval mining
           const wait = () =>
-            // unref, so we don't hold the chain open if nothing can interact with it
             (this.#timer = setTimeout(next, minerOpts.blockTime * 1e3));
           // when interval mining, only one block should be mined. the block
           // can, however, be filled
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -353,12 +353,15 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         if (instamine) {
           // insta mining
           // whenever the transaction pool is drained mine the txs into blocks
+          // only one transaction should be added per block
           txPool.on("drain", mineAll.bind(null, Capacity.Single));
         } else {
           // interval mining
           const wait = () =>
             // unref, so we don't hold the chain open if nothing can interact with it
             unref((this.#timer = setTimeout(next, minerOpts.blockTime * 1e3)));
+          // when interval mining, only one block should be mined. the block
+          // can, however, be filled
           const next = () => mineAll(Capacity.FillBlock, true).then(wait);
           wait();
         }
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -586,11 +586,6 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     };
   };
 
-  awaitBlockSaving = async () => {
-    await this.#blockBeingSavedPromise;
-    return;
-  };
-
   #isPaused = () => {
     return (this.#state & Status.paused) !== 0;
   };
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -346,8 +346,10 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
 
         //#region automatic mining
         const nullResolved = Promise.resolve(null);
-        const mineAll = (maxTransactions: Capacity) =>
-          this.#isPaused() ? nullResolved : this.mine(maxTransactions);
+        const mineAll = (maxTransactions: Capacity, onlyOneBlock = false) =>
+          this.#isPaused()
+            ? nullResolved
+            : this.mine(maxTransactions, null, onlyOneBlock);
         if (instamine) {
           // insta mining
           // whenever the transaction pool is drained mine the txs into blocks
@@ -357,7 +359,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
           const wait = () =>
             // unref, so we don't hold the chain open if nothing can interact with it
             unref((this.#timer = setTimeout(next, minerOpts.blockTime * 1e3)));
-          const next = () => mineAll(Capacity.FillBlock).then(wait);
+          const next = () => mineAll(Capacity.FillBlock, true).then(wait);
           wait();
         }
         //#endregion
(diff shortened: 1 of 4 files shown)


src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -346,18 +346,25 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
 
         //#region automatic mining
         const nullResolved = Promise.resolve(null);
-        const mineAll = (maxTransactions: Capacity) =>
-          this.#isPaused() ? nullResolved : this.mine(maxTransactions);
+        const mineAll = (maxTransactions: Capacity, onlyOneBlock = false) =>
```
MicaiahReid, line 349: update the mineAll function to actually utilize `blockchain.mine`'s `onlyOneBlock` flag

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -346,18 +346,25 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
           // interval mining
           const wait = () =>
             // unref, so we don't hold the chain open if nothing can interact with it
-            unref((this.#timer = setTimeout(next, minerOpts.blockTime * 1e3)));
-          const next = () => mineAll(Capacity.FillBlock).then(wait);
+            (this.#timer = setTimeout(next, minerOpts.blockTime * 1e3));
+          // when interval mining, only one block should be mined. the block
+          // can, however, be filled
+          const next = () => {
+            mineAll(Capacity.FillBlock, true).then(wait);
```
MicaiahReid, line 366: When interval mining, pass flag to only mine one block.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -346,18 +346,25 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+            : this.mine(maxTransactions, null, onlyOneBlock);
         if (instamine) {
           // insta mining
           // whenever the transaction pool is drained mine the txs into blocks
+          // only one transaction should be added per block
           txPool.on("drain", mineAll.bind(null, Capacity.Single));
         } else {
           // interval mining
           const wait = () =>
             // unref, so we don't hold the chain open if nothing can interact with it
```
```suggestion
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -346,18 +346,24 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+            : this.mine(maxTransactions, null, onlyOneBlock);
         if (instamine) {
           // insta mining
           // whenever the transaction pool is drained mine the txs into blocks
+          // only one transaction should be added per block
           txPool.on("drain", mineAll.bind(null, Capacity.Single));
         } else {
           // interval mining
           const wait = () =>
-            // unref, so we don't hold the chain open if nothing can interact with it
```
MicaiahReid, line 358: `unref` would prevent promises waiting on the blockchain from actually being awaited. i.e.
```TypeScript
await new Promise(resolve => {
  blockchain.on("block", () => {
    resolve(void 0);
  });
);
```
would not be properly awaited. We already kept track of the latest timer in `this.#timer`, and we `clearTimeout(this.#timer)` when we `stop` the blockchain, so we can unref with no negative side affects.

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -500,7 +500,14 @@ export default class Miner extends Emittery<{
 
   #reset = () => {
     this.#origins.clear();
-    this.#priced.clear();
+    const priced = this.#priced;
```
MicaiahReid, line 512: This is a slightly hacky fix.

When the priced heap is `reset`, meaning we're clearing out the heap
and origins list to be set again when the miner is called, loop over
the priced heap transactions and "unlock" them (set `tx.locked = false`)

The real fix would include fixing the use of `locked`, as it's
currently overloaded to mean "is in priced heap" and also "is being
mined".

src/chains/ethereum/ethereum/tests/blockchain.test.ts
```diff
@@ -0,0 +1,114 @@
...
+              off();
+              resolve(true);
+            }
+          } catch (e) {
+            off();
+            reject(e);
+          }
+        });
+      });
+      const success = (await assertBlocks) as any;
```
davidmurdoch: you can remove the `as any` by specifying the type the Promise above will return:

```
const assertBlocks = new Promise<boolean>((resolve, reject) => {
```

src/chains/ethereum/ethereum/tests/blockchain.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      assert.equal(success, true, success);
```
davidmurdoch: It's like a mini haiku! This should probably be `assert(success, <some message about what we expected but didn't get here>)`?

src/chains/ethereum/ethereum/src/miner/miner.ts
davidmurdoch, line 512:
```suggestion
    // HACK: see: https://github.com/trufflesuite/ganache/issues/3093
    //
    //When the priced heap is reset, meaning we're clearing out the heap
    // and origins list to be set again when the miner is called, loop over
    // the priced heap transactions and "unlock" them (set tx.locked = false)
    //
    // The real fix would include fixing the use of locked, as it's
    // currently overloaded to mean "is in priced heap" and also "is being
    // mined".
    const priced = this.#priced;
```

davidmurdoch (changes requested):
Looking good. Just a couple of suggestions!

src/chains/ethereum/ethereum/tests/blockchain.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      // assert that second block's timestamp is at least `blockTime` greater
+      // than the first block's. meaning, these blocks weren't mined one after
+      // the other
+      assert(timestamps[1] >= timestamps[0] + blockTime);
```
jeffsmale90, line 112: I'm pretty sure there's no scenario where this can _not_ be the case right? 

The "wait" timer starts _after_ the first block is mined, so even under strange performance / load conditions this should be correct?

davidmurdoch: I think you're right @jeffsmale90 . Though the check Is nice because the miner has options to mine many blocks while enforcing 1 transaction per block. So while someone would really have too break multiple things in the mine call to trigger this issue, it could still catch that sort of mistake. Maybe we should also check that the second block is < 2x the block time?

MicaiahReid: @jeffsmale90 This test does not pass without the associated fix in the blockchain file. Previously, when the `blockTime` first elapsed, we'd mine _all transactions in the pool_, regardless of how many blocks it took. So in this test scenario, two blocks would be mined back to back.

The blockchain fix ensures that we only one mine one block every `blockTime`. This test confirms that fix. So, you're right that this should always pass. Unless we break something, then this will catch it :smile: 

Unless I'm misunderstanding you?

davidmurdoch (approved):
```
                                         |
                                         |
                                         |
                                         |
   _______                   ________    |
  |ooooooo|      ____       | __  __ |   |
  |[]+++[]|     [____]      |/  \/  \|   |
  |+ ___ +|     ]()()[      |\__/\__/|   |
  |:|   |:|   ___\__/___    |[][][][]|   |
  |:|___|:|  |__|    |__|   |++++++++|   |
  |[]===[]|   |_|_/\_|_|    | ______ |   |
_ ||||||||| _ | | __ | | __ ||______|| __|
  |_______|   |_|[::]|_|    |________|   \
              \_|_||_|_/                  \
                |_||_|                     \
               _|_||_|_                     \
      ____    |___||___|                     \
     /  __\          ____                     \
     \( oo          (___ \                     \
     _\_o/           oo~)/
    / \|/ \         _\-_/_
   / / __\ \___    / \|/  \
   \ \|   |__/_)  / / .- \ \
    \/_)  |       \ \ .  /_/
     ||___|        \/___(_/
     | | |          | |  |
     | | |          | |  |
     |_|_|          |_|__|
     [__)_)        (_(___]
```

> I approve of this

src/chains/ethereum/ethereum/tests/blockchain.test.ts
jeffsmale90: I agree with the change - I was considering whether there is a scenario where this will fail due to inconsistent timing.

I believe that is _not_ the case, assuming the interval timer is started _after_ mining is complete, which I believe is the case.

ie. if the interval was created with `setInterval` rather than `setTimeout` _after_ the block is mined, this test would be inconsistent.

jeffsmale90 (approved):

Issue #411, opened 2019-04-11
`evm_revert` does not emit `"changed"` events.


xavierlepretre:
* I asked about this on [gitter](https://gitter.im/ConsenSys/truffle): [ ]
Although I searched for an answer in Gitter

I run the following sequence of actions:

* Listen to `myWeb3Contract.events.allEvents`
    * register `.on("data", console.log)`
    * register `.on("changed", console.log)`
* Call `evm_snapshot`, keep the id
* Make my contract emit an event
* Call `evm_revert` to the snapshot id

## Expected Behavior

I expect to see 2 things happen:

* `.on("data"` to be called once with the event
* `.on("changed"` to be called once with the same event

## Current Behavior

Instead, only one thing happens:

* `.on("data"` is called once with the event

But `.on("changed"` is never called.

## Steps to Reproduce (for bugs)

You can run `truffle test` here: https://github.com/xavierlepretre/on-changed-event

I would expect both tests to pass.

## Your Environment
* Version used: Ganache CLI v6.4.1 (ganache-core: 2.5.3)
* Version of Truffle/Remix/Other tools used: v5.0.8 (core: 5.0.8)
* Web3 Version: v1.0.0-beta.37
* NodeJS Version: [ ] 6.x, [ ] 7.x (unsupported), [ ] 8.x, [ ] 9.x, [x] v10.15.3
* Operating System and version (include distro if Linux): Linux ubuntu-xenial 4.4.0-145-generic
* Link to your project or repro gist:
* Commit hash to use with above link for reproduction:
* I intend to submit a pull request to fix this issue: [ ]

xavierlepretre:
Our use-case:

As part of our Ethereum developer course at B9lab, we task our students with a small project. This contract keeps a `mapping(address => uint) balances;`. There are 2 events, `LogDeposit` (increase balance) and `LogWithdraw` (decrease balance).
A sub-task of the project is to rebuild the contract state with a history of all events, and only events. So the student creates a small utility that keeps a `balances` mapping equivalent in Javascript with the values going up or down depending on the events.

Now that Web3 provides a convenient notification of fork reorganisation, it is also possible for this Javascript utility to backtrack and keep in sync with the chain.
I used `revert` in an effort, and hope, to simulate a reorg.

That is how I came across this "bug".

PR #3054, opened 2022-05-12
feat: fire reorg events on snapshot/revert

### LINKED ISSUE #91: Handle chain reorgs in newHeads event
See expected behavior here. Today Ganache doesn't do this.
https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB#newheads

benjamincburns:
This change makes it so that ganache behaves as if the chain has been reorged when `evm_revert` is called. When `evm_revert` rolls back the current head of the chain, the following things happen:

1. Subscribers to `newHeads` events receive a message indicating that the head has been rolled back to the requested snapshot.
2. Subscribers to `logs` events receive a message indicating the removal of any logs that were emitted by transactions that were rolled back by the `evm_revert`.
2. Filters are updated to indicate the removal of any logs that were previously emitted by transactions that were rolled back by the `evm_revert`.

fixes #91 and #411

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -818,7 +818,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
       })
     );
 
-    await this.#database.batch(async () => {
+    await this.#database.batch(() => {
       const { blocks, transactions, transactionReceipts, blockLogs } = this;
       // point to the new "latest" again
       blocks.updateLatestIndex(newLatestBlockNumber);
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -416,7 +416,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         this.transactionReceipts.set(hash, encodedReceipt);
 
         // collect block logs
-        tx.getLogs().forEach(blockLogs.append.bind(blockLogs, index, tx.hash));
+        tx.getLogs().forEach(blockLogs.append.bind(blockLogs, false, index, tx.hash));
 
         // prepare log output
         logOutput.push(
@@ -807,7 +807,18 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         blocksToDelete[0].header.number
       );
     }
-    await this.#database.batch(() => {
+    
+    // notify subscribers that logs have been rolled back
+    await Promise.all(
+      blocksToDelete.map(async block => {
+        const blockNum = block.header.number.toBuffer();
+
+        const logsToDelete = (await this.blockLogs.get(blockNum)).asRemoved();
+        await this.emit("blockLogs", logsToDelete);
+      })
+    );
+
+    await this.#database.batch(async () => {
       const { blocks, transactions, transactionReceipts, blockLogs } = this;
       // point to the new "latest" again
       blocks.updateLatestIndex(newLatestBlockNumber);
@@ -818,7 +829,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
           transactions.del(txHash);
           transactionReceipts.del(txHash);
         });
+
         const blockNum = block.header.number.toBuffer();
+
         blocks.del(blockNum);
         blocks.del(block.hash().toBuffer());
         blockLogs.del(blockNum);
@@ -936,6 +949,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         snapshotHeader.stateRoot.toBuffer()
       );
       blocks.latest = snapshotBlock;
+
+      // notify subscribers of reorg
+      await this.emit("block", snapshotBlock);
     }
 
     // put our time adjustment back
--- a/src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
@@ -12,6 +12,7 @@ describe("api", () => {
       let context = {} as any;
       let startingBalance;
       let snapshotId;
+      let newHeadsSubscription;
 
       beforeEach("Set up provider and deploy a contract", async () => {
         const contract = compile(join(__dirname, "./snapshot.sol"));
@@ -22,7 +23,7 @@ describe("api", () => {
         const accounts = await p.send("eth_accounts");
         const from = accounts[3];
 
-        await p.send("eth_subscribe", ["newHeads"]);
+        newHeadsSubscription = await p.send("eth_subscribe", ["newHeads"]);
 
         const transactionHash = await p.send("eth_sendTransaction", [
           {
@@ -89,6 +90,11 @@ describe("api", () => {
 
         // Now checkpoint.
         snapshotId = await send("evm_snapshot");
+      });
+
+      afterEach("clean up newHeads subscription", async () => {
+        const { send } = context;
+        await send("eth_unsubscribe", [newHeadsSubscription]);
       });
 
       it("rolls back successfully", async () => {
@@ -398,6 +404,7 @@ describe("api", () => {
           16
         );
         const snapShotId = await send("evm_snapshot");
+        const snapshotBlockHash = (await send("eth_getBlockByNumber", ["latest", false])).hash;
 
         // increment value for each transaction so the hashes always differ
         let value = 1;
@@ -406,14 +413,24 @@ describe("api", () => {
         const revertedTx = await send("eth_sendTransaction", [
           { from, to, value: value++ }
         ]);
+
         await provider.once("message");
+
+        let reverted = false;
 
         // revert while these transactions are being mined
         const revertPromise = send("evm_revert", [snapShotId]);
 
         const txsMinedProm = new Promise(resolve => {
           let count = 0;
-          const unsubscribe = provider.on("message", _ => {
+          const unsubscribe = provider.on("message", m => {
+            if (!reverted && m.data.result.hash === snapshotBlockHash) {
+              // Ignore the `newHeads` event that is triggered by the revert
+              // operation. We don't include this in the count because in theory
+              // it could either happen here, or in the gotTxsProm below
+              reverted = true;
+              return;
+            }
             if (++count === 2) {
               unsubscribe();
               resolve(null);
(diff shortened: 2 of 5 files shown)


src/chains/ethereum/utils/src/things/blocklogs.ts
```diff
@@ -84,12 +109,13 @@ export class BlockLogs {
...
   public append(
-    /*removed: boolean, */ transactionIndex: Quantity,
+    removed: boolean,
+    transactionIndex: Quantity,
     transactionHash: Data,
     log: TransactionLog
   ) {
     this[_raw][1].push([
-      BUFFER_ZERO, // `removed`, TODO: this is used for reorgs, but we don't support them yet
+      removed ? RPCQUANTITY_ONE.toBuffer() : BUFFER_ZERO, // `removed`
```
davidmurdoch, line 118: It's probably time to make a `BUFFER_ONE` in `constants.ts` and use that here (and update anywhere else that also uses `RPCQUANTITY_ONE.toBuffer()`)?

davidmurdoch: Note: there is a `ONE_BUFFER` defined in the `@ganache/ethereum-transaction` package that should be replaced with this new constant.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -807,7 +807,18 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
     }
-    await this.#database.batch(() => {
+    
+    // notify subscribers that logs have been rolled back
+    await Promise.all(
+      blocksToDelete.map(async block => {
+        const blockNum = block.header.number.toBuffer();
+
+        const logsToDelete = (await this.blockLogs.get(blockNum)).asRemoved();
+        await this.emit("blockLogs", logsToDelete);
```
davidmurdoch, line 817: I'm wondering if we need to emit these events only after the block logs are deleted, because at this point in time they actually aren't. Maybe we should move it to beneath your `// notify subscribers of reorg` comment?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -936,6 +949,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         snapshotHeader.stateRoot.toBuffer()
       );
       blocks.latest = snapshotBlock;
+
+      // notify subscribers of reorg
+      await this.emit("block", snapshotBlock);
```
davidmurdoch, line 954: I think i'm 👍 on the await here. Curious of other reviewers have any thoughts.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -807,7 +807,18 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
       );
     }
-    await this.#database.batch(() => {
+    
+    // notify subscribers that logs have been rolled back
+    await Promise.all(
+      blocksToDelete.map(async block => {
+        const blockNum = block.header.number.toBuffer();
+
+        const logsToDelete = (await this.blockLogs.get(blockNum)).asRemoved();
```
davidmurdoch, line 816: So `evm_snapshot` can cause memory to grow unbounded, so there _is_ a problem with what I'm about to suggest...

What if we tracked all `blockLogs` in `this.#snapshots` so we didn't have to query the database for them when we revert? Maybe we only keep track of the "raw" snapshot data if it saves time and space? 🤔

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -807,7 +807,18 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+        await this.emit("blockLogs", logsToDelete);
+      })
+    );
+
+    await this.#database.batch(async () => {
```
```suggestion
    await this.#database.batch(() => {
```

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: We "emit block logs first so filters can pick them up before block listeners are notified" (per [this comment](https://github.com/trufflesuite/ganache/blob/917503b0ff9635241cc179bc2763ac94b4dfa859/src/chains/ethereum/ethereum/src/blockchain.ts#L493-L494)). I'm thinking this was because a user can install a filter (like for logs or blocks, or whatever), subscribe to `newHeads`, get notified of a new block, and then if they immediately check the contents of the filter (like via `eth_getFilterLogs`, maybe?) the block might not yet exist in the filter and the contents would be incomplete.

So this question is: should we reverse the order for a reorg and emit `block` and then `blockLogs`? I haven't yet given it much thought myself, but it might be worth thinking about!

src/chains/ethereum/ethereum/tests/api/evm/snapshot.test.ts
```diff
@@ -466,9 +483,16 @@ describe("api", () => {
         // and mine one more block just to force the any executable transactions
         // to be immediately mined
 
-        const gotTxsProm = new Promise(resolve => {
+        const gotTxsProm = new Promise((resolve, reject) => {
```
davidmurdoch:
```suggestion
        const gotTxsProm = new Promise(resolve => {
```

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: Additionally, I'm wondering if we should emit these in reverse order?

src/chains/ethereum/transaction/src/transaction-receipt.ts
```diff
@@ -153,7 +153,7 @@ export class TransactionReceipt {
     const transactionHash = transaction.hash;
     const transactionIndex = transaction.index;
     blockLog.blockNumber = blockNumber;
-    raw[3].forEach(l => blockLog.append(transactionIndex, transactionHash, l));
+    raw[3].forEach(l => blockLog.append(blockLog[0], transactionIndex, transactionHash, l));
```
davidmurdoch, line 156: Is `blockLog[0]` in this case is always `undefined` here?

davidmurdoch (changes requested):
I think we may need to re-think our approach, and also make reorg notifications opt-in for three reasons:

1. it breaks currently true assumptions about how ganache works (it doesn't emit reorg events)
2. every truffle test file uses snapshot/revert, and `evm_revert` just got a whole lot more expensive!
3. Reorgs need _new_ blocks (and technically _more_ of them)

There are a few places I can think of that we could have users opt-in to reorg behavior:

1. Something like `evm_snapshot({ enableReorgs: true })` (defaults to `false`)
2. Something like `evm_revert(snapshotId, { reorg: true })` defaults to `false`
3. As an option, like: `Ganache.provider({chain: { enableReorgs: true }})` defaults to `false`

And/or We could allow the user to supply new blocks/transactions via:

1. Something like `evm_revert(snapshotId, { newBlocks: Block[] })`
2. I don't think I like this as much, but it worth bringing up: `evm_reorg(toBlockNumber, newBlocks)`

Of course, we'll also need to spend some time thinking about the appropriate option naming in any case, since public interfaces are forever :-) .

Here is geth's reorg algorithm: https://github.com/ethereum/go-ethereum/blob/acd7b369994887b342be03f8cf11cf617d8ac278/core/blockchain.go#L1924

Additionally... we need to think more about uncles; we don't need to implement the feature right away, however, we need to ensure the behavior we decide now with will allow us to implement them in the future.

I know this is a huge ask and you probably have other things to do. So feel free to punt it back over to us of you'd like.

Thoughts?

src/chains/ethereum/ethereum/src/blockchain.ts
benjamincburns: > I'm wondering if we should emit these in reverse order

I considered that, but since the docs I could find don't specify any order, I figured that it's probably better for testing purposes if they're in forward order (as there aren't any guarantees of order in production). Also it looks like per the geth link you shared that geth also emits them in forward order, unless there's some logic elsewhere that I missed that reverses them when the `removed` flag is set.

src/chains/ethereum/utils/src/things/blocklogs.ts
benjamincburns: I could add that easily enough if you'd like? I suppose whether it's worthwhile to do so would depend on the other higher-level concerns you raised in your summary, however.

src/chains/ethereum/ethereum/src/blockchain.ts
benjamincburns: Maybe I'm misunderstanding the concern, but I think the order would still need to be `blockLogs` and then `block`, for the same reason that you described? If the order were reversed, the application code might see a rollback on `block` first, but the `blockLogs` with the `removed` flag set might not be present in their filter.

Not that this is reason to not address any ordering issues here, but I wonder how often this occurs in practice. I'd expect if someone is using pub/sub they'd use a logs subscription rather than `eth_getFilterLogs`, but I'm sure there's some code out there that does this, as devs generally aren't super disciplined about swapping things over all at once when doing things like moving from polling to pub/sub.

src/chains/ethereum/ethereum/src/blockchain.ts
benjamincburns: Do you have any rough idea as to how much of a speedup this might be?

My intuition is that it wouldn't be huge, as even if there's no explicit caching in the database, I'd imagine that during an automated test run the data would still be resident in the OS disk cache. That said, if decoding is inefficient/expensive, I could imagine it being impactful. At the very least it'd shave off a few ticks of the ol' event loop (likely where the bulk of the savings would come from).

src/chains/ethereum/ethereum/src/blockchain.ts
benjamincburns: Good catch

src/chains/ethereum/transaction/src/transaction-receipt.ts
benjamincburns: It looks to be, yes. Honestly I can't remember why I wrote it this way instead of just passing `false`.

Passing booleans as positional args kinda makes me squirm from a readability standpoint, though. It makes me inclined to suggest that `append` should be refactored such that `removed` is an optional argument that defaults to `false`, but then that would make the argument ordering inconsistent with the ordering of the values in `BlockLog[_raw][1]`... 🤔🤔

src/chains/ethereum/utils/src/things/blocklogs.ts
jeffsmale90: Probably not going to directly impact this change, but in https://github.com/trufflesuite/ganache/pull/2983 I have moved `RPCQUANTITY_ONE` to `Quantity.One`.

PR #3102, opened 2022-05-19
DO NOT MERGE: testing mining race condition fail case

MicaiahReid:


davidmurdoch:
Do you mind if I convert this back to a draft?

MicaiahReid:
I've converted it back! I didn't know if the checks would run in a draft state.

PR #3119, opened 2022-05-23
fix: `eth_getTransactionByBlockNumberAndIndex` and `eth_getTransactionByBlockHashAndIndex` to respect non-zero transaction index

### LINKED ISSUE #3118: eth_getTransactionByBlockNumberAndIndex and eth_getTransactionByBlockHashAndIndex always return transaction with index 0
Both functions `eth_getTransactionByBlockNumberAndIndex` and `eth_getTransactionByBlockHashAndIndex` erroneously parse the index parameter to 0, regardless of the value.

The tests of this function currently pass, because they only test for calls to these functions with index value of 0.

jeffsmale90:
Previously both `eth_getTransactionByBlockNumberAndIndex` and `eth_getTransactionByBlockHashAndIndex` would incorrectly return the first transaction in the block, regardless of the value passed as `index`.

With this fix, any transaction can be retrieved by providing its index within the given block.

### DIFF
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -1242,7 +1242,7 @@ export default class EthereumApi implements Api {
       .catch<Block>(_ => null);
     if (!block) return null;
     const transactions = block.getTransactions();
-    return transactions[parseInt(Quantity.from(index).toString(), 10)].toJSON(
+    return transactions[Quantity.from(index).toNumber()].toJSON(
       blockchain.common
     );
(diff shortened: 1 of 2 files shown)


davidmurdoch (approved):
Nice catch!

This needs a PR description that can be used in release before release.

jeffsmale90:
I can update this to use the newer `Quantity.toNumber` style that's in #2983 once one of these is accepted to merge. Fixes #3118

MicaiahReid (approved):
Good catch!

PR #3130, opened 2022-05-25
fix: make forking work in the browser

davidmurdoch:
Ganache's zero-config forking feature now works in the browser!


src/packages/ganache/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ganache",
-  "version": "7.1.0",
+  "version": "7.1.0-clairvoyance.4",
```
tenthirtyone: Sneaky!

davidmurdoch:
You can test this by checking out this branch and then running:

```console 
$ npm run reinstall && INFURA_KEY=<YOUR INFURA KEY  HERE> npm run build && npm run docs.build && npm run docs.preview
```

then paste the following in any of the editors:

```javascript
const p = ganache.provider({fork:"mainnet"});
console.log(await p.request({method:"debug_traceTransaction", params:["0x560900937892cd888a685c836f666804e05e04f3d161e1751ffb5ab7809c55d9"]}));
```

then wait a while for the result to appear (it can takes a lot of time).

Additionally, you should test the http handler code (the above tests WebSockets):

```javascript
const p = ganache.provider({fork:{url: "<your infura url goes here>"}});
console.log(await p.request({method:"debug_traceTransaction", params:["0x560900937892cd888a685c836f666804e05e04f3d161e1751ffb5ab7809c55d9"]}));
```


Example:

image

index.html
```diff
@@ -0,0 +1,163 @@
+<!DOCTYPE html>
```
jeffsmale90: Should this file be deleted?

src/chains/ethereum/ethereum/src/forking/handlers/http-handler.ts
```diff
@@ -119,17 +128,24 @@ export class HttpHandler extends BaseHandler implements Handler {
         const { headers } = res;
 
         let buffer: Promise<Buffer>;
-        // if we have a transfer-encoding we don't care about "content-length"
-        // (per HTTP spec). We also don't care about invalid lengths
-        if ("transfer-encoding" in headers) {
+        // in the browser we can't detect if the response is compressed (gzip),
```
jeffsmale90, line 119: Does this imply that we cannot trust the `content-length` in the browser (because it may have been compressed), so we _must_ handle as a chunked response?

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -596,7 +596,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     await this.#blockBeingSavedPromise;
     return {
       transactions,
-      blockNumber: nextBlock.header.number.toBuffer()
+      blockNumber: nextBlock.header.number.toArrayLike(Buffer)
```
davidmurdoch, line 597: We already polyfill Buffer in the browser, but BN doesn't handle it well, so we have to do it this way .

src/chains/ethereum/ethereum/src/forking/handlers/http-handler.ts
davidmurdoch: Right, but I'd word it slightly differently: we can't trust it to mean the length of the decompressed content, as originally implemented.

I will have a follow up PR that adds compression handling to the Node side which handles the Content-Length + Content-Encoding: gzip case by falling back to the chunked handler. It just still needs tests :-)

jeffsmale90 (approved):
🚢 it!

MicaiahReid (dismissed):
Soooo cooooooool! All it needs is a release-ready PR description!

I've also pushed two changes to the README to this branch if you could review them. Let me know and I can remove this commits from your branch, I just knew you were out of town and thought it would be easier to push than have you go in and update the README.

davidmurdoch:
blocked by https://github.com/trufflesuite/ganache/pull/3275

MicaiahReid (approved):

PR #3131, opened 2022-05-25
feat: add `miner.timestampIncrement` option

davidmurdoch:
By default ganache uses the system clock time for automatically mined blocks. This behavior is convenient, but results in nondeterminism, invalid block timestamps*, and can cause issues with relative block timestamp dependent contracts and tests. There were workarounds for all of these issues (by manually mining via `evm_mine({timestamp})`) but a built in solution was clearly necessary.

Ganache now permits the use of a new `--miner.timestampIncrement` option which can be used to set the amount of seconds between consecutive blocks, regardless of how much system time has passed. The default value, "clock", retains the previous default behavior of using the system time for each block.

The following example will start ganache instructing it to increment each new block's timestamp by `1` second from 1649736000000:

```console
$ ganache --time 1649736000000 --miner.timestampIncrement=1
```

\* _The [Ethereum Yellowpaper](https://ethereum.github.io/yellowpaper/paper.pdf) defines:_ $H_S > P(H)_{H_S}$  _where_ $H_S$ _is the timestamp of Block_ $H$ _and_ $P(H)$ _is the parent block._


src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -548,14 +551,28 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+    const minerOptions = this.#options.miner;
+    let qTimestamp: Quantity;
+    if (timestamp != null) {
+      qTimestamp = Quantity.from(timestamp);
+    } else {
+      const { timestampIncrement } = minerOptions;
+      if (timestampIncrement === "clock") {
+        qTimestamp = Quantity.from(this.#currentTime());
+      } else {
+        qTimestamp = Quantity.from(
```
jeffsmale90: I've got some changes incoming that will allow us to change this to something like:

`qTimestamp = previousHeader.timestamp.add(timestampIncrement)`

I'll make a note to pick that up in my change.

src/chains/ethereum/ethereum/src/forking/fork.ts
```diff
@@ -223,6 +227,13 @@ export class Fork {
       this.blockNumber.toBigInt()
     );
     this.block = new Block(BlockManager.rawFromJSON(block, common), common);
+    if (!chainOptions.time && minerOptions.timestampIncrement !== "clock") {
+      chainOptions.time = new Date(
```
jeffsmale90, line 231: I'll pick this up in my change also

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,154 @@ describe("provider", () => {
...
+        parseInt(block.timestamp),
+        +time / 1000 + timestampIncrement
+      );
+      await provider.disconnect();
+    });
+
+    it("uses the timestampIncrement for the first block when forking", async () => {
+      const time = new Date("2019-01-01T00:00:00.000Z");
+      const timestampIncrement = 5;
+      const fakeMainnet = await getProvider({
```
jeffsmale90, line 147: This is pretty cool - forking ganache with ganache.

Is `fakeMainnet` the best name, as forking isn't just used for forking mainnet 🤷 

Should it rather be `upstream` or something?

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,154 @@ describe("provider", () => {
...
+        chain: { time }
+      });
+      const provider = await getProvider({
+        fork: { provider: fakeMainnet as any },
+        miner: { timestampIncrement }
+      });
+      const block = await provider.request({
```
jeffsmale90, line 154: I think I read somewhere that a fork is created immediately on forking - is that why we can assume that a block has been created?

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,154 @@ describe("provider", () => {
...
+        params: ["latest", false]
+      });
+      assert.strictEqual(
+        parseInt(block.timestamp),
+        +time / 1000 + timestampIncrement
+      );
+      await provider.disconnect();
+      await fakeMainnet.disconnect();
+    });
+    it("uses the `time` option for the first block even when when `timestampIncrement` is not 'clock' when forking", async () => {
```
jeffsmale90:
```suggestion
    it("uses the `time` option for the first block even when `timestampIncrement` is not 'clock' when forking", async () => {
```

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,154 @@ describe("provider", () => {
...
+      const hash = await provider.request({
+        method: "eth_sendTransaction",
+        params: [tx]
+      });
+      const receipt = await provider.request({
+        method: "eth_getTransactionReceipt",
+        params: [hash]
+      });
+      assert.notStrictEqual(receipt, null);
+      const sentBefore = Math.floor(Date.now() / 1000);
```
jeffsmale90: Because this is an upper bound, should it be `Math.ceil`'ed, or can we depend on blocktime also being `Math.floor`'ed?

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,154 @@ describe("provider", () => {
...
+        fork: { provider: fakeMainnet as any },
+        chain: { time: time2 },
+        miner: { timestampIncrement }
+      });
+      const block = await provider.request({
+        method: "eth_getBlockByNumber",
+        params: ["latest", false]
+      });
+      assert.strictEqual(parseInt(block.timestamp), +time2 / 1000);
+      await provider.disconnect();
```
jeffsmale90, line 183: What's the implication of _not_ disconnecting here?

I see that we don't disconnect any of the other providers in this file.

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: If it isn't clear to you what it is attempting to mock, then no, it's not the best name. 😁 Not sure about "upstream", but the name may be worth mulling over (we use this same pattern in many fork tests, all with different names 😅).

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: Yup. We do create a block immediately upon forking. We'll add an option to _not_ create a new block in the future, because it's actually really annoying. I could see Ganache v8 defaulting to this no new empty immediate block behavior.

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: We really should always disconnect, but for forking it's more important due to the persistent cache being unable to be shared between instances.

jeffsmale90:
Should we open a PR on https://github.com/mds1/convex-shutdown-simulation to use this new configuration?

davidmurdoch:
> Should we open a PR on https://github.com/mds1/convex-shutdown-simulation to use this new configuration?

We will update it, yes.

src/chains/ethereum/ethereum/tests/provider.test.ts
jeffsmale90: I guess it kinda _is_ clear what it's doing, but it seemed strange to me.

In retrospect, maybe it's not so bad - although you could argue that it's not _truly_ a fake mainnet, it _is_ reasonably clear in the test what it is, and how it's being used.

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: It's also floored: https://github.com/trufflesuite/ganache/blob/develop/src/chains/ethereum/ethereum/src/blockchain.ts#L776

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,154 @@ describe("provider", () => {
...
+      // (floored), and mining happens much quicker than 1 second.
+      assert(
+        parseInt(block.timestamp) >= sentAfter,
+        `block wasn't mined at the right time, should have been on or after ${sentAfter}, was ${parseInt(
+          block.timestamp
+        )}`
+      );
+      assert(
+        parseInt(block.timestamp) <= sentBefore,
+        `block wasn't mined at the right time, should have been on or before ${sentAfter}, was ${parseInt(
```
MicaiahReid:
```suggestion
        `block wasn't mined at the right time, should have been on or before ${sentBefore}, was ${parseInt(
```
Got the variable names swapped here

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,154 @@ describe("provider", () => {
...
     });
+
+    it("uses the default 'clock' time for the `timestampIncrement` option", async () => {
+      const provider = await getProvider();
+      const options = provider.getOptions();
+      assert(options.miner.timestampIncrement, "clock");
+
+      const accounts = await provider.send("eth_accounts");
+      const tx = { from: accounts[0], gas: "0xfffff" };
+      const sentAfter = Math.floor(Date.now() / 1000);
```
MicaiahReid: The `sentAfter`/`sentBefore` variable names here are very confusing to me. Maybe consider naming them after what the time represents rather than how the block's timestamp should compare to it? Something like `timeBeforeMiningBlock`/`timeAfterMiningBlock`?

That will also allow the failure statements to read like a sentence:
> block wasn't mined at the right time, should have been on or after ${timeBeforeMiningBlock}

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,154 @@ describe("provider", () => {
...
+    });
+
+    it("uses the timestampIncrement option", async () => {
+      const time = new Date("2019-01-01T00:00:00.000Z");
+      const timestampIncrement = 5;
+      const provider = await getProvider({
+        chain: { time },
+        miner: { timestampIncrement }
+      });
+      const accounts = await provider.send("eth_accounts");
```
MicaiahReid: With these tests, is there an advantage to sending a transaction rather than just using `evm_mine`? If not, we could remove a good bit of code.

MicaiahReid: If so, a comment to explain why we need to send a transaction in this test would be helpful.

MicaiahReid (changes requested):

jeffsmale90 (approved):

MicaiahReid (approved):

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -777,32 +785,50 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+      milliseconds = 0;
     }
-    return (this.#timeAdjustment += seconds);
+    return (this.#timeAdjustment += milliseconds);
   }
 
   /**
-   * @param seconds -
+   * @param newTime - the number of milliseconds to adjust the time by. Can be negative.
+   * @param sourceTime - the current time in milliseconds
```
MicaiahReid: Isn't this the current time _or_ the timestamp of the latest block we're using as a reference?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -550,19 +551,28 @@ export default class EthereumApi implements Api {
...
-        t = Quantity.from(time).toNumber();
+        timestamp = Quantity.from(time).toNumber();
         break;
     }
-    return Math.floor(this.#blockchain.setTime(t) / 1000);
+    const blockchain = this.#blockchain;
+    // when using clock time use Date.now(), otherwise use the timestamp of the
+    // current latest block
+    const currentTime =
+      this.#options.miner.timestampIncrement === "clock"
```
MicaiahReid: Would it make more sense for this logic to get the `sourceTime` to be inside the `setTimeDiff` function?
i.e.
```TypeScript
public setTimeDiff(newTime: number) {
  const sourceTime = this.#options.miner.timestampIncrement === "clock"
        ? Date.now()
        : this.blocks.latest.header.timestamp.toNumber() * 1000;
  return (this.#timeAdjustment = newTime - sourceTime);
}
```

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,185 @@ describe("provider", () => {
...
+        params: ["latest", false]
+      });
+      assert.strictEqual(
+        parseInt(block.timestamp),
+        Math.floor(+time / 1000) + timestampIncrement
+      );
+      await provider.disconnect();
+    });
+
+    it("uses time adjustment after evm_setTime when timestampIncrement is used", async () => {
```
MicaiahReid:
```suggestion
    it("uses time adjustment after `evm_setTime` when `timestampIncrement` is used", async () => {
```
Just to be consistent with the other tests

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,185 @@ describe("provider", () => {
...
+        params: ["latest", false]
+      });
+      const expectedTime =
+        Math.floor((fastForward + +time) / 1000) + timestampIncrement;
+      assert.strictEqual(parseInt(block.timestamp), expectedTime);
+
+      await provider.disconnect();
+    });
+
+    it("uses time adjustment after evm_increaseTime when timestampIncrement is used", async () => {
```
MicaiahReid:
```suggestion
    it("uses time adjustment after `evm_increaseTime` when `timestampIncrement` is used", async () => {
```

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,185 @@ describe("provider", () => {
...
+        method: "eth_getBlockByNumber",
+        params: ["latest", false]
+      });
+      const expectedTime =
+        Math.floor(+time / 1000) + fastForward + timestampIncrement;
+      assert.strictEqual(parseInt(block.timestamp), expectedTime);
+      await provider.disconnect();
+    });
+
+    it("uses the timestampIncrement for the first block when forking", async () => {
```
MicaiahReid:
```suggestion
    it("uses the `timestampIncrement` for the first block when forking", async () => {
```

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -30,6 +30,185 @@ describe("provider", () => {
...
+      const block = await provider.request({
+        method: "eth_getBlockByNumber",
+        params: ["latest", false]
+      });
+      assert.strictEqual(parseInt(block.timestamp), +time2 / 1000);
+      await provider.disconnect();
+      await fakeMainnet.disconnect();
+    });
+
+    it("uses the timestampIncrement option when interval mining", async () => {
```
MicaiahReid:
```suggestion
    it("uses the `timestampIncrement` option when interval mining", async () => {
```

MicaiahReid (changes requested):

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: Good call! Done!

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: removed per your other comment.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -294,14 +294,17 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
-        let firstBlockTime: number;
-        if (options.chain.time != null) {
-          // If we were given a timestamp, use it instead of the `_currentTime`
-          const t = options.chain.time.getTime();
-          firstBlockTime = Math.floor(t / 1000);
-          this.setTime(t);
-        } else {
-          firstBlockTime = this.#currentTime();
+
+        // if we don't have a time form the user get one now
```
jeffsmale90:
```suggestion
        // if we don't have a time from the user get one now
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -294,14 +294,17 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+        if (options.chain.time == null) options.chain.time = new Date();
+
+        const timestamp = options.chain.time.getTime();
+        const firstBlockTime = Math.floor(timestamp / 1000);
+
+        // if we are using clock time we need to record the time offset so
+        // other blocks can have timestamps relative to our initial time.
+        if (options.miner.timestampIncrement === "clock") {
+          this.setTimeDiff(timestamp, Date.now());
```
jeffsmale90: Should we use `timestamp` instead of `Date.now()`?

davidmurdoch: I've refactored this part so I don't think this is an issue anymore.

MicaiahReid (approved):

PR #3151, opened 2022-05-28
chore: add relevant keywords to ganache's package.json

### LINKED ISSUE #3143: add `ganache-core` and `ganache-cli` to the `ganache` package's package.json `keywords` array.
Reason: the `ganache` package doesn't seem to show up in npm search results for `ganache-core` and `ganache-cli`.

davidmurdoch:
Users reported that it was difficult to find the `ganache` package in the npm registry because they were used to searching for `ganache-cli` and `ganache-core`. This change adds those terms to our `package.json`'s `"keywords"` list, which hopefully alleviates the problem.

Fixes #3143

### DIFF
--- a/src/packages/ganache/package.json
+++ b/src/packages/ganache/package.json
@@ -39,6 +39,10 @@
   },
   "keywords": [
     "ganache",
+    "ganache-core",
+    "ganache-cli",
+    "@trufflesuite/ganache",
+    "@truffle/ganache",
     "ethereum",
     "evm",
     "blockchain",


jeffsmale90:
Looks good to me. Is there anything else we need to do in order to push users over from`ganache-cli`?

jeffsmale90 (approved):

davidmurdoch:
> Looks good to me. Is there anything else we need to do in order to push users over from`ganache-cli`?

We've already deprecated the package and pointed users here. Another issue is that the publisher of this package is a "truffle-cicd" without an avatar and it looks a bit strange. We've received feedback that the npm user doesn't look legitimate. An update for that is in the works.

MicaiahReid (approved):
Truly groundbreaking work. :exploding_head:

PR #3164, opened 2022-06-01
improvement: changes to internal RPC Quantity type to make basic mathematic operations cleaner

jeffsmale90:
Add addition and multiplication methods  to internal RPC-JSON `Quantity` data type.

This change impacts Ganache internals only, except for some minor performance improvements.


jeffsmale90:
This change went back and forward a little bit - initially I implemented buffer level addition in `buffer-math.ts`, which was somewhat promising, but ultimately I believe that the minor performance improvements that we gained were not worth the increased complexity - in the case of adding a `number`, `Buffer`, or `Quantity` type, this was up to ~10% faster, and 1-2% faster (wooo) if adding a `bigint`.

I deleted those changes, but left them in the history for anyone who is curious, or have better ideas about how we could approach this.

https://github.com/trufflesuite/ganache/commit/4cad19352e19e2884c91d3ccc0783fccc068df03 is changes to consumers of `.add()` and `.multiply()`, plus a couple of changes to use the static `Quantity.toXXX()` methods (which I'm happy to remove if anyone feels they don't belong here. [View just the first 6 commits](https://github.com/trufflesuite/ganache/pull/3164/files/de1a6e545e23b5e31cd7a0e26defc3e9b31bd6d2) to see the changes in isolation to the consumer changes.

Oh and I used property-based testing with `fast-check`. I might look to expand this to the rest of the `JSON-RPC` tests at some stage.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -950,7 +950,7 @@ export default class EthereumApi implements Api {
       }
 
       const block = new RuntimeBlock(
-        Quantity.from((parentHeader.number.toBigInt() || 0n) + 1n),
+        parentHeader.number.add(1),
```
```suggestion
        parentHeader.number.add(1n),
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1067,9 +1066,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         buf: caller
       } as any);
       fromAccount.nonce.iaddn(1);
-      const txCost = new BN(
-        (gasLimit * transaction.gasPrice.toBigInt()).toString()
-      );
+      const txCost = new BN(transaction.gasPrice.multiply(gasLimit).toBigInt().toString());
```
davidmurdoch: `BN` actually allows a `Buffer` in its constructor. I wonder if we should change it.

src/chains/ethereum/ethereum/tests/api/eth/getCode.test.ts
```diff
@@ -123,7 +123,7 @@ describe("api", () => {
           it("should return the no code at the previous block number", async () => {
             const code = await provider.send("eth_getCode", [
               contractAddress,
-              Quantity.from(blockNumber.toBigInt() - 1n).toString()
+              Quantity.toString(blockNumber.toBigInt() - 1n)
```
davidmurdoch: Can we do subtraction?, i.e., `blockNumber.add(-1n)`?

jeffsmale90:
This has become funky after rebasing the upstream branch, so I will have to address that in the morning.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -115,6 +117,77 @@ export class Quantity extends BaseJsonRpcType {
...
+        throw new Error(`Cannot wrap ${argument} as a json-rpc type.`);
+      }
+      argumentValue = BigInt(argumentNumber);
+    }  else if (argumentType === "bigint") {
+      if (argument < 0n) {
+        throw new Error("Cannot wrap a negative number as a JSON-RPC type.");
+      }
+      argumentValue = argument as bigint;
+    } else if (argumentType === "string") {
+      if ((argument as string).slice(2).toLowerCase() !== "0x") {
```
MicaiahReid: Do we need to do any verification that the characters are only hexadecimal, or is this safe because it's for internal use only?

jeffsmale90: We rely on `BigInt(value)`'s validation here. We have to explicitly check the first two chars, otherwise it may parse as decimal. I've added a few more test cases: https://github.com/trufflesuite/ganache/pull/3164/commits/e5b747bc9e6abb3858b2d098c7afd14bfb916ee8

MicaiahReid (approved):

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
MicaiahReid: Ahh there already were tests with invalid hex, so you probably didn't need to add those. My miss!

davidmurdoch (changes requested):
just a quick partial review

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: Yeah, I like that, it's cleaner and it's less work.

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Done!

src/chains/ethereum/ethereum/tests/api/eth/getCode.test.ts
jeffsmale90: We do not presently support subtraction. I intentionally did not support it, as (IIRC) there's only one case of this, and it simplifies the implementation to use the same parsing / validation logic as the Quantity constructor (which doesn't support negatives).

We _could_ implement this, but I'm not sure whether it's worth the effort. Happy to hear differing opinion though.

jeffsmale90:
There's a bunch of conflicts on here, I'll rebase to develop and resolve those tomorrow.

src/chains/ethereum/ethereum/tests/api/eth/getCode.test.ts
jeffsmale90: Actually it's not that hard, I implemented support for -ve addend (so long as the sum >= 0). But only support 
 for `number | bigint` type addends (as other types don't have a clear way to represent negative numbers).

And I found an issue when testing this! I opened #3328 to fix an issue with converting buffer to number (I put a fix in this PR also, but I'll resolve any divergence when I merge).

As commented on the related PR, I think I'll convert some more of the tests in here into FC tests after I've merged a couple of the outstanding tests in this area.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,98 @@ export class Quantity extends BaseJsonRpcType {
...
+    }
+
+    const thisValue = this.toBigInt();
+
+    const productBuffer = bigIntToBuffer(thisValue * multiplierValue);
+    return new Quantity(productBuffer, this._nullable);
+  }
+
+  static toBuffer(value: JsonRpcInputArg): Buffer;
+  static toBuffer(value: JsonRpcInputArg, nullable?: boolean): Buffer | null {
```
jeffsmale90, line 211: I added overrides to the `toXXXX()` methods so that calling the function without supplying `nullable` (defaults to `false`), returns a non-nullable type. It'd be nice to do this on the instance methods too, but I think that will be a more significant undertaking.

src/chains/ethereum/ethereum/tests/api/eth/getCode.test.ts
jeffsmale90: As commented on the related PR, I think I'll convert some more of the tests in here into FC tests after I've merged a couple of the outstanding tests in this area.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
...
+      throw new Error(`Cannot multiply a Quantity by ${multiplier}`);
+    }
+
+    const thisValue = this.toBigInt();
+
+    const productBuffer = bigIntToBuffer(thisValue * multiplierValue);
+    return new Quantity(productBuffer, this._nullable);
+  }
+
+  static toBuffer(value: JsonRpcInputArg, nullable?: false): Buffer;
```
jeffsmale90, line 242: I added overrides to the `toXXXX()` methods so that calling the function `nullable: false`, returns a non-nullable type. It'd be nice to do this on the instance methods too, but I think that will be a more significant undertaking, that I'll pick up outside of here.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
...
+      }
+      argumentValue = BigInt(argumentNumber);
+    }  else if (argumentType === "bigint") {
+      argumentValue = argument as bigint;
+    } else if (argumentType === "string") {
+      if ((argument as string).slice(0, 2).toLowerCase() !== "0x") {
+        throw new Error(`Cannot wrap "${argument}" as a JSON-RPC type; strings must be hex-encoded and prefixed with "0x".`);
+      }
+      argumentValue = BigInt(argument as string);
+    } else if (argument instanceof Quantity) {
```
davidmurdoch: This check will fail if the `Quantity` object passed in is not an instance of _this_ `Quantity` class. We should probably check for `"toBigInt" in argument` or something like that.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
     return firstNonZeroByte;
   }
 
-  static toBuffer(value: JsonRpcInputArg, nullable?: boolean): Buffer {
+  private static getArgumentAsBigInt(argument: JsonRpcInputArg | Quantity): bigint {
+    const argumentType = typeof argument;
+    let argumentValue: bigint;
+
+    if (argumentType === "number") {
```
davidmurdoch: If you use a `switch` as `switch (typeof argument) {` you won't have to `as number`, etc, as typescript will narrow the type automatically:

```
switch (typeof argument) {
  case "number":
    if (argument % 1) {
       // ...
    break;
  case "string":
    // ...
  default:
    if (Buffer.isBuffer(argument)) //... etc
}
```

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -217,9 +217,10 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
     }
     const transactor = await transactorPromise;
 
-    const cost =
-      transaction.gas.toBigInt() * transaction.maxGasPrice().toBigInt() +
-      transaction.value.toBigInt();
+    const cost = transaction.gas.multiply(transaction.maxGasPrice())
```
davidmurdoch, line 220: This new approach will much more expensive because each intermediate step gets converted back to a Buffer.

I worry these new methods are too expensive to be so easy to use. haha

src/chains/ethereum/ethereum/tests/api/eth/getCode.test.ts
```diff
@@ -123,7 +123,7 @@ describe("api", () => {
           it("should return the no code at the previous block number", async () => {
             const code = await provider.send("eth_getCode", [
               contractAddress,
-              Quantity.from(blockNumber.toBigInt() - 1n).toString()
+              Quantity.toString(blockNumber.add(-1n).toBigInt())
```
davidmurdoch:
```suggestion
              blockNumber.add(-1n).toString()
```

src/chains/ethereum/ethereum/tests/api/eth/getCode.test.ts
```diff
@@ -169,14 +169,12 @@ describe("api", () => {
           });
 
           it("should return a `header not found` error for requests to non-existent blocks", async () => {
-            const nextBlockNumber =
-              Quantity.from(await provider.send("eth_blockNumber")).toBigInt() +
-              1n;
+            const nextBlockNumber = Quantity.from(await provider.send("eth_blockNumber")).add(1);
```
davidmurdoch:
```suggestion
            const nextBlockNumber = Quantity.from(await provider.send("eth_blockNumber")).add(1n);
```

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
...
+  static toNumber(value: JsonRpcInputArg, nullable?: true): number | null;
+  static toNumber(value: JsonRpcInputArg, nullable?: boolean): number | null {
     return Quantity.from(value, nullable).toNumber();
   }
 
-  static toBigInt(value: JsonRpcInputArg, nullable?: boolean): bigint {
+  static toBigInt(value: JsonRpcInputArg, nullable?: false): bigint;
+  static toBigInt(value: JsonRpcInputArg, nullable?: true): bigint | null;
+  static toBigInt(value: JsonRpcInputArg, nullable?: boolean): bigint | null {
     return Quantity.from(value, nullable).toBigInt();
```
davidmurdoch, line 263: worth updating this to the new static Quantity helper in the case that `value` is _not_ null?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
...
+    return new Quantity(sumBuffer, this._nullable);
+  }
+
+  public multiply(multiplier: JsonRpcInputArg | Quantity): Quantity {
+    if (this.bufferValue == null ||this.bufferValue.length === 0) {
+      return this._nullable ? Quantity.Empty : Quantity.Zero;
+    }
+
+    const multiplierValue = Quantity.getArgumentAsBigInt(multiplier);
+    if (multiplierValue < 0) {
```
davidmurdoch:
```suggestion
    if (multiplierValue < 0n) {
```

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
...
+      throw new Error(`Cannot multiply a Quantity by a negative multiplier`);
+    }
+
+    if (multiplierValue === undefined) {
```
davidmurdoch, line 219: Why do we need to guard against this case? `getArgumentAsBigInt` can't return `undefined`, right?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
...
+  }
+
+  public add(addend: JsonRpcInputArg | Quantity): Quantity {
+    if (this.bufferValue == null || this.bufferValue.length === 0) {
+      return addend instanceof Quantity ? addend : Quantity.from(addend, this._nullable);
+    }
+
+    const addendValue = Quantity.getArgumentAsBigInt(addend);
+
+    if (addendValue === undefined) {
```
davidmurdoch: Why do we need to guard against this case? `getArgumentAsBigInt` can't return `undefined`, right?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
...
+      throw new Error(`Cannot add ${addend} to a Quantity`);
+    }
+
+    const sum = this.toBigInt() + addendValue;
+
+    if (sum < 0n) {
+      throw new Error(`Cannot add ${addend} to a a Quantity of ${this}, as it results in a negative value`);
```
davidmurdoch:
```suggestion
      throw new Error(`Cannot add ${addend} to a Quantity of ${this}, as it results in a negative value`);
```

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -2,8 +2,10 @@ import { bufferToBigInt } from "../../utils/buffer-to-bigint";
 import { BaseJsonRpcType } from "./json-rpc-base-types";
 import { JsonRpcInputArg } from "./input-parsers";
 import { BUFFER_EMPTY, BUFFER_ZERO } from "../../utils/constants";
+import { bigIntToBuffer } from "../../utils/bigint-to-buffer";
 
 export class Quantity extends BaseJsonRpcType {
+  public static Null = Quantity.from(null, true);
```
davidmurdoch: It doesn't look like this is used. Am I wrong?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -131,19 +134,100 @@ export class Quantity extends BaseJsonRpcType {
...
+    } else if (argumentValue == null) {
+      argumentValue = 0n;
+    }
+
+    return argumentValue;
+  }
+
+  public add(addend: JsonRpcInputArg | Quantity): Quantity {
+    if (this.bufferValue == null || this.bufferValue.length === 0) {
+      return addend instanceof Quantity ? addend : Quantity.from(addend, this._nullable);
```
davidmurdoch: This check will fail if the `Quantity` object passed in is not an instance of this `Quantity` class. We should probably check for `"toBigInt" in argument` or something like that.

davidmurdoch (changes requested):
partial review I wanted to get over real quick!

src/chains/ethereum/ethereum/src/transaction-pool.ts
jeffsmale90: I've added 31161d0c29ca0ac93bd8ed0dc7ba116effa0fd5a which defers conversion to `Buffer` and will use the raw numeric value where possible.

This means that `Quantity.from(1).toBigInt()` will never convert to `Buffer`, as will the code that you linked.

Keen to hear your thoughts.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: I've never been a huge fan of switch statements, but I guess they're no worse than big if / else chains. Also there's obvious benefits in cases like this in TS.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: It's also effectively the same as `Quantity.Empty`

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: Agreed - although I think that's a bug!

I've changed it to not coerce invalid types to 0, and moved the error into the getArgumentAsBigInt function.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: Same as below.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: I <3 `switch` statements (they're usually faster, too)

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -91,59 +113,153 @@ export class Quantity extends BaseJsonRpcType {
...
     }
     return firstNonZeroByte;
   }
 
-  static toBuffer(value: JsonRpcInputArg, nullable?: boolean): Buffer {
+  private static getArgumentAsBigInt(argument: JsonRpcInputArg | Quantity): bigint {
+    const argumentType = typeof argument;
+    let argumentValue: bigint;
+
+    switch(argumentType) {
```
davidmurdoch, line 162: looks like these files all need to be prettiered

davidmurdoch, line 162: Do this:

```suggestion
    let argumentValue: bigint;

    switch (typeof argument) {
```

and then remove the `as number`s below.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -91,59 +113,153 @@ export class Quantity extends BaseJsonRpcType {
...
+          throw new Error(`Cannot wrap "${argument}" as a JSON-RPC type; strings must be hex-encoded and prefixed with "0x".`);
+        }
+        argumentValue = BigInt(argument as string);
+        break;
+      default:
+        if (Quantity.isQuantity(argument)) {
+          argumentValue = argument.toBigInt();
+        } else if (Buffer.isBuffer(argument)) {
+          argumentValue = bufferToBigInt(argument);
+        } else if (argument == null) {
```
davidmurdoch, line 187: this can't happen (per the types for `argument`)

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: Clarification: I meant  `Quantity.getArgumentAsBigInt(value)` when we know `value != null`

src/packages/utils/package.json
```diff
@@ -56,6 +56,7 @@
     "@types/mocha": "9.0.0",
     "@types/seedrandom": "3.0.1",
     "cross-env": "7.0.3",
+    "fast-check": "^3.0.0",
```
davidmurdoch, line 59: This should be pinned to an exact version.

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -1,20 +1,43 @@
...
+  }
+}
+
+const LAZY_BUFFER_EMPTY = Lazy.of(BUFFER_EMPTY);
+
 /**
- * Parse and validate a {@link number} to {@link Buffer} as internal representation for a JSON-RPC data type.
+ * Parses and validates {@link number} to {@link Buffer} as internal representation for a JSON-RPC data type.
  * @param {number} input - must be a positive, finite integer, or null.
  * @returns {Buffer}
```
davidmurdoch, line 38: This doesn't return a `Buffer` anymore.

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -25,31 +48,36 @@ export function parseAndValidateNumberInput(input: number): Buffer {
     throw new Error(`Cannot wrap a ${input} as a json-rpc type`);
   }
 
-  return uintToBuffer(input as number);
+  if (input === 0) {
+    return LAZY_BUFFER_EMPTY;
+  }
+
+  return Lazy.from(() => uintToBuffer(input as number));
```
davidmurdoch, line 55: `input` is already a `number`

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -25,31 +48,36 @@ export function parseAndValidateNumberInput(input: number): Buffer {
...
   if (input < 0n) {
     throw new Error("Cannot wrap a negative value as a json-rpc type");
   }
-  return bigIntToBuffer(input as bigint);
+
+  if (input === 0n) {
+    return LAZY_BUFFER_EMPTY;
+  }
+
+  return Lazy.from(() => bigIntToBuffer(input as bigint));
```
davidmurdoch, line 72: `input` is already a bigint

src/chains/ethereum/ethereum/src/transaction-pool.ts
davidmurdoch: `31161d0` reminds me a lot of the original `Quantity`/`Data` implementation (though this way is simpler as it only ). I'm very curious about the performance here.

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
```diff
@@ -38,25 +47,26 @@ export class BaseJsonRpcType {
...
   }
 
   toBuffer(): Buffer {
-    return this.bufferValue;
+    return this._lazyBufferValue.getValue();
   }
 
   valueOf(): any {
-    return this.bufferValue;
+    return this._lazyBufferValue.getValue();
```
davidmurdoch, line 62: Any reasons not to use this here:

```suggestion
    return this.toBuffer();
```

src/packages/utils/src/things/json-rpc/json-rpc-base-types.ts
```diff
@@ -38,25 +47,26 @@ export class BaseJsonRpcType {
...
   }
 
   toJSON(): string | null {
     return this.toString();
   }
 
   isNull() {
-    return this.bufferValue == null;
+    return this._lazyBufferValue.getValue() == null;
```
davidmurdoch, line 70: Any reason not to use this here:


```suggestion
    return this.toBuffer() == null;
```

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
davidmurdoch: We still don't need to guard against it, right?

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -91,59 +113,153 @@ export class Quantity extends BaseJsonRpcType {
...
     return firstNonZeroByte;
   }
 
-  static toBuffer(value: JsonRpcInputArg, nullable?: boolean): Buffer {
+  private static getArgumentAsBigInt(argument: JsonRpcInputArg | Quantity): bigint {
+    const argumentType = typeof argument;
+    let argumentValue: bigint;
+
+    switch(argumentType) {
+      case "number":
```
davidmurdoch, line 163: Can you run a test to determine the best order for these `switch` cases? They should be ordered from mostly likely to be used to least likely.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -924,7 +924,7 @@ export default class EthereumApi implements Api {
       }
 
       const block = new RuntimeBlock(
-        Quantity.from((parentHeader.number.toBigInt() || 0n) + 1n),
+        parentHeader.number.add(1n),
```
davidmurdoch, line 927: Since we `add(1n)` more than anything else I wonder if a highly optimized `.increment()` function would be worth while. Maybe we do all the Math in a Buffer directly or something. :thinking: (this probably can/should be saved for another PR)

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -13,8 +14,9 @@ export class Quantity extends BaseJsonRpcType {
 
   _nullable: boolean = false;
 
-  public static from(value: JsonRpcInputArg, nullable = false) {
-    if (value instanceof Quantity) return value;
+  public static from(value: JsonRpcInputArg, nullable = false): Quantity {
+    if (Quantity.isQuantity(value))
+      return value;
```
davidmurdoch, line 19: This actually can't happen (per the types). Can we remove this check altogether?

```suggestion
```

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -91,59 +113,153 @@ export class Quantity extends BaseJsonRpcType {
...
+    return new Quantity(productBuffer, this._nullable);
+  }
+
+  /**
+   * Returns true if the instance passed conforms to the Quantity interface. This is a best guess effort,
+   * and doesn't confirm that the instance passed is an instance of the Quantity class.
+   * @param  {any} instance
+   * @returns boolean whether the instance conforms to the Quantity interface
+   */
+  private static isQuantity(instance: any): instance is Quantity {
```
davidmurdoch, line 234: This is used in two places, and i've suggested we remove it in one of them, leaving us with the `toBigInt` check. Which means we really only need to check if the thing passed in has a `toBigInt` function, as that's all the object really needs anyway. Checking just the one property would definitely be faster than checking all of these methods.

`"toBigInt" in value && typeof value.toBigInt === "function` could replace `Quantity.isQuantity(value)`.

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: I'll look into that. Initially I did all of the math directly in the buffer, and got somewhat promising results, but I considered the complexity to be unreasonable for the minimal gains. It is likely much more significant for a simple increment() function however, as the conversion to bigint and back is a more singificant percentage of the entire operation. I will look into it.

Edit: I created #3390 to cover this.

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -1,20 +1,43 @@
 import { bigIntToBuffer } from "../../utils/bigint-to-buffer";
 import { uintToBuffer } from "../../utils/uint-to-buffer";
-
-const BUFFER_EMPTY = Buffer.allocUnsafe(0);
+import { BUFFER_EMPTY } from "../../utils/constants";
 
 export type JsonRpcInputArg = number | bigint | string | Buffer;
 
+export class Lazy<T> {
+  private _resolver?: () => T;
```
MicaiahReid, line 8: Is there any difference between `private _resolver` and `#resolver`? If not, we should switch to be consistent with the rest of the repo.

src/packages/utils/src/things/json-rpc/input-parsers.ts
```diff
@@ -1,20 +1,43 @@
 import { bigIntToBuffer } from "../../utils/bigint-to-buffer";
 import { uintToBuffer } from "../../utils/uint-to-buffer";
-
-const BUFFER_EMPTY = Buffer.allocUnsafe(0);
+import { BUFFER_EMPTY } from "../../utils/constants";
 
 export type JsonRpcInputArg = number | bigint | string | Buffer;
 
+export class Lazy<T> {
```
MicaiahReid, line 7: I think a js-doc comment explaining this class would be worthwhile

src/packages/utils/src/things/json-rpc/input-parsers.ts
jeffsmale90: @davidmurdoch has pointed out that # prefixed private class fields perform significantly worse, in part due to how webpack polyfills these (I think?)

I don't know whether there's a reasonable fix to this, but I've updated #3240 as a tech-debt to cover this issue.

src/chains/ethereum/ethereum/src/transaction-pool.ts
jeffsmale90: Right? I had flashbacks.

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: Correct!

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: That's fair enough. I think I need to get out of thinking of these as part of a public api...

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
jeffsmale90: That's true, but we do use this in the `persistent-cache` key generation. 

Blocknumber is passed in as either string or Quantity instance. I've pushed that differentiation back out into `persistent-cache`, so that the `Quantity` implementation is cleaner.

jeffsmale90:
I've done some performance analysis on this, and created the benchmark suite here: https://github.com/jeffsmale90/quantity-perf

I think that @davidmurdoch 's concern is correct - adding these helper methods make the performance of chained math functions very poor. The work arounds I have added show some promise, but are adding complexity (maybe there's something in there that could be used to benefit Quantity performance generally).

I believe that this is no longer a valuable thing to pursue, so will put this on hold, and may return to it at a later date.

PR #3188, opened 2022-06-06
chore: create CODEOWNERS

davidmurdoch:
This is just sets up an initial `CODEOWNERS` file so we can improve our code review process in the near future.


jeffsmale90 (approved):
Seems pretty reasonable. Should we consider a group rather than individual?

MicaiahReid (approved):

PR #3199, opened 2022-06-08
feat: add `eth_getProof` RPC method

### LINKED ISSUE #382: Implement `eth_getProof` RPC message
Would be really amazing to provide support for the [`eth_getProof`](https://github.com/ethereum/EIPs/issues/1186) RPC message. I actually needed to migrate bunch of tests to geth to make them work :)

Please let me know if I can help with it.

### LINKED ISSUE #3076: No longer able to access state trie
I am working on a long-overdue upgrade to the latest ganache at [Colony](https://github.com/joincolony/colonynetwork), and believe I've hit a blocker now that that the internals of the provider are no longer accessible. We have [smoke tests](https://github.com/JoinColony/colonyNetwork/blob/develop/test-smoke/colony-storage-consistent.js#L67) that alert us if the state trie hash associated with a contract after the tests have been run has changed. The worst-case scenario that this would alert us to is the storage layout changing unexpectedly, but normally it's a change associated with the addition of a variable during an upgrade, or the address of a deployed contract that is stored changing due to the number of transactions the deploying address makes changing.

Getting the state hash for an address captures the entire storage layout, avoiding the need to check every slot that we think should be set - but even this wouldn't capture the change of a storage slot we weren't expecting to be set.

An interface making this possible, or a workaround, would be greatly appreciated! Note that read-only access (in this case) would be adequate.

jeffsmale90:
[eip-1186](https://eips.ethereum.org/EIPS/eip-1186) proposes a new `eth_getProof` RPC method, which returns a number of useful values, including hashes and proofs associated with the given address and storage keys.

Fixes: #3076 and #382


jeffsmale90:
Fixes #3199 

WIP - created draft PR to associate changes with issue.

jeffsmale90:
I'd like to look into leveraging dependency inversion in order to gain access to the underlying world trie and compute the expected values, rather than comparing with static values: #3303.

I've deferred those changes so that we can get this out and resolve the associated issues.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1060,6 +1060,96 @@ export default class EthereumApi implements Api {
...
+   * const proof = await provider.request({ method: "eth_getProof", params: [txReceipt.contractAddress, ["0x0", "0x1"], "latest"] });
+   * console.log(proof);
+   * ```
+   */
+  @assertArgLength(3)
+  async eth_getProof(
+    address: DATA,
+    slots: DATA[],
+    blockNumber: QUANTITY | Ethereum.Tag
+  ): Promise<any> {
```
gnidan: this should probably be typed better than `any`

jeffsmale90: Looking into this I found that I probably need to map the return type, so I'll have another pass. Thanks @gnidan !

src/chains/ethereum/ethereum/tests/api/eth/getProof.test.ts
```diff
@@ -0,0 +1,258 @@
...
+        "0xf869a03b2c95fd2a6e48d2dcb37be554d03b55e31ec582b450211db4e9c3883ffac678b846f8440180a0fe26ce2b5dafd303bdbd9adbd1bbc1ef59fceea4f81c282158ab624e2fd1787da067b59c91b38bdcd65d2f8129925a939a1469b88b66ef65395d22f74476582a5f"
+      ];
+      const emptyMerkelRoot = Data.toString(new SecureTrie().root);
```
gnidan: you spelled merkle wrong

jeffsmale90: So it's _not_ named after https://en.wikipedia.org/wiki/Angela_Merkel???

src/chains/ethereum/ethereum/src/api-types.ts
```diff
@@ -194,4 +194,22 @@ export namespace Ethereum {
 
   // whisper
   export type WhisperPostObject = UtilTypes.WhisperPostObject;
+
+  //#region eth_getProof
+  export type StorageProof = {
```
jeffsmale90: I've aligned this with the Geth implementation, which differs from the specification in [EIP-1186](https://eips.ethereum.org/EIPS/eip-1186#specification).

The specification defines the key as `Quantity` type. When I call `eth_getProof` on infura, specifying storage keys, I receive the keys back as a fixed length, (the value _is_ returned as a `Quantity` type as expected [the returned value is not forced to an even number of octets]). 

ie.
```
{
  key: "0x0000000000000000000000000000000000000000000000000000000000000001",
  value: "0x1",
  proof: [
    "0xf8518080a0a75fd430c15087d0a2c54dcfe4bfa47c13129638e5ec6ea8236bd5df2ad66ac68080808080808080a0f4984a11f61a2921456141df88de6e1a710d28681b91af794c5a721e47839cd78080808080",
    "0xe2a0310e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf601"
  ]
}
```

src/chains/ethereum/ethereum/tests/api/eth/getProof.test.ts
```diff
@@ -0,0 +1,316 @@
...
+
+      const contractCode =
+        "0x6080604052607b60005560018060006101000a81548173ffffffffffffffffffffffffffffffffffffffff021916908373ffffffffffffffffffffffffffffffffffffffff16021790555034801561005657600080fd5b50610169806100666000396000f3fe608060405234801561001057600080fd5b50600436106100365760003560e01c80636e8d4ab51461003b5780637104ddb214610059575b600080fd5b610043610077565b60405161005091906100bc565b60405180910390f35b61006161007d565b60405161006e9190610118565b60405180910390f35b60005481565b600160009054906101000a900473ffffffffffffffffffffffffffffffffffffffff1681565b6000819050919050565b6100b6816100a3565b82525050565b60006020820190506100d160008301846100ad565b92915050565b600073ffffffffffffffffffffffffffffffffffffffff82169050919050565b6000610102826100d7565b9050919050565b610112816100f7565b82525050565b600060208201905061012d6000830184610109565b9291505056fea26469706673582212209c21580c5505f87f9862db81d9f3aed499a5b4966d1a626e72d1a4e999576cb064736f6c634300080b0033";
+      // todo: Generate contractCode from GetProof.sol. The generated code is operating system dependent (due to metadata). Differing code generates a different proof.
```
jeffsmale90, line 25: I've some thoughts about how to resolve this, but I wanted to explore them outside of this PR so that we can get the feature out.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1060,6 +1061,103 @@ export default class EthereumApi implements Api {
     return block ? block.toJSON<IncludeTransactions>(transactions) : null;
   }
 
+  /**
```
MicaiahReid, line 1049: Can you rebase from develop so that I can test this example in the interactive docs? @davidmurdoch's latest forking in the browser fix actually fixes "eager" instamine mode in the browser.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1060,6 +1061,103 @@ export default class EthereumApi implements Api {
...
+    }
+
+    const blockchain = this.#blockchain;
+    const targetBlock = await blockchain.blocks.get(blockNumber);
+    if (targetBlock == null) throw new Error("header not found");
+
+    const stateTrie = this.#blockchain.trie.copy(false);
+    stateTrie.setContext(
+      targetBlock.header.stateRoot.toBuffer(),
+      null,
```
MicaiahReid: `setContext`'s second parameter is `address: Buffer`. While this is only actually used in the `/forking/trie` version of `setContext`, and this feature isn't supported yet with forking, can we still set this value? It feels weird to set the block number in the next param but not the address. You could move up the definition of `ganacheAddress` and `ethereumJsAddress` to here and use `ethereumJsAddres.buf`

src/chains/ethereum/ethereum/tests/api/eth/getProof.test.ts
```diff
@@ -0,0 +1,362 @@
...
+          Data.from(contractAccount.stateRoot).toString(),
+          "Unexpected storageHash"
+        );
+        assert.deepEqual(
+          result.accountProof,
+          expectedContractAccountProof,
+          "Unexpected accountProof"
+        );
+
+        const expectedStorageProof = [
```
MicaiahReid, line 241: I was playing around with this on my own and couldn't get it to work. I am deploying a contract with the same code you're using for this contract. I'd get the transaction receipt to find the contract address, and call `eth_getProof` with:
 `[ contractAddress, ["0x0000000000000000000000000000000000000000000000000000000000000000", "0x0000000000000000000000000000000000000000000000000000000000000001"], "latest"]`. 

I don't get any error, but the storage proof is: `[{"key":"0x0000000000000000000000000000000000000000000000000000000000000000","proof":[],"value":"0x0"},{"key":"0x0000000000000000000000000000000000000000000000000000000000000001","proof":[],"value":"0x0"}]`

Any thoughts on why the value and proof is empty for my attempts?

MicaiahReid (changes requested):

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Absolutely I can!

jeffsmale90: Done!

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Done!

src/chains/ethereum/ethereum/tests/api/eth/getProof.test.ts
jeffsmale90: It looks like the create contract transaction is failing with `out of gas`. I passed `"gasLimit":"0x15f900"` and got the expected proof.

MicaiahReid: Makes sense, thanks!

MicaiahReid (approved):

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: Actually, I don't know if I've fully understood the repercussions of adding that parameter to `setContext`. I just did this in my own change where I had to clone the VM, and it broke a lot of things. Maybe this should be undone (unless your confident in the change) and @davidmurdoch can teach us what's going on with that function.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1047,104 @@ export default class EthereumApi implements Api {
...
+   * * `storageHash`: `DATA` - 32 Bytes - SHA3 of the StorageRoot. All storage will deliver a
+   *   MerkleProof starting with this rootHash.
+   * * `accountProof`: `ARRAY` - Array of rlp-serialized MerkleTree-Nodes, starting with the
+   *   stateRoot-NODE, following the path of the SHA3 (address) as key.
+   * * `storageProof`: `ARRAY` - Array of storage-entries as requested. Each entry is an object
+   *   with the following properties:
+   *   * `key`: `DATA` - the requested storage key.
+   *   * `value`: `QUANTITY` - the storage value.
+   *   * `proof`: `ARRAY` - Array of rlp-serialized MerkleTree-Nodes, starting with the
+   *     storageHash-Node, following the path of the SHA3 (key) as path.
```
davidmurdoch: Reworded some stuff:

```suggestion
   * Returns the details for the account at the specified address and block
   * number, the account's Merkle proof, and the storage values for the
   * specified storage keys with their Merkle-proofs.
   *
   * @param address - Address of the account
   * @param slots - Array of storage keys to be proofed.
   * @param blockNumber - A block number, or the string "earliest", "latest", or
   * "pending".
   * @returns An object containing the details for the account at the specified
   * address and block number, the account's Merkle proof, and the
   * storage-values for the specified storage keys with their Merkle-proofs:
   * * `balance`: `QUANTITY` - the balance of the account.
   * * `codeHash`: `DATA` - 32 Bytes - hash of the account. A simple account
   *   without code will return
   *   `"0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"`
   * * `nonce`: `QUANTITY` - the nonce of the account.
   * * `storageHash`: `DATA` - 32 Bytes - SHA3 of the StorageRoot. All storage
   *   will deliver a MerkleProof starting with this rootHash.
   * * `accountProof`: `Array` - Array of rlp-serialized MerkleTree-Nodes,
   *   starting with the stateRoot-NODE, following the path of the SHA3
   *   (address) as key.
   * * `storageProof`: `Array` - Array of storage entries as requested. Each
   *   entry is an object with the following properties:
   *   * `key`: `DATA` - the requested storage key.
   *   * `value`: `QUANTITY` - the storage value.
   *   * `proof`: `Array` - Array of rlp-serialized MerkleTree-Nodes, starting
   *     with the storageHash-Node, following the path of the SHA3 (key) as
   *     path.
```

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1047,104 @@ export default class EthereumApi implements Api {
...
+   * await provider.once("message"); // Note: `await provider.once` is non-standard
+   * const txReceipt = await provider.request({ method: "eth_getTransactionReceipt", params: [txHash] });
+   * const proof = await provider.request({ method: "eth_getProof", params: [txReceipt.contractAddress, ["0x0", "0x1"], "latest"] });
+   * console.log(proof);
+   * ```
+   */
+  @assertArgLength(3)
+  async eth_getProof(
+    address: DATA,
+    slots: DATA[],
```
davidmurdoch: Should we rename `slots` to `storageKeys`?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1047,104 @@ export default class EthereumApi implements Api {
...
+    blockNumber: QUANTITY | Ethereum.Tag
+  ): Promise<Ethereum.Proof> {
+    if (this.#blockchain.fallback) {
+      throw new Error(
+        "eth_getProof is not supported on a forked network. See https://github.com/trufflesuite/ganache/issues/3234 for details."
+      );
+    }
+
+    const blockchain = this.#blockchain;
```
```suggestion
    const blockchain = this.#blockchain;

    if (blockchain.fallback) {
      throw new Error(
        "eth_getProof is not supported on a forked network. See https://github.com/trufflesuite/ganache/issues/3234 for details."
      );
    }
```

davidmurdoch: Private field access is super slow (100-1000x slower than public field access) so I like to only do it once where possible.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1047,104 @@ export default class EthereumApi implements Api {
...
+    );
+
+    const vm = await blockchain.createVmFromStateTrie(
+      stateTrie,
+      this.#options.chain.allowUnlimitedContractSize,
+      false, // precompiles have already been initialized in the stateTrie
+      blockchain.common
+    );
+
+    const ethereumJsAddress = new EthereumJsAddress(bufferAddress);
```
davidmurdoch: I think this is the first use of `EthereumJsAddress` address in our code base (other than as a typescript type).

Historically, this was because `EthereumJsAddress` was doing a bunch of unnecessary conversions and work when instantiated. Whenever an `@ethereumjs/*` package needs an account it usually only needs `{ buf: buffer }` or sometimes `{ buf: buffer, equals: (a: { buf: Buffer }) => to.equals(buffer) }`.

Perhaps we should reevaluate our internal (mis)use of `{ buf: buffer }` from a performance perspective to see if it is ever worth it, and if not, maybe we could switch to `new EthereumJSAddress` everywhere we do the hacky version.

If it _is_ worth it perhaps helper classes to create "light" versions of `EthereumJSAddress` are in order, as seen in [this commit](https://github.com/trufflesuite/ganache/blob/910cd1a684ba0c6d8ed7c2b93ddcdf94184c83b8/src/chains/ethereum/ethereum/src/helpers/simulation-handler.ts#L257-L262). But we could do that later.

I suspect it is worth it to switch to EthereumJsAddress: https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/address.ts

All that to suggest that for now this PR could switch to the hacky `{ buf }` version, and follow it up with a "tech debt" Issue to reevaluate our `{ bug: Buffer }` hacks.

src/chains/ethereum/ethereum/tests/api/eth/getProof.test.ts
davidmurdoch: :+1:

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1047,104 @@ export default class EthereumApi implements Api {
...
+
+    const slotBuffers = slots.map(slotHex => Data.toBuffer(slotHex, 32));
+    const proof = await vm.stateManager.getProof(
```
davidmurdoch: Did you look into the viability of doing something like the following?

```
const sm = this.#blockchain.vm.stateManager.copy();
await sm.setStateRoot(targetBlock.header.stateRoot.toBuffer());
const proof = await sm.getProof(
```

This would avoid having to create a new VM instance.

src/chains/ethereum/ethereum/tests/api/eth/getProof.test.ts
```diff
@@ -0,0 +1,362 @@
...
+      // todo: Generate contractCode from GetProof.sol. The generated code is operating system dependent (due to metadata). Differing code generates a different proof.
+      // This should be solved by generating the expected Proof direclty. Once this is done, most of the static assertions can be removed, as well as the deterministic wallet.
```
davidmurdoch:
```suggestion
      // This should be solved by generating the expected Proof directly. Once this is done, most of the static assertions can be removed, as well as the deterministic wallet.
```

src/chains/ethereum/ethereum/tests/api/eth/getProof.test.ts
```diff
@@ -0,0 +1,362 @@
...
+      });
+
+      it("gets the proof for an externally owned account", async () => {
+        const result = await provider.send("eth_getProof", [
+          ownerAddress,
+          [],
+          "latest"
+        ]);
+
+        assert.equal(result.address, ownerAddress, "Unexpected address");
```
davidmurdoch: These tests should probably be using `strictEqual` and `strictDeepEqual`; I don't see any obvious ways there will be a difference in test results, but it feels weird to be doing "==" comparisons :-)

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: (or just leave it as is)

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1047,104 @@ export default class EthereumApi implements Api {
...
+  ): Promise<Ethereum.Proof> {
+    if (this.#blockchain.fallback) {
+      throw new Error(
+        "eth_getProof is not supported on a forked network. See https://github.com/trufflesuite/ganache/issues/3234 for details."
+      );
+    }
+
+    const blockchain = this.#blockchain;
+    const targetBlock = await blockchain.blocks.get(blockNumber);
+    if (targetBlock == null) throw new Error("header not found");
```
davidmurdoch: if the header isn't found (is `null` or `undefined`) `blockchain.blocks.get` throws a `"header not found"` error, so this check is redundant.

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Yes. They're referred to by a couple different names (slots, positions, storageKeys), but I think `storageKeys` is the best.

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: I made #3359 as a tech debt issue to resolve directly.

I changed this reference to use the hacky "light" EJsAddress, for consistencies sake.

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Oh that's real nice

src/chains/ethereum/ethereum/src/api-types.ts
```diff
@@ -194,4 +194,22 @@ export namespace Ethereum {
...
+  export type Proof = {
+    address: Data;
+    balance: Quantity;
+    codeHash: Data;
+    nonce: Quantity;
+    storageHash: Data;
+    accountProof: Data[];
+    storageProof: StorageProof[];
+  };
+  //#endregion eth_getProof
```
davidmurdoch: Types in this file are public and user-facing. These types shouldn't export Data/Quantity, as they will be actual Ethereum JSON-RPC values (0x prefixed strings).

See `ganache/test/types.test.ts` for some types tests.

See the `TraceTransactionResult<P extends PublicPrivate = "public">` type in this same file for an how to manage a type that is different internally than it is externally.

src/chains/ethereum/ethereum/src/api-types.ts
```diff
@@ -194,4 +194,22 @@ export namespace Ethereum {
...
   export type WhisperPostObject = UtilTypes.WhisperPostObject;
+
+  //#region eth_getProof
+  export type StorageProof = {
+    key: Data;
+    proof: Data[];
+    value: Quantity;
+  };
+
+  export type Proof = {
```
davidmurdoch: We should discuss if this is the best name for this (since it is part of our public interface). This object isn't really a "Proof"; though it does contain a bunch of proofs. :thinking:

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/api-types.ts
jeffsmale90: It's a little confusing initially, but I think I got it 🤞

src/chains/ethereum/ethereum/src/api-types.ts
jeffsmale90: Renamed to `AccountProof`

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1046,99 @@ export default class EthereumApi implements Api {
...
+   * const txReceipt = await provider.request({ method: "eth_getTransactionReceipt", params: [txHash] });
+   * const proof = await provider.request({ method: "eth_getProof", params: [txReceipt.contractAddress, ["0x0", "0x1"], "latest"] });
+   * console.log(proof);
+   * ```
+   */
+  @assertArgLength(3)
+  async eth_getProof(
+    address: DATA,
+    storageKeys: DATA[],
+    blockNumber: QUANTITY | Ethereum.Tag
```
davidmurdoch: Per https://twitter.com/atdavidmurdoch/status/1549825055353356289 let's make give this a default `Tag.latest` (and don't forget to update `assertArgLength(2, 3)`, too)! :-D

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1046,99 @@ export default class EthereumApi implements Api {
...
+  ): Promise<Ethereum.AccountProof<"private">> {
+    if (this.#blockchain.fallback) {
+      throw new Error(
+        "eth_getProof is not supported on a forked network. See https://github.com/trufflesuite/ganache/issues/3234 for details."
+      );
+    }
+    const blockchain = this.#blockchain;
```
davidmurdoch: Don't access the private member `#blockchain` twice; it's slow.

```suggestion
    const blockchain = this.#blockchain;
    if (blockchain.fallback) {
      throw new Error(
        "eth_getProof is not supported on a forked network. See https://github.com/trufflesuite/ganache/issues/3234 for details."
      );
    }
```

src/chains/ethereum/utils/src/index.ts
```diff
@@ -18,3 +18,4 @@ export * from "./types/shh";
 export * from "./types/subscriptions";
 export * from "./types/trace-transaction";
 export * from "./types/tuple-from-union";
+export * from "./types/account-proof";
```
davidmurdoch: files should have a trailing newline (prettier should automatically add them)

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1046,6 +1046,99 @@ export default class EthereumApi implements Api {
...
+    const targetBlock = await blockchain.blocks.get(blockNumber);
+
+    const ganacheAddress = Address.from(address);
+    const bufferAddress = ganacheAddress.toBuffer();
+    const ethereumJsAddress = { buf: bufferAddress, equals: (a:{buf:Buffer}) => a.buf.equals(bufferAddress)} as any;
```
davidmurdoch: I don't think the`getProof` method will ever use the `equals`, so we don't need to create it (code coverage doesn't hit this function).

```suggestion
    const ethereumJsAddress = { buf: bufferAddress } as any;
```

Also, it looks like these files needs to be prettiered.

davidmurdoch (changes requested):
ALMOST THERE!

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Done! I also created #3396 to address other methods.

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Missed that one!

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: I didn't think there were any others.

davidmurdoch (approved):

PR #3226, opened 2022-06-13
fix: ensure `COINBASE` instruction is correct in `debug_traceTransaction`

robmcl4:
Previously, the instruction `COINBASE` was returning the address of the prior block's miner, which resulted in an incorrect debug trace. This corrects it to the miner of the traced transaction's block.

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -1394,7 +1394,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     const newBlock = new RuntimeBlock(
       Quantity.from((parentBlock.header.number.toBigInt() || 0n) + 1n),
       parentBlock.hash(),
-      parentBlock.header.miner,
+      targetBlock.header.miner,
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
This makes sense to me, but I'd definitely like to get @davidmurdoch's approval, as he knows much more about `debug_traceTransaction` than I do. He's out at a company retreat, so it might be a bit before he has a chance to take a look. 

Thank you for the PR! I inevitably learn something new from your contributions.

robmcl4:
Actually, it's also necessary to change the difficulty as well -- or else the `DIFFICULTY` instruction will run incorrectly. For example, it causes this transaction to `debug_traceTransaction` incorrectly: [`0x4fdc0efb59dbc47759bb23069eb5eec83079068818ade9578f57c13adb2b6290`](https://etherscan.io/tx/0x4fdc0efb59dbc47759bb23069eb5eec83079068818ade9578f57c13adb2b6290)

davidmurdoch:
Awesome. Thanks for finding these issues and proposing fixes. I'd like to add some tests for this before merging. We might want to update this PR to fix DIFFICULTY as well.

PR #3227, opened 2022-06-13
fix: EIP-1559 access list transaction gas cost

robmcl4:
EIP-1559 'Access List' transactions were not counting gas correctly. There are two issues here:

1. up-front gas cost did not factor in the cost of of the access list itself
2. access list addresses were not being properly marked as warmed by the VM, because address buffers were improperly extended to 32 bytes when the EVM expected 20 byte addresses


robmcl4:
For an example of such a transaction, see https://etherscan.io/tx/0xc33ffe3a45e7da5d7806ac370244c50abbd758c2a6e492826ca536961986c052

src/chains/ethereum/transaction/src/access-lists.ts
```diff
@@ -43,7 +43,7 @@ export class AccessLists {
       const json: AccessList = [];
       for (let i = 0; i < bufferAccessList.length; i++) {
         const data = bufferAccessList[i];
-        const address = Data.from(data[0], 32).toString();
+        const address = Data.from(data[0], 20).toString();
```
MicaiahReid: It won't let me add a comment to the line (I guess cause it's too far from the line that actually changed?) but line 30 above needs this change as well. This is for the case where we are creating an address buffer from an address string, we also are incorrectly padding to 32 bytes. Great find!

MicaiahReid (changes requested):

src/chains/ethereum/transaction/src/access-lists.ts
robmcl4: followed-up to fix that one too, thanks!

MicaiahReid (approved):

davidmurdoch (approved):

PR #3232, opened 2022-06-14
feat: build a real pending block

### LINKED ISSUE #772: Build a real pending block!
https://github.com/trufflesuite/ganache-core/blob/e3614bc4b446469bea4522f7007b4f2477c5fdc3/src/chains/ethereum/ethereum/src/data-managers/block-manager.ts#L57

MicaiahReid:
This change expands `eth_getBlockByNumber` so that Ganache actually returns a `pending` block when you call it and pass the "pending" tag. _shhhhh, don't tell anyone but in the past we'd just return the `latest` block._

So, if you retrieve the `pending` block, Ganache will now return a block that is _after_ the `latest` block and which contains the transactions that are still waiting in the transaction pool to be mined. It's almost like we can see into the 
<p align="center">
  future
</p>

Sadly our clairvoyance does have its limitations, so don't expect transactions that haven't made it into the transaction pool yet to be in the `pending` block.

Fixes #772, #3481, #3529


src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -370,7 +370,7 @@ export default class Miner extends Emittery<{
             // notice: when `maxTransactions` is `-1` (AKA infinite), `numTransactions === maxTransactions`
             // will always return false, so this comparison works out fine.
             if (
-              blockGasLeft <= Params.TRANSACTION_GAS ||
+              blockGasLeft < Params.TRANSACTION_GAS ||
```
MicaiahReid, line 378: Writing my tests lead me to this bug - if the block's remaining gas was _equal_ to the amount of gas it takes to run an empty transaction, we'd break out. If a block had used `gasLimit - 21000`, it wouldn't mine one more empty transaction. This can be another PR if needed.

MicaiahReid:
For the reviewers - I have a few notes on design decisions.

When I first started this PR about a year ago, David and I went back and forth on how to handle some race conditions. 

 - What do we do if there are constantly transactions coming into the pool that alter the results of the pending block? Do we fix the pending block, or return one that we know to be wrong?
 - What do we do if a block is being mined as the user requests the pending block? Do we wait to mine the new block, then make a pending block off of that one? Or do we just make a pending block off of the current latest block?

Like most things in life, my answer is to take the easy way. When a pending block is requested, we clone the state of the vm and executable transaction pool _as is, right now_, and make a pending block off of it.
>What do we do if there are constantly transactions coming into the pool that alter the results of the pending block? Do we fix the pending block, or return one that we know to be wrong?

We return a pending block that is already outdated and incorrect. There will always be the possibility of the pending block being incorrect, so I don't think it's worth spending too much energy to make sure it's accurate in this race condition. In the real world, requesting a pending block is just one node's opinion on the next block, based off of the transactions in their pool. So it's even less likely that a pending block is correct in the real world.

>What do we do if a block is being mined as the user requests the pending block? Do we wait to mine the new block, then make a pending block off of that one? Or do we just make a pending block off of the current latest block?

In this case we still make a pending block based off of the latest block at the time the request is made. If mining has already started as that request is made, because creating the pending block requires mining all of the same transactions that will be included in the already started block, the pending block will finish second to the already started block. We considered pausing the real miner to make sure the pending block returns first, but it seems silly to slow down real mining for this.

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -97,6 +99,288 @@ describe("api", () => {
...
+            [from, to] = await provider.send("eth_accounts");
+          });
+
+          afterEach(async () => {
+            provider && (await provider.disconnect());
+          });
+
+          it(`has one transaction regardless of the number of executables in the pool and is equal to next mined block`, async () => {
+            // send 1 more transaction than will fit in the block
+            for (let i = 0; i < emptyTransactionsPerBlock + 1; i++) {
```
tenthirtyone, line 247: I think this is fine but without `miner_stop` would this be a bit race-y at higher values?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,71 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+
+    // we don't want any events from mining to be
+    // caught by `this.vm`, so we need to make a new one.
+    const vm = await this.createVmFromStateTrie(
+      this.trie.copy(false),
+      options.chain.allowUnlimitedContractSize,
+      true
+    );
+
+    const miner = new Miner(minerOpts, executables, vm, this.#readyNextBlock);
```
tenthirtyone: So it looks like `mine` is destructive and the way to navigate preserving state with making an accurate prediction is spinning up a miner with a cloned state? It seems like the least-invasive way. 

Is there a way to create the pending block using reads? I understand the `mine` function is destructive and may make that difficult

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -452,6 +453,40 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
...
+
+    if (pending.size > 0) {
+      pending.forEach(txsFromOrigin => {
+        txsFromOrigin.array.forEach(tx => {
+          tx.locked = false;
+        });
+      });
+    }
+
+    if (inProgress.size > 0) {
```
tenthirtyone: Good defensive programming but I don't think you need this here to protect the `forEach`.

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -452,6 +453,40 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
...
+  /**
+   * Deep clones and resets the transaction pool's `executables` such that `inProgress`
+   * transactions are moved back to `pending` and all transactions are unlocked.
+   * @returns Cloned and reset copy of transaction pool's `executables`.
+   */
+  public cloneAndResetExecutables() {
+    const executables = cloneDeep(this.executables);
+    const { inProgress, pending } = executables;
+
+    if (pending.size > 0) {
```
tenthirtyone: Same as below, I don't think you need this check

tenthirtyone (changes requested):

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: Too true, fixed!

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: Fixed!

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: I think there probably is, but there'd be a major trade-off on the accuracy of the block. I figured the miner already has the infrastructure to generate an accurate block, so why not reuse it ¯\_(ツ)_/¯

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
MicaiahReid: Yes, I think you're right. That's why I limit it to only a few transactions :)

I suppose for this test I could just send two transactions, because I'm just proving that it will only add one transaction to the block. Hmm

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,71 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+    // the number of transactions that should be included in this block depends
+    // on a few factors. in "instamine" mode (meaning blockTime = 0), we usually
+    // only mine 1 transaction per block. However, if the miner is currently
+    // paused (from miner_stop being called), whenever they restart the miner
+    // all executable transactions will be mined, so the pending block should
+    // contain all executables that will fit in a block. if we're not in
+    // instamine mode, the miner will always try to fill the block, regardless
+    // of whether the miner is paused or not
+    const paused = this.#isPaused();
+    const maxTransactions = instamine
```
jeffsmale90: nit: this could be simplified to a single conditional:
```javascript
const maxTransactions = instamine && !paused ? Capacity.Single : Capacity.FillBlock;
```

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -97,6 +99,288 @@ describe("api", () => {
...
+      describe(`passing "pending" tag`, () => {
+        let from, to: string;
+        const emptyTransactionsPerBlock = 2;
+        const blockGasLimit = "0xA410"; // enough gas for 2 empty txs
+
+        /**
+         * Fetches a block with number `pendingBlock.number` and asserts that
+         * the resultant block is deep strict equal to the `pendingBlock`.
+         *
+         * NOTE: The `hash` and `timestamp` properties are removed from the
```
jeffsmale90: I think the hash should be null for `pending` block: https://github.com/trufflesuite/ganache/blob/90b3f2b743974b13d23ec72e713350021655c6f9/src/chains/ethereum/ethereum/src/api.ts#L1007

I validated that with this change `pending` block has a _non-null_ hash, and that geth returns a null hash.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,71 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+    const options = this.#options;
+    const minerOpts = options.miner;
+    // "instamine" mode means blockTime = 0
+    const instamine = this.#instamine;
+    let timestamp: number = null;
+    if (!instamine) {
+      // we can't always get the pending block's timestamp right because they
+      // could have the miner paused, but if blockTime > 0, blocks are mined at
+      // regular intervals, so we can at least make an educated guess
+      timestamp =
```
jeffsmale90: This will need to be changed once #3317 is merged - I'll take that as a follow up action.

Even without #3317 - I think this is only correct if a couple of assumptions hold true:
- `miner.timestampIncrement` is "clock"
- the offset has not been changed via `eth_setTime` or `eth_increaseTime` since the last block was mined

Would it be more reasonable to return the timestamp as if the block were mined immediately (ie not pass a value into `#readyNextBlock` ? Although the next block will probably be mined after `miner.blockTime` seconds, that might not be true, for instance if `evm_mine` is called.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,71 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+        previousBlock.header.timestamp.toNumber() + minerOpts.blockTime;
+    }
+    const nextBlock = this.#readyNextBlock(previousBlock, timestamp);
+    // If the executables has inProgress transactions, we were already
+    // in the middle of mining a block when the pending block was requested.
+    // The block that is currently being mined is what we want to return as pending,
+    // so we'll reset the executables to be what it was when mining started,
+    // i.e. all inProgress executables are moved to the front of the pending executables.
```
jeffsmale90, line 601: The way I understand this is that any transactions that have been removed from the transaction pool (for creating the next block), but not mined into `latest` yet, are sitting in `executables` - is this correct?

So there's no way for a transaction to be removed from the pool, and _not_ make it into `pending` because it's in the process of being mined?

jeffsmale90 (changes requested):
Partial review, I got pulled down a rabbit hole :)

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: That's _wayyy_ better, thanks!

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,71 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+    const maxTransactions = instamine
+      ? paused
+        ? Capacity.FillBlock
+        : Capacity.Single
+      : Capacity.FillBlock;
+
+    // we don't want any events from mining to be
+    // caught by `this.vm`, so we need to make a new one.
+    const vm = await this.createVmFromStateTrie(
+      this.trie.copy(false),
```
jeffsmale90: This is more a query for my own learning:

`this.trie.copy()` tells me that it `Returns a copy of the underlying trie`, which calls `this.db.copy()` which `Returns a copy of the DB instance, with a reference to the **same** underlying leveldb instance.`

So this tells me that any difference between the instances will be ephemeral data that isn't persisted to the database? 

I did some investigations, and the trie root _is_ different after creating a "pending" block, _if_ we don't create a copy, so obviously there is something different between the original instance and the copy.

Do you know what that is? Or am I missing something entirely?

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Hmm, I think there's some misunderstandings on how the transaction pool works here, but I could be just misunderstanding you. In researching this PR, I wrote up [some notes](https://hackmd.io/FgAwge5kS26budoVrwn8DA) on hackmd that might be helpful with definitions, but I can elaborate here.

A transaction enters the pool and stays there _until it is included on a block and `"finalized"`_. Up until that point, it stays in the pool.

When a transaction enters the pool, it can be either `executable` or `queued`. Executable transactions are valid transactions that have a nonce of `sender.transactionCount + 1`. Queued transactions are valid transactions that have a nonce that's higher than that.

The executable transactions are further divided in to `pending` and `inProgress`. Pending transactions are waiting to be run. In progress transactions have been run (meaning `vm.runTx` has been called on the transaction), but the block hasn't been finalized yet (maybe cause the miner is still running more transactions to go on that block), so it's kinda in limbo still.

So, for Ganache, a transaction lifecycle is something like:

`pending` => `runTx called` => `removed from pending and added to inProgress` => `block finalized and removed from inProgress`

The transaction is in the transaction pool until the last step.

MicaiahReid: After typing all of that up and rereading your question a few more times, I think I now understand your question. I should listen more and talk less :sweat_smile: 

My interpretation of your question:
If a transaction has been moved from the executable's `pending` transactions to `inProgress` transactions and the `pending` block is requested, will that transaction always be in the `pending` block?

I _think_ the answer here is almost always yes. The transaction will be run again on a new block with identical fields except potentially the `timestamp`. So if one of those transactions that is already mined is a contract call that is somehow dependent on the block's `timestamp`, it could potentially fail on the `pending` block when it didn't originally.

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
MicaiahReid: Good catch! I didn't realize that geth's pending block has a null hash.

This is surprisingly difficult with how a block's hash is implemented. I've added a test for this case and a fix [here](https://github.com/trufflesuite/ganache/pull/3232/files#diff-eb13881c7bcdaf19f3b7fe3f7c0b32570442528d3511e250621734dbbcb74992R643), but I'm not particularly happy with it.

The block doesn't actually have a `hash` field (in Ganache). It has a `hash` function, which returns a private cashed `_hash` (if it's previously been computed), or a hash of the block data, with type `Data`. So I can't directly set the block's `hash` (doesn't exist) or its `_hash` (private), so I have to set a `hash()` function. Even worse, `Data` types aren't nullable, so I'm resorting to:
```TypeScript
pendingBlock.hash = () => {
  return Quantity.from(null, true);
};
```

Let me know if you have any better alternatives :thinking:

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: Awesome, thanks for the clarification! Definitely helped me.

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
jeffsmale90: Erg - that's unfortunate on both counts.

I wonder if setting a flag (I'll leave it to you to figure out a nicer implementation) to indicate whether the block has been mined or not (or some other flag) would be tidier than monkey patching / duck punching the ash function.

```javascript
  public isMined = true;
  private _hash: Data;
  hash() {
    if (this.isMined) {
      return (
        this._hash || (this._hash = Data.from(keccak(encode(this._raw)), 32))
      );
    } else {
      return Quantity.Empty;
    }
  }
```

Also `Quantity.Empty` is a constant null `Quantity`. Maybe it should be called `Quantity.Null`, but it's not 🤷

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,68 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+        serialized: Buffer;
+        storageKeys: StorageKeys;
+        transactions: TypedTransaction[];
+      }) => {
+        pendingBlock = blockData.block;
+      }
+    );
+    // await so we're sure the above line has had time to set the pending block
+    // even though we don't actually need the result of `miner.mine`
+    await miner.mine(nextBlock, maxTransactions, true);
```
jeffsmale90: I'm guessing that we can be confident that event handlers complete before resolving the return value of `miner.mine`, but would it be better to directly await the block?

Something like:
```javascript
const blockPromise = new Promise<Block>((resolve, reject) => {
miner.on(
      "block",
      async (blockData: {
        block: Block;
        serialized: Buffer;
        storageKeys: StorageKeys;
        transactions: TypedTransaction[];
      }) => {
        resolve(blockData.block);
      }
    );
});
miner.mine(nextBlock, maxTransactions, true);
return await blockPromise;
````

File: src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts [hunk 12]
```diff
@@ -97,6 +99,299 @@ describe("api", () => {
...
+         * block for comparison, since the `timestamp` of the block is largely
+         * dependent on CI speeds and the `hash` will be impacted by the
+         * `timestamp`.
+         * @param pendingBlock The pending block to compare.
+         */
+        const assertPendingEqualsMined = async (
```
jeffsmale90, line 119: It's tricky without being able to interrogate the trie directly, but it'd be nice to be able to validate that mining the `pending` block doesn't affect the state of the trie.

Something like:
```javascript
const rootBefore = provider.#blockChain.trie.root.toString("hex");
//... create a pending block
const rootAfter = provider.#blockChain.trie.root.toString("hex");
assert.deepStrictEqual(rootBefore, rootAfter);
```

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Fair point. I was torn on adding this piece, and I hadn't even thought to consider `evm_mine` being used. It seems there are too many caveats to being able to make an "educated guess" on the timestamp, so I agree, I will omit passing a timestamp to `#readyNextBlock`.

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
MicaiahReid: That's part of what's tricky here - the `pending` block _is_ mined. We clone the vm/executables/miner, and mine the block. So when the block is returned and `block.toJson` is called, there's currently no difference between this block and a regular block. Perhaps we could add a field that's passed into a block's `toJson` function to indicate that this is a pending block, I'd just hate to see more code getting mucked up for the small case of requesting a pending block. @davidmurdoch, thoughts?

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Wayyyyy better, fixed!

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
MicaiahReid: I've added a test to do this, and it lead to me finding that we _were_ altering the state of things! Good suggestion. I've fixed this by adding `await vm.stateManager.checkpoint()`

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
MicaiahReid: Side note, I also found that the `stateRoot` for the pending block should be `"0x0000000000000000000000000000000000000000000000000000000000000000"`

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,63 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     );
   };
 
+  createPendingBlock = async (previousBlock: Block) => {
+    const options = this.#options;
+    const minerOpts = options.miner;
```
davidmurdoch: you could `const {miner: minerOptions, chain: chainOptions} = this.#options;` if you'd like.

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -162,12 +156,23 @@ export default class BlockManager extends Manager<Block> {
   }
 
   getEffectiveNumber(
-    tagOrBlockNumber: QUANTITY | Buffer | Tag = typeof Tag.latest
+    tagOrBlockNumber: QUANTITY | Buffer | Tag = Tag.latest
```
davidmurdoch: Oh look at that little bug just sitting here waiting. Great find! We should probably remove the default "Tag" here since we don't use it anyway, which would turn this into:

```
  getEffectiveNumber(
    tagOrBlockNumber: QUANTITY | Buffer | Tag
  ): Quantity
```

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -146,14 +141,13 @@ export default class BlockManager extends Manager<Block> {
...
     switch (tag) {
-      case "latest":
+      case Tag.latest:
         return this.latest;
-      case "pending":
-        // TODO: build a real pending block!
-        return this.latest; // this.createBlock(this.latest.header);
-      case "earliest":
+      case Tag.pending:
+        return await this.createPendingBlock(this.latest);
```
davidmurdoch: I was originally against making `getBlockByTag` `async`, but I think it's fine, since `getEffectiveNumber` was still able to be synchronous.

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -162,12 +156,23 @@ export default class BlockManager extends Manager<Block> {
...
 
   getEffectiveNumber(
-    tagOrBlockNumber: QUANTITY | Buffer | Tag = typeof Tag.latest
+    tagOrBlockNumber: QUANTITY | Buffer | Tag = Tag.latest
   ): Quantity {
     if (typeof tagOrBlockNumber === "string") {
-      const block = this.getBlockByTag(tagOrBlockNumber as Tag);
-      if (block) {
-        return block.header.number;
+      // having another switch to get the number rather than using
```
davidmurdoch: Since `getBlockByTag` could easily be refactored into something that is not a `switch`, I think the comment referencing its internals here should be changed/removed. Great comment and change otherwise!

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -222,7 +227,7 @@ export default class BlockManager extends Manager<Block> {
 
   async get(tagOrBlockNumber: QUANTITY | Buffer | Tag) {
     if (typeof tagOrBlockNumber === "string") {
-      const block = this.getBlockByTag(tagOrBlockNumber as Tag);
+      const block = await this.getBlockByTag(tagOrBlockNumber as Tag);
```
davidmurdoch, line 237: :+1:

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -333,4 +338,8 @@ export default class BlockManager extends Manager<Block> {
       }
     }
   }
+
+  async createPendingBlock(previousBlock: Block) {
+    return await this.#blockchain.createPendingBlock(previousBlock);
```
davidmurdoch: since we only use `blockManager's createPendingBlock` once, and that use is within itself, let's remove this public instance function and just call blockchain's `createPendingBlock` directly.

src/chains/ethereum/ethereum/src/miner/miner.ts
davidmurdoch, line 378: Can you also create a separate Issue for this that this PR will close? It'll make it a little easier to point to difference in behavior before/after this change (in the unlikely event this changes things for anyone)

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -452,6 +453,37 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
     }
     return null;
   }
+  /**
+   * Deep clones and resets the transaction pool's `executables` such that `inProgress`
+   * transactions are moved back to `pending` and all transactions are unlocked.
+   * @returns Cloned and reset copy of transaction pool's `executables`.
+   */
+  public cloneAndResetExecutables() {
+    const executables = cloneDeep(this.executables);
```
davidmurdoch: I'm not sure how `cloneDeep` can successfully clone class instances, functions, etc. Maybe we just get lucky here?

It does make me a bit uncomfortable. It'd be nice if each transaction should have a `clone` or `copy` method that returns a new instance of the transaction. Then we could have a safe Transaction and Heap instances, while also only looping over everything once. I wrote what that could potentially look like (this probably misses important details):

```typescript
  public cloneAndResetExecutables() {
    const executables: Executables = {
      pending: new Map(),
      inProgress: new Set()
    };
    this.executables.inProgress.forEach(transaction => {
      const copy = transaction.copy();
      const origin = copy.from.toString();
      const txsFromOrigin = executables.pending.get(origin);
      if (txsFromOrigin) txsFromOrigin.push(copy);
      else executables.pending.set(origin, Heap.from(copy, byNonce));
    });
    this.executables.pending.forEach(({ array, length }, from) => {
      let newOrigin = executables.pending.get(from);
      if (!newOrigin) {
        newOrigin = new Heap(byNonce);
        executables.pending.set(from, newOrigin);
      }
      for (let i = 0; i < length; i++) newOrigin.push(array[i].copy());
    });
    return executables;
 }
```

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -452,6 +453,37 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
...
+    const { inProgress, pending } = executables;
+
+    pending.forEach(txsFromOrigin => {
+      txsFromOrigin.array.forEach(tx => {
```
davidmurdoch: You shouldn't loop over the whole array, just `0` through `txsFromOrigin.length`.

A `Heap` may have junk items at the end it hasn't bothered to remove yet.

Actually, i'm now unsure if this is true (and we may have bugs that rely on this assumption). Does the Heap guarantee that the left-most items in its storage `array` are the relevant items? Hmmmm

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -452,6 +453,37 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
...
+      });
+    });
+
+    inProgress.forEach(transaction => {
+      transaction.locked = false;
+      const origin = transaction.from.toString();
+      if (pending.has(origin)) {
+        const txsFromOrigin = pending.get(origin);
+        txsFromOrigin.push(transaction);
+        pending.set(origin, txsFromOrigin);
```
davidmurdoch: since `pending` already `has` `txsFromOrigin` you don't need to `set` it here. It's also faster to skip the `has` and only do the `get` if you are going to `get` it anyway.

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -452,6 +453,37 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
     }
     return null;
   }
+  /**
+   * Deep clones and resets the transaction pool's `executables` such that `inProgress`
+   * transactions are moved back to `pending` and all transactions are unlocked.
+   * @returns Cloned and reset copy of transaction pool's `executables`.
+   */
+  public cloneAndResetExecutables() {
```
davidmurdoch: Feel free to shoot this idea down post haste... what if we refactor executables and move this function to a new `Executables` class?

```typescript
public readonly executables: new Executables();

// ...

 class Executables {
    public readonly inProgress: new Set(); // should this be called `purgatory`? :-D
    public readonly pending: new Map();
    public cloneAndReset() { // not my favorite name for this /shrug
      const executables = new Executables();
      // ... do the things
      return executables;
    }
  }
  ```
  
  Actually, now that I think about it... should Executables have a method to `removeBestFromOrigin(best, origin)`? :thinking:

File: src/chains/ethereum/ethereum/tests/helpers/compare-chain-state.ts [hunk 23]
```diff
@@ -0,0 +1,70 @@
...
+ * @param blockchain
+ * @param addresses
+ * @param testFunction
+ * @param equal
+ */
+export const compareBlockchainState = async (
+  blockchain: Blockchain,
+  addresses: EthereumJsAddress[],
+  testFunction: () => void,
+  equal: boolean = true
```
davidmurdoch: `compare` functions (at least in JavaScript) typically return a `number` where 0 means they are equal, negative means `a` is less than `b`, and positive means `a` is greater than `b`. I think this function should be renamed.

I'm not a fan of the `equal` param here. Myabe this function could return `true`/`false` instead?

src/chains/ethereum/ethereum/tests/helpers/compare-chain-state.ts
```diff
@@ -0,0 +1,70 @@
...
+import Blockchain from "../../src/blockchain";
+import { GanacheTrie } from "../../src/helpers/trie";
+
+/**
+ * Gets all db data from a trie.
+ * @param trie
+ * @returns
+ */
+const getDbData = async (trie: GanacheTrie) => {
+  let dbData: (string | Buffer)[] = [];
```
davidmurdoch:
```suggestion
  const dbData: (string | Buffer)[] = [];
```

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -97,6 +99,294 @@ describe("api", () => {
           });
         });
       });
+
+      describe(`passing "pending" tag`, () => {
+        let from, to: string;
```
davidmurdoch:
```suggestion
        let from: string, to: string;
```

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -97,6 +99,294 @@ describe("api", () => {
...
+              "latest"
+            ]);
+            assert.strictEqual(
+              Quantity.toNumber(pendingBlock.number),
+              Quantity.toNumber(latestBlock.number) + 1,
+              `Pending block doesn't have expected number.`
+            );
+          });
+
+          it(`has the correct \`baseFeePerGas\``, async () => {
```
davidmurdoch:
```suggestion
          it("has the correct `baseFeePerGas`", async () => {
```

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -162,12 +156,23 @@ export default class BlockManager extends Manager<Block> {
...
-        return block.header.number;
+      // having another switch to get the number rather than using
+      // `getBlockByTag` allows us to run this synchronously since we don't have
+      // to make a pending block
+      const tag = tagOrBlockNumber as Tag;
+      switch (tag) {
+        case Tag.latest:
+          return this.latest.header.number;
+        case Tag.pending:
+          return Quantity.from(this.latest.header.number.toNumber() + 1);
```
davidmurdoch: It is a bit silly, since the number of blocks number would be ridiculously high... but I think `toBigInt() + 1n` is safer (should probably update test's use of `toNumber()`, too).

I think that tecchhniically someone could start a genesis block at block number <insert very large number here> if they wanted to. And we will likely allow genesis.json block's soon. Maybe that's not possible? hmm

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,63 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+    const vm = await this.createVmFromStateTrie(
+      this.trie.copy(false),
+      options.chain.allowUnlimitedContractSize,
+      true
+    );
+    // checkpoint to prevent writing any changes to the database
+    await vm.stateManager.checkpoint();
+    const miner = new Miner(minerOpts, executables, vm, this.#readyNextBlock);
+    const pendingBlockPromise = new Promise<Block>(resolve => {
+      miner.on(
```
davidmurdoch: can you use `const { block } = miner.once("block")` here instead of wrapping this in a Promise?

src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
```diff
@@ -908,87 +907,57 @@ describe("api", () => {
...
-          // copy the trie, its database, the vm, and the accounts again for comparison
-          const after = await getBlockchainState();
-
-          // simulating a transaction does not change any of the data
-          assert.deepStrictEqual(before, after);
+          const testFunction = async () => {
+            // simulate the transaction
+            await blockchain.simulateTransaction(simTx, parentBlock, {});
+          };
+          await compareBlockchainState(blockchain, addresses, testFunction);
```
davidmurdoch: neat! <3

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -97,6 +99,294 @@ describe("api", () => {
...
+              "pending"
+            ]);
+            assert.strictEqual(
+              pendingBlock.transactions.length,
+              0,
+              `Pending block isn't empty when the transaction pool was empty.`
+            );
+          });
+
+          it(`has \`hash\` of \`null\``, async () => {
```
davidmurdoch:
```suggestion
          it("has `hash` of `null`", async () => {
```

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -97,6 +99,294 @@ describe("api", () => {
...
+              "pending"
+            ]);
+            assert.strictEqual(
+              pendingBlock.hash,
+              null,
+              `Pending block doesn't have a \`hash\` of \`null\`. Actual: ${pendingBlock.hash}`
+            );
+          });
+
+          it(`has \`stateRoot\` of \`"0x0000000000000000000000000000000000000000000000000000000000000000"\``, async () => {
```
davidmurdoch:
```suggestion
          it('has `stateRoot` of `"0x0000000000000000000000000000000000000000000000000000000000000000"`', async () => {
```

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
davidmurdoch: If we use `memdown` for our db in this test I think it would be (more) deterministic -- at least in regards to the timing of when internal promises are resolved, and would no longer be at the whims of disk IO. If this test could possibly fail due to faster/slower I/O then we should remedy that.

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
```diff
@@ -97,6 +99,294 @@ describe("api", () => {
...
+          const subId = await provider.send("eth_subscribe", ["newHeads"]);
+          // this won't do anything if the miner isn't paused, so no harm in
+          // calling
+          await provider.send("miner_start");
+          await provider.once("message");
+          await provider.send("eth_unsubscribe", [subId]);
+        };
+
+        const itFillsBlock = () =>
+          it(`has all executable transactions in the pool that will fit within the block \`gasLimit\` and is equal to next mined block`, async () => {
```
davidmurdoch:
```suggestion
          it("has all executable transactions in the pool that will fit within the block `gasLimit` and is equal to next mined block", async () => {
```

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
davidmurdoch: One more note: `getEffectiveNumber`'s `Tag.pending` branch isn't tested. Can you add a test for this?

davidmurdoch: image

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Done! Though I had to rearrange a bit to not await `miner.mine`, is that what you intended?

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -452,6 +453,37 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
...
+    inProgress.forEach(transaction => {
+      transaction.locked = false;
+      const origin = transaction.from.toString();
+      if (pending.has(origin)) {
+        const txsFromOrigin = pending.get(origin);
+        txsFromOrigin.push(transaction);
+        pending.set(origin, txsFromOrigin);
+      } else {
+        const newHeap = Heap.from(transaction, byNonce);
+        pending.set(origin, newHeap);
```
davidmurdoch: I don't think this gets hit by any tests. Can you add some?

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Done!

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
MicaiahReid: Changed to `toBigInt()`!

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: Yeah, probably. And i mean to `await miner.once` in my previous comment. My hope was the change would make it simpler and more readable.

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
MicaiahReid: 🎉

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
davidmurdoch: I think I like:

```
block.hash = () => Quantity.Empty;
```

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
MicaiahReid: Fixed and added a test!

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
MicaiahReid: I _think_ I fixed it inline with your comment, resolve if you agree or let me know if not.

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
MicaiahReid: Done!

src/chains/ethereum/ethereum/src/miner/miner.ts
MicaiahReid: Done! #3529

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
MicaiahReid: Done!

src/chains/ethereum/ethereum/tests/helpers/compare-chain-state.ts
MicaiahReid: Renamed to `statesAreDeepStrictEqual`, updated to return their comparison result, and removed the `equal` param.

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: Fixed to only loop up to `txsFromOrigin.length`

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: Fixed, thanks!

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: I like it! I have a couple of follow up questions.
1. Do you think adding this before #1330 will further entrench us in tech debt in our transaction classes?
  - If so, do you think it's worth making a separate issue for this and linking it to #1330 to be worked in conjunction?
2. Since we add all `inProgress` transactions to `pending`, wouldn't it be more performant to look through `pending` transactions first and then the `inProgress` transactions? It seems like we're doing extra work, unless I'm missing something.
3. Why do spell checkers think that performant is not a word? I feel like they're gas lighting me when they underline that word in red.

davidmurdoch: `3`. I Google "performant" every time. I've finally done `Right Click` -> `Add to dictionary`. Thanks!
 `2`. My _hunch_ is that `inProgress` first would be more performant because the transactions in `inProgress` will always sort _before_ the transactions in `pending`, and insertion queues (`Heap`) is fastest when the items being inserted are already ordered (I have NOT actually tested this though).
 `1`. It might be adding some tech debt, but I'm worried that `deepClone` has hidden/subtle bugs, since we are cloning class instances, not just plain objects -- and hidden/subtle bugs probably worse.

MicaiahReid: @davidmurdoch Thanks for the response. Your backwards ordering is deeply concerning.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,55 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+    const miner = new Miner(
+      minerOptions,
+      executables,
+      vm,
+      this.#readyNextBlock
+    );
+    miner.mine(nextBlock, maxTransactions, true);
+
+    const { block } = await miner.once("block");
+    block.hash = () => {
```
jeffsmale90: `Quantity.Empty` will give you a static "null" Quantity instance. In retrospect, I'm not a huge fan of the name, and maybe should have called it `Quantity.Null`.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -586,6 +586,55 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+      return Quantity.from(null, true);
+    };
+    block.header.stateRoot = Data.from("0x", 32);
```
jeffsmale90: NIT: I don't like that we are instantiating this every time (especially from a string), but alternatives may be worse.

We use 32 byte length a lot with `Data` types - I wonder if we want to encode that length as a static - `Data.WordByteLength`, and maybe even some instance statics - `Data.NullWord` (although that's not super meaningful by itself) :shrug:

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -22,6 +22,8 @@ const NOTFOUND = 404;
 
 const EMPTY_BUFFER = Buffer.from([]);
 
+type RawOrBlock<T> = T extends true ? Buffer : Block;
```
jeffsmale90: In order to understand what `T` means, I need to look at the conditional type.

Could we rename it to something like `IsRaw` ?

Additionally - we should constraint it's type to boolean.

```suggestion
type RawOrBlock<IsRaw extends boolean> = IsRaw extends true ? Buffer : Block;
```

src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
```diff
@@ -908,87 +907,62 @@ describe("api", () => {
...
         });
 
-        it("does not persist changes to vm or state trie when overrides are set", async () => {
-          // copy the trie, its database, the vm, and the accounts
-          const before = await getBlockchainState();
-
-          // simulate the transaction, also setting overrides
-          const overrides = {
-            [from]: { balance: "0x0", nonce: "0xfff", code: "0x12345678" }
+        it.only("does not persist changes to vm or state trie when overrides are set", async () => {
```
jeffsmale90:
```suggestion
        it("does not persist changes to vm or state trie when overrides are set", async () => {
```

src/chains/ethereum/ethereum/tests/helpers/compare-chain-state.ts
```diff
@@ -0,0 +1,73 @@
...
+    blockchain.common
+  );
+  const addressStates = [];
+  for (const address of addresses) {
+    addressStates.push(await vm.stateManager.getAccount(address));
+  }
+  return { root: trie.root, db: trieDbData, addressStates };
+};
+
+/**
```
jeffsmale90, line 56: Duplicate comment!

src/chains/ethereum/ethereum/tests/helpers/compare-chain-state.ts
```diff
@@ -0,0 +1,73 @@
+import { Address as EthereumJsAddress } from "ethereumjs-util";
+import { isDeepStrictEqual } from "util";
+import Blockchain from "../../src/blockchain";
+import { GanacheTrie } from "../../src/helpers/trie";
+
+/**
+ * Gets all db data from a trie.
```
jeffsmale90: I think there's a nuance here that _is_ captured in your code, but isn't initially apparent:

You can have an identical `trie` with different data in the underlying database (which I assume is why we are extracting all of the kvps here and not only comparing the trie roots).

I wonder if this justifies a comment?

jeffsmale90:
I don't know if there's other cases where this is going to be problematic, but at least for `eth_getProof`, we can pass in `pending` block and throw an internal error: `State trie does not contain state root`. Note Geth does support this.

On https://github.com/trufflesuite/ganache/blob/83d195bac65f17165ab11f5025aed879ee3a5dd3/src/chains/ethereum/ethereum/src/api.ts#L1121 we call `StateManager.setStateRoot` with the state root of the target block. If we pass in `pending`, this will be a 32 byte `00` buffer.

```
curl -s http://localhost:8545 -X POST -H "ContentType: application/json" -d '{"method":"eth_getProof","params":["0x1895a79379d99a344983745bf3737d058bbce3ac",["0x0","0x1"],"pending"],"id":1,"jsonrpc":"2.0"}'

{"id":1,"jsonrpc":"2.0","error":{"message":"State trie does not contain state root","stack":"Error: State trie does not contain state root\n    at DefaultStateManager.setStateRoot (/Users/jeffsmale/g/src/chains/ethereum/ethereum/node_modules/@ethereumjs/vm/src/state/stateManager.ts:442:15)\n    at async EthereumApi.eth_getProof (/Users/jeffsmale/g/src/chains/ethereum/ethereum/src/api.ts:1124:5)","code":-32700}}
```

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: @davidmurdoch  I've made `Executables` a class! I've also made #3539 to move `removeBestFromOrigin` to that class.

src/chains/ethereum/ethereum/src/transaction-pool.ts
MicaiahReid: Done!

src/chains/ethereum/ethereum/tests/helpers/compare-chain-state.ts
MicaiahReid: Fixed!

src/chains/ethereum/ethereum/tests/helpers/compare-chain-state.ts
MicaiahReid: I've added a _little_ more here, but I also added a supporting comment for where this is used. Since the function just _gets_ the data and isn't explicitly used only for comparison purposes.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Switched to `Quantity.Empty`, thanks!

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: I've created an issue for this: https://github.com/trufflesuite/ganache/issues/3542. I'm not adding a TODO here, because it doesn't seem like we're 100% set on doing this, and I'd rather not have the TODO sticking around forever. If someone chooses to work this issue, it should be easy enough to find this occurence.

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
MicaiahReid: Great call, thank you!

src/chains/ethereum/ethereum/tests/api/eth/getBlockByNumber.test.ts
MicaiahReid: I've implemented `memdown` as suggested

src/chains/ethereum/block/src/block.ts
```diff
@@ -116,21 +119,17 @@ export class Block {
     }
   }
 
-  static fromParts(
```
MicaiahReid, line 119: This wasn't used anywhere.

src/chains/ethereum/block/src/block.ts
```diff
@@ -29,12 +30,14 @@ export class Block {
   protected _common: Common;
   protected _rawTransactions: TypedDatabaseTransaction[];
   protected _rawTransactionMetaData: GanacheRawBlockTransactionMetaData[];
+  protected _serialized: Buffer;
```
jeffsmale90: I'm curious what the real world benefit of storing the _serialized value here is. I'm concerned that, because the `block` is mutable, this could result in unexpected results.

src/chains/ethereum/block/src/block.ts
```diff
@@ -116,21 +119,27 @@ export class Block {
...
-    common: Common
-  ): Block {
-    const block = new Block(null, common);
-    block._raw = rawHeader;
-    block._rawTransactions = txs;
-    block.header = makeHeader(rawHeader, totalDifficulty);
-    block._rawTransactionMetaData = extraTxs;
-    block._size = size;
-    return block;
+  getParts() {
```
jeffsmale90, line 120: It'd be nice if we could make `getParts()` protected so as not to leak it into the public interface.

src/chains/ethereum/block/src/pending-block.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Used to override the `stateRoot` value of the return value of the `toJSON`
+   * function.
+   */
+  #toJSONStateRootOverride = Data.from("0x", 32);
+  /**
+   * Used to override the `hash` value of the return value of the `toJSON`
+   * function.
+   */
+  #toJSONHashOverride = Quantity.Empty;
```
jeffsmale90: These could both be static members

src/chains/ethereum/block/src/pending-block.ts
```diff
@@ -0,0 +1,60 @@
...
+   */
+  toJSON<IncludeTransactions extends boolean>(
+    includeFullTransactions: IncludeTransactions
+  ) {
+    const json = super.toJSON(includeFullTransactions);
+    json.stateRoot = this.#toJSONStateRootOverride;
+    json.hash = this.#toJSONHashOverride;
+
+    return json;
+  }
```
jeffsmale90: I really like how you've extended the `Block` class, but I feel that overriding the `toJSON` behaviour is insufficient - can we rather override the properties `stateRoot` and `hash`, so that we can do things like:

```
const block: PendingBlock = ...;
assert.strictEqual(block.hash(), AN_EMPTY_HASH);
assert.strictEqual(block.stateRoot, A_NULL_STATEROOT);
```

src/chains/ethereum/block/src/pending-block.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * @returns The trie at the time the pending block was made.
+   */
+  getTrie(): GanacheTrie {
```
jeffsmale90: Why not just make the trie property `public readonly`, rather than wrap it in a function?

src/chains/ethereum/block/src/pending-block.ts
```diff
@@ -0,0 +1,60 @@
...
+  static fromBlock(block: Block, common: Common, trie: GanacheTrie) {
+    const pendingBlock = new PendingBlock(common, trie);
+    const { rawHeader, txs, header, extraTxs, size } = block.getParts();
+    pendingBlock._raw = rawHeader;
+    pendingBlock._rawTransactions = txs;
+    pendingBlock.header = header;
+    pendingBlock._rawTransactionMetaData = extraTxs;
+    pendingBlock._size = size;
+    return pendingBlock;
+  }
```
jeffsmale90, line 53: > It'd be nice if we could make getParts() protected so as not to leak it into the public interface.

In support of the above, maybe this part of `fromBlock()` could become a private instance method.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1269,7 +1268,9 @@ export default class EthereumApi implements Api {
     const blockNum = await blocks.getNumberFromHash(hash);
     if (!blockNum) return null;
 
-    const rawBlock = await blocks.getRawByBlockNumber(Quantity.from(blockNum));
+    const rawBlock = await blocks.getRawByBlockNumberOrTag(
+      Quantity.from(blockNum)
```
jeffsmale90: Looks like `getRawByBlockNumberOrTag` accepts a buffer, so I _think_ you can do away with the `Quantity.from` call.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1782,12 +1775,11 @@ export default class EthereumApi implements Api {
...
-    const addressData = await trie.get(addressBuf);
-    // An address's stateRoot is stored in the 3rd rlp entry
-    const addressStateRoot = decode<EthereumRawAccount>(addressData)[2];
-    trie.setContext(addressStateRoot, addressBuf, blockNum);
-    const value = await trie.get(paddedPosBuff);
+    const value = await accounts.getStorageAt(
+      Address.from(address),
+      paddedPosBuff,
+      blockNumber
+    );
```
jeffsmale90, line 1780: This is so much cleaner :)

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -145,7 +150,7 @@ function makeTrie(blockchain: Blockchain, db: LevelUp | null, root: Data) {
   if (blockchain.fallback) {
     return new ForkTrie(db, root ? root.toBuffer() : null, blockchain);
   } else {
-    return new GanacheTrie(db, root ? root.toBuffer() : null, blockchain);
+    return new GanacheTrie(db, root ? root.toBuffer() : null);
```
jeffsmale90, line 153: Now that we've dropped node 12 support, can we finally do `root?.toBuffer()`?

#NotYourCode

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
```diff
@@ -9,86 +9,96 @@ import { Quantity, Data } from "@ganache/utils";
...
-    address: Address,
-    blockNumber: Buffer | Tag = Tag.latest
-  ): Promise<Account | null> {
-    const raw = await this.getRaw(address, blockNumber);
-    if (raw == null) return null;
-    return Account.fromBuffer(raw);
+  /**
+   * Creates a copy of the trie that was used to mine the specified
+   * `blockNumber`. Sets the context of the trie to the state root and number
+   * of `blockNumber`.
```
jeffsmale90: I'm not sure of the wording here - "... that was used to mine the ..."  seems ambiguous. Maybe an explanation as to _why_ we have a different trie for Pending blocks would be valuable to have here.

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
```diff
@@ -9,86 +9,96 @@ import { Quantity, Data } from "@ganache/utils";
...
+   * @param blockNumber
+   * @returns
```
jeffsmale90:
```suggestion
   * @returns { trie: GanacheTrie }
```



Additionally, every time we call this, we immediately destructure the return object. Can we just return `GanacheTrie`?

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
```diff
@@ -9,86 +9,96 @@ import { Quantity, Data } from "@ganache/utils";
...
+   */
+  public async getTrieAt(blockNumber: string | Buffer | Tag = Tag.latest) {
+    const { trie, blocks } = this.#blockchain;
+
+    const block = await blocks.get(blockNumber);
+    // a pending block has a copy of the trie at the time the block was made.
+    // for a regular block we can use the current trie
+    const trieCopy =
+      block instanceof PendingBlock ? block.getTrie() : trie.copy(false);
```
jeffsmale90: @davidmurdoch will have a better idea, but I think because PendingBlock is defined outside of this package, the _version_ of the package (and therefore the reference to the class) _could_ be different between where it's instantiated and where the type is checked.

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
```diff
@@ -9,86 +9,96 @@ import { Quantity, Data } from "@ganache/utils";
...
   }
 
-  public async getRaw(
+  public async getRawAccountAndTrie(
     address: Address,
     blockNumber: string | Buffer | Tag = Tag.latest
-  ): Promise<Buffer | null> {
-    const { trie, blocks } = this.#blockchain;
-
+  ): Promise<{ raw: Buffer | null; trie: GanacheTrie }> {
```
jeffsmale90: can we rename `raw` to either `address` or `rawAddress` or similar? `raw` a little ambiguous.

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
```diff
@@ -9,86 +9,96 @@ import { Quantity, Data } from "@ganache/utils";
...
   ): Promise<Quantity> {
-    const data = await this.getRaw(address, blockNumber);
+    const { raw } = await this.getRawAccountAndTrie(address, blockNumber);
 
-    if (data == null) return Quantity.Zero;
+    if (raw == null) return Quantity.Zero;
 
-    const [nonce] = decode<EthereumRawAccount>(data);
+    const [nonce] = decode<EthereumRawAccount>(raw);
     return nonce.length === 0 ? Quantity.Zero : Quantity.from(nonce);
```
jeffsmale90: I think we should drop these conditionals - it's might actually be slower, as we do this check within the Quantity constructor as well.

#NotYourChange

src/chains/ethereum/ethereum/tests/api/eth/contracts/EthCall.sol
```diff
@@ -12,6 +12,10 @@ contract EthCall {
     return block.basefee;
   }
 
+  function getBlockNumber() public view virtual returns (uint256) {
+    return block.number;
+  }
+
   function doARevert() public pure {
     revert("you are a failure");
```
jeffsmale90, line 20: ouch!

src/chains/ethereum/ethereum/tests/api/eth/eth.test.ts
```diff
@@ -138,6 +138,31 @@ describe("api", () => {
...
+          "pending"
+        ]);
+        await provider.send("eth_subscribe", ["newHeads"]);
+        await provider.send("miner_start");
+        await provider.once("message");
+        const balanceAfter = await provider.send("eth_getBalance", [
+          accounts[0],
+          "latest"
+        ]);
+        assert.strictEqual(pendingBalance, balanceAfter);
```
jeffsmale90, line 164: I feel a constant assertion would be better here, as we _could_ mess something up with creating the pending block, and result in an incorrect `balanceAfter`

src/chains/ethereum/ethereum/tests/api/eth/eth.test.ts
```diff
@@ -217,6 +242,24 @@ describe("api", () => {
...
+      it("should get the transaction count of the pending block", async () => {
+        const tx = {
+          from: accounts[0],
+          to: accounts[1],
+          value: "0x1"
+        };
+
+        // send one tx and don't mine, then check the count
+        await provider.send("miner_stop");
+        await provider.send("eth_sendTransaction", [{ ...tx }]);
```
jeffsmale90: do we need to decompose `tx` here?

src/chains/ethereum/ethereum/tests/api/eth/eth.test.ts
```diff
@@ -372,6 +415,40 @@ describe("api", () => {
       );
       assert.strictEqual(count, "0x1");
     });
+    it("eth_getBlockTransactionCountByNumber for pending tag", async () => {
```
jeffsmale90, line 419:
```suggestion

    it("eth_getBlockTransactionCountByNumber for pending tag", async () => {
```

src/chains/ethereum/transaction/src/runtime-transaction.ts
```diff
@@ -258,4 +268,10 @@ export abstract class RuntimeTransaction extends BaseTransaction {
 
   protected abstract toVmTransaction();
   protected abstract updateEffectiveGasPrice(baseFeePerGas?: Quantity);
+  /**
+   * Creates a copy of the transaction and all of the fields that are filled
+   * _before_ the transaction is mined onto the block. (Note: block and receipt
+   * data is not added to the copy.)
+   */
+  protected abstract copy();
```
jeffsmale90: very low importance, but I prefer the name `clone()` here, as we use the term `copy` for transferring data between instances.

src/chains/ethereum/ethereum/src/miner/executables.ts
```diff
@@ -1,7 +1,42 @@
 import { TypedTransaction } from "@ganache/ethereum-transaction";
 import { Heap } from "@ganache/utils";
+import { byNonce } from "../transaction-pool";
 
-export type Executables = {
-  inProgress: Set<TypedTransaction>;
-  pending: Map<string, Heap<TypedTransaction>>;
-};
+export class Executables {
```
jeffsmale90: Just a though - feel free to dismiss: Would this make sense to be named `ExecutableTransactionContainer` or something more explicit?

src/chains/ethereum/ethereum/src/miner/executables.ts
```diff
@@ -1,7 +1,42 @@
...
+  public inProgress: Set<TypedTransaction> = new Set();
+  public pending: Map<string, Heap<TypedTransaction>> = new Map();
```
jeffsmale90: I assume that the key on the map is the `from` address. Would it be worthwhile naming the property `pendingBySender` or in some other way indicating this?

src/chains/ethereum/ethereum/src/miner/executables.ts
```diff
@@ -1,7 +1,42 @@
...
+
+  /**
+   * Deep clones the executables, moving all `inProgress` transactions to the
```
jeffsmale90: I initially didn't think that this clones the transactions themselves, but after looking at the code I see that it does.

Maybe worth making this more clear in the comment - e.g., `... cloning all transactions into "pending" queue ..."

src/chains/ethereum/ethereum/src/miner/executables.ts
```diff
@@ -1,7 +1,42 @@
...
+    const executables = new Executables();
+    const { inProgress, pending } = this;
+
+    inProgress.forEach(transaction => {
+      const copy = transaction.copy();
+      copy.locked = false;
+      const origin = copy.from.toString();
+      const txsFromOrigin = executables.pending.get(origin);
+      if (txsFromOrigin) txsFromOrigin.push(copy);
+      else executables.pending.set(origin, Heap.from(copy, byNonce));
```
jeffsmale90: Erg - I'm not a huge fan of this if/else styling from prettier

src/chains/ethereum/ethereum/src/miner/executables.ts
```diff
@@ -1,7 +1,42 @@
...
+    });
+
+    pending.forEach(({ array, length }, from) => {
```
jeffsmale90: array here is transactions right? can we have a more meaningful name?

src/chains/ethereum/ethereum/tests/executables.test.ts
```diff
@@ -0,0 +1,138 @@
...
+      assert.strictEqual(
+        clonedTransactions.length,
+        originalTransactions.length
+      );
+      for (const originalTransaction of originalTransactions) {
+        const clonedTransaction = clonedTransactions.find(transaction => {
+          return transaction.hash === originalTransaction.hash;
+        });
+        // there is a transaction with the same hash
+        assert(clonedTransaction);
```
jeffsmale90:
```suggestion
        assert(clonedTransaction !== undefined);
```
I'd rather this were explicit, purely for the sake of readability.

jeffsmale90: Can we also assert that, although the hashes match, it is, infact, a clone and not the same instance?

jeffsmale90 (changes requested):

src/chains/ethereum/block/src/block.ts
MicaiahReid: It doesn't look like this is possible, at least with the `protected` keyword. This gives the error `Property 'getParts' is protected and only accessible through an instance of class 'PendingBlock'. This is an instance of class 'Block'`.

src/chains/ethereum/block/src/block.ts
MicaiahReid: We serialize the block when we mine it, so this prevents us from having to do it again any time a block's raw data is needed. We need this raw data when fetching a raw block in the block manager.

All of the fields used in serialization are private and aren't currently modified, but I suppose it's true they could be. This seems worthy of adding to a discussion to weigh the risk/reward.

src/chains/ethereum/block/src/pending-block.ts
MicaiahReid: I like it, will change!

src/chains/ethereum/block/src/pending-block.ts
MicaiahReid: We are overriding the toJSON function because we still _need_ the actual `stateRoot` and `hash` values of the pending block for use cases that traverse the trie. For example, if we were to override the `stateRoot`, this function in your `eth_getProof` RPC method wouldn't work: 
```
 await stateManagerCopy.setStateRoot(
      targetBlock.header.stateRoot.toBuffer()
    );
```

However, geth returns empty data for the `toJSON` representation of the pending block, hence this "half-measure" override of the data.

src/chains/ethereum/block/src/pending-block.ts
MicaiahReid: Done!

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: <3

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: Good eye, fixed!

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: I've created an issue for this! https://github.com/trufflesuite/ganache/issues/3791

#DontWannaDoIt :)

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
MicaiahReid: Fixed!

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
MicaiahReid: Done!

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
MicaiahReid: Ahh good find! I think originally this returned multiple values so I threw it in an object. Fixed!

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
MicaiahReid: Fixed, let me know if that makes more sense.

src/chains/ethereum/ethereum/src/miner/executables.ts
MicaiahReid: Fixed

src/chains/ethereum/ethereum/src/miner/executables.ts
MicaiahReid: Yeah I don't love it either, and prettier doesn't force it. Fixed.

src/chains/ethereum/ethereum/src/miner/executables.ts
MicaiahReid: Clarified!

src/chains/ethereum/ethereum/src/miner/executables.ts
MicaiahReid: I like it, renamed.

src/chains/ethereum/ethereum/src/miner/executables.ts
MicaiahReid: Hmm, yeah, now that this is a class maybe `Executables` isn't the best name. I'm cool with that name, fixing!

src/chains/ethereum/ethereum/tests/api/eth/contracts/EthCall.sol
MicaiahReid: :laughing:

src/chains/ethereum/ethereum/tests/api/eth/eth.test.ts
MicaiahReid: Nope! Fixed.

src/chains/ethereum/transaction/src/runtime-transaction.ts
MicaiahReid: I'm convinced, changed.

src/chains/ethereum/ethereum/tests/api/eth/eth.test.ts
MicaiahReid: I don't know if I agree. In the future gas price mechanics could change and the balance after would change. This asserts that the pending balance is _actually_ what it will be after the latest block. Open to discussion.

src/chains/ethereum/ethereum/tests/executables.test.ts
MicaiahReid: We can't assert (without a good bit more code) that they are exact clones (because they aren't, some of the data doesn't get cloned). But I did make an assertion that they aren't the same instance.

src/chains/ethereum/block/src/pending-block.ts
MicaiahReid: As per our discussion, this is a little tricker than originally expected, and probably isn't worth it.

src/chains/ethereum/block/src/block.ts
MicaiahReid: As per our discussion, this is a little tricker than originally expected, and probably isn't worth it.

jeffsmale90 (approved):

PR #3274, opened 2022-06-23
chore: remove unused build file

davidmurdoch:
Just removing some totally unused code. Nothing to see here!


MicaiahReid (approved):
We should do this more often!

tenthirtyone (approved):

PR #3275, opened 2022-06-23
fix: reduce bundle size and check size in CI

davidmurdoch:
[jsDelivr](https://www.jsdelivr.com/) has an unpacked bundle size limit of 100MB, which we recently exceeded, breaking our CDN recommendation in our README.md.

This change reduces the bundle size to under 100MB and also adds a check to our pr.yml CI Action that will fail if the new bundle size is greater than 100 MB.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -59,6 +59,7 @@ jobs:
           cd ./src/packages/ganache &&
           npm pack &&
           size="$(zcat ganache-*.tgz | wc -c)" &&
-          test $size -gt 100000000 && exit $size || exit 0
+          echo "Bundle size: $size" &&
+          test $size -lt 100000000
         env:
           INFURA_KEY: ${{ secrets.INFURA_KEY }}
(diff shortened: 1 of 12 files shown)


.github/workflows/pr.yml
```diff
@@ -44,3 +44,22 @@ jobs:
...
+        # 4. test to make sure the tarball is less than 99MB because jsDelivr
+        #    CDN doesn't allow bundles greater than 100MB.
+        # 5. exit with non-zero exit code if the test fails.
+        run: |
+          npm run build &&
+          cd ./src/packages/ganache &&
+          npm pack &&
+          size="$(zcat ganache-*.tgz | wc -c)" &&
+          echo "Bundle size: $size" &&
+          test "$size" -lt 99000000
```
cds-amal: A little extra, perhaps?

```suggestion
          echo "Bundle size: $size" &&
          echo "Bundle size is $([[ "$size" -lt 99000000 ]] && echo "ok" || echo "not ok")" &&
          test "$size" -lt 99000000
```

cds-amal (commented):
Is it still possible for the bundle to be larger than the threshold, and what happens, Ganache process wise, if too large?

MicaiahReid (approved):

cds-amal (approved):
Looks good to me, bash wise :)

PR #3282, opened 2022-06-24
ci: use test infura key in bundle  size check

davidmurdoch:


davidmurdoch:
this should actually be a dummy key in the bundle size check, since it doesn't need to be valid, and we want this size check to work even with external contributor PRs that don't have access to environment secrets.

Issue #3240, opened 2022-06-15
Standardise private members


jeffsmale90:
Throughout the codebase we use javascript private class fields [[see docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields)], and typescript `private` modifiers to signify private members.

```
private _myPrivateThing;
vs
#myActualPrivateThing;
```

The Javascript private class fields have a performance cost - in part due to how they are implemented in v8 (although this may not be the case in more modern versions), and in part because how we webpack. We work around this in some cases by pointing a local variable to the private field.

Ideally we would standardise on a single case which would perform well, without need for workarounds, and would provide reasonable encapsulation.

As an example: Quantity, Data, Account etc all use private constructor parameters.

This introduces some issues with typings, as this changes the type definitions such that JSON-RPC types are no longer compatible.

PR #3317, opened 2022-06-30
fix: if a timestamp is passed to `evm_mine`, subsequent blocks reflect the change in time

### LINKED ISSUE #3265: calling `evm_mine` with a `timestamp` does not persist time offset
### Discussed in https://github.com/trufflesuite/ganache/discussions/3264

I'm opening as  an issue as suggested by @davidmurdoch

<sup>Originally posted by **msuscens** June 21, 2022</sup>
Hello, 

I have Truffle tests that successfully pass with Ganache v6 but fail with Ganache v7.3.1.  The truffle tests in question use the ganache-time-traveler utility to advance time.  Specifically, the issue is with functions that need a certain amount of time to pass before they will execute (without reverting).  I, therefore, use ganache-time-traveler to advance the time before executing the function.  Unfortunately, with ganache 7.3.1 these functions revert even though time has been advanced by the required amount.   Here is an example:

```
describe("After Incubation: Hatch single egg", () => {

        let snapshotId
        beforeEach("Advance Time to End of Incubation", async() => {
            let snapshot = await timeMachine.takeSnapshot()
            snapshotId = snapshot['result']

            // Incubate for required time
            await truffleAssert.passes(
                timeMachine.advanceTimeAndBlock(INCUBATION_DURATION),
                "Failed to advance time and block"
            )
        })

        afterEach("Revert To Start of Incubation Time", async() => {
            await timeMachine.revertToSnapshot(snapshotId)
        })

        it("should allow egg owner to hatch it, emitting an Hatched event ", async () => {
            let txResult
            await truffleAssert.passes(
                txResult = await eggToken.hatch(
                    [0], //eggids
                    {from: accounts[2]}
                )
            )
            truffleAssert.eventEmitted(txResult, 'Hatched', (ev) => {
                return equals(toNumbers(ev.hatchlingIds), [0]) &&
                    equals(toNumbers(ev.eggIds),[0]) &&
                    ev.owner == accounts[2]
            }, "Event Hatched has incorrect parameter values!")
        })
```
etc ...

This test passes with Ganche v6 but `eggToken.hatch` reverts under v7.3.1 with the following message:

```
After Incubation: Hatch single egg
         should allow egg owner to hatch it, emitting an Hatched event :
     Transaction: 0x20883a2a3329b9925a86499b4fe1e9727a029f197d00f084ed920e94eb1954d6 exited with an error (status 0). Reason given: hatch: Egg not incubated!.
     Please check that the transaction:
     - satisfies all conditions set by Solidity `require` statements.
     - does not trigger a Solidity `revert` statement.
 
  StatusError: Transaction: 0x20883a2a3329b9925a86499b4fe1e9727a029f197d00f084ed920e94eb1954d6 exited with an error (status 0). Reason given: hatch: Egg not incubated!.
      at Context.<anonymous> (test/2.eggToken/6.eggToken_hatch_gen0_test.js:217:43)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

I'd appreciate any help/advice on how to resolve this issue.

I'm on a 2022 MacBook Pro (M1 Pro processor): MacOS 12.4
Truffle v5.5.3 (core: 5.5.3)
Ganache v7.0.1
Solidity - 0.8.13 (solc-js)
Node v14.15.5
Web3.js v1.5.3

Note: Even though it says "Ganache v7.0.1" above (from 'truffle version'), I've installed v7.3.1 and when I start ganache (via CL) it says "ganache v7.3.1 (@ganache/cli: 0.4.1, @ganache/core: 0.4.1)"
</div>

### LINKED ISSUE #3271: "should return the `timeAdjustment` value via `evm_increaseTime` when provided as a number" test is flaky
Presumably due to the combination of Math.floor and using the system clock for our tests.

### LINKED ISSUE #3330: `evm_setTime` doesn't work correctly with `miner.timestampIncrement`
When using `miner.timestampIncrement`, calling `evm_setTime` can result in unexpected values, and potentially with an error being thrown.

This is because the `timeAdjustment` is applied to the previous blocktime, so this offset is applied for every block that is mined.

With the following example, the initial block is mined correctly, as is the first block after setting the time, but subsequent blocks will have a descending block time, until the value becomes negative.

```javascript
const p = await getProvider({
  miner: {timestampIncrement: 1},
  chain: {time : 10000}
});

const mineAndShowBlock = async () => {
  await p.send("evm_mine");
  const {number, timestamp} = await p.send("eth_getBlockByNumber", ["latest"]);

  console.log({number, timestamp: Quantity.toNumber(timestamp)});  
}
await mineAndShowBlock();

await p.send("evm_setTime", [5000]);

while(true) {
  await mineAndShowBlock();
}
```

```
{ number: '0x1', timestamp: 11 }
{ number: '0x2', timestamp: 6 }
{ number: '0x3', timestamp: 1 }

Error: Cannot wrap a negative value as a json-rpc type
```

jeffsmale90:
This fixes a [regression introduced in Ganche v7](https://github.com/trufflesuite/ganache/issues/3265), where providing a timestamp in a call to `evm_mine` would result in a block with the correct timestamp, but the internal time-offset would not be updated, so subsequent blocks would not reflect the change in time.

Fixes: #3265 #3271


src/chains/ethereum/ethereum/tests/api/evm/evm.test.ts
```diff
@@ -132,6 +132,51 @@ describe("api", () => {
...
+              );
+              const subsequentBlockTimestamp =
+                Quantity.toNumber(subsequentBlockTime);
+
+              // add 2 seconds in case testing is slow
+              assert(
+                between(
+                  subsequentBlockTimestamp,
+                  miningTimestamp,
+                  miningTimestamp + 2
```
MicaiahReid, line 186: I think, if I'm following the test correctly, we could use the `--miner.timestampIncrement` to know exactly what the `subsequentBlockTimestamp` should be without having any room for slow testing having an impact

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -297,33 +297,34 @@ export default class EthereumApi implements Api {
...
+      numBlocks = arg.blocks;
+      timestamp = arg.timestamp;
     } else {
+      timestamp = arg as number | null;
+    }
+
+    if (numBlocks == null) {
+      numBlocks = 1;
+    }
+    if (timestamp) {
```
MicaiahReid:
```suggestion
    if (timestamp !== null) {
```
I know this is a silly edge case, but what if they set the timestamp to 0? This would get skipped.

MicaiahReid (changes requested):
Just two small changes!

BTW, I've installed this version and tested against the repro I had set up from #3265, and it fixes the issue as expected!

src/chains/ethereum/ethereum/tests/api/evm/evm.test.ts
jeffsmale90: Good call! Unfortunately (fortunately??) while testing this, I found an existing issue which I'm going to include in this PR: #3330

MicaiahReid: Good find!

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,97 @@
+export interface BlockTime {
```
jeffsmale90: I've extracted the logic for block timestamp into BlockTime, which allows us to more clearly describe the behaviour, and test more thoroughly.

By settings this as a dependency, it will allow us to decouple our tests from the system clock (without deferring to `timestampIncrement`), for the likes of #3271 (which I'll pick up after this).

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: I fixed it :)

jeffsmale90:
I've standardised all timestamp units to milliseconds, with the exceptions:

- A block timestamp is obviously seconds.
- Some API parameters are in seconds.

I ensure that all interactions below `api.ts` (with the exception of a block's timestamp) are ms.

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: And added an explicit test case

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,97 @@
...
+  #getReferenceClockTime: () => number | Date;
+  #timeOffset: number;
+
+  private static validateTimestamp(timestamp: number | Date) {
+    if (timestamp < 0) {
+      throw new Error(`Invalid timestamp: ${timestamp}. Value must be positive.`);
+    }
+  }
+
+  constructor(startTime: number | Date, getReferenceClockTime: () => number | Date) {
```
davidmurdoch: Use `number` here (and elsewhere) instead of `number | Date`. The `Date` object isn't used internally and we are casting everywhere we used the referenced clock time. `Date.now()` is about a bajillion or so times faster than `new Date()`

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -297,33 +297,36 @@ export default class EthereumApi implements Api {
...
+      this.#blockchain.setTime(timestamp * 1000);
+    }
+
+    // TODO(perf): add an option to mine a bunch of blocks in a batch so
+    // we can save them all to the database in one go.
+    // Developers like to move the blockchain forward by thousands of blocks
+    // at a time and doing this would make it way faster
+    for (let i = 0; i < numBlocks; i++) {
+      const timestampMilliseconds =
+        timestamp == undefined ? undefined : timestamp * 1000;
```
davidmurdoch: We shouldn't compute this in the loop; add this logic to the `timestamp != undefined` condition above.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -297,33 +297,36 @@ export default class EthereumApi implements Api {
     const blockchain = this.#blockchain;
     const options = this.#options;
     const vmErrorsOnRPCResponse = options.chain.vmErrorsOnRPCResponse;
+
+    let numBlocks, timestamp;
```
davidmurdoch: These need types.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -297,33 +297,36 @@ export default class EthereumApi implements Api {
...
-      }
+      numBlocks = arg.blocks;
+      timestamp = arg.timestamp;
     } else {
+      timestamp = arg as number | null;
+    }
+
+    if (numBlocks == undefined) {
+      numBlocks = 1;
+    }
```
davidmurdoch: Any objection to checking this condition in the `if` above?

```
numBlocks = arg.blocks == null ? 1 : arg.blocks;
```

and then in the `else` assign `numBlocks` to `1`.

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,97 @@
...
+  createBlockTimestampInSeconds(timestamp?: number | Date): number;
+}
+
+export class ClockBasedBlockTime implements BlockTime {
+  #getReferenceClockTime: () => number | Date;
+  #timeOffset: number;
+
+  private static validateTimestamp(timestamp: number | Date) {
+    if (timestamp < 0) {
+      throw new Error(`Invalid timestamp: ${timestamp}. Value must be positive.`);
```
davidmurdoch: Why must the timestamp be positive? This breaks `ganache --time`, and `evm_setTime`.

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  getOffset(): number {
+    return this.#timeOffset;                              
+  }
+
+  setOffset(offset: number) {
+    const referenceTimestamp = this.#getReferenceClockTime();
+    if (offset < -referenceTimestamp) {
+      throw new Error(`Invalid offset: ${offset}, value must be greater than the negative of the current reference clock timestamp: ${referenceTimestamp}`);
+    }
```
davidmurdoch: Why do we need to guard against whatever this is guarding against?

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const seconds = Math.floor(milliseconds / 1000);
+    return seconds;
+  }
+}
+
+/*
+  A BlockTime implementation that increments it's reference time by the duration specified by incrementMilliseconds,
+  every time createBlockTimestampInSeconds() is called. The timestamp returned will be impacted by the timestamp offset,
+  so can be moved forward and backwards by calling .putOffset() independent of the start time, and increments caused
```
davidmurdoch: There is no `putOffset`, should this be `setOffset`?

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -162,9 +162,26 @@ export class EthereumProvider
       providerOptions.fork.url ||
       providerOptions.fork.provider ||
       providerOptions.fork.network;
-    const fallback = fork ? new Fork(providerOptions, accounts) : null;
+
+    const startTime = this.#options.chain.time || new Date();
```
davidmurdoch:
```suggestion
    const startTime = providerOptions.chain.time != null ? +providerOptions.chain.time : Date.now();
```

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -162,9 +162,26 @@ export class EthereumProvider
...
+    const blockTime =
+      this.#options.miner.timestampIncrement === "clock"
```
davidmurdoch:
```suggestion
      providerOptions.miner.timestampIncrement === "clock"
```

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -162,9 +162,26 @@ export class EthereumProvider
...
+        ? new ClockBasedBlockTime(startTime, () => new Date())
```
davidmurdoch:
```suggestion
        ? new ClockBasedBlockTime(startTime, () => Date.now())
```

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -162,9 +162,26 @@ export class EthereumProvider
...
+        : new IncrementBasedBlockTime(
+            startTime,
+            this.#options.miner.timestampIncrement.toNumber() * 1000
```
davidmurdoch:
```suggestion
            providerOptions.miner.timestampIncrement.toNumber() * 1000
```

src/chains/ethereum/ethereum/src/forking/fork.ts
```diff
@@ -227,11 +234,13 @@ export class Fork {
...
-      chainOptions.time = new Date(
-        (this.block.header.timestamp.toNumber() +
-          minerOptions.timestampIncrement.toNumber()) *
-          1000
+    if (!chainOptions.time) {
+      // this needs to set the time of the _previous_ block,
+      // whereas in incrementBasedBlockTime we set the time of the _next_ block
+      // maybe when we call setTime, we increment the value passed in.
+      const blockTimestampMs = this.block.header.timestamp.toNumber() * 1000;
+      this.block.header.timestamp = Quantity.from(
```
davidmurdoch, line 245: `chainOptions.time` is no longer set and should be. The CLI startup output uses this updated `time` value when it outputs its startup details.

You can check this by running `ganache --fork mainnet --fork.blockNumber 15073145 --miner.timestampIncrement 1`

ganache 7.3.2:
```
Forked Chain
==================
Location:        Ethereum Mainnet, via 丕Infura
Block:           15073145
Network ID:      1
Time:            Sun Jul 03 2022 21:41:07 GMT-0400 (Eastern Daylight Time)
```

this PR:
```console
Forked Chain
==================
Location:        Ethereum Mainnet, via 丕Infura
Block:           15073145
Network ID:      1
Time:            Tue Jul 19 2022 12:41:59 GMT-0400 (Eastern Daylight Time)
```

We really should add tests for the CLI output one day :sweat_smile:

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -297,33 +297,36 @@ export default class EthereumApi implements Api {
...
+    if (timestamp != undefined) {
+      this.#blockchain.setTime(timestamp * 1000);
+    }
+
+    // TODO(perf): add an option to mine a bunch of blocks in a batch so
+    // we can save them all to the database in one go.
+    // Developers like to move the blockchain forward by thousands of blocks
+    // at a time and doing this would make it way faster
+    for (let i = 0; i < numBlocks; i++) {
```
davidmurdoch, line 319: Some backstory:

Before we accepted a `blocks` param many devs would just call `evm_mine` in a loop... for like tens to hundreds of thousands of blocks.

I made an effort to avoid negatively affecting the performance of those existing usages by separating out the `blocks` param's loop logic path from the original logic.

I think it's fine to do this now... as any performance difference is probably negligible anyway. Just wanted to mention it (feel free to resolve this comment)

src/chains/ethereum/ethereum/src/block-time.ts
jeffsmale90: I opened PR #3389 to address #3271 but have also fixed the issue in this PR, and was able to test the issue in this PR (at least on the underlying provider, not so easy at the higher level integration test).

I figured I would leave both PRs open, so that we can get a quick fix in with #3389, but am happy to close that PR if this one goes through quickly.

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,100 @@
...
+  #getReferenceClockTime: () => number | Date;
+  #timeOffset: number = 0;
+
+  private static validateTimestamp(timestamp: number | Date) {
+    if (timestamp < 0) {
+      throw new Error(`Invalid timestamp: ${timestamp}. Value must be positive.`);
+    }
+  }
+
+  constructor(getReferenceClockTime: () => number | Date, startTime: number | Date | undefined) {
```
jeffsmale90: I made the constructor explicitly accept a `startTime` of value `undefined`, and switched the argument order to make it more intuitive. I didn't make the argument optional, because the only time that we don't pass it (except for providing a variable of value `undefined`) is in tests.

This allows us to instantiate a `ClockBasedBlockTime` with no offset, without jumping through hoops.

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Create a timestamp for a new block in seconds. Can specify the timestamp for the block, in milliseconds, which will be reflected in future calls to this function.
+   * @param  {number} timestamp? - Optional - The timestamp for the block that will be mined, in milliseconds.
+   * @returns number - the timestamp in seconds
+   */
+  createBlockTimestampInSeconds(timestamp?: number | Date): number;
+}
+
+export class ClockBasedBlockTime implements BlockTime {
+  private _getReferenceClockTime: () => number | Date;
```
jeffsmale90: I changed all of the # prefixed private class fields to _ prefixed typescript privates. The former have performance implications, and I don't see the benefit until we can resolve this.

#3240 addresses this more generally.

davidmurdoch (changes requested):
quick partial review

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: 👍 thanks for the explanation!

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Yeah definitely!

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Not at all - I think that makes sense. It splits out the two use cases (args vs timestamp) nicely.

src/chains/ethereum/ethereum/src/block-time.ts
jeffsmale90: Yep - it _was_ `putOffset` originally.

src/chains/ethereum/ethereum/src/forking/fork.ts
jeffsmale90: Ah. Good to know.

I've set that value, and validated (note the timezone difference!):
```
Forked Chain
==================
Location:        Ethereum Mainnet, via 丕Infura
Block:           15073145
Network ID:      1
Time:            Mon Jul 04 2022 13:41:06 GMT+1200 (New Zealand Standard Time)

Chain Id
==================
1337
```

davidmurdoch: Except the timestamp is now incorrect if ganache is using "clock" mode (since it doesn't use the clock for the first block after the fork block).

Note: I don't particularly like that this is ganache's behavior in "clock" mode.

src/chains/ethereum/block/tests/index.test.ts
```diff
@@ -52,7 +53,11 @@ describe("@ganache/ethereum-block", async () => {
         },
         "london"
       );
-      blockchain = new Blockchain(options, fromAddress);
+      blockchain = new Blockchain(
+        options,
+        fromAddress,
+        new ClockBasedBlockTime(Date.now, undefined)
```
davidmurdoch: No need to pass `undefined` here.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -297,33 +297,34 @@ export default class EthereumApi implements Api {
...
     } else {
+      numBlocks = 1;
+      timestamp = arg as number | null;
+    }
+
+    let timestampMilliseconds: number;
+    if (timestamp != undefined) {
+      timestampMilliseconds = timestamp * 1000;
+      this.#blockchain.setTime(timestampMilliseconds);
+    }
```
davidmurdoch: I don't think we need to `setTime` here because `blockchain.mine` already does this if a timestamp is passed in. I think this can be simplified further into something like:


```typescript
    let numBlocks: number, timestampMilliseconds: number;
    if (arg == null) {
      numBlocks = 1;
    } else if (typeof arg === "object") {
      numBlocks = typeof arg.blocks === "number" ? arg.blocks : 1;
      if (typeof arg.timestamp === "number") {
        timestampMilliseconds = arg.timestamp * 1000;
      }
    } else {
      numBlocks = 1;
      timestampMilliseconds = arg * 1000;
    }
```

That said, we are converting our time from seconds to milliseconds here, just to convert it back to seconds in `readyNextBlock`. Maybe this is fine and is the best we can do because the precision of our options in various places is all over the place (some in seconds, some milliseconds).

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -107,8 +108,7 @@ describe("provider", () => {
         method: "eth_getBlockByNumber",
         params: ["latest", false]
       });
-      const expectedTime =
-        Math.floor((fastForward + +time) / 1000) + timestampIncrement;
+      const expectedTime = Math.floor((fastForward + +time) / 1000) + timestampIncrement;
```
davidmurdoch: Looks like this may need to be prettieried.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -214,6 +216,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
   constructor(
     options: EthereumInternalOptions,
     coinbase: Address,
+    blockTime: BlockTime,
```
davidmurdoch, line 219: Should we give this a default? Something like `blockTime: BlockTime = new ClockBasedBlockTime(Date.now),`. I think it could simplify a bunch of tests.

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,100 @@
...
+      milliseconds = this._getReferenceClockTime() + this._timeOffset;
+    }
+
+    const seconds = Math.floor(milliseconds / 1000);
+    return seconds;
+  }
+}
+
+/*
+  A BlockTime implementation that increments it's reference time by the duration specified by incrementMilliseconds,
```
davidmurdoch:
```suggestion
  A BlockTime implementation that increments its reference time by the duration specified by incrementMilliseconds,
```

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,100 @@
...
+  every time createBlockTimestampInSeconds() is called. The timestamp returned will be impacted by the timestamp offset,
+  so can be moved forward and backwards by calling .setOffset() independent of the start time, and will increment when
```
davidmurdoch: Can you check this comment for accuracy/consistency? I'm having a hard time following it.

src/chains/ethereum/ethereum/src/block-time.ts
```diff
@@ -0,0 +1,100 @@
...
+  createBlockTimestampInSeconds() is called.
+*/
+export class IncrementBasedBlockTime extends ClockBasedBlockTime {
+  private readonly _tickReferenceClock: () => void;
+
+  constructor(startTime: number, incrementMilliseconds: number) {
+    let referenceTime = startTime;
```
davidmurdoch: I don't think this `referenceTime` is reset on `evm_revert`, but I think it should be.

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/forking/fork.ts
jeffsmale90: I've fixed and added some explicit tests around this behaviour to `forking.test.ts`

src/chains/ethereum/ethereum/src/block-time.ts
jeffsmale90: Yeah, it's a bit mucky, I've made it more concise and easier to follow. 🤞

src/chains/ethereum/ethereum/src/block-time.ts
jeffsmale90: As discussed separately, although a negative timestamp is valid, the block `timestamp` property must be a positive integer value (it is of type `Quantity`). Presently Ganache will accept a negative value (programatically, but not via cli), but will result in undesired block time. 

I've added a comment explaining why we are validating. I don't think this should be considered a breaking change, as it's aligning Ganache with the expected behaviour, _but_ it could cause problems for our users, so I'm happy to push that change out to milestone 8 if desired.

src/chains/ethereum/ethereum/src/block-time.ts
jeffsmale90: Same as above, but I've changed it to leverage the `validateTimestamp()` function.

src/chains/ethereum/ethereum/src/block-time.ts
jeffsmale90: That's a really good point. I rethought the approach and went back and forward a couple of times. I've settled on a slight modification, which I think makes sense. The IncrementBasedBlockTime now holds the reference clock at `startTime` and increments it's _offset_ rather than reference clock.

Offset wasn't working correctly with increment based clocks prior to this change, (as `blocktime = previousBlocktime + increment + offset` will result in compounding the offset). This change fixes that.

It does mean that the meaning of "offset" wrt increment based clocks is now strictly the elapsed duration (in block time) since the "start time" (plus / minus any modifications made by calls to `evm_setTime` / `evm_increaseTime`).

Happy to hear opinions about the above, or we can set up a call to beat out the details.

src/chains/ethereum/block/tests/index.test.ts
jeffsmale90: That's intentional - https://github.com/trufflesuite/ganache/pull/3317#discussion_r925294713

I wanted to make it explicit. In real usage in `provider.ts` we explicitly pass a variable with value `undefined`. I wanted the tests to explicitly match this, so I didn't make the argument optional.

I'm open to changing that if you feel strongly about it.

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: Yeah, you are correct. I've simplified as you suggested.

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: I prefer to not specify a default in cases such as this, as it encourages more implicit code which is prone to miss-understandings. I also feel that the implementation should be optimised for the "real" use case (while obviously being testable).

I've added a `fromSystemClock()` static constructor. I intentionally didn't just give `getReferenceClockTime` argument a default, because I wanted to make it explicit that it's using the system clock.

I think this gives a nice balance, and also tidies up the real world consumption - where we can just call `fromSystemClock(startTime)` and have the `Date.now` part abstracted away.

src/chains/ethereum/block/tests/index.test.ts
jeffsmale90: As per https://github.com/trufflesuite/ganache/pull/3317/#discussion_r934043870

I added `fromSystemClock()` static constructor method which means the only time we call this constructor directly is in tests, so I've done as you suggested.

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -162,9 +162,30 @@ export class EthereumProvider
...
+        ? Date.now()
+        : +providerOptions.chain.time;
+    // if startTime is undefined, BlockTime will determine 0 offset from the reference time
+    const blockTime =
+      providerOptions.miner.timestampIncrement === "clock"
+        ? BlockTime.fromSystemClock(startTime)
+        : new IncrementBasedBlockTime(
+            startTime,
+            providerOptions.miner.timestampIncrement.toNumber() * 1000
+          );
```
davidmurdoch: A `BlockTime` static that abstracts this logic away would really help make the BlockTime logic more cohesive. I'm not a fan of adding all this additional time logic to the provider constructor.

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: Hm, there _is_ a default, you just put it in a different place. :-)

I'm fine with it... but I can foresee someone creating a test helper that applies actual defaults in the future (like the `getProvider()` helper does already).

davidmurdoch (commented):
since you told me to hold off on the rest of this review I thought I'd just send off the two comments I have so far

jeffsmale90:
This PR is now redundant

PR #3327, opened 2022-07-02
feat: add the ability to use `console.log` from Solidity

davidmurdoch:
Ganache now supports `console.log` in Solidity!

More details and how-tos will be written up in Truffle's upcoming `console.log` release. So stay tuned for that!


davidmurdoch:
~I still need to figure out if I need to warm the console.sol precompile address (à la https://github.com/ethereumjs/ethereumjs-monorepo/blob/cfd7b7754490b072a035cceaba59c3dfb517effd/packages/vm/src/runTx.ts#L176) for post-berlin blocks and if we really need to add this account to the state tree on start up (this is what is causing the new test case failures, as "activating" it causes the initial state to change)~ We do not; adding precompile accounts on startup is a gas optimization for the CALL opcode, `console.sol` uses the `STATICCALL` opcode so the gas optimization doesn't apply.

src/chains/ethereum/ethereum/src/helpers/precompiles.ts
```diff
@@ -9,76 +9,11 @@ const NUM_PRECOMPILES = 18;
  * An account with a balance of 1
  */
 const SERIALIZED_PRECOMPILE = Uint8Array.from([
```
davidmurdoch, line 11: prettier did this

src/chains/ethereum/ethereum/src/helpers/precompiles.ts
```diff
@@ -88,7 +23,7 @@ const accountCache: Address[] = [];
 const makeAccount = (i: number): Address => {
   if (accountCache[i]) return accountCache[i];
 
-  // 10 bytes, the first 19 are 0, the last byte is the address
+  // 20 bytes, the first 19 are 0, the last byte is the address
```
davidmurdoch, line 26: not worth a separate PR :-)

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -452,7 +451,10 @@ export default class Miner extends Emittery<{
     this.emit("ganache:vm:tx:before", { context });
     // we always listen to the step event even if `#emitStepEvent` is false in
     // case the user starts listening in the middle of the transaction.
-    const stepListener = event => {
+    const stepListener = (event: InterpreterStep) => {
+      const logs = maybeGetLogs(event);
+      if (logs) this.emit("ganache:vm:tx:console.log", { context, logs });
```
davidmurdoch, line 456: Notice that we don't call `logger.log` here, even though we emit the event. I didn't think it was correct to log for `debug_traceTransaction`, but we DO emit the other `ganache:vm:tx` events already, so I felt like we should still emit console.log events here.

src/packages/ganache/README.md
```diff
@@ -511,9 +511,15 @@ const contexts = new Map();
 provider.on("ganache:vm:tx:before", (event: { context: {} }) => {
   contexts.set(event.context, []);
 });
-provider.on("ganache:vm:tx:step", (event: StepEvent) => {
+provider.on("ganache:vm:tx:step", (event: { context: {}; data: StepEvent }) => {
```
davidmurdoch, line 515: this is just a drive by fix. can remove if you'd prefer.

davidmurdoch:
Important note: contracts that use this console.sol implementation won't entirely work in Remix and Hardhat Network until https://github.com/ethereum/remix-project/issues/2589 and https://github.com/NomicFoundation/hardhat/issues/2666 are fixed.

src/chains/ethereum/console.log/README.md
```diff
@@ -0,0 +1,59 @@
...
+Use it in Solidity just like any other library, e.g.,
+
+```solidity
+// SPDX-License-Identifier: MIT
+pragma solidity >= 0.4.22 <0.9.0;
+
+import "console.sol";
+
+contract MyContract {
+  constructor(){
```
jeffsmale90:
```suggestion
  constructor() {
```

src/chains/ethereum/console.log/README.md
```diff
@@ -0,0 +1,59 @@
...
+// ... initialize vm ...
+
+// vm should be an instance of @ethereumjs/vm
+vm.on("step", event => {
+  const logs: ConsoleLogs | null = maybeGetLogs(event); // (string | bigint | boolean)[] | null
+  // do something with `logs`
+});
+```
+
+Note: Ganache CLI, as well as `Ganache.server` and `Ganache.provider`, already
```
jeffsmale90, line 58: Is it worth mentioning here the `ganache:vm:tx:console.log` event and how a consumer of `Ganache.provider` might use that?

Although I guess this is an internal package, and so maybe these comments are a little too detailed...

src/chains/ethereum/console.log/README.md
```diff
@@ -0,0 +1,59 @@
...
+implement the logger parser and will automatically log calls to `console.sol`.
```
jeffsmale90: ... will automatically log calls to `console.sol` to standard output ?

src/chains/ethereum/console.log/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "mocha": "9.1.3",
+    "nyc": "15.1.0",
+    "solc": "0.8.11",
+    "ts-node": "10.4.0",
+    "typescript": "4.6.4"
+  },
+  "dependencies": {
+    "@ganache/utils": "0.3.0",
+    "ethereumjs-util": "7.1.5",
+    "fast-check": "3.0.1"
```
jeffsmale90: I think fast-check should be dev-dependency?

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,322 @@
...
+    * // Prints: count: 5, to stdout
+    * console.log('count:', count);
+    * // Prints: count: 5, to stdout
+    * \`\`\`
+    *
+    * See \`util.format()\` for more information.
+    */`;
+
+/**
+ * Cache used to ensure we do not accidentally generate 4 byte collisions.
```
jeffsmale90: I didn't understand this comment when I first read it. Is it correct that this map is only used to _detect_ collisions, and throws if they are?

Something like this might be more clear
> Cache used only to detect 4 byte collisions

also, I don't see where we put the signatures in signatureCache?

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,322 @@
...
+ * params.
+ * @param params
+ * @param solidityFunctionName
+ * @returns
+ */
+function getSignature(
+  params: SolidityType[],
+  solidityFunctionName = "log"
+): SignatureDetail {
+  const abiParams = params.map(type => type.replace(" memory", ""));
```
jeffsmale90, line 96: Maybe a bit of an edge case, but is something like this possible as a value for param?

`string storage memoryLocation` which would convert to `string storageLocation` ?

[Edit] looking further into this, I assume that param is actually just the type and modifier for the param, so would be `string memory` in the case below. Is that correct?

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,322 @@
...
+  const abiSignatureString = `log(${abiParams.join(",")})`;
+  // the solidity "4-bytes" signature:
+  const signature = keccak(Buffer.from(abiSignatureString)).subarray(0, 4);
+
+  // we store the signature as an int on the JS side:
+  const signatureInt = parseInt(signature.toString("hex"), 16);
```
jeffsmale90, line 102: buffer.toString accepts a `start` and `end` param, so instead of creating a subarray (granted this is cheap, and might be more clear), we could go direct with:

`keccak(Buffer.from(abiSignatureString)).toString("hex", 0, 4)`

jeffsmale90 (commented):
Partial review. This is gonna take a while 🙂

src/chains/ethereum/console.log/README.md
davidmurdoch: Good catch. I also forgot to indent with 4 spaces, so i'm doing that in a new commit.

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: Yikes! It wasn't being used. I've fixed it.

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: The `type` here is a solidity type (plus the data location, like "memory", in two cases) and is restricted (by TypeScript) to one of `"address" | "bool" | "string memory" | "uint256" | "bytes memory" | "int256" | "bytes11" | "bytes12" | "bytes13" | "bytes14" | "bytes15" | "bytes16" | "bytes17" | "bytes18" | "bytes10" | "bytes19" | "bytes21" | "bytes22" | "bytes23" | "bytes24" | "bytes25" | "bytes26" | "bytes27" | "bytes28" | "bytes20" | "bytes29" | "bytes30" | "bytes31" | "bytes32" | "bytes2" | "bytes1" | "bytes3" | "bytes4" | "bytes5" | "bytes6" | "bytes7" | "bytes8" | "bytes9" | "uint" | "int"`

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: I just tried it and the downside is that prettier made it look like this:

```
  // the solidity "4-bytes" signature:
  const signature = keccak(Buffer.from(abiSignatureString)).toString(
    "hex",
    0,
    4
  );

  // we store the signature as an int on the JS side:
  const signatureInt = parseInt(signature, 16);
  ```
 
   :vomiting_face: 
  
I could bikeshedding some variable names if you feel strongly. :-)

File: src/chains/ethereum/console.log/package.json (hunk 8 above)
Comment (davidmurdoch) [reply src/chains/ethereum/console.log/package.json]: good catch. moved it!

File: src/chains/ethereum/console.log/README.md (hunk 6 above)
Comment (davidmurdoch) [reply src/chains/ethereum/console.log/README.md:58]: Eh, probably not important here. The `console.log` package itself doesn't implement events, as that's all left up to the Ethereum package in Ganache.

Comment (MicaiahReid) [reply src/chains/ethereum/console.log/README.md:58]: I think these details have been written up before, maybe it could just be linked to?

File: src/chains/ethereum/console.log/index.ts [hunk 12]
```diff
@@ -0,0 +1,8 @@
+/*!
+ * @ganache/console.log
+ *
+ * @author David Murdoch
+ * @license MIT
+ */
+
+export { ConsoleLogs, maybeGetLogs } from "./src/maybe-get-logs";
```
MicaiahReid, line 8: I don't have any better suggestions, but I don't _love_ the name `maybeGetLogs` for something that we're exporting. Not a deal breaker for me, but thought I'd mention this in case anyone else agrees.

src/chains/ethereum/console.log/console.sol
```diff
@@ -0,0 +1,2799 @@
...
+
+library console {
+    address constant CONSOLE_ADDRESS = address(0x000000000000000000636F6e736F6c652e6c6f67);
+
+    function _sendLogPayload(bytes memory payload) private view {
+        address consoleAddress = CONSOLE_ADDRESS;
+        assembly {
+            let argumentsLength := mload(payload)
+            let argumentsOffset := add(payload, 32)
+            pop(staticcall(gas(), consoleAddress, argumentsOffset, argumentsLength, 0, 0))
```
MicaiahReid, line 12: Can you explain the need for `pop` here?

src/chains/ethereum/console.log/scripts/helpers.ts
jeffsmale90: I do _not_ feel strongly about this.

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+
+  /* istanbul ignore if */
```
jeffsmale90, line 104: not really 100% coverage!!!! 💔

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+  yield getSignature([solidityType], logName);
+
+  if (hardhatTypeAliases.get(solidityType)) {
+    const alias = hardhatTypeAliases.get(solidityType);
+    const aliasLogName = getNamedLogFunctionName(alias);
+    const solditiySignature = getSignature([solidityType], aliasLogName);
+    delete solditiySignature.javascript;
+    const javascriptSignature = getSignature([alias], aliasLogName);
+    delete javascriptSignature.solidity;
+    yield solditiySignature;
```
jeffsmale90:
```suggestion
    yield soliditySignature;
```

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+ */
+function* getNamedSignature(solidityType: SolidityType) {
+  const logName = getNamedLogFunctionName(solidityType);
+  yield getSignature([solidityType], logName);
+
+  if (hardhatTypeAliases.get(solidityType)) {
+    const alias = hardhatTypeAliases.get(solidityType);
+    const aliasLogName = getNamedLogFunctionName(alias);
+    const solditiySignature = getSignature([solidityType], aliasLogName);
+    delete solditiySignature.javascript;
```
jeffsmale90:
```suggestion
    delete soliditySignature.javascript;
```

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+ * @param solidityType
+ */
+function* getNamedSignature(solidityType: SolidityType) {
+  const logName = getNamedLogFunctionName(solidityType);
+  yield getSignature([solidityType], logName);
+
+  if (hardhatTypeAliases.get(solidityType)) {
+    const alias = hardhatTypeAliases.get(solidityType);
+    const aliasLogName = getNamedLogFunctionName(alias);
+    const solditiySignature = getSignature([solidityType], aliasLogName);
```
jeffsmale90:
```suggestion
    const soliditySignature = getSignature([solidityType], aliasLogName);
```

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+    )
+  );
+
+  /* istanbul ignore if */
+  if (errors && errors.some(error => error.severity === "error")) {
+    throw new Error(errors.map(e => e.formattedMessage).join("\n\n"));
+  }
+
+  const contract = contracts[CONTRACT_NAME][CONTRACT_NAME];
+  delete contracts[CONTRACT_NAME];
```
jeffsmale90: Why delete [CONTRACT_NAME] here? are we not just discarding the `contracts` object

jeffsmale90 (commented):
> 9/25 files viewed

Getting there 😅

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+ */
+function* permute<T>(array: T[] | readonly T[]) {
+  for (let i = 0; i < array.length; i++) {
+    const length = Math.pow(array.length, i + 1);
+    for (let j = 0; j < length; j++) {
+      const parameters: SolidityType[] = [];
+
+      for (let k = i; k >= 0; k--) {
+        const denominator = Math.pow(array.length, k);
+        const index = Math.floor(j / denominator) % combinatorTypes.length;
```
MicaiahReid:
```suggestion
        const index = Math.floor(j / denominator) % array.length;
```
The only use of `permute` passes in `combinatorTypes`, but we use `array.length` everywhere else. Is there a reason to use `combinatorTypes.length` here?

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+        const type = combinatorTypes[index];
```
MicaiahReid:
```suggestion
        const type = array[index];
```
Same as above

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: :sweat_smile:

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: for fun of course! (i've committed a fix for this)

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: fixed!

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+ * contract ConsoleLogTest {
+ *   function testLog(string memory value) public view {
+ *     console.log(value);
+ *   }
+ * }
+ * ```
+ *
+ * @returns
+ */
+export const createContract = (functions: FunctionDescriptor[]) => {
```
MicaiahReid: It looks like this function is unused.

src/chains/ethereum/console.log/scripts/helpers.ts
```diff
@@ -0,0 +1,337 @@
...
+      pop(staticcall(gas(), consolePrecompile, 999, 0, 0, 0)) // out of bounds memory read (offset)
+      pop(staticcall(gas(), consolePrecompile, 0, 999, 0, 0)) // out of bounds memory read (length)
+      pop(staticcall(gas(), randomAddress, 0, 0, 0, 0)) // ensure we don't listen to other contract addresses
+    }
+  }
+${functionStrings.join("\n\n")}
+}`;
+};
+
+export const compileContract = (contractSource: string) => {
```
MicaiahReid: This function also seems to be unused.

MicaiahReid (changes requested):
Still working through a full review but here's a start.

src/chains/ethereum/console.log/src/handlers.ts
```diff
@@ -0,0 +1,152 @@
...
+ * Calling this function when the type is not dynamic will not work. You have
+ * been warned.
+ *
+ * @param memory
+ * @param offset The offset into the given memory from which to start parsing.
+ * @returns The starting position of the value in `memory`, and the `end` position of the value.
+ */
+function getDynamicDataMarkers(memory: Buffer, offset: number) {
+  // A Buffer's length maxes out at UInt32 max, 4294967295, (2**(8 * 4)) - 1) it
+  // is safe to decode the `start` and `length` values as UInt32s.
```
MicaiahReid:
```suggestion
  // A Buffer's length maxes out at UInt32 max, 4294967295, (2**(8 * 4)) - 1) so
  // it is safe to decode the `start` and `length` values as UInt32s.
```

src/chains/ethereum/console.log/src/handlers.ts
```diff
@@ -0,0 +1,152 @@
...
+  // If we had a start position "word" of:
+  //   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 60
+  // we know we really only need to read the last 4 bytes, since we know that
+  // a Buffer's length can't be greater than UInt32 (which can be represented by
+  // up to 4 bytes). So we only need to look at:
+  // 00 00 00 60
+  // `offset + WORD_SIZE - 4` effectively tells `readUInt32BE` to read only
+  // these last 4 bytes of the word that represent the `start`
+  // value.
+  // We add WORD_SIZE to this value because without it it doesn't work.
```
MicaiahReid: If I'm understanding correctly, `memory.readUInt32BE(offset + WORD_SIZE - 4)` gets us the last 4 bytes (and only relevant bytes) of the word after the offset, which contains the location of the data we actually care about here. At that location is a single word containing the length of the data, plus our actual data.

By adding ` + WORD_SIZE + 4`, we're skipping the word that represents the length (and, as your comment says, the signature) so we actually start at the data. Conveniently, we're only looking for the last 4 bytes of the "length" word, so we can just subtract 4 to get the length of the data. Neat stuff!

If this is correct, can we update the 
> We add WORD_SIZE to this value because without it it doesn't work.

comment to indicate that we are skipping the length of the data to get to the actual start?

Also, `4` is a magic number here that comes up a lot, but it's more a coincidence that it's 4 in different cases. For example, we add 4 because this is the size of the signature. We subtract 4 because we move to the _end_ of the word, then backtrack to the relevant data. Could we perhaps be more explicit about what is happening and where data lives?

```TypeScript
  const SIGNATURE = 4;
  const SHORTCUT_TO_LAST4 = WORD_SIZE - 4;
  const startOfLength = memory.readUInt32BE(offset + SHORTCUT_TO_LAST4) + SIGNATURE;
  const start = startOfLength + WORD_SIZE; // the data starts after the length, which is one word
  const length = memory.readUInt32BE(startOfLength + SHORTCUT_TO_LAST4);
  return { start, end: start + length };
```

or a smaller change that only gives the signature a name:
```TypeScript
  const SIGNATURE = 4;
  const start = memory.readUInt32BE(offset + WORD_SIZE - 4) + WORD_SIZE + SIGNATURE;

  // the same optimization is applied here.
  // The word that immediately precedes the data itself is the data's `length`:
  const length = memory.readUInt32BE(start - 4);
  return { start, end: start + length };
```

MicaiahReid (changes requested):

src/chains/ethereum/console.log/src/handlers.ts
```diff
@@ -0,0 +1,152 @@
...
+ * Handle decoding of solidity's `address` type from EVM memory
+ * @param memory
+ * @param offset
+ * @returns A 20 byte hex encoding string, prefixed with 0x
+ */
+export const address = (memory: Buffer, offset: number) =>
+  `0x${memory
+    .subarray(offset + 12, offset + WORD_SIZE)
+    .toString("hex")
+    .padStart(40, "0")}`;
```
MicaiahReid: Why do we need to `padStart`?  Won't this always be 20 bytes converted to a hex string?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1077,6 +1085,12 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
       vm.stateManager.checkpoint();
 
       vm.on("step", (event: InterpreterStep) => {
+        const logs = maybeGetLogs(event);
+        if (logs) {
+          options.logging.logger.log(...logs);
```
MicaiahReid, line 1090: `logging.logger.log(...logs)` is in my top 10 favorite lines of code in ganache :laughing:

MicaiahReid (changes requested):

src/chains/ethereum/console.log/src/maybe-get-logs.ts
```diff
@@ -0,0 +1,71 @@
...
+  stack: BN[];
+}): ConsoleLogs | null => {
+  if (event.opcode.name !== "STATICCALL") return null;
+
+  // STATICCALL, which is the OPCODE that is used to initiate a console.log, has
+  // 6 params, but we only care about these 3:
+  const stack = event.stack as LogsStack;
+  const [inLength, inOffset, toAddress] = stack.slice(-4, -1);
+
+  // only if the toAddress is our console address we should try parsing
```
MicaiahReid, line 47: why talking like yoda are we?

src/chains/ethereum/console.log/src/maybe-get-logs.ts
```diff
@@ -0,0 +1,71 @@
...
+  opcode: { name: string };
+  memory: Buffer;
+  stack: BN[];
+}): ConsoleLogs | null => {
+  if (event.opcode.name !== "STATICCALL") return null;
+
+  // STATICCALL, which is the OPCODE that is used to initiate a console.log, has
+  // 6 params, but we only care about these 3:
+  const stack = event.stack as LogsStack;
+  const [inLength, inOffset, toAddress] = stack.slice(-4, -1);
```
MicaiahReid: The `inLength` and `inOffset` are very weird names to me, I get confused when I see them. Would you be open to `dataLength` or just `length` and `dataStart` or `start` or something like that?

src/chains/ethereum/console.log/src/maybe-get-logs.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  // only if the toAddress is our console address we should try parsing
+  if (!toAddress.eq(CONSOLE_ADDRESS)) return null;
+
+  // STATICCALL allows for passing in invalid pointers and lengths so we need to
+  // guard against failures with a try/catch
+  try {
+    const memoryStart = inOffset.toNumber();
+    const memoryEnd = memoryStart + inLength.toNumber();
+    const values: Buffer = event.memory.subarray(memoryStart, memoryEnd);
```
MicaiahReid: If we pulled out the method from the values, it would slightly simplify having to keep track of the length of the signature, especially for the `string` handler. Would you consider:

```TypeScript
const method = event.memory.subarray(memoryStart, 4).readUInt32BE(0); // 4 bytes wide
const values: Buffer = event.memory.subarray(memoryStart + 4, memoryEnd)
```
It's more expensive, but a bit simpler, your call!

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,526 @@
...
+
+    const { status, contractAddress } = await provider.send(
+      "eth_getTransactionReceipt",
+      [transactionHash]
+    );
+    assert.strictEqual(status, "0x1", "Contract was not deployed");
+    return contractAddress;
+  }
+
+  function encode(params: Param[]) {
```
MicaiahReid: Can this be pulled out of the describe into the global scope since it doesn't depend on a provider or anything?

MicaiahReid:
```suggestion
  function encode(params: Param[] | null) {
```

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,526 @@
...
+  }
+
+  function assertLogs(logs: any[], expectedParams: Param[] | null) {
+    if (expectedParams == null) {
+      // if params is null we shouldn't have collected any logs
+      assert.strictEqual(logs.length, 0);
+    } else {
+      // ensure we logged the right things
+      assert.deepStrictEqual(
+        logs[0],
```
MicaiahReid: I'm not quite following why it's only the 1st one we care about here. Is it possible for us to be more explicit about the type of logs here and in the below `watchLogs` function? Shouldn't the logs be type `ConsoleLogs`?

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,526 @@
...
+      }).timeout(60000);
+    }
+  });
+
+  describe("static", () => {
+    function generateBytesN(n: number): [string, any[]] {
+      return [`bytes${n}`, ["0x" + "00".padEnd(n * 2, "0")]];
+    }
+
+    // `staticValues` genenerates 1000s of tests, adding a single additional
```
MicaiahReid:
```suggestion
    // `staticValues` generates 1000s of tests, adding a single additional
```

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,526 @@
...
+      const contractSource = createContract(functions);
+      const code = compileContract(contractSource);
+      contractAddress = await deploy(code);
+    });
+
+    let counter = 0;
+    for (const { params, name } of getSignatures()) {
+      // don't test signatures with int and uint because our `console.sol`
+      // doesn't use them as these types are only for hardhat's console.sol
+      // compatability.
```
MicaiahReid:
```suggestion
      // compatibility.
```

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,526 @@
...
+            }
+          ]);
+          assert.notEqual(result, "0x");
+          assert.strictEqual(allLogs.length, 0);
+        } finally {
+          logger.log = () => {};
+        }
+      });
+
+      it("does NOT log when `console.log` is called within an `debug_traceTransaction`", async () => {
```
MicaiahReid:
```suggestion
      it("does NOT log when `console.log` is called within a `debug_traceTransaction`", async () => {
```

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,526 @@
...
+          // we need to send two transactions in one block, so we stop the miner
+          // so we can add them to the pool:
+          await provider.send("miner_stop");
+        });
+
+        afterEach("start the miner", async () => {
+          await provider.send("miner_start");
+        });
+
+        it("does NOT log when `console.log` is called within an `debug_storageRangeAt`", async () => {
```
MicaiahReid:
```suggestion
        it("does NOT log when `console.log` is called within a `debug_storageRangeAt`", async () => {
```

MicaiahReid (changes requested):
I do not envy whoever has to debug these tests at some point :sweat_smile: Suuuuper cool stuff though!

src/chains/ethereum/console.log/index.ts
davidmurdoch: I'm don't feel strongly about keeping it. The name is borrowed from functional programming: https://en.wikipedia.org/wiki/Monad_(functional_programming)#An_example:_Maybe

src/chains/ethereum/console.log/console.sol
davidmurdoch: The result of the call to the staticcall must be handled somehow. The `pop` removes the result from the stack. An alternative is to assign it to a variable:

```solidity
let trash := staticcall(gas(), consoleAddress, argumentsOffset, argumentsLength, 0, 0)
```


Which basically does the same thing (it might even end up writing the same op codes, I haven't checked) and cost the same amount of gas either way.

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: Good catch. This was just a bug waiting to happen!

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: It is used by `index.test.ts`

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: It is used by `index.test.ts`

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: It is used by `index.test.ts`

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: It is used by `index.test.ts`

src/chains/ethereum/console.log/src/maybe-get-logs.ts
davidmurdoch: gotta spice things up a  bit!

src/chains/ethereum/console.log/src/handlers.ts
davidmurdoch: Good call. Simplified!

src/chains/ethereum/console.log/src/maybe-get-logs.ts
davidmurdoch: Done!

src/chains/ethereum/console.log/tests/index.test.ts
davidmurdoch: Good catch. We collect _all_ logs over the course of the transactions, and `console.log` could happen multiple times in one transaction. If that happens there could be multiple "sets" of logs, e.g.

```
console.log("Hello", "World");
console.log(1234, 42);
```
would generate:

```
[
  ["Hello", "World"],
  [1234, 32]
]
```

I've updated the test to allow for testing contracts that test multiple logs, however we don't currently test contracts that log multiple times in one transaction.

src/chains/ethereum/console.log/src/maybe-get-logs.ts
davidmurdoch: Good call. Renamed to `memoryLength` and `memoryOffset`. I hope you like it.

src/chains/ethereum/console.log/tests/index.test.ts
davidmurdoch: done

src/chains/ethereum/console.log/src/handlers.ts
davidmurdoch: I'm testing removing the `padStart` locally now.

davidmurdoch: I _can_ think of a way to make things get weird here by executing a malformed `STATICCALL` manually (outside of our console.sol), but I don't think that is real concern here. Good find!

src/chains/ethereum/ethereum/tests/forking/cache/arbitraries.ts
```diff
@@ -112,7 +112,7 @@ export const Networks = (): fc.Arbitrary<Model> =>
     .tuple(
       Commands.AddNetwork(),
       fc.array(
-        fc.frequency(
+        fc.oneof(
```
davidmurdoch, line 115: this was because I updated fast-check to a new major version and `frequency` was replaced by `oneof`.

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: I've moved this to a helpers.ts in the tests directory.

src/chains/ethereum/console.log/scripts/helpers.ts
davidmurdoch: I've moved this to a helpers.ts in the tests directory.

src/chains/ethereum/console.log/tests/index.test.ts
davidmurdoch: I've added a test to test multiple console.log statements in a single transaction.

src/chains/ethereum/console.log/console.sol
MicaiahReid: Neat! Might be worth a comment :shrug:

src/chains/ethereum/console.log/src/handlers.ts
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Returns the hex representation of the bytes in `memory`.
+ *
+ * @param memory
+ * @param start The start of the data in the given memory.
+ * @param end The end of the data in the given memory (not inclusive).
+ * @returns
+ */
+function handleBytes(memory: Buffer, start: number, end: number) {
+  return `0x${memory.subarray(start, end).toString("hex")}`;
```
jeffsmale90:
```suggestion
  return `0x${memory.toString("hex", start, end)}`;
```

I checked, and Prettier doesn't seem upset about this one.

src/chains/ethereum/console.log/src/handlers.ts
```diff
@@ -0,0 +1,151 @@
...
+};
+
+/**
+ * Handle decoding of solidity's `bool` type from EVM memory
+ * @param memory
+ * @param offset
+ * @returns
+ */
+export const bool = (memory: Buffer, offset: number) =>
+  memory[offset + 31] !== 0;
```
jeffsmale90:
```suggestion
  memory[offset + WORD_SIZE - 1] !== 0;
```

src/chains/ethereum/console.log/src/handlers.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Handle decoding of solidity's `address` type from EVM memory.
+ * @param memory
+ * @param offset
+ * @returns A 20 byte hex encoded string, prefixed with 0x.
+ */
+export const address = (memory: Buffer, offset: number) =>
+  `0x${memory.subarray(offset + 12, offset + WORD_SIZE).toString("hex")}`;
```
jeffsmale90:
```suggestion
  `0x${memory.toString("hex", offset + 12, offset + WORD_SIZE)}`;
```

src/chains/ethereum/console.log/src/maybe-get-logs.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!handlers) return null;
+
+    // we skip the first 4 bytes, as that is our signature
+    const values = memory.subarray(4);
+    return handlers.map((handler, index) => {
+      const offset = index * WORD_SIZE;
+      return handler(values, offset);
+    });
+  } catch {
+    return null;
```
jeffsmale90, line 67: Should we log something here? I'm thinking if users are trying to call console.log incorrectly and it's silently failing, it may be difficult to debug.

src/chains/ethereum/ethereum/src/provider-events.ts
```diff
@@ -75,6 +76,11 @@ export type VmAfterTransactionEvent = {
   readonly context: EvmStepContext;
 };
 
+export type VmConsoleLogEvent = {
+  readonly context: EvmStepContext;
+  readonly logs: ConsoleLogs | null;
```
jeffsmale90: Is it not the case that `logs` should be of type `ConsoleLogs`, as we check whether `logs`?
```
if (logs) this.emit("ganache:vm:tx:console.log", { context, logs });
```

https://github.com/trufflesuite/ganache/pull/3327/files#diff-32b463201dcf328f35f745f3b3866381848e899dda327b4d23e4cd1b0ee37db2R456

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -268,18 +271,14 @@ export default class Miner extends Emittery<{
       // of the storage key and save it to the db along with it's keccak hashed version of
       // the storage key. Why you might ask? So we can reference the raw version in
       // debug_storageRangeAt.
-      const stepListener = (
-        event: InterpreterStep,
-        next: (error?: any, cb?: any) => void
```
jeffsmale90, line 273: I don't understand the removal of the next argument, do you have a one liner explanation, or a link you can share?

src/chains/ethereum/ethereum/src/helpers/precompiles.ts
jeffsmale90: Classic Prettier

src/chains/ethereum/console.log/tests/helpers.ts
```diff
@@ -0,0 +1,173 @@
...
+      );
+}
+/**
+ * Generates the 4-byte signature for the given solidity signature string
+ * e.g. `log(address)` => `2c2ecbc2`
+ * @param signature
+ * @returns
+ */
+export function get4ByteForSignature(signature: string) {
+  return `${keccak(Buffer.from(signature)).subarray(0, 4).toString("hex")}`;
```
jeffsmale90:
```suggestion
  return keccak(Buffer.from(signature)).toString("hex", 0, 4);
```

Unless I'm missing something about the string interpolation?

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,543 @@
...
+      miner: { blockGasLimit: "0xfffffffff" },
+      logging: { logger },
+      chain: { allowUnlimitedContractSize: true },
+      // using memdown for performance
+      database: { db: memdown() }
+    });
+  });
+
+  before("get our account address", function () {
+    [from] = Object.keys(provider.getInitialAccounts());
```
jeffsmale90, line 44: handy!

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,543 @@
...
+    method: string,
+    contractAddress: string
+  ) {
+    // send our logging transaction
+    return provider.send("eth_sendTransaction", [
+      {
+        from,
+        to: contractAddress,
+        data:
+          "0x" + method + encode(params ? params.flat() : null).toString("hex")
```
jeffsmale90, line 102:
```suggestion
          `0x${method}${encode(params ? params.flat() : null).toString("hex")}`
```

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -0,0 +1,543 @@
...
+          }),
+          { numRuns: 5, endOnFailure: true }
+        );
+      }).timeout(60000);
+    }
+  });
+
+  describe("static", () => {
+    function generateBytesN(n: number): [string, any[]] {
+      return [`bytes${n}`, ["0x" + "00".padEnd(n * 2, "0")]];
```
jeffsmale90:
```suggestion
      return [`bytes${n}`, ["0x" + "00".repeat(n)]];
```

src/chains/ethereum/console.log/src/maybe-get-logs.ts
davidmurdoch: The users of our console.sol can't call it incorrectly.

They could use assembly and do it wrong though. I feel like that's on them though.

Or maybe HH, foundry, vyper, etc extend their own console.sol and we don't understand those extensions yet, in which case this `return null` behaves like it does in stable Ganache (ignores it).

We also ignore the case where the signature is "missing", which is pretty much the same issue, but we check for its existence instead of letting it throw.

Anyway, I was pretty much thinking about invalid data as adversarial, not mistakes.

src/chains/ethereum/console.log/src/handlers.ts
```diff
@@ -32,27 +32,29 @@ const OR_WITH_TWOS_COMPLEMENT =
...
-  // We add WORD_SIZE to this value because without it it doesn't work.
-  // We add `4` because `memory` includes the function signature, which is 4
-  // bytes, but the `start` position doesn't.
-  const start = memory.readUInt32BE(offset + WORD_SIZE - 4) + WORD_SIZE + 4;
+  // We add WORD_SIZE to this value because we are skipping over the location of
+  // the length and referencing the start of the location of the data itself.
+  const UINT32_BYTE_LENGTH = 4;
+  const nextWordOffset = offset + WORD_SIZE;
+  const startUint32MemoryPosition = nextWordOffset - UINT32_BYTE_LENGTH;
+  const start = memory.readUInt32BE(startUint32MemoryPosition) + WORD_SIZE;
```
MicaiahReid, line 52:
```suggestion
  const UINT32_BYTE_LENGTH = 4;
  const nextWordOffset = offset + WORD_SIZE;
  const startUint32MemoryPosition = nextWordOffset - UINT32_BYTE_LENGTH;
  // We add WORD_SIZE to this value because we are skipping over the location of
  // the length and referencing the start of the location of the data itself.
  const start = memory.readUInt32BE(startUint32MemoryPosition) + WORD_SIZE;
```
Moving comment to be over the line it's actually referencing

src/chains/ethereum/console.log/src/maybe-get-logs.ts
```diff
@@ -0,0 +1,72 @@
...
+);
+
+/**
+ * The purpose of this type is to help maintainers visualize the layout of
+ * the `stack`.
+ */
+type LogsStack = [
+  ...rest: BN[],
+  inLength: BN,
+  inOffset: BN,
```
MicaiahReid:
```suggestion
  memoryLength: BN,
  memoryOffset: BN,
```
Just for consistency with the name changing elsewhere.

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -159,20 +98,55 @@ describe("@ganache/console.log", () => {
...
       }
     ]);
   }
 
-  function assertLogs(logs: any[], expectedParams: Param[] | null) {
-    if (expectedParams == null) {
+  /**
+   * Throws if the given logs aren't a match for the given `expectedParamGroups`
+   *
+   * If `expectedParamGroups` is `null` `logs` is expected to be an empty array
```
MicaiahReid:
```suggestion
   * If `expectedParamGroups` is `null`, `logs` is expected to be an empty array
```

src/chains/ethereum/console.log/tests/index.test.ts
MicaiahReid: @davidmurdoch Sweet I see the fix and the new test! 

I still wonder if the type of `logs` passed to `assertLogs` should be `ConsoleLogs[]` rather than `any[][]`. This change could propagate down to the `watchLogs` function, which should have `allLogs` be of type `ConsoleLogs[]` and 
```TypeScript
    logger.log = (...logs: ConsoleLogs) => {
      allLogs.push(logs);
    };
```

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -432,6 +406,49 @@ describe("@ganache/console.log", () => {
...
+    console.log(value1);
+    console.log(value2);
+  }
+}`);
+
+        const contractAddress = await deploy(code);
+        const method = get4ByteForSignature(`testLog(string,uint256)`);
+        await runTxTest(params, method, contractAddress);
+
+        // const logsProm = watchLogs();
```
MicaiahReid: Did you mean to uncomment this section, or is this currently not working?

MicaiahReid (changes requested):

src/chains/ethereum/ethereum/src/provider-events.ts
davidmurdoch: Good catch. Fixed!

src/chains/ethereum/ethereum/src/miner/miner.ts
davidmurdoch: ethereumjs-vm uses https://www.npmjs.com/package/async-eventemitter which  doesn't require the second argument.

src/chains/ethereum/console.log/tests/index.test.ts
davidmurdoch: I meant to remove it. It was just a prototype of the test I forgot to remove.

src/chains/ethereum/console.log/src/maybe-get-logs.ts
davidmurdoch: :+1: It might be worth understanding that the names here came from the EthereumJS VM, which generates the `event` this type is applied to.

src/chains/ethereum/console.log/src/handlers.ts
davidmurdoch: If this line is moved I think the comment above that needs to be reworded and moved. The whole comment block describes what will happen and why; it currently isn't "anchored" to any of the code lines.

src/chains/ethereum/console.log/console.sol
davidmurdoch: Maaybbeee... it just that this is: 

 * assembly in a stack machine VM (last in first out)
 * and I already find `mload` and `add` to be way more confusing :sweat_smile:

src/chains/ethereum/ethereum/src/blockchain.ts
tenthirtyone, line 1090: Proposal:
`options.logging.logger.log.logged(searchString)` for historic lookups.

tenthirtyone (changes requested):
When I tested as a drop in for a hardhat boilerplate project i do not get the logs in Hardhat.

1. `npm link` the @ganache/console.log package
2. npx hardhat, create basic javascript project
3. Install @ganache/console.log package in hardhat
4. Uncomment the `console.log` line in the boilerplate
5. Run `hardhat test`

ER: `hardhat/console.log` logs to cli
image

AR: Match/Pass

6. Swap in @ganache/console.log
image

7. Rerun `hardhat test`

ER: `@ganache/console.log` to log to cli
AR: Tests run, no console.log

image


The function:
``` solidity
function withdraw() public {
        // Uncomment this line to print a log in your terminal
        console.log(
            "Unlock time is %o and block timestamp is %o",
            unlockTime,
            block.timestamp
        );

        require(block.timestamp >= unlockTime, "You can't withdraw yet");
        require(msg.sender == owner, "You aren't the owner");

        emit Withdrawal(address(this).balance, block.timestamp);

        owner.transfer(address(this).balance);
    }
```

When I remove the `unlockTime` and `block.timestamp` parameters I can confirm that just the string value will log. 

image


I tossed some random combinations at the log. It appears that this only affects `uint` variations. The following also did not log:
`
uint160 test = uint160(address(0));
        // Uncomment this line to print a log in your terminal
        console.log(
            "Unlock time is and block timestamp is",
            "Unlock time is and block timestamp is 123",
            test
        );
`

tenthirtyone:
Bah, I have more testing to do but that last comment decided to just add a review without me being finished

davidmurdoch:
> When I tested as a drop in for a hardhat boilerplate project i do not get the logs in Hardhat.

That is because Hardhat's generated signatures aren't ABI compliant, but the console.log methods implemented here are.

Ganache does understand's HardHat's invalid ABI signatures though.

src/chains/ethereum/console.log/src/handlers.ts
MicaiahReid: Ahh, I see your point. Either is fine with me

MicaiahReid (approved):

tenthirtyone (approved):

src/chains/ethereum/console.log/src/maybe-get-logs.ts
jeffsmale90: Sounds good

jeffsmale90 (approved):

PR #3328, opened 2022-07-04
fix: issue where internal RPC data representation is converted to the incorrect numerical value

jeffsmale90:
A regression in [v7.3.1](https://github.com/trufflesuite/ganache/releases/tag/v7.3.1) introduced an issue where (in a fairly narrow edge case) an incorrect value was returned from an internal RPC data representation. 

The return value of `Quantity.toNumber()` was incorrect in the case where the input buffer contains at least one leading `00` byte, and the significant byte length is greater than 6.

This would result in values being returned much lower than they should have been (ie right shifted by the number of leading `00` bytes).

ie. `[0x00, 0x10, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]` would be converted to `[0x10, 0x00, 0x00, 0x00, 0x00, 0x00]`
resulting in: ```17_592_186_044_416``` instead of ```4_503_599_627_370_496```.


jeffsmale90:
I would like to convert the tests to fast-check tests, as I found this issue writing fast-check tests for the quantity math PR, but I'll pick that up separately (I've got a couple outstanding PRs now in this area)

The regression comes about because I incorrectly pass a `bytelength` parameter to `subarray()`, where it expects an `end` parameter 😱

src/packages/utils/src/things/json-rpc/json-rpc-quantity.ts
```diff
@@ -92,7 +92,7 @@ export class Quantity extends BaseJsonRpcType {
       const trimmedBuffer =
         firstNonZeroByte === 0
           ? this.bufferValue
-          : this.bufferValue.subarray(firstNonZeroByte, length);
```
jeffsmale90, line 95: The second argument here is `end`, which should not be confused with the argument `byteLength`  passed to `readUIntBE()` on line :108 (_I_ confused the two, which is why this regression exists).

MicaiahReid (approved):
Good find!

tenthirtyone (approved):
verified on cli change corrects:
image

davidmurdoch (approved):

PR #3355, opened 2022-07-12
chore: enforce `concurrency: production` in Release workflow

davidmurdoch:
This ensures we don't trigger multiple Releases at once. See https://docs.github.com/en/actions/deployment/about-deployments/deploying-with-github-actions#using-concurrency for details on how this works.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,8 @@
 name: Release
 
+# allow only one "Release" workflow to run at a time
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):

tenthirtyone (approved):

jeffsmale90 (approved):

cds-amal (approved):
Nice, only one production workflow at a time.

PR #3361, opened 2022-07-13
fix: add an upper limit to # of accounts that can be generated by ganache

### LINKED ISSUE #736: Add an upper limit to # of accounts that can be generated by ganache
1000? 10000? Ganache can handle it, but huge numbers of accounts likely accidental and affect start up and time memory. We primarily want to optimize memory in this case, as we need memory for the future in-memory cache.

tenthirtyone:
Issue https://github.com/trufflesuite/ganache/issues/736 calls for a cap on the `totalAccounts` generated by ganache. This change sets a cap at 1000 accounts. 1000 was chosen because it was the largest power of ten that kept its unit tests under < 1s. 

Fixes #736


src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -4,7 +4,7 @@ import { entropyToMnemonic } from "bip39";
 import { Definitions, DeterministicSeedPhrase } from "@ganache/options";
 
 const unseededRng = seedrandom();
-
+const MAX_ACCOUNTS = 1000;
```
tenthirtyone: I expect we'll tune this at some point but for now it was a quick/easy issue

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -229,7 +229,7 @@ export const WalletOptions: Definitions<WalletConfig> = {
     conflicts: ["totalAccounts"]
   },
   totalAccounts: {
-    normalize,
+    normalize: rawInput => (rawInput > MAX_ACCOUNTS ? MAX_ACCOUNTS : rawInput),
```
jeffsmale90: Rather than normalizing this to MAX_ACCOUNTS, should we validate and throw?

jeffsmale90 (commented):
Other than my one query, I give this the 👍

src/chains/ethereum/options/src/wallet-options.ts
tenthirtyone: It's already setup to handle `NaN` so I went for the least impact.

E.g.
``` bash
$ ganache -a roflcopter 
Available Accounts
==================
(no accounts unlocked)

```

But I def agree this is the place to do it if we want to throw.

davidmurdoch: I think we should `console.log` a warning.

src/chains/ethereum/options/src/wallet-options.ts
davidmurdoch: I think we should put this number much much higher, like 100000 accounts, maybe? We aren't so concerned with start up time here (users should notice extreme start-up performance issues) as we are with memory.

We want to switch to a lazy database model, where initial writes are to memory only, and only when memory starts getting full due to start writing to disk. Currently we just always write to disk. Once we switch to this lazy DB model the more memory we have available after start up, the better!

It looks like memory scales rather linearly, after initial startup garbage collection. Here is a graph of accounts (x axis) and memory (y axis) on my machine: https://www.wolframalpha.com/input?i=plot+%7C+%7B%7B10%2C+81%7D%2C+%7B100%2C+82%7D%2C+%7B1000%2C+88%7D%2C+%7B10000%2C+118%7D%2C+%7B25000%2C+181%7D%2C+%7B50000%2C300%7D%2C+%7B100000%2C+515%7D%7D

So what ever limit we choose is is arbitrary. /shrug

davidmurdoch (changes requested):

src/chains/ethereum/options/src/wallet-options.ts
tenthirtyone: :+1: https://github.com/trufflesuite/ganache/pull/3361/files#diff-b465f04541f0c5e2ad842f50433df5af81c1bdd98b7921c558dbd30d635ca25dR7

src/chains/ethereum/options/src/wallet-options.ts
tenthirtyone: :+1: https://github.com/trufflesuite/ganache/pull/3361/files#diff-b465f04541f0c5e2ad842f50433df5af81c1bdd98b7921c558dbd30d635ca25dR234

src/chains/filecoin/options/src/wallet-options.ts
```diff
@@ -66,7 +67,13 @@ export type WalletConfig = {
...
 export const WalletOptions: Definitions<WalletConfig> = {
   totalAccounts: {
-    normalize,
+    normalize: rawInput => {
+      if (rawInput >= MAX_ACCOUNTS)
+        console.log(
+          `wallet.totalAccounts set above MAX_ACCOUNTS: ${MAX_ACCOUNTS} and may affect performance.`
+        );
+      return rawInput;
+    },
```
davidmurdoch: Let's not change filecoin's settings.

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -229,7 +229,13 @@ export const WalletOptions: Definitions<WalletConfig> = {
     conflicts: ["totalAccounts"]
   },
   totalAccounts: {
-    normalize,
+    normalize: rawInput => {
+      if (rawInput >= MAX_ACCOUNTS)
+        console.log(
+          `wallet.totalAccounts set above MAX_ACCOUNTS: ${MAX_ACCOUNTS} and may affect performance.`
```
davidmurdoch: What about something like?

```suggestion
          `The wallet.totalAccounts value of ${rawInput} is greater than ${MAX_ACCOUNTS} which may affect performance.`
```

(if we change it to something like this then `rawInput >= MAX_ACCOUNTS` should change to `rawInput > MAX_ACCOUNTS`)

davidmurdoch (changes requested):

src/chains/ethereum/options/src/wallet-options.ts
jeffsmale90: Shall we console.warn? There's precedent for this, including in similar cases such as:

https://github.com/trufflesuite/ganache/blob/90b3f2b743974b13d23ec72e713350021655c6f9/src/chains/ethereum/options/src/fork-options.ts#L281

src/chains/filecoin/options/src/wallet-options.ts
tenthirtyone: :+1: https://github.com/trufflesuite/ganache/pull/3361/commits/11bdfdb0aa0de9fd1a50f3291a77b0bde982b68a

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -229,7 +229,13 @@ export const WalletOptions: Definitions<WalletConfig> = {
     conflicts: ["totalAccounts"]
   },
   totalAccounts: {
-    normalize,
+    normalize: rawInput => {
+      if (rawInput >= MAX_ACCOUNTS)
+        console.log(
```
jeffsmale90: this should probably be `console.warn`

jeffsmale90 (changes requested):

src/chains/ethereum/options/src/wallet-options.ts
davidmurdoch: The validation should be moved to the wallet, where it would use logger.log

davidmurdoch (changes requested):

src/chains/ethereum/options/src/wallet-options.ts
MicaiahReid: @davidmurdoch are you sure? Don't we log other startup data to `console.log`?

davidmurdoch: I'm pretty sure. Where else do we use `console.log`?

MicaiahReid: When we start up the CLI: https://github.com/trufflesuite/ganache/blob/32a78143744b9191971f1c1bf029f7725d081abe/src/packages/cli/src/initialize/ethereum.ts#L129

davidmurdoch: The user can't control the `logger` on the CLI.

davidmurdoch: Note: we don't respect the "quiet" flag (which sets the logger to a noop) for the startup CLI output. So an argument could be made for this specific case that the warning should _also_ not respect the `quiet` flag. :thinking:

MicaiahReid: Yes, @davidmurdoch, that was the argument I was attempting to make!

MicaiahReid: I can always count on you to make my arguments more intelligible :heart:

davidmurdoch: But it was a trick; this message is not CLI-only output. :-p This log output would also appear in programmatic use, so I think we should respect their user's logger (or quiet flag) 🤔

MicaiahReid: @tenthirtyone do what that guy (^) says, he seems smart

tenthirtyone:
Since I have to back out the changes in the cli and move them to wallet && I borked the conventional commit, I am going to close this pr/del branch and re-open with the appropriate conventional commit.

davidmurdoch:
Not sure why you need to close. The conversations here may be valuable in the future and it would be nice to be able to reference this if we need to; that becomes more difficulty in a closed PR simple because it's not as visible (since you'd have to link to it from another PR to know to get to it.)

tenthirtyone:
@davidmurdoch Alrighty.

tenthirtyone:
Summary: 

* We decided to move this change out of options and into `wallet.ts`
* Nix `console.log` for `logger.log`
* Nix a soft/hard cap in favor of a warning
* Set the warning at `100000`

Cool.

I thought `logger.log` was available in `wallet.ts` but `EthereumInternalOptions["wallet"]` does not include `logger` so I passed `options.logging` to the constructor. I expect that to change. Can I get some clarity on `logger.log` inside `wallet`?

Specifically, should I pass the whole `EthereumOptions` in and pull out wallet/logger or what is preferred ?

jeffsmale90:
> should I pass the whole `EthereumOptions` in and pull out wallet/logger or what is preferred ?

My preference is to pass in _just_ what is required, narrow the interface and all that.

tenthirtyone:
> My preference is to pass in just what is required, narrow the interface and all that.

Since `EthereumOptions` object is the options for multiple classes and the logger was previously unavailable to the wallet, part of the conversation is do we update all those class constructors as well or leave `wallet` as a snowflake/unicorn? 

My assumption is uniformity. So do we update every instantiation of `chain`, `database`, `miner`, and `fork` to bolt on the extra logger param, so they will now match `wallet` even though they have no reason to be coupled to `logger`. Or do we let the class constructors describe the interface with destructuring and now the instantiation is uniform but classes that do not need `logger` are not unnecessarily coupled.

[here](https://github.com/trufflesuite/ganache/pull/3361/files#diff-18efefc92e43bf72c47fc73dde8d4c4a13f1c5646e254482683b2aa25ede9524R33)
``` javascript
const wallet = new Wallet(options.wallet); // becomes
const wallet = new Wallet(options.wallet, options.logging);
```

And the constructor becomes

[here](https://github.com/trufflesuite/ganache/pull/3361/files#diff-aec7e0ea6cf2648413b278ca937c3e772944972d94af20a499e1baa07db1a828R123)
``` javascript
  constructor(opts: EthereumInternalOptions["wallet"]) { // becomes
  constructor(
    opts: EthereumInternalOptions["wallet"],
    logging: EthereumInternalOptions["logging"]
  ) {
```


or, use destructuring

``` javascript
      const wallet = new Wallet(options.wallet); // becomes
      const wallet = new Wallet(options); 
```

and the constructors
``` javascript
  constructor({
    wallet: opts,
    logging // only for classes that need logger
  }: {
    wallet: EthereumInternalOptions["wallet"];
    logging: EthereumInternalOptions["logging"];
  }) {
```

Or my assumption is wrong about uniformity. Or something else.

MicaiahReid:
My preference would be to pass in only what is required (so we'd now be passing in `options.wallet, options.logging` rather than expanding it to `options`).

And I'd extend that preference to pass in only what is required to all of the classes over having uniformity between those classes. I don't see the need to pass logging options into `chain`, `database`, etc. just because we're using them for the wallet.

davidmurdoch:
I think we should leave it as you have it right now `new Wallet(options.wallet, options.logging);`; if only so we can finally close this issue. :-p

tenthirtyone:
> finally close this issue. :-p

yes.

src/chains/ethereum/ethereum/tests/miner/miner.test.ts
```diff
@@ -38,10 +38,11 @@ describe("miner", async () => {
         miner: {
           blockGasLimit: "0x5208" // 21000, or sending one empty tx
         },
-        chain: { chainId: 1337 }
+        chain: { chainId: 1337 },
+        logging: { logger: console }
```
davidmurdoch: I think `EthereumOptionsConfig.normalize` will set this to console.log automatically. Was that not happening here? :thinking:

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -231,7 +241,8 @@ export default class Wallet {
   };
 
   #initializeAccounts = (
-    options: EthereumInternalOptions["wallet"]
+    options: EthereumInternalOptions["wallet"],
+    logging: EthereumInternalOptions["logging"]
```
davidmurdoch: is `logging` needed here?

tenthirtyone: Yeah that's unnecessary now. I was originally being specific and [doing the check here](https://github.com/trufflesuite/ganache/pull/3361/files/8f787a25df718f8c37a16a91f35125f8398580d5#diff-aec7e0ea6cf2648413b278ca937c3e772944972d94af20a499e1baa07db1a828R285) since I saw the conditionals but the options exclusions made it possible to just do in the constructor.

src/chains/ethereum/ethereum/tests/miner/miner.test.ts
tenthirtyone: Nope, my mistake from when I was mass updating to match the new constructor

jeffsmale90 (approved):
I love it!

src/chains/ethereum/ethereum/src/wallet.ts
```diff
@@ -119,7 +120,16 @@ export default class Wallet {
   readonly unlockedAccounts = new Map<string, Data>();
   readonly lockTimers = new Map<string, NodeJS.Timeout>();
 
-  constructor(opts: EthereumInternalOptions["wallet"]) {
+  constructor(
+    opts: EthereumInternalOptions["wallet"],
+    logging: EthereumInternalOptions["logging"]
+  ) {
+    // checkForConflicts in @ganache/options will prevent
```
MicaiahReid: I don't understand this comment. Can you elaborate or remove?

MicaiahReid (changes requested):

src/chains/ethereum/ethereum/src/wallet.ts
tenthirtyone: :+1: Yup, partial note to myself. 

I did not want to create a condition where the message could fire unnecessarily. I was originally doing the check in `#initializeAccounts` but `checkForConflicts` will prevent a user from specifying both given accounts and a `totalAccounts`. So it should be safe in the constructor and reduce passing the logger around more than is needful.

MicaiahReid (approved):

davidmurdoch (commented):
LGTM lol

PR #3389, opened 2022-07-20
fix: use a consistent "now" time when setting up blockchain

### LINKED ISSUE #3271: "should return the `timeAdjustment` value via `evm_increaseTime` when provided as a number" test is flaky

jeffsmale90:
Previously when setting up the blockchain, slightly different start times could be used in different places. This caused calls to `evm_increaseTime` to return unexpected values.

Fixes: #3271

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -294,18 +294,21 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
       );
 
       {
-        // create first block
+        // Grab current time once to be used in all references to "now", to avoid
+        // any discrepancies. See https://github.com/trufflesuite/ganache/issues/3271
+        const startTime = new Date();
 
         // if we don't have a time from the user get one now
-        if (options.chain.time == null) options.chain.time = new Date();
+        if (options.chain.time == null) options.chain.time = startTime;
 
+        // create first block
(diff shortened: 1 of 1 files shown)


jeffsmale90:
Time elapsing between [setting the chain time](https://github.com/trufflesuite/ganache/pull/3389/files#diff-eb13881c7bcdaf19f3b7fe3f7c0b32570442528d3511e250621734dbbcb74992L300) and [calculating #timeAdjustment](https://github.com/trufflesuite/ganache/pull/3389/files#diff-eb13881c7bcdaf19f3b7fe3f7c0b32570442528d3511e250621734dbbcb74992L308) was resulting in a negative initial value being set for #timeAdjustment (-1ms in the reproduction that I made).

We now grab `new Date()` once, and use that in both cases.

jeffsmale90:
I also fixed this in #3317 but figured I would leave this open so that we can get a fix in quickly, in the case that #3317 takes longer to get resolved.

MicaiahReid (approved):
Looks good to me! I ran the broken test a few thousand times locally and it seems to work!

davidmurdoch (approved):

tenthirtyone (approved):
Of all the PRs where I've read the title and gone 'oh wow' :in fear: and then got into the code and gone "oh wow" :in relief: I think this is the best one.

PR #3404, opened 2022-07-22
fix: add `" "` after the HTTP/1.1 Response Status-Code

### LINKED ISSUE #3400: http server is emitting an invalid HTTP/1.1 status line
The http server is emitting an invalid HTTP/1.1 status line, eg.:

```
HTTP/1.1 400
```

The RFC is quite clear, however (https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.1):

```
        Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF
```

The Reason-Phrase is allowed to be an empty string, but the separating space is not optional.

I'm unsure whether to open another issue for this, or if https://github.com/uNetworking/uWebSockets.js is where the buck stops.

_Originally posted by @mxork in https://github.com/trufflesuite/ganache/issues/2320#issuecomment-1034381933_

davidmurdoch:
This change brings Ganache into strict compliance with the [HTTP/1.1 Status-Line specification](https://datatracker.ietf.org/doc/html/rfc2616#section-6.1). Nearly all established HTTP/1.1 response parsers handle HTTP/1.1 Status-Lines that omit the Status-Code trailing space character, but it doesn't hurt to follow the specification strictly.

This change fixes #3400

### DIFF
--- a/src/packages/core/src/servers/utils/http-response-codes.ts
+++ b/src/packages/core/src/servers/utils/http-response-codes.ts
@@ -1,3 +1,19 @@
+/**
+ * HTTP/1.1 Response Status-Codes, including the _required_ space character.
+ *
+ * e.g., `"200 "` or `"404 "`
+ *
+ * RFC Grammar:
+ *
+ * ```ebnf
+ * Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF
+ * ```
+ *
+ * The Status-Codes defined here fullfill the `Status-Code SP` part of the above
+ * grammar.
+ *
+ * See https://datatracker.ietf.org/doc/html/rfc2616#section-6.1 for details.
+ */
 enum HttpResponseCodes {
   OK = "200 ",
   NO_CONTENT = "204 ",
--- a/src/packages/core/src/servers/utils/http-response-codes.ts
+++ b/src/packages/core/src/servers/utils/http-response-codes.ts
@@ -1,9 +1,9 @@
 enum HttpResponseCodes {
-  OK = "200",
-  NO_CONTENT = "204",
-  BAD_REQUEST = "400",
-  NOT_FOUND = "404",
-  METHOD_NOT_ALLOWED = "405",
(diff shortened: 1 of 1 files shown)


src/packages/core/src/servers/utils/http-response-codes.ts
```diff
@@ -1,9 +1,25 @@
...
-  BAD_REQUEST = "400",
-  NOT_FOUND = "404",
-  METHOD_NOT_ALLOWED = "405",
-  IM_A_TEAPOT = "418"
+  OK = "200 ",
+  NO_CONTENT = "204 ",
+  BAD_REQUEST = "400 ",
+  NOT_FOUND = "404 ",
+  METHOD_NOT_ALLOWED = "405 ",
+  IM_A_TEAPOT = "418 "
```
davidmurdoch, line 23: Testing this is difficult, and the payoff for doing so it little to none. I _did_ manually test it by stepping through the lifecycle of a request sent to the Ganache server and the additional space character was present:

```
HTTP/1.1 200 <-- look! an extra space character!
Content-Type: application/json
Content-Length: 42

{"id":"1","jsonrpc":"2.0","result":"1234"}'
```

You can verify this by changing the following line in `./src/packages/core/node_modules/superagent/lib/node/index.js`:

```javascript
  this.req = mod.request(options);
```

with:

```javascript
  var req = this.req; // set tcp no delay
  req.on('socket', (socket) => {
    socket.on("data", data => {
console.log(`|${Buffer.from(data).toString("utf8").split("\r\n")[0]}|`);
    });
  });
```

and then running our `@ganache/core` tests.

jeffsmale90 (approved):
Is it worth adding a reason phrase also?

davidmurdoch:
> Is it worth adding a reason phrase also?

I don't think so :-D

MicaiahReid (approved):
:ship:

tenthirtyone (approved):

PR #3407, opened 2022-07-22
ci: drop macos-11, add macos-12

davidmurdoch:
Just updating CI to run the newer macos-12 runner instead of macos-11.


jeffsmale90 (approved):
Seems reasonable to me!

MicaiahReid (approved):

PR #3408, opened 2022-07-22
test: make bundle-size-check test run on external contributor PRs

### LINKED ISSUE #3340: Use a dummy key in the bundle size check
Use a dummy key in the Github Action's bundle-size check, since it doesn't need to be valid, and we want this size check to work even with external contributor PRs that don't have access to environment secrets.

davidmurdoch:
A [recent addition](https://github.com/trufflesuite/ganache/commit/783ae6325d617afaf633fc8b045e524669ed9c54) to our CI tests that ensures we do not exceed our bundle size limit causes external contributor PRs to fail this check. The reason for the failure is because this check requires an environment secret that is not available to the GitHub Action Environment that runs external contributor pull requests. This change replaces the secret with a fake value that allows the test to run as if the real secret was supplied.

Fixes #3340

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -65,4 +65,4 @@ jobs:
         env:
           # use a fake infura key for the bundle size check so this test will
           # run successfully on external contributor Pull Requests
-          INFURA_KEY: 00000000000000000000000000000000
+          INFURA_KEY: "deadc0dedeadc0dedeadc0dedeadc0de"
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -65,4 +65,4 @@ jobs:
         env:
           # use a fake infura key for the bundle size check so this test will
           # run successfully on external contributor Pull Requests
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -63,4 +63,6 @@ jobs:
           echo "Bundle size is $([[ "$size" -lt 99000000 ]] && echo "ok" || echo "not ok")" &&
           test "$size" -lt 99000000
         env:
-          INFURA_KEY: ${{ secrets.INFURA_KEY }}
+          # use a fake infura key for the bundle size check so this test will
+          # run successfully on external contributor Pull Requests
+          INFURA_KEY: DUMMY_INFURA_KEY_FOR_SIZE_CHECK!
```
cds-amal: Nice.

cds-amal (approved):
Looks good to me. Seems harmless, is documented and will be quickly tested. :)

.github/workflows/pr.yml
davidmurdoch: if only it worked :-(

jeffsmale90 (approved):

PR #3433, opened 2022-07-27
fix: after `provider.disconnect()` is called, Ganache should stop serving requests

### LINKED ISSUE #3293: Ganache Provider does not shut down in eager mode
Hi 👋, I notice some confusing behavior when using `Ganache.provider` with `instamine: eager`. If I call `await provider.disconnect()`, the process does not exit, however the process exits when the disconnect method is removed.

I created a [reproduction repo](https://github.com/cds-amal/ganache-no-exit) for your convenience.

- node: v16.15.1
- platform: OS X 12.3.1 m1, Ubuntu 20.04.4 LTS

jeffsmale90:
Previously, when `provider.disconnect()` was called, a lot of the internals of Ganache would be stopped, but just enough was kept alive to serve requests. This could stop consumers from shutting down cleanly.

Now all pending requests, and any new requests will be rejected with a helpful error.

For example, before this change, the following would successfully return the block number:
```javascript
const provider = Ganache.provider();
await provider.disconnect();
const blockNumber = await provider.send("eth_blockNumber");
```

But now:
```
Error: Cannot process request, Ganache is disconnected.
```

Caveat: any request that has _started_ to process, but not yet complete, may fail with an unexpected error. See #3499 for details.

Fixes #3293

### DIFF
--- a/src/chains/ethereum/ethereum/src/provider.ts
+++ b/src/chains/ethereum/ethereum/src/provider.ts
@@ -35,7 +35,6 @@ import {
   MessageEvent,
   VmConsoleLogEvent
 } from "./provider-events";
-import { ConsoleLogs } from "@ganache/console.log";
 
 declare type RequestMethods = KnownKeys<EthereumApi>;
 
@@ -141,8 +140,8 @@ export class EthereumProvider
 {
   #options: EthereumInternalOptions;
   #api: EthereumApi;
-  #executor: Executor;
   #wallet: Wallet;
+  readonly #executor: Executor;
   readonly #blockchain: Blockchain;
 
   constructor(
@@ -419,9 +418,22 @@ export class EthereumProvider
     }
   };
 
+  /**
+   * Disconnect the provider instance. This will cause the underlying blockchain to be stopped, and any pending
+   * tasks to be rejected. Await the returned Promise to ensure that everything has been cleanly shut down
+   * before terminating the process.
+   * @return Promise<void> - indicating that the provider has been cleanly disconnected
+   */
   public disconnect = async () => {
-    this.#executor.disconnect();
+    const coordinator = this.#executor.getCoordinator();
+
+    // We make a best effort to resolve any currently executing tasks, before rejecting pending tasks. This relies on
+    // this.#blockchain.stop() waiting to resolve until after all executing tasks have settled. Executor does not
+    // guarantee that no tasks are currently executing, before rejecting any remaining pending tasks.
+    coordinator.stop();
     await this.#blockchain.stop();
+    coordinator.rejectPendingTasks();
+
     this.emit("disconnect");
   };
 
--- a/src/chains/ethereum/ethereum/src/provider.ts
+++ b/src/chains/ethereum/ethereum/src/provider.ts
@@ -138,9 +138,9 @@ export class EthereumProvider
   }>
   implements Provider<EthereumApi>
 {
-  readonly #options: EthereumInternalOptions;
-  readonly #api: EthereumApi;
-  readonly #wallet: Wallet;
+  #options: EthereumInternalOptions;
+  #api: EthereumApi;
+  #wallet: Wallet;
   readonly #executor: Executor;
   readonly #blockchain: Blockchain;
 
@@ -419,22 +419,19 @@ export class EthereumProvider
   };
 
   /**
-   * Disconnect the provider, the underlying blockchain will be stopped. Any tasks currently executing will be completed,
-   * any pending tasks will be rejected. The returned Promise will resolve when the provider is disconnected, and a
-   * "disconnect" event will be emitted.
-   * @returns Promise<void> - resolves when the provider has disconnected
+   * Disconnect the provider instance. This will cause the underlying blockchain to be stopped, and any pending
+   * tasks to be rejected. Await the returned Promise to ensure that everything has been cleanly shut down
+   * before terminating the process.
+   * @return Promise<void> - indicating that the provider has been cleanly disconnected
    */
   public disconnect = async () => {
     const executor = this.#executor;
-
-    // we await executor.stop() here to ensure that any currently executing tasks are complete before pulling the
-    // rug out by stopping the blockchain.
-    await executor.stop();
-
-    // we call rejectPendingTasks() _after_ executor.stop() has resolved, to ensure that all tasks are executed in
-    // FIFO order
-    executor.rejectPendingTasks();
+    // We make a best effort to resolve any currently executing tasks, before rejecting pending tasks. This relies on
+    // this.#blockchain.stop() waiting to resolve until after all executing tasks have settled. Executor does not
+    // guarantee that no tasks are currently executing, before rejecting any remaining pending tasks.
+    executor.stop();
     await this.#blockchain.stop();
+    executor.rejectPendingTasks();
 
     this.emit("disconnect");
   };
--- a/src/chains/ethereum/ethereum/src/provider.ts
+++ b/src/chains/ethereum/ethereum/src/provider.ts
@@ -425,14 +425,13 @@ export class EthereumProvider
    * @return Promise<void> - indicating that the provider has been cleanly disconnected
    */
   public disconnect = async () => {
-    const coordinator = this.#executor.getCoordinator();
-
+    const executor = this.#executor;
     // We make a best effort to resolve any currently executing tasks, before rejecting pending tasks. This relies on
     // this.#blockchain.stop() waiting to resolve until after all executing tasks have settled. Executor does not
     // guarantee that no tasks are currently executing, before rejecting any remaining pending tasks.
-    coordinator.stop();
+    executor.stop();
     await this.#blockchain.stop();
-    coordinator.rejectPendingTasks();
+    executor.rejectPendingTasks();
 
     this.emit("disconnect");
   };
--- a/src/chains/ethereum/ethereum/src/provider.ts
+++ b/src/chains/ethereum/ethereum/src/provider.ts
@@ -138,9 +138,9 @@ export class EthereumProvider
   }>
   implements Provider<EthereumApi>
 {
-  #options: EthereumInternalOptions;
-  #api: EthereumApi;
-  #wallet: Wallet;
+  readonly #options: EthereumInternalOptions;
+  readonly #api: EthereumApi;
+  readonly #wallet: Wallet;
   readonly #executor: Executor;
   readonly #blockchain: Blockchain;
 
@@ -419,19 +419,20 @@ export class EthereumProvider
   };
 
   /**
-   * Disconnect the provider instance. This will cause the underlying blockchain to be stopped, and any pending
-   * tasks to be rejected. Await the returned Promise to ensure that everything has been cleanly shut down
-   * before terminating the process.
-   * @return Promise<void> - indicating that the provider has been cleanly disconnected
+   * Disconnect the provider, the underlying blockchain will be stopped. Any tasks currently executing will be completed,
+   * any pending tasks will be rejected. The returned Promise will resolve when the provider is disconnected, and a
+   * "disconnect" event will be emitted.
+   * @returns Promise<void> - resolves when the provider has disconnected
    */
   public disconnect = async () => {
     const executor = this.#executor;
-    // We make a best effort to resolve any currently executing tasks, before rejecting pending tasks. This relies on
-    // this.#blockchain.stop() waiting to resolve until after all executing tasks have settled. Executor does not
-    // guarantee that no tasks are currently executing, before rejecting any remaining pending tasks.
 
-    // await executor.stop() to ensure that all currently processing tasks are complete
+    // we await executor.stop() here to ensure that any currently executing tasks are complete before pulling the
+    // rug out by stopping the blockchain.
     await executor.stop();
+
+    // we call rejectPendingTasks() _after_ executor.stop() has resolved, to ensure that all tasks are executed in
+    // FIFO order
     executor.rejectPendingTasks();
     await this.#blockchain.stop();
 
--- a/src/chains/ethereum/ethereum/src/provider.ts
+++ b/src/chains/ethereum/ethereum/src/provider.ts
@@ -420,9 +420,9 @@ export class EthereumProvider
   };
 
   public disconnect = async () => {
+    this.#executor.disconnect();
     await this.#blockchain.stop();
     this.emit("disconnect");
-    return;
   };
 
   //#region legacy
--- a/src/chains/ethereum/ethereum/src/provider.ts
+++ b/src/chains/ethereum/ethereum/src/provider.ts
@@ -420,9 +420,8 @@ export class EthereumProvider
 
   /**
    * Disconnect the provider instance. This will cause the underlying blockchain to be stopped, and any pending
-   * tasks to be rejected. Await the returned Promise to ensure that everything has been cleanly shut down before
-   * terminating the process.
-   * @return Promise<void> - indicating that the provider has been cleanly disconnected
+   * tasks to be rejected. Emits a `disconnect` event once successfully disconnected.
+   * @returns Fullfills with `undefined` once the provider has been disconnected.
    */
   public disconnect = async () => {
     // executor.stop() will stop accepting new tasks, but will not wait for inflight tasks. These may reject with
@@ -430,7 +429,7 @@ export class EthereumProvider
     this.#executor.stop();
     await this.#blockchain.stop();
 
-    this.#executor.rejectPendingTasks();
+    this.#executor.end();
     this.emit("disconnect");
   };
 
--- a/src/chains/ethereum/ethereum/src/provider.ts
+++ b/src/chains/ethereum/ethereum/src/provider.ts
@@ -420,19 +420,17 @@ export class EthereumProvider
 
   /**
    * Disconnect the provider instance. This will cause the underlying blockchain to be stopped, and any pending
-   * tasks to be rejected. Await the returned Promise to ensure that everything has been cleanly shut down
-   * before terminating the process.
+   * tasks to be rejected. Await the returned Promise to ensure that everything has been cleanly shut down before
+   * terminating the process.
    * @return Promise<void> - indicating that the provider has been cleanly disconnected
    */
   public disconnect = async () => {
-    const executor = this.#executor;
-    // We make a best effort to resolve any currently executing tasks, before rejecting pending tasks. This relies on
-    // this.#blockchain.stop() waiting to resolve until after all executing tasks have settled. Executor does not
-    // guarantee that no tasks are currently executing, before rejecting any remaining pending tasks.
-    executor.stop();
+    // executor.stop() will stop accepting new tasks, but will not wait for inflight tasks. These may reject with
+    // (unhelpful) internal errors. See https://github.com/trufflesuite/ganache/issues/3499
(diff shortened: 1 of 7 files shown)


src/packages/utils/src/utils/request-coordinator.ts
```diff
@@ -74,6 +77,23 @@ export class RequestCoordinator {
...
+    this.queue = async () => {
+      throw new Error("Cannot process request, Ganache is disconnected.");
+    };
+    // ensure that processing cannot be resumed
+    this.resume = () => {
+      throw new Error("Cannot resume processing requests, Ganache is disconnected.");
+    }
+
+    while (this.pending.length > 0) {
+      const current = this.pending.shift();
```
cds-amal: If the order of this flush doesn't matter, does it make sense to use `pop` instead of `shift` in this case? `pop` updates a property (length likely), `shift` would have to shift every remaining element by 1.

src/packages/utils/src/utils/request-coordinator.ts
```diff
@@ -74,6 +77,23 @@ export class RequestCoordinator {
     }
   };
 
+  public disconnect() {
+    this.pause();
+    // make this async to force a Promise return type
```
cds-amal: Perhaps add why, like the comment for `this.resume`.

cds-amal (commented):
Thanks, @jeffsmale90. I spent a moment to marvel at the fix and left some minor nits.

src/packages/utils/src/utils/request-coordinator.ts
jeffsmale90: I thought about that - and it _probably_ doesn't matter, but I figured respecting the FIFO nature of the queue makes sense. 

It may be marginally faster, but this is only happening during disconnection _if_ there's anything queued up, so I don't imagine it'll have much of an impact 🤷

jeffsmale90: I added a comment explaining

MicaiahReid (approved):
Cool stuff!

src/packages/utils/src/utils/request-coordinator.ts
davidmurdoch: Hm.... I don't think we need to purge pending tasks since we can't actually stop them anyway, i'm pretty sure `blockchain.stop()` waits until all pending operations are complete before resolving, so forcing the requests to be `reject`ed early in the RequestCoordinator doesn't shut us down any faster.

davidmurdoch (changes requested):

src/packages/utils/src/utils/request-coordinator.ts
jeffsmale90: We're only rejecting pending tasks, which haven't been executed yet.

Once they get to the head of the queue, they're `shift`ed from `pending` before execution, meaning they're unavailable to the purging function.

davidmurdoch: Ah, I was mistaken!

src/packages/utils/src/utils/request-coordinator.ts
```diff
@@ -97,6 +124,7 @@ export class RequestCoordinator {
           reject(e);
         }
       };
+      executor.reject = reject;
```
davidmurdoch: I'm not fond of extending the function itself with an additional property. What are your thoughts about doing something like:

```
this.pending.push({ executor, reject });
```

instead?

src/packages/utils/src/utils/request-coordinator.ts
davidmurdoch: Mowing the lawn again, thinking about code, per usual, and I had a realization: these shouldn't be rejected until after Ganache in flight requests have been fulfilled. Ganache lets users opt out of asyncRequestProcessing which means requests will be processed in the order they are received. I think this should hold true even when Ganache is shutting down.

So I think we need to pause the coordinator, shut down everything else (which already waits for in-flight requests to be fulfilled), and _then_ reject these. Feel free to try to change my mind though!

src/packages/utils/src/utils/request-coordinator.ts
jeffsmale90: Yep, I like that.

src/packages/utils/src/utils/request-coordinator.ts
jeffsmale90: My understanding is that `database.shutdown()`  will wait until open connections close, before closing the database.

This does not guarantee that all currently processing Ganache requests/tasks are complete - I can imagine a scenario where a single Ganache request requires multiple queries to the database before we can return, and we close the underlying database between these queries.

This PR does not address this. It's possible for an in-flight request to reject with `Error: Database is not open`.

In order to ensure that all currently processing requests are complete before starting shutdown, we would need to keep track of these inflight requests (currently we track a count of them only), or implement a `queueEmpty` event or handler.

src/packages/utils/src/utils/executor.ts
```diff
@@ -12,6 +12,15 @@ export class Executor {
     this.#requestCoordinator = requestCoordinator;
   }
 
+  public stop() {
```
jeffsmale90, line 19: I'd rather not pass these functions through, but I figured this was the lesser of several evils, having considered:
- exposing the RequestCoordinator dependency from executor
- passing RequestCoordinator _and_ Executor as dependencies into Provider
- combining RequestCoordinator and Executor into a single dependency

Happy to hear opinions / alternatives.

src/packages/utils/src/utils/request-coordinator.ts
```diff
@@ -84,7 +124,7 @@ export class RequestCoordinator {
   ) => {
     return new Promise<{ value: ReturnType<typeof fn> }>((resolve, reject) => {
       // const executor is `async` to force the return value into a Promise.
```
davidmurdoch: need to update this comment to match the code change

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -419,10 +418,22 @@ export class EthereumProvider
     }
   };
 
+  /**
+   * Disconnect the provider instance. This will cause the underlying blockchain to be stopped, and any pending
+   * tasks to be rejected. Await the returned Promise to ensure that everything has been cleanly shut down
+   * before terminating the process.
+   * @return Promise<void> - indicating that the provider has been cleanly disconnected
```
davidmurdoch: Maybe something like:


```suggestion
   * @returns Fullfills with `undefined` once the provider has been disconnected.
```

(note that we use `@returns` instead of `@return`)

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -419,10 +418,22 @@ export class EthereumProvider
     }
   };
 
+  /**
+   * Disconnect the provider instance. This will cause the underlying blockchain to be stopped, and any pending
+   * tasks to be rejected. Await the returned Promise to ensure that everything has been cleanly shut down
+   * before terminating the process.
```
davidmurdoch: Can you reword this a bit? Something like "Disconnects the provider from ... ", or "The `provider.disconnect()` method ... "?

src/packages/utils/src/utils/executor.ts
davidmurdoch: Alternatively, what if we:

 1. call/implement `executor.stop()`
    * which calls `requestCoordinator.stop()`
 2. call `await blockchain.stop()`
 3. call a new `executor.end()`
   * which calls a new `requestCoordinator.end()` which would do what `rejectPendingTasks()` does now.

I think this may be slightly different behavior from what you have now, but seems acceptable and the naming makes sense to me?

I may have missed some details, but it seems like a simple approach that isn't too dissimilar from what you have now.


---

I do think RequestCoordinator+Executor could use a small refactoring, but I'd rather not do that now.

davidmurdoch (changes requested):

src/packages/utils/src/utils/request-coordinator.ts
davidmurdoch: Ah, that's unfortunate. In that case I think this PR should only do what it set out to do.

src/packages/utils/src/utils/executor.ts
jeffsmale90: I've adopted your naming for simplicity :)

jeffsmale90:
I've narrowed the scope of this PR as suggested by @davidmurdoch. It now _just_ stops processing requests, and rejects pending tasks.

It makes a "best effort" attempt to resolve in order, but will fail in most cases, because we do not wait for inflight requests to resolve before shutting down the blockchain. See #3499 to cover those changes.

src/packages/utils/tests/request-coordinator.test.ts
```diff
@@ -0,0 +1,98 @@
...
+describe("request-coordinator", () => {
+  const noop = () => undefined;
+  let coordinator: RequestCoordinator;
+
+  beforeEach("instantiate RequestCoordinator", () => {
+    coordinator = new RequestCoordinator(0);
+  });
+
+  describe("stop()", () => {
+    it("should pause processing", () => {
```
davidmurdoch: The test doesn't actually test if processing is paused, it just tests that calling `stop` sets the `paused` property to `true`.

```suggestion
    it("should set `paused` property to `true`", () => {
```

src/packages/utils/tests/request-coordinator.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        () => coordinator.resume(),
+        new Error("Cannot resume processing requests, Ganache is disconnected.")
+      );
+    });
+
+    it("should not allow new requests to be queued", async () => {
+      coordinator.stop();
+
+      await assert.rejects(
+        coordinator.queue(noop, this, []),
```
davidmurdoch: Instead of `this` can you use something like `{}` or `null`? `this` is an actual thing and at a quick glance it seems like it matters.

src/packages/utils/tests/request-coordinator.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  describe("end()", () => {
+    it("should reject pending requests in the order that they were received", async () => {
+      coordinator.pause();
+
+      let nextRejectionIndex = 0;
+      const pendingAssertions: Promise<any>[] = [];
+
+      for (let taskIndex = 0; taskIndex < 10; taskIndex++) {
+        const task = coordinator.queue(noop, this, []);
```
davidmurdoch: I like this test. :+1:

src/packages/utils/tests/request-coordinator.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        pendingAssertions.push(
+          task.catch(_ => {
```
davidmurdoch: Would this be better written as `.finally(() => {`, since we explicitly check the rejection case below, or is it better to only test the `catch` here?

src/packages/utils/src/utils/request-coordinator.ts
```diff
@@ -74,6 +85,32 @@ export class RequestCoordinator {
...
+  public end() {
+    for (const current of this.pending) {
+      current.reject(
+        new Error("Cannot process request, Ganache is disconnected.")
+      );
+    }
+
+    if (this.pending.length > 0) {
+      this.pending.splice(0, this.pending.length);
+    }
```
davidmurdoch: What do you think of:

```suggestion
    while (this.pending.length) {
      this.pending.shift().reject(
        new Error("Cannot process request, Ganache is disconnected.")
      );
    }
```

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -461,4 +461,92 @@ describe("provider", () => {
...
+    [true, false].forEach(asyncRequestProcessing => {
+      describe(`asyncRequestProcessing: ${asyncRequestProcessing}`, () => {
+        beforeEach("Instantiate provider", async () => {
+          provider = await getProvider({
+            chain: { asyncRequestProcessing }
+          });
+        });
+
+        it("rejects requests after disconnect() is called", async () => {
+          provider.disconnect();
```
davidmurdoch, line 477: Should we `await` `disconnect` here?

Without `await`ing it seems that we are guaranteeing that `disconnect` immediately and _synchronously_ stops accepting new requests. I don't _think_ that's what we really want to test here (I think we _should_ `await`).

davidmurdoch (changes requested):

src/packages/utils/src/utils/request-coordinator.ts
jeffsmale90: So I originally had that, but shift is more expensive, so opted for this pattern.

In saying that, I doubt it makes a material impact for any users, unless they have 100s of 1000s of tasks queued up, which probably isn't a likely case.

davidmurdoch: Oh. Neat! Did you measure `shift()`?

jeffsmale90: I did - my benchmark is pretty flawed, as I do setup _within_ the benchmark, but it's the same setup for both runs, so it just dilutes the difference.

```javascript
function codeUnderTest() {
    const data = [...new Array(n)].fill({});
    let item = data.shift();
    while (item !== undefined)
      item = data.shift();
}
```

vs

```javascript
function codeUnderTest() {
    const data = [...new Array(n)].fill({});
    for (const item of data) {}
    data.splice(0, data.length);
}
```

For n = 100, this reports iterating is ~5x faster.

Although:
```javascript
function codeUnderTest() {
    const data = [...new Array(n)].fill({});
    for (let i = 0; i < data.length; i++) {
      const item = data[i];
    }
    data.length = 0;
}
```
is > 2x faster again

src/packages/utils/tests/request-coordinator.test.ts
jeffsmale90: Good point!

src/packages/utils/tests/request-coordinator.test.ts
jeffsmale90: Absolutely - that makes a lot of sense.

src/packages/utils/tests/request-coordinator.test.ts
jeffsmale90: yeah, I think it almost doesn't matter, but finally is maybe more resilient. I can imagine a change to the test meaning that we no longer assert that the task rejects (as we do below), meaning the order check doesn't do anything at all if the tasks don't reject.

src/packages/utils/src/utils/request-coordinator.ts
jeffsmale90: Either way, I don't think that optimizing this for performance makes sense, over readability - using .shift() also aligns with the above where we are processing normally.

Assuming a case where the queue length is 1000 items (and that the total cost is the runtime of my benchmark, including allocating the data), using shift(), we would see a total of ~5ms spent iterating the array :shrug:

davidmurdoch: Cool. Go with whatever feels more readable then!

src/chains/ethereum/ethereum/tests/provider.test.ts
jeffsmale90: That's _exactly_ what we are testing.

Because the disconnect logic is synchronous up until we start to shutdown the blockchain, as I understand, the `provider.request` in this test should be executed in the same event as the synchronous portion of `disconnect()` (the call to `coordinator.stop()`). Happy to learn more if you disagree.

Maybe the test should have a better name and / or comment to explain this? Keen to hear your thoughts?

src/packages/utils/tests/request-coordinator.test.ts
jeffsmale90: I reverted this change after thinking about it some more after the test failed :/

I think we can make this a single promise to push into the `pendingAssertions` array, by chaining:

```javascript
const task = coordinator.queue(...).finally(.../*assert order*/)
```

But then it's only checking the rejection order based on the error message, which I don't like.

Alternatively, we could _not_ chain them, but have two assertion arrays, one that all will reject (the `task.finally(...)` ones), and one that all will resolve (`assert.rejects(...)`).

But I think in retrospect I prefer the original approach. Happy to hear alternatives if you have them.

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: The new comment clears it up. My suggestion here was really that I don't think we need to guarantee this sort of synchronous behavior. I'm okay with the guarantee, just not _convinced_ we need to go that far. Feel free to Resolve this comment if you are okay with it too! (as I suspect you are)

src/packages/utils/tests/request-coordinator.test.ts
davidmurdoch: :+1: on your original approach!

src/packages/utils/src/utils/request-coordinator.ts
```diff
@@ -82,9 +115,15 @@ export class RequestCoordinator {
     thisArgument: any,
     argumentsList: OverloadedParameters<T>
   ) => {
+    if (this.#stopped) {
```
davidmurdoch: I thought I left a comment about this yesterday, but I liked your [original approach](https://github.com/trufflesuite/ganache/pull/3433/commits/0638bf651a0302cbd88939f2b83c04bc0ab81be6#diff-de6272ed4a53f5d059f4c98d25691912a50df13d02ec23f909958cc9a7760ffeR83-R89) where you replaced `queue` and `resume` with functions that reject. Doing it this way made it so that the `this.#stopped` property wasn't unnecessarily checked on every request. What was the reason for changing it?

davidmurdoch (commented):
Just a question and a comment for you!

src/chains/ethereum/ethereum/tests/provider.test.ts
jeffsmale90: Maybe we don't need to guarantee it, but if we change the behaviour, we probably want to look deep into what the implications are - don't want to leak internal errors (such as the db is not connected).

src/packages/utils/src/utils/request-coordinator.ts
jeffsmale90: I liked that for the same reason. I changed it however, because there could be events already scheduled referencing the original functions, which would end up throwing internal errors if they execute after `stop()` is called.

Alternatively, I could have added an _internal_ function that I switched out, but then I'm tossing up between a boolean value check vs an additional function call on the stack 🤷

davidmurdoch: Ah. I didn't think about other code peeling off the methods and using them later. Do we do that anywhere currently?

I'm not sure we should be that concerned with the behavior of a function after it has been removed from its original call context.

```
const queue = requestCoordinator.queue.bind(requestCoordinator);
// then later...
queue() // <-- could we get away with calling this sort of pattern "undefined behavior"
```

jeffsmale90: Yeah, especially because the function is called only in `Executor` which is basically "pass-through", I think it's fine. I liked the idea of making the interface as "correct" as possible, but maybe not worth it.

I'm going to add a comment explaining the edge case to ensure I don't forget and break it in a future change :P

davidmurdoch (approved):
⭐⭐⭐⭐⭐

Would review again!

PR #3449, opened 2022-07-28
draft : Plugin support

satyajeetkolhapure:
To support plugin within ganache, cli and flavors packages
dynamically load the plugin package depending on the argument sent to cli command. E.g
ganache @ganache/filecoin command has @ganache/filecoin flavor as well as package
name. The code looks for the package on the local machine and then looks for certain exports
from packages such as callback method, server default configuration etc.

Suggested release plan -
1st phase - Filecoin package reference will not be removed from cli and flavors packages.
Filecoin code will still remain in the same repository as ganache. Even though there is
hardcoded reference to filecoin package, the code is dynamically loading the filecoin plugin
package. The reason to keep reference is because of dependency on the options package.
2nd phase - Once new packages versions are available in production, move filecoin code into
its separate repository. Use the newly released options package version in filecoin repository
packages. Remove hardcoded filecoin package references from cli and flavors packages.


davidmurdoch:
moved this here: https://github.com/trufflesuite/ganache/pull/4362

PR #3460, opened 2022-08-01
Feat/eth create access list

MicaiahReid:
Just to compare before/after


src/chains/ethereum/ethereum/src/helpers/simulations.ts
```diff
@@ -566,18 +564,42 @@ export const createAccessList = async ({
...
       return { accessList, gasUsed };
     } else {
       // we are going to run this loop again so revert our changes
       await stateManager.revert();
+      stateManager.clearWarmedAccounts();
       previousAccessList = accessList;
     }
     iterations++;
   } while (iterations < MAX_ITERATIONS);
+  // we've tried too many times, but so return our latest attempt
```
davidmurdoch:
```suggestion
  // we've tried too many times, so return our latest attempt
```

PR #3466, opened 2022-08-02
fix: switch to user-provided logger rather than `console.log`

MicaiahReid:
The Ganache options allow you to set a custom logger to handle any logging that Ganache does*. When using Ganache programatically, this can be done as follows:

```TypeScript
const logger = {
  log: (message?: any, ...optionalParams: any[]) => {
    // whatever custom log stuff you want
  }
};

const provider = Ganache.provider({
  logging: { logger }
});
```

However, there were still a few hold-out `console.log`s in our code, which logged directly to the console rather than the user-provided log function. This change fixes this issue.

_*Pro Tip: You can actually update that `logger` function while Ganache is running to temporarily change how you handle logs, if you're into that sort of thing._

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -618,7 +618,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
 
   resume(_threads: number = 1) {
     if (!this.#isPaused()) {
-      console.log("Warning: startMining called when miner was already started");
+      this.#options.logging.logger.log(
+        "Warning: startMining called when miner was already started"
+      );
       return;
     }
 
--- a/src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
+++ b/src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
@@ -23,7 +23,10 @@ export class WsHandler extends BaseHandler implements Handler {
   constructor(options: EthereumInternalOptions, abortSignal: AbortSignal) {
     super(options, abortSignal);
 
-    const { url, origin } = options.fork;
+    const {
+      fork: { url, origin },
+      logging
+    } = options;
 
     this.connection = new WebSocket(url.toString(), {
       origin,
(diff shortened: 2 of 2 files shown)


src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -618,7 +618,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
 
   resume(_threads: number = 1) {
     if (!this.#isPaused()) {
-      console.log("Warning: startMining called when miner was already started");
```
jeffsmale90, line 621: In other places we `console.warn()`.

should we consider extending the logger type to add some sort of log level awareness?

```javascript
export type Logger = {
  log(message?: any, ...optionalParams: any[]): void;
  info(message?: any, ...optionalParams: any[]): void;
  warn(message?: any, ...optionalParams: any[]): void;
  error(message?: any, ...optionalParams: any[]): void;
};
```

When normalizing we could default info/warn/error back to log to avoid this being a breaking change.

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
```diff
@@ -98,7 +98,7 @@ export class WsHandler extends BaseHandler implements Handler {
     }
   }
 
-  private connect(connection: WebSocket) {
+  private connect(connection: WebSocket, options: EthereumInternalOptions) {
```
jeffsmale90: Can we require just the `LoggingConfig` here rather than the entire config object?

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
MicaiahReid: I'm glad you have an opinion because I kept going back and forth on this. Fixed.

Though the types here are a bit confusing - I think I got it right with `EthereumInternalOptions["logging"]` rather than `LoggingConfig`.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: I think it would be great to have a more robust logging system to do this, and I believe it's on the docket (though I couldn't find an issue for it). This PR was meant to just be a quick fix though, so I'd rather get this through. _Especially_ because I hardly consider this a "warning" - there are no consequences to calling `miner_start` when the miner isn't paused, so what are we warning them against?

jeffsmale90: Although... it _does_ start with the word "Warning:" 🦖

jeffsmale90 (approved):

davidmurdoch (approved):

PR #3470, opened 2022-08-03
fix: make `debug_storageRangeAt` return all known keys at transaction index

### LINKED ISSUE #3338: `debug_storageRangeAt` fails to find storage when the slot was created earlier in the same block
## Issue

In some cases, when the transaction index is greater than 0 (i.e. you're not looking at the first transaction in a block), `debug_storageRangeAt` will fail to find storage that was there at the start of the transaction.

This doesn't happen in all cases.  If there was storage at that slot at the start of the block, it seems like it works fine; it will correctly report the storage for that index.  If however there was no storage at that slot at the start of the block, but storage was set in transaction 0, then doing `debug_storageRangeAt` with index 1 will fail to find the storage that was set there in transaction 0.

(At least, that's how the problem looks to me.)

## Reproduction steps

I found this using the `solc8-magic-square` project in the `solidity-test-cases` repo, but I'll admit that this would probably be more easily reproduced with a `SimpleStorage` contract.  So, modify as appropriate.  The relevant fact here is that doing `generateMagicSquare(n)` will set storage slot 0 to the number `n`, at least when `n` is small and odd (you might want to use `SimpleStorage` instead if you don't like being restricted to small odd numbers :P ).

I compiled the project.  I ran `ganache -p 7545 -b 5`.  I then did `truffle migrate` to migrate the project, then `truffle console`.  I then did the following:
```
it = await MagicSquare.deployed()
it.generateMagicSquare(1), it.generateMagicSquare(3) //send both txs, hope they end up in same block
block = await web3.eth.getBlock(/* block number I saw in output */)
block //confirm there are exactly two txs in the block
promisify = require("util").promisify
provider = web3.currentProvider
send = promisify(provider.send.bind(provider))
slothash = web3.utils.soliditySha3("0x" + "00".repeat(32))
await send({ jsonrpc: "2.0", method: "debug_storageRangeAt", id: Date.now(), params: [block.hash, 0, it.address, slothash, 1]}) //no storage returned, as expected
await send({ jsonrpc: "2.0", method: "debug_storageRangeAt", id: Date.now(), params: [block.hash, 1, it.address, slothash, 1]}) //no storage returned... uh-oh!
```

No storage was returned for index 1, despite the slot having been set in index 0.  If you do an `eth_getStorageAt` for the block, you'll be able to see that the slot is correctly set to 3 at the end of the block.

I also tested this with three transactions in the block instead of two; the result was that no storage was returned for any of the three indices.

## Non-reproduction steps

I also did the same thing, but with an extra step of doing `it.generateMagicSquare(5)` in a separate block before doing the two transactions above.  In this case, the problem does not occur; the first `debug_storageRangeAt` shows the correct value of 5, and the second `debug_storageRangeAt` shows the correct value of 1.  The problem only seems to occur when there was no storage present at that slot at the beginning of the block.

(Of course, it's possible I have the exact conditions of the bug incorrect, but I thought this was worth pointing out all the same.)

## Environment

Version: This problem probably exists in all released versions of Ganache 7, since it exists on both 7.0.0 and on 7.3.2.

### LINKED ISSUE #3468: event contexts during debug_traceTransaction are shared between transactions

davidmurdoch:
`debug_storageRangeAt` would sometimes fail to find storage that was there at the start of the transaction if the storage slots were written to in the same block as the target transaction. Ganache now waits to read storage slots until after the requisite transactions in the block have been processed.

Fixes #3338, also fixes #3468


src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2931,11 +2931,12 @@ export default class EthereumApi implements Api {
    * console.log(transactionTrace);
    * ```
    */
+  @assertArgLength(1, 2)
```
davidmurdoch, line 3138: I noticed this was missing from both `debug_` methods so I added it. I can move it to another PR if anyone objects to it being in here.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2931,11 +2931,12 @@ export default class EthereumApi implements Api {
...
   async debug_traceTransaction(
     transactionHash: DATA,
-    options?: Ethereum.TraceTransactionOptions
+    options: Ethereum.TraceTransactionOptions = {}
   ): Promise<Ethereum.TraceTransactionResult<"private">> {
-    return this.#blockchain.traceTransaction(transactionHash, options || {});
+    return this.#blockchain.traceTransaction(transactionHash, options);
```
davidmurdoch, line 3143: Just a drive-by refactor

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2981,14 +2982,22 @@ export default class EthereumApi implements Api {
...
     transactionIndex: number,
     contractAddress: DATA,
     startKey: DATA,
     maxResult: number
   ): Promise<Ethereum.StorageRangeAtResult<"private">> {
-    return this.#blockchain.storageRangeAt(
+    const blockchain = this.#blockchain;
+    if (blockchain.fallback) {
+      throw new Error(
+        "debug_storageRangeAt is not supported on a forked network. See https://github.com/trufflesuite/ganache/issues/3488 for details."
```
davidmurdoch, line 3201: see the linked issue for why this has to be there.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1164,17 +1162,11 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
-    newBlock: RuntimeBlock & { transactions: VmTransaction[] },
-    options: TraceTransactionOptions,
-    keys?: Buffer[],
-    contractAddress?: Buffer
-  ): Promise<TraceTransactionResult> => {
-    let currentDepth = -1;
-    const storageStack: TraceStorageMap[] = [];
-
+    block: RuntimeBlock & { transactions: VmTransaction[] }
+  ): Promise<VM & { stateManager: GanacheStateManager }> => {
```
davidmurdoch, line 1212: `traceTransaction` previously had `debug_storageRangeAt`-specific logic in it, which was just gross and weird. I've refactored our the common code into `fastForwardVM`.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1187,21 +1179,60 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+      const tx = block.transactions[i] as any;
+      const transactionEventContext = {};
+      this.emit("ganache:vm:tx:before", {
+        context: transactionEventContext
+      });
+      await vm.runTx({ tx, block: block as any });
+      this.emit("ganache:vm:tx:after", {
+        context: transactionEventContext
+      });
+    }
```
davidmurdoch: This is pretty much the same as before, but now we don't have weird branching just for `storageRangeAt`

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1324,67 +1356,15 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
-      await vm.stateManager.revert();
-    }
+    vm.on("step", stepListener);
+    this.emit("ganache:vm:tx:before", {
+      context: transactionEventContext
+    });
+    await vm.runTx({ tx: transaction as any, block: newBlock as any });
+    this.emit("ganache:vm:tx:after", {
+      context: transactionEventContext
+    });
```
davidmurdoch: This part needed to be added here, as it was previously handled inside the weird loop that had the condition specifically for `storageRangeAt`, here:  https://github.com/trufflesuite/ganache/pull/3470/files#diff-eb13881c7bcdaf19f3b7fe3f7c0b32570442528d3511e250621734dbbcb74992L1379

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1398,7 +1378,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
   #prepareNextBlock = (
     targetBlock: Block,
     parentBlock: Block,
-    transactionHash: Buffer
+    transactionIndex: number
```
davidmurdoch, line 1455: Instead of checking if we are done by comparing hashes let's just loop up to (and including) the index the transaction is at. No hash comparisons needed!

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1540,102 +1508,53 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
-
-    // #3 - use the contractAddress storage trie to get relevant hashed keys
-    const getStorageKeys = () => {
-      const storageTrie = trie.copy(false);
-      // An address's stateRoot is stored in the 3rd rlp entry
-      storageTrie.setContext(
-        decode<EthereumRawAccount>(addressData)[2],
+    let storageTrie: Trie;
+    if (txIndex === 0) {
+      // there are no transactions to run, so let's just grab what we need
```
davidmurdoch, line 1593: an :sparkles: optimization :sparkles:

src/chains/ethereum/ethereum/src/data-managers/account-manager.ts
```diff
@@ -40,22 +40,6 @@ export default class AccountManager {
     return await trieCopy.get(address.toBuffer());
   }
```
davidmurdoch, line 42: `getStorageAt` wasn't used anywhere

src/chains/ethereum/ethereum/src/state-manager.ts
```diff
@@ -1,20 +1,8 @@
 import SM from "@ethereumjs/vm/dist/state/stateManager";
+import { SecureTrie as Trie } from "merkle-patricia-tree";
 
-export class StateManager extends SM {
-  /**
```
davidmurdoch, line 4: removed some unused "code"

src/chains/ethereum/ethereum/src/state-manager.ts
```diff
@@ -1,20 +1,8 @@
...
-  //     for (const address of touchedArray) {
-  //       const empty = await this.accountIsEmpty(address);
-  //       if (empty) {
-  //         this._cache.del({ buf: address } as any);
-  //       }
-  //     }
-  //   }
-  //   this._touched.clear();
-  // }
+export class GanacheStateManager extends SM {
```
davidmurdoch: This whole class wasn't being used. So I commandeered it for my own nefarious purposes...

...which is just to get at the base class's private method `_getStorageTrie` without typescript complaining about it.

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -317,146 +279,151 @@ describe("api", () => {
...
         }
       });
 
       it("should throw an error for transaction indexes out of range", async () => {
         const message = `transaction index 3 is out of range for block ${blockHash}`;
+        // txIndex out of range
+        await assert.rejects(debugStorage(blockHash, 3, 2), { message });
+      });
+
+      it("throws with a forked network", async () => {
```
davidmurdoch, line 365: the only new test!

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -134,141 +147,108 @@ describe("api", () => {
...
+              "0x0000000000000000000000000000000000000000000000000000000000000001"
+          }
         };
 
         assert.deepStrictEqual(result.storage, storage);
         assert.strictEqual(result.nextKey, null);
       });
 
-      it("should return correct storage given different transaction indexes", async () => {
+      it("should return correct storage given different storage slot changes at different transaction indexes in same block", async () => {
```
davidmurdoch, line 237: The updated test!

It now sets a value for the very first time in the middle of a block. This is the exact issue this PR is fixing.

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -2,33 +2,76 @@ import assert from "assert";
 import { EthereumProvider } from "../../../src/provider";
 import getProvider from "../../helpers/getProvider";
 import compile from "../../helpers/compile";
-import path from "path";
+import { join } from "path";
```
davidmurdoch, line 5: The previous tests had tons of duplication, so I tried cleaning it up a bit. Hopefully its not worse.

There are (hopefully, ha) only three real changes to this file, and I've left a review comment on all three:

1. edited an existing test to test for the right thing
2. added a test to make sure the forking error is tested.
3. changed a static value in a contract, which changed the expected value in the test.

src/chains/ethereum/ethereum/tests/contracts/DebugStorage.sol
```diff
@@ -5,20 +5,27 @@ contract DebugStorage {
...
   string public valueThree;
+  uint public value4;
 
   event ValueSet(uint);
 
   constructor() payable {
     value = 5;
     valueTwo = 1;
-    valueThree = "hello rpdr";
+    valueThree = "hello world";
```
davidmurdoch, line 15: rdpr? huh? I just made this make sense.

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -40,92 +83,62 @@ describe("api", () => {
...
-            }
+          [VALUE1_KEY.hashKey]: {
+            key: VALUE1_KEY.key,
+            value:
+              "0x0000000000000000000000000000000000000000000000000000000000000005"
+          },
+          [VALUE3_KEY.hashKey]: {
+            key: VALUE3_KEY.key,
+            value:
+              "0x68656c6c6f20776f726c64000000000000000000000000000000000000000016"
```
davidmurdoch, line 113: I updated the contract value from "hello rdpr" to "hello world", and this change reflects that.

src/chains/ethereum/utils/src/types/debug-storage.ts
```diff
@@ -1,18 +1,14 @@
...
+export type StorageRecords = Record<string, StorageRecord>;
 
 export type StorageRangeAtResult = {
   nextKey: Data | null;
   storage: StorageRecords;
 };
 
 export type StorageKeys = Map<string, { key: Buffer; hashedKey: Buffer }>;
-
-export type RangedStorageKeys = { keys: Buffer[]; nextKey: Data };
```
davidmurdoch, line 18: After all the refactoring I noticed this wasn't used anymore /shrug

src/chains/ethereum/ethereum/tests/contracts/DebugStorage.sol
tenthirtyone: This should be a separate pr 

jk

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1187,21 +1179,60 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
       : this.common;
 
     const vm = await VM.create({
-      state: trie,
       activatePrecompiles: false,
       common,
       allowUnlimitedContractSize: this.vm.allowUnlimitedContractSize,
       blockchain,
       stateManager: this.fallback
         ? new ForkStateManager({ common, trie: trie as ForkTrie })
```
tenthirtyone: With `debug_storageRangeAt` [throwing](https://github.com/trufflesuite/ganache/pull/3470/files#diff-33bf12664f527a9e8ee226d4340fc6c3255c690db880c187f0a69166dd24e55fR2995) for the mentioned reason, would `ForkStateManager` be reachable? 

`traceTransaction` is the only other place to use `fastForwardVM` but it has an existing TODO for mainnet forking. I'm guessing this is benign for now but something we eventually want on mainnet forking.

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -2,33 +2,76 @@ import assert from "assert";
...
+  hashKey: "0x405787fa12a823e0f2b7631cc41b3ba8828b3321ca811111fa75cd3aa3bb5ace",
+  key: "0x0000000000000000000000000000000000000000000000000000000000000002"
+};
+const VALUE4_KEY = {
+  hashKey: "0xc2575a0e9e593c00f959f8c92f12db2869c3395a3b0502d05e2516446f71f85b",
+  key: "0x0000000000000000000000000000000000000000000000000000000000000003"
+};
 
 describe("api", () => {
   describe("storageRangeAt", () => {
```
tenthirtyone, line 25: This was old/stale, I left it in but it may not be relevant - Per our chat, it might be worthwhile it to add a unit test to ensure zeroed out storage is being properly handled. I believe it is working just fine but we may not explicitly test for it

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: What linked issue? The only linked issue I could find was [this one](https://github.com/trufflesuite/ganache/issues/3338), and it didn't mention forking.

src/chains/ethereum/ethereum/src/state-manager.ts
MicaiahReid: Should we standardize to using `GanacheStateManager` across the repo? Currently the only uses of `DefaultStateManager` and `StateManager` are in `precompiles.ts`, but my create access list PR adds a few uses.

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
```diff
@@ -0,0 +1,129 @@
...
+            nextKey: null,
+            values
+          });
+        }
+      })
+      .on("error", reject);
+  });
+}
+
+export async function getStorage(
```
MicaiahReid, line 111: jsdoc would be nice

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
```diff
@@ -0,0 +1,129 @@
...
+        };
+      })
+    );
+  }
+  await Promise.all(promises);
+
+  return storage;
+}
+
+export async function dumpTrieStorageDetails(
```
MicaiahReid, line 150: jsdoc would be nice

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1164,17 +1162,11 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
-    keys?: Buffer[],
-    contractAddress?: Buffer
-  ): Promise<TraceTransactionResult> => {
-    let currentDepth = -1;
-    const storageStack: TraceStorageMap[] = [];
-
+    block: RuntimeBlock & { transactions: VmTransaction[] }
+  ): Promise<VM & { stateManager: GanacheStateManager }> => {
     const blocks = this.blocks;
     // ethereumjs vm doesn't use the callback style anymore
```
MicaiahReid, line 1214: seems like an unnecessary comment at this point, can we remove? (I know you didn't add it in this PR, but since we're here)
```suggestion
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1187,21 +1179,60 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     const common = this.fallback
       ? this.fallback.getCommonForBlockNumber(
           this.common,
-          BigInt(newBlock.header.number.toString())
+          BigInt(block.header.number.toString())
         )
       : this.common;
 
     const vm = await VM.create({
-      state: trie,
```
MicaiahReid: For future reviewers: the `state` field is ignored if `stateManager` is passed, so this wasn't ever necessary

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1540,102 +1500,53 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
-      parentBlock,
-      transactionHashBuffer
-    );
-    // get storage data given a set of keys
-    const options = {
-      disableMemory: true,
-      disableStack: true,
-      disableStorage: false
-    };
+      // ethereumjs-vm has a dumpStorage(account) method, but we need to honor
```
MicaiahReid: I was a little confused about the placement of this comment. Should it be over `dumpTrieStorageDetails`?

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -134,141 +147,108 @@ describe("api", () => {
 
       it("should return account doesn't exist error when debugging a deployment transaction", async () => {
```
MicaiahReid, line 148:
```suggestion
      it("should return "account doesn't exist" error when debugging a deployment transaction", async () => {
```
pls

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
```diff
@@ -134,141 +147,108 @@ describe("api", () => {
...
 
         assert.deepStrictEqual(result.storage, storage);
         assert.strictEqual(result.nextKey, null);
       });
 
-      it("should return correct storage given different transaction indexes", async () => {
+      it("should return correct storage given different storage slot changes at different transaction indexes in same block", async () => {
         /* 
           Strategy for this test:
             1. Create snapshot
```
MicaiahReid, line 240: What is the purpose of the snapshot/revert in this test?

MicaiahReid (dismissed):

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2981,14 +2982,22 @@ export default class EthereumApi implements Api {
    * console.log(storage);
    * ```
    */
+  @assertArgLength(5)
```
jeffsmale90, line 3190: I can't comment on the correct line, because you didn't change it. The example here doesn't work, and I think it's because the 4 byte signature for the `value()` accessor should be `0x3fa4f245` not `0x55241077`.

At least that's what I get when I compile the contract, and it works if I change :
```javascript
const data = "0x552410770000000000000000000000000000000000000000000000000000000000000019";`
```
to
```javascript
const data = "0x3fa4f2450000000000000000000000000000000000000000000000000000000000000019";`
```

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: https://github.com/trufflesuite/ganache/issues/3488

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1187,21 +1179,60 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     const common = this.fallback
```
jeffsmale90, line 1224: Another "you didn't do it, but..."

```
const block = await blocks.get(number.toBuffer()).catch(_ => null);
```

The `catch` here might be extraneous, as we catch "NOTFOUND" exceptions in `manager.ts` https://github.com/trufflesuite/ganache/blob/3a74e07ea9a90ec7d6749cfad5703d679ba0dac1/src/chains/ethereum/ethereum/src/data-managers/manager.ts#L30

Do with that thought what you will.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1164,17 +1162,11 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     }
   }
 
-  #traceTransaction = async (
-    transaction: VmTransaction,
+  #fastForwardVM = async (
```
jeffsmale90: This _creates_ a fastForward vm right? would it be worthwhile naming it `createFastForwardVm`? A comment explaining what it does might be in order, as I took a while to untangle this and `prepareNextBlock` below.

My understand is that this will run _up to but not including_ the provided transaction index - would be good to indicate that (especially because `prepareNextBlock` _includes_ the provided `transactionIndex`.

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: I made a similar comment regarding `fastForwardVM` - a comment explaining what this function does (including the inclusiveness of transactionIndex) might be helpful.

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
```diff
@@ -0,0 +1,129 @@
...
+
+type TrieData = { key: Buffer; value: Buffer };
+type HashedKeysWithValues = {
+  length: number;
+  values: Buffer[];
+  keys: Buffer[];
+  nextKey: Data | null;
+};
+
+const compare = (a: Buffer, b: Buffer) => a.compare(b) < 0;
```
jeffsmale90, line 29: I guess this is mostly just moving stuff around, but it seems unnecessary to provide a comparator, when we don't provide anything but the default buffer comparator. I'd rather just have this function hard coded within the `findInsertPosition` function.

src/chains/ethereum/ethereum/tests/contracts/DebugStorage.sol
```diff
@@ -5,20 +5,27 @@ contract DebugStorage {
   uint public value;
   uint public valueTwo;
   string public valueThree;
+  uint public value4;
```
jeffsmale90, line 8: is there a reason that this is `value4` and not `valueFour` like the others?

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: @MicaiahReid The link was in the `Error`.

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: Great catch! `55241077` is `setValue`  :facepalm:

davidmurdoch: fixed in https://github.com/trufflesuite/ganache/pull/3470/commits/6197023e85faf135a13adcd61668a00965f83ca2

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
davidmurdoch: The only reason I'll push back on this a tiny bit is because `findInsertPosition` was ported from C++, and that implementation takes a comparator, and it can compare _anything_, not just Buffers. If we hardcode the comparator we'll have to update `findInsertPosition` to be monomorphic (Buffers only).

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
davidmurdoch: added in https://github.com/trufflesuite/ganache/pull/3470/commits/950e7b94261acdc56d4a6aa4e22c928b1e3cf836

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
davidmurdoch: added in https://github.com/trufflesuite/ganache/pull/3470/commits/950e7b94261acdc56d4a6aa4e22c928b1e3cf836

src/chains/ethereum/ethereum/src/state-manager.ts
davidmurdoch: I don't think we should update precompiles.ts, since it is using those just for the types.

_Theoretically_ this could be useful if we obtained a `stateManager` instance from something like `vm.stateManager`, and that instance's type returned only `StateManager` (despite it being a full `GanacheStateManager` with whatever _extra_ properties that may include). Since precompiles doesn't _need_ the extras from `GanacheStateManager`, using the narrower type is usually preferred.

src/chains/ethereum/ethereum/tests/contracts/DebugStorage.sol
davidmurdoch: no reason. fixed in https://github.com/trufflesuite/ganache/pull/3470/commits/d2da0f52f5562526235dc7b005da76dfb7dd79e7

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
davidmurdoch: This suggestion would break things because it isn't quoted correctly; also, the error message isn't _exactly_ `"account doesn't exist"`, so quoting it is a bit odd.

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
jeffsmale90: Defs not worthwhile holding up this PR, but it might be worthwhile considering that especially if performance is important for this function. I _guess_ the comparator can be invoked a lot?

jeffsmale90 (approved):

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: You guys are so dang smart.

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
MicaiahReid: Ahh, you're so right. Reading that `it` statement took me about 5 attempts to understand it, but based off of my other review suggestions from that day, I think I was a little slow when I reviewed this :sweat_smile:

src/chains/ethereum/utils/src/types/debug-storage.ts
```diff
@@ -1,18 +1,14 @@
...
-  {
-    key: Data;
-    value: Data;
-  }
->;
+type StorageRecord = {
+  key: Data;
+  value: Data;
+};
+export type StorageRecords = Record<string, StorageRecord>;
```
tenthirtyone, line 7: I was like "Wouldn't `StorageRecord**s**` be a collection for `StorageRecord` not `Record`" and then I was like "Where is Record coming from"? 

And that's when learning occurred. Their utility naming does force a double take from the uninitiated (me).

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
```diff
@@ -0,0 +1,152 @@
...
+    const values: Buffer[] = [];
+
+    trie
+      .createReadStream()
+      .on("data", function onData({ key, value }: TrieData) {
+        // ignore anything that comes before our starting point
+        if (startKey.compare(key) > 0) return;
+
+        // insert the key exactly where it needs to go in the array
+        const position = findInsertPosition(keys, key, compare);
```
tenthirtyone, line 63: totally unnecessary thought on optimizing.

If we hold `highestKey` in memory and `highestKey` === the highest key value in the `keys` array 
could we compare `keys.length === maxKeys && key > highestKey` return? This would guarantee we only find the insert position for keys that could be in the array once it is full by automatically dropping keys that are too high


I think that would guarantee you always have the keys you want but it would only insertion sort `maxKeys` instead of keys greater than `startKey` 

This _might_ eliminate the need for the `length > maxKeys` in the `end`

tenthirtyone (commented):
Just comments no changes requested

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
davidmurdoch: I originally thought you were on to something (and you still might be) and started implementing this, but I realized ( I think) that if `key` ends up being less than `highestKey`, we still will need to insert into the array, which will make the array too long; so we still need to trim it down to size somehow. So we don't eliminate the `length > maxKeys` check.

Additionally, we _need_ to collect one more key than `maxKeys`, in order to find and return the `nextKey` value. So while there may be something to your optimization idea, implementing it started getting a bit complex so I've bailed and am here writing this comment now. Let me know if you want to pair on this and make we can make the optimization make more sense!

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: I've moved it!

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
davidmurdoch: These tests share a `provider`. It's just the way it was already, so I didn't update it to use `beforeEach`.

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: made an issue for it here: https://github.com/trufflesuite/ganache/issues/3841

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: done!

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: done!

src/chains/ethereum/ethereum/tests/api/debug/debug_storageRangeAt.test.ts
davidmurdoch: made an issue for it here: https://github.com/trufflesuite/ganache/issues/3842

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
tenthirtyone: Much appreciation to the time put in. 

I [made a branch](https://github.com/trufflesuite/ganache/blob/storageRangePrune/src/packages/utils/tests/utils.test.ts) with unit tests to show existing vs suggested. 

The TLDR is [one line](https://github.com/trufflesuite/ganache/blob/storageRangePrune/src/packages/utils/tests/utils.test.ts#L101).

I believe the existing code is the 'corner office' (best) solution.

It _is_ possible for the suggested custom solution to work without the `if (position > maxKeys) return;` however performance suffers greatly. 

The only time I saw the linked suggestion bench faster is when there were >1000 keys but I believe this is only due to the test conditions -> I simplified it to test against integers. The custom should prune operations when `position` equals `maxLength-1`. E.g. a key collision for the final item in the results array. 

I don't believe that is possible in production. Sharing code/thoughts for code/thoughts sake.

davidmurdoch: So would you like to chase this optimization further or is the code good as is?

tenthirtyone: > I believe the existing code is the 'corner office' (best) solution.

Good to go

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
MicaiahReid: I'm not seeing it as part of this branch. Is there some weird git stuff going on?

src/chains/ethereum/ethereum/src/forking/state-manager.ts
```diff
@@ -1,13 +1,13 @@
...
+import { decode } from "@ganache/rlp";
 import { ForkCache } from "./cache";
 import { ForkTrie } from "./trie";
 import { GanacheStateManager } from "../state-manager";
 
 /**
  * Options for constructing a [[GanacheStateManager]].
  */
-export interface GanacheStateManagerOpts {
+export interface ForkStateManagerOpts {
```
MicaiahReid, line 10: I changed this name - it doesn't really have anything to do with the merge. I just thought that it was weird that we were calling it `GanacheStateManagerOpts` when it specifies a trie of type `ForkTrie`.

src/chains/ethereum/ethereum/src/forking/state-manager.ts
```diff
@@ -42,7 +43,7 @@ export class ForkStateManager extends GanacheStateManager {
    * at the last fully committed point, i.e. as if all current
    * checkpoints were reverted.
    */
-  copy(): GanacheStateManager {
+  copy(): ForkStateManager {
```
MicaiahReid, line 46: Another one that I changed that's unrelated to the merge. But figured the types should line up and I was here so...

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -734,7 +733,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
       const { miner: minerOptions } = this.#options;
 
       // commit accounts, but for forking.
-      const stateManager = <GanacheStateManager>this.vm.stateManager;
+      const stateManager = <DefaultStateManager>this.vm.stateManager;
```
MicaiahReid: Most of the time I'm changing from `DefaultStateManager` to `GanacheStateManager`. Here I'm doing the reverse. This is in the `initializeGenesisBlock` function, and as far as I can tell there's no reason for this to be a `GanacheStateManager`.

src/chains/ethereum/ethereum/src/helpers/storage-range-at.ts
davidmurdoch: Fixed in https://github.com/trufflesuite/ganache/pull/3470/commits/b36eb4b3f0865351ba4c981dc7604c1c37167f54

MicaiahReid (approved):

jeffsmale90 (approved):

PR #3494, opened 2022-08-04
chore: update `prettier` and `husky`, add `husky` to `prepare` script

davidmurdoch:
This is the first of 3 PRs to apply prettier to Ganache.

* https://github.com/trufflesuite/ganache/pull/3496
* https://github.com/trufflesuite/ganache/pull/3495

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "docs.preview": "lerna run docs.preview",
     "postinstall": "cd scripts && ts-node postinstall",
     "reinstall": "npm run clean && npm install",
+    "prepare": "husky install",
     "prepublishOnly": "cd scripts && ts-node filter-shrinkwrap.ts \"$(lerna list --parseable --scope ganache)\"/npm-shrinkwrap.json",
     "postpublish": "git restore \"$(lerna list --parseable --scope ganache)\"/npm-shrinkwrap.json",
     "start": "lerna exec --loglevel=silent --scope ganache -- npm run start --silent -- ",
@@ -27,7 +28,7 @@
     "@types/marked": "4.0.8",
     "@types/mocha": "9.0.0",
     "@types/node": "17.0.0",
-    "@types/prettier": "2.4.2",
+    "@types/prettier": "2.6.4",
     "@types/yargs": "16.0.0",
     "camelcase": "6.2.1",
     "chalk": "4.1.2",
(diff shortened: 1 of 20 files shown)


MicaiahReid (approved):
:dog:

package.json
```diff
@@ -55,7 +56,7 @@
   "license": "MIT",
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged"
+      "pre-commit": "npx pretty-quick --staged"
```
jeffsmale90: Why `npx`, AFAIK scripts will use local packages _anyways_, although I see that `npx` will:

> installing any packages needed in order for <command> to run

Is there some general "Thou shalt use npx" dictate / reasoning?

tenthirtyone: Do we always need/want latest?

davidmurdoch: Husky says to do it this way. See: https://prettier.io/docs/en/precommit.html#option-2-pretty-quickhttpsgithubcomazzpretty-quick

I don't know why though. :thinking:

davidmurdoch: I wonder if it is so it still works if you didn't `npm install` before committing? If that is the reason, it should probably be `npx pretty-quick@3.1.3 ...`. BUTTTT if pretty-quick isn't installed then `prettier` probably isn't either... so it'd probably fail anyway. :thinking:  :thinking:

davidmurdoch: Anyone have opinions on what we should do here? Probably need to pin the dependency version and we'll be set? Or should we go against the recommendation?

davidmurdoch: pretty-quick's own docs say to do

```
"husky": {
  "hooks": {
    "pre-commit": "pretty-quick --staged"
  }
}
```

https://github.com/azz/pretty-quick#pre-commit-hook

hm

davidmurdoch: I think I'm more comfortable _without_ `npx`.

tenthirtyone: I think I would side with the `pretty-quick` docs over `husky` on that.

tenthirtyone (approved):
Cool with either solution you chose for the `npx` or not part.

package.json
davidmurdoch: Taking @tenthirtyone's recommendation to use pretty-quick's docs over husky's:

```suggestion
      "pre-commit": "pretty-quick --staged"
```

jeffsmale90 (approved):

PR #3498, opened 2022-08-04
fix: nonce generation race conditions

### LINKED ISSUE #2489: Automatic nonce generation race condition
### Discussed in https://github.com/trufflesuite/truffle/discussions/4774

<sup>Originally posted by **theblock-dev** February 24, 2022</sup>
Hi All
While running the truffle migrate --reset command, I am getting into nonce issue. The migration code seems to be fine and I can not seem to find any issue with it. Here is the error I am getting

image

Is there a way to reset the nonce for the truffle? 

Thanks</div>

### LINKED ISSUE #3794: Ganache mines two different transactions with the same "from" and "nonce"
[Reproduction](https://replit.com/@someguy321/MoralWaterySeahorse#index.ts). You need to fork repl (top right corner) to see the console.

Ethereum does not allow two different transactions with the same from and nonce. But ganache does. In the repro, we can verify that it indeed confirmed those two transactions because the balance of the destination address is `3` (`1 + 2`). 

I want simulate transaction replacement behaviour locally but I can't because of this bug.

MicaiahReid:
#2489 documents a race condition in the transaction pool when nonces are automatically being assigned by Ganache. This ended up running pretty deep in the causes of the race condition.

### Check `inProgress` queue for highest nonce
Before this change, we would look at the pending pool first to assign a highest nonce. This is good. The pending transactions are up next to be run, so if we fetched the account's nonce from the chain instead, we'd be assigning this transaction a nonce that is about to be out of date.

However, we wouldn't check the `inProgress` pool. When a transaction is _run_ in the miner, it is removed from the `pending` pool and added to the `inProgress` pool. However, at this point, the block hasn't been saved so the account's nonce hasn't been updated yet. The `pending` pool won't contain transactions from this origin, and the account nonce is out of date.

To resolve this, I've added the `#getLatestInProgressFromOrigin` function, which will search the `inProgress` queue for transactions from this origin and get the one with the highest nonce. This lead to changing the shape of the `inProgress` pool a bit - previously it was a `Set<TypedTransaction>`, which we'd have to loop over to get all transactions from this origin. Now it is `Map<string, Set<TypedTransaction>>` so that we can reduce the number of transactions to look through - now we only check the transactions from this origin.

As a note, we don't care about the `inProgress` pool if there _are_ transactions in the `pending` pool. The `pending` pool will have the account's highest nonces, followed by the `inProgress` pool, and finally the account. So that's the priority we put on finding the nonce.

### Use Semaphore to queue same-origin transactions
If transactions come into the pool quickly enough, it's possible that multiple transactions are looking to have a nonce generated before any of them have been added to the pool. All of these transactions will have the same pool conditions - the same (or no) transactions in the `pending` and `inProgress` pools, so they will all be assigned the same nonce. This will cause all but the first to be mined to throw with an incorrect nonce error.

To solve this, we now use a semaphore to queue transactions from the same origin. Unique-origins transactions can still be prepared in parallel, but same-origin transactions are prepared in series.

### Cache account balance with `inProgress` transactions
Regardless of the above enhancement, we'd still have to make a slow trip to the database to get the account's balance. This is to verify that the sender has sufficient funds to pay for the transaction. This async call was causing a lot of race conditions, so I did my best to minimize them.

Now, when a transaction is run in the miner, we fetch the account balance (which is still in memory from running the transaction) and store it with the transaction in the `inProgress` pool. This lead to further modifying the shape of the `inProgress` pool - from `Map<string, Set<TypedTransaction>>` to `Map<string, Set<{transaction: TypedTransaction, originBalance: Quantity}>>`

Fixes #2489 and Fixes #3794

### DIFF
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -371,7 +371,7 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
     if (secretKey) {
       transaction.signAndHash(secretKey.toBuffer());
     }
-    console.log(`txNonce ${txNonce}, origin: ${origin}`);
+
     switch (transactionPlacement) {
       case TriageOption.Executable:
         // if it is executable add it to the executables queue
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -158,10 +158,11 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
     secretKey?: Data
   ) {
     const origin = transaction.from.toString();
+    const originsQueue = this.#originsQueue;
     let queueForOrigin: Semaphore;
-    if (!(queueForOrigin = this.#originsQueue.get(origin))) {
+    if (!(queueForOrigin = originsQueue.get(origin))) {
       queueForOrigin = Semaphore(1);
-      this.#originsQueue.set(origin, queueForOrigin);
+      originsQueue.set(origin, queueForOrigin);
     }
     await new Promise(resolve => queueForOrigin.take(resolve));
     try {
--- a/src/chains/ethereum/ethereum/src/transaction-pool.ts
+++ b/src/chains/ethereum/ethereum/src/transaction-pool.ts
@@ -12,7 +12,7 @@ import {
   INSUFFICIENT_FUNDS
 } from "@ganache/ethereum-utils";
 import { EthereumInternalOptions } from "@ganache/ethereum-options";
-import { Executables } from "./miner/executables";
+import { Executables, InProgressData } from "./miner/executables";
 import Semaphore from "semaphore";
 import { TypedTransaction } from "@ganache/ethereum-transaction";
 
@@ -77,17 +77,14 @@ function shouldReplace(
     return true;
   }
 }
-function findHighestNonceByOrigin(set: Set<TypedTransaction>, origin: string) {
-  let highestNonce: bigint = null;
-  for (const transaction of set) {
-    if (
-      transaction.from.toString() === origin &&
-      (highestNonce === null || transaction.nonce.toBigInt() > highestNonce)
-    ) {
-      highestNonce = transaction.nonce.toBigInt();
-    }
+
+function validateSufficientFunds(cost: bigint, balance: bigint) {
+  if (balance < cost) {
+    throw new CodedError(
+      INSUFFICIENT_FUNDS,
+      JsonRpcErrorCode.TRANSACTION_REJECTED
+    );
   }
-  return highestNonce;
 }
 
 function byNonce(values: TypedTransaction[], a: number, b: number) {
@@ -144,14 +141,10 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
     this.#priceBump = options.priceBump;
   }
   public readonly executables: Executables = {
-    inProgress: new Set(),
+    inProgress: new Map(),
     pending: new Map()
   };
   public readonly origins: Map<string, Heap<TypedTransaction>>;
-  readonly #accountPromises = new Map<
-    string,
-    Promise<{ balance: Quantity; nonce: Quantity }>
-  >();
 
   public async prepareTransaction(
     transaction: TypedTransaction,
@@ -202,18 +195,6 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
 
     const origin = from.toString();
 
-    // We await the `transactorNoncePromise` async request to ensure we process
-    // transactions in FIFO order *by account*. We look up accounts because
-    // ganache fills in missing nonces automatically, and we need to do it in
-    // order.
-    // The trick here is that we might actually get the next nonce from the
-    // account's pending executable transactions, not the account...
-    // But another transaction might currently be getting the nonce from the
-    // account, if it is, we need to wait for it to be done doing that. Hence:
-    let transactorPromise = this.#accountPromises.get(origin);
-    if (transactorPromise) {
-      await transactorPromise;
-    }
     // if the user called sendTransaction or sendRawTransaction, effectiveGasPrice
     // hasn't been set yet on the tx. calculating the effectiveGasPrice requires
     // the block context, so we need to set it outside of the transaction. this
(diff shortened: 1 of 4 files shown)


src/chains/ethereum/ethereum/tests/transaction-pool.test.ts
```diff
@@ -95,421 +96,714 @@ describe("transaction pool", async () => {
...
     };
-  });
-  beforeEach(async function () {
-    // for each test, we'll need a fresh set of origins
+
     origins = new Map();
-  });
+  };
 
-  it("rejects transactions whose gasLimit is greater than the block gas limit", async () => {
```
MicaiahReid, line 104: This big block is just rearranging to prevent subsequent tests from relying on previous ones being run (`.only` was broken before this)

src/chains/ethereum/ethereum/tests/transaction-pool.test.ts
```diff
@@ -95,421 +96,714 @@ describe("transaction pool", async () => {
...
-      secretKey
-    );
-    assert(!isExecutable); // our transaction is not executable
-    // our non executable transaction should be found in the origins queue
-    const found = findIn(futureNonceTx.hash.toBuffer(), origins);
-    assert.strictEqual(
-      found.serialized.toString(),
-      futureNonceTx.serialized.toString()
-    );
+  describe("nonce generation and validation", async () => {
```
MicaiahReid, line 592: This is the new describe block. I don't know what github is doing here with this diff, I'd just look at this new describe block in your editor instead of here.

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -130,14 +146,45 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
...
+    // the semaphore will hold on `take` until previous uses have resolved
+    await new Promise(resolve => queueForOrigin.take(resolve));
+    try {
+      // check if transaction is in origins, if not add it
+      // if so, use semaphore to call prepareTransaction
+      // if not, just call prepare transaction
+      return await this._prepareTransaction(transaction, secretKey);
+    } finally {
+      // free up the semaphore once we finish
+      queueForOrigin.leave();
```
MicaiahReid, line 185: Note: I think we could use the [semaphore's `available()` function](https://github.com/abrkn/semaphore.js/blob/67e14f8599563a37de1a8ca411d501b8aadb388b/lib/semaphore.js#L80) to see if there's any transaction waiting to be prepared in _this_ semaphore. If not, we could dispose of this semaphore. But then, if we get another transaction from that origin we have to make a whole new semaphore for it. 

So currently we don't drain the `originsQueue`. I'm interested in hearing y'all's thoughts on if we should drain it.

It's easy to imagine someone sending a transaction at regular intervals from the same origin, and this interval being slower than it takes to queue up the previous transaction. In that case, we'd be making a new semaphore for every transaction, which is why I decided against draining.

However, someone could also send a bunch of transactions, each from a unique origin. In that case, this `originsQueue` will just keep getting wider and wider without taking advantage of its purpose.

MicaiahReid:
Here's a summary of intent on what's changed for this PR.

#2489 documents a race conditions in the transaction pool when nonces are automatically being assigned by Ganache. This ended up running pretty deep in the causes of the race condition.

### Check `inProgress` queue for highest nonce
Before this change, we would look at the pending pool first to assign a highest nonce. This is good. The pending transactions are up next to be run, so if we fetched the account's nonce from the chain instead, we'd be assigning this transaction a nonce that is about to be out of date.

However, we wouldn't check the `inProgress` pool. When a transaction is _run_ in the miner, it is removed from the `pending` pool and added to the `inProgress` pool. However, at this point, the block hasn't been saved so the account's nonce hasn't been updated yet. The `pending` pool won't contain transactions from this origin, and the account nonce is out of date.

To resolve this, I've added the `#getLatestInProgressFromOrigin` function, which will search the `inProgress` queue for transactions from this origin and get the one with the highest nonce. This lead to changing the shape of the `inProgress` pool a bit - previously it was a `Set<TypedTransaction>`, which we'd have to loop over to get all transactions from this origin. Now it is `Map<string, Set<TypedTransaction>>` so that we can reduce the number of transactions to look through - now we only check the transactions from this origin.

As a note, we don't care about the `inProgress` pool if there _are_ transactions in the `pending` pool. The `pending` pool will have the account's highest nonces, followed by the `inProgress` pool, and finally the account. So that's the priority we put on finding the nonce.

### Use Semaphore to queue same-origin transactions
If transactions come into the pool quickly enough, it's possible that multiple transactions are looking to have a nonce generated before any of them have been added to the pool. All of these transactions will have the same pool conditions - the same (or no) transactions in the `pending` and `inProgress` pools, so they will all be assigned the same nonce. This will cause all but the first to be mined to throw with an incorrect nonce error.

To solve this, we now use a semaphore to queue transactions from the same origin. Unique-origins transactions can still be prepared in parallel, but same-origin transactions are prepared in series.

### Cache account balance with `inProgress` transactions
Regardless of the above enhancement, we'd still have to make a slow trip to the database to get the account's balance. This is to verify that the sender has sufficient funds to pay for the transaction. This async call was causing a lot of race conditions, so I did my best to minimize them.

Now, when a transaction is run in the miner, we fetch the account balance (which is still in memory from running the transaction) and store it with the transaction in the `inProgress` pool. This lead to further modifying the shape of the `inProgress` pool - from `Map<string, Set<TypedTransaction>>` to `Map<string, Set<{transaction: TypedTransaction, originBalance: Quantity}>>`

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -280,23 +293,51 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
         // origin's heap we are executable.
         transactionPlacement = TriageOption.Executable;
       }
+      // we've gotten the account's nonce the synchronous way (by looking at
+      // the pending queue), but we still need to look up the account's balance
+      const { balance } = await this.#blockchain.accounts.getNonceAndBalance(
```
MicaiahReid, line 298: In the case that we get the nonce from the pending transaction pool, we still need to fetch the account's balance. I've moved this to be _after_ we find the nonce so that finding the nonce is all synchronous.

src/chains/ethereum/ethereum/tests/transaction-pool.test.ts
```diff
@@ -95,421 +98,795 @@ describe("transaction pool", async () => {
...
+      const { pending } = txPool.executables;
+      // our executable transaction should be found in the pending queue after the drain event
+      const found = findIn(transaction.hash.toBuffer(), pending);
+      assert.strictEqual(
+        found.serialized.toString(),
+        transaction.serialized.toString()
+      );
+    });
+
+    /**
```
MicaiahReid, line 519: This "in series" and the next "in parallel" `it`s are new - review these.

PR #3527, opened 2022-08-11
fix: return exactly 32 bytes from `eth_getStorageAt`

### LINKED ISSUE #3526: Leading zeros are trimmed in getStorageAt result
Here is the example of request made to a normal Ethereum client:

```shell
 curl -X POST --data '{"jsonrpc":"2.0","method": "eth_getStorageAt","params": ["0xC62ad447cb9Ccb9875935B438A1520662a7B863a","0x123123123","latest"],"id": 53}' http://127.0.0.1:8545

{
  "jsonrpc" : "2.0",
  "id" : 53,
  "result" : "0x0000000000000000000000000000000000000000000000000000000000000000"
}
```

And here is request made to a ganache client:

```shell
$ curl -X POST --data '{"jsonrpc":"2.0","method": "eth_getStorageAt","params": ["0xC62ad447cb9Ccb9875935B438A1520662a7B863a","0x123123123","latest"],"id": 53}' http://127.0.0.1:8555

{"id":53,"jsonrpc":"2.0","result":"0x"}
```

Leading zeros are trimmed, which should not happen.

Tested with version 7.4.0 of ganache.

davidmurdoch:
`eth_getStorageAt` was returning compact representations of data, e.g., `"0x5"` instead of `"0x0000000000000000000000000000000000000000000000000000000000000005"`. We now always return the padded whole 32 byte word.

Fixes #3526

### DIFF
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -1781,7 +1781,7 @@ export default class EthereumApi implements Api {
     const addressStateRoot = decode<EthereumRawAccount>(addressData)[2];
     trie.setContext(addressStateRoot, addressBuf, blockNum);
     const value = await trie.get(paddedPosBuff);
-    return Data.from(decode(value));
+    return Data.from(decode(value), 32);
   }
 
   /**
--- a/src/chains/ethereum/ethereum/tests/api/eth/getStorageAt.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/getStorageAt.test.ts
@@ -53,7 +53,10 @@ describe("api", () => {
           contractAddress,
           "0x1"
         ]);
-        assert.strictEqual(result2, "0x01");
+        assert.strictEqual(
+          result2,
+          "0x0000000000000000000000000000000000000000000000000000000000000001"
+        );
       });
 
       it("returns the value at the 32-byte hex position", async () => {
@@ -66,7 +69,10 @@ describe("api", () => {
           contractAddress,
           "0x" + THIRTY_TWO_BYES.slice(-1) + "1"
         ]);
-        assert.strictEqual(result2, "0x01");
+        assert.strictEqual(
+          result2,
+          "0x0000000000000000000000000000000000000000000000000000000000000001"
+        );
       });
 
       it("returns the value even when hex positions exceeds 32-bytes", async () => {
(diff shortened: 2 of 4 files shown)


MicaiahReid (approved):

src/chains/ethereum/ethereum/tests/api/eth/getStorageAt.test.ts
```diff
@@ -66,7 +69,10 @@ describe("api", () => {
           contractAddress,
           "0x" + THIRTY_TWO_BYES.slice(-1) + "1"
         ]);
-        assert.strictEqual(result2, "0x01");
+        assert.strictEqual(
+          result2,
+          "0x0000000000000000000000000000000000000000000000000000000000000001"
```
jeffsmale90, line 73: I don't like that sometimes we have a big long string literal and sometimes we construct it from "0x" and `THIRTY_TWO_BYES` (granted, one is an argument, and one is an assert).

In tests, my personal preference is probably the string literal, even though sometimes it seems silly, it's less prone to error.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1781,7 +1781,7 @@ export default class EthereumApi implements Api {
     const addressStateRoot = decode<EthereumRawAccount>(addressData)[2];
     trie.setContext(addressStateRoot, addressBuf, blockNum);
     const value = await trie.get(paddedPosBuff);
-    return Data.from(decode(value));
+    return Data.from(decode(value), 32);
```
jeffsmale90, line 1784: I made the same comment on @MicaiahReid 's review - wonder if we should encode the word length as a constant somewhere 🤷 ?

Obviously not worth holding the review up for this - keen to hear your opinions if you have some?

jeffsmale90 (changes requested):

tenthirtyone (approved):
Agree with @jeffsmale90 about a `const` for word length somewhere.

MicaiahReid:
> Agree with @jeffsmale90 about a `const` for word length somewhere.

By the way, I've made issue #3542 for this so we don't have to hold this up.

src/chains/ethereum/ethereum/tests/api/eth/getStorageAt.test.ts
davidmurdoch: fixed in https://github.com/trufflesuite/ganache/pull/3527/commits/8adeb549d6e25afe1bb4455d05168b09f9bd8e71 !

jeffsmale90 (approved):
Boom!

PR #3531, opened 2022-08-12
fix: calling `evm_mine` with a `timestamp` argument should reflect the change of time in subsequent blocks

### LINKED ISSUE #3265: calling `evm_mine` with a `timestamp` does not persist time offset

jeffsmale90:
Previously, calling `evm_mine` with a `timestamp` argument would result in a block with the specified timestamp, but subsequent blocks would have a timestamp that didn't reflect this change in time. This only occurred when `miner.timestampIncrement` is unspecified or `clock`. 

```javascript
const provider = ganache.provider();

await provider.request({ method: "evm_mine", params: ["0x1"] });
const { timestamp: firstTimestamp } = await provider.request({
  method: "eth_getBlockByNumber",
  params: ["latest", false],
});

// wait 1 second before mining the second block
await new Promise((resolve, reject) => setTimeout(resolve, 1000));

await provider.request({ method: "evm_mine", params: [] });
const { timestamp: secondTimestamp } = await provider.request({
  method: "eth_getBlockByNumber",
  params: ["latest", false],
});

console.log({
  firstTimestamp,
  secondTimestamp,
});
```

Will output something like:
```
{ firstTimestamp: '0x1', secondTimestamp: '0x5e0be0ff' }
```

Where `secondTimestamp` is the current time in seconds, but should be `0x2`.

With this change, blocks mined after providing a `timestamp` parameter to `evm_mine`, will have timestamps that reflect the change in time.

Fixes: #3265


src/chains/ethereum/ethereum/tests/blockchain.test.ts
```diff
@@ -112,7 +112,12 @@ describe("blockchain", async () => {
       // assert that second block's timestamp is at least `blockTime` greater
       // than the first block's. meaning, these blocks weren't mined one after
       // the other
-      assert(timestamps[1] >= timestamps[0] + blockTime);
+      assert(
+        timestamps[1] >= timestamps[0] + blockTime,
+        `Unexpected timestamp - expected >= ${timestamps[0] + blockTime}, got ${
```
jeffsmale90, line 117: I updated this message as I was debugging the change, and the error message wasn't helpful

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -833,7 +841,8 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
   }
 
   /**
-   * @param newTime - the number of milliseconds to adjust the time by. Can be negative.
```
jeffsmale90, line 842: I found this comment misleading - happy to hear thoughts on whether my change is better or not.

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -141,6 +141,84 @@ describe("provider", () => {
...
+
+        assert.strictEqual(
+          specifiedBlock,
+          `0x${timeArgumentSeconds.toString(16)}`,
+          "Unexpected timestamp for block mined with specified timestamp"
+        );
+
+        assert(
+          +unspecifiedBlock >= timeArgumentSeconds &&
+            +unspecifiedBlock <= timeArgumentSeconds + 1000,
```
MicaiahReid: If CI is running slowly, could this break? I don't know what the solution is, and I don't feel it's worth holding up this PR for, but just wanted to throw it out there in case you have a solution on-hand.

MicaiahReid (approved):

src/chains/ethereum/ethereum/tests/provider.test.ts
jeffsmale90: It _could_ do. Because we are specifying a timestamp in seconds, it would need to take an entire second to mine and fetch 2 blocks.

I feel like I'm building a lot of pressure on the time refactor - but it'll let us inject a `now` provider, which allow us to be much more resilient to slow build agents.

🤞

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -360,7 +360,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         const mineAll = (maxTransactions: Capacity, onlyOneBlock = false) =>
           this.#isPaused()
             ? nullResolved
-            : this.mine(maxTransactions, null, onlyOneBlock);
+            : this.mine(maxTransactions, undefined, onlyOneBlock);
```
davidmurdoch: Passing `undefined` means we have defined the value as `undefined`, and thus it is defined, with the value `undefined`. I think that is what `null` is for :-) Thoughts?

Also, I think passing `undefined` like this will not be allowed once we enable typescript's "strict" mode (https://github.com/trufflesuite/ganache/issues/3555).

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: As discussed, this is probably up for a refinement of the `blockchain` interface. In the meantime, I've reordered the arguments to avoid needing to pass `undefined` or `null` explicitly.

(I also named the arguments within `evm_mine()` in api.ts because I found them confusing - open to dissenting opinions though!)

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -608,6 +608,15 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
       onlyOneBlock
     );
     await this.#blockBeingSavedPromise;
+
+    if (
+      timestamp !== undefined &&
+      this.#options.miner.timestampIncrement === "clock"
```
davidmurdoch: There are two time-related blocks of code in this function that both check `this.#options.miner.timestampIncrement === "clock"`. Can these two blocks be joined together? If not, can you cache `this.#options.miner.timestampIncrement === "clock"` in a variable so we don't do it twice?

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -141,6 +141,83 @@ describe("provider", () => {
       await provider.disconnect();
     });
 
+    describe("uses timestamp adjustment in subsequent blocks after calling `evm_mine` with a `timestamp` argument", () => {
```
davidmurdoch: Our `describe`s are usually simple words or phrases like "api", "eth_chainId", "server". This is worded more like an `it`.

Can you rephrase the `describe` and `it`s here to be more inline with our other tests?

We use `describe` to define _structure_ to our tests, so maybe a bunch of these tests in this file should already be under `describe("timestampIncrement", ...)` (don't need to make a change like this in this same PR, of course), and then you could have another `describe` under that one like `describe("`evm_mine` with `timestamp`", ...)` (just an example -- I didn't think it through very much).

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -141,6 +141,83 @@ describe("provider", () => {
...
+        await provider.request({
+          method: "evm_mine",
+          params: []
+        });
+        const unspecifiedBlock = await provider.request({
+          method: "eth_getBlockByNumber",
+          params: ["latest", false]
+        });
+
+        return [specifiedBlock.timestamp, unspecifiedBlock.timestamp];
```
davidmurdoch: If you coerce these to numbers here it'd make the tests slightly simpler.

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -141,6 +141,83 @@ describe("provider", () => {
...
+            timeArgumentSeconds + 1000
+          }, got ${+unspecifiedBlock}`
+        );
+
+        await provider.disconnect();
+      });
+
+      it("should work with a numeric `timestampIncrement`", async () => {
+        const provider = await getProvider({
+          miner: { timestampIncrement }
```
davidmurdoch: Since this `timestampIncrement` is only used in this function is there a reason to define it in the `describe`?

davidmurdoch (changes requested):

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: Good call!

src/chains/ethereum/ethereum/tests/provider.test.ts
jeffsmale90: I intentionally didn't do this, thinking that we should be asserting the _exact_ returned value. Otherwise, there's a potential issue that could slip through where the return type isn't what's expected, but _coerces to_ what's expected.

Thinking about it in light of your comment, I think that we _should_ be ensuring the above, but maybe in a specific test for that.

I've updated my tests to what you've suggested, keen to hear your thoughts on that.

src/chains/ethereum/ethereum/tests/provider.test.ts
jeffsmale90: I just flattened that describe down, and was more explicit in the test names. lmk what you think.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -357,10 +357,10 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
 
         //#region automatic mining
         const nullResolved = Promise.resolve(null);
-        const mineAll = (maxTransactions: Capacity, onlyOneBlock = false) =>
+        const mineAll = (maxTransactions: Capacity, onlyOneBlock?: boolean) =>
```
davidmurdoch, line 360: A decent alternative would be to change `mineAll` to `(maxTransactions: Capacity, onlyOneBlock: boolean) => `

and then all use of `mineAll` would need to pass the values it needs, basically just:

```javascript
txPool.on("drain", () => mineAll(Capacity.Single, false));
```

This might make this code a bit clearer. Or not. I don't think I care either way.

davidmurdoch (approved):
Approved! (with one half-hearted suggestion/thought)

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: I left it as is because we set a default on the `mine` function, and don't feel that requiring it for `mineAll`, but not for `mine` feels nice - we might as well leverage the default that we have specified.

PR #3547, opened 2022-08-16
ci: remove docker from automated release workflow

### LINKED ISSUE #3546: remove docker publish from automated release workflow
Publishing to docker is the most unreliable part of our release workflow. If it fails, the whole release fails. Currently, this means:
 - we have a red X next to that commit in our history
 - we don't ever merge `master` back into `develop` so we have to run that part of the release manually
 - the release takes way longer than it needs to

MicaiahReid:
Previously our automated release process would require successfully publishing to Docker to complete. Since publishing to Docker can sometimes be unreliable, this would occasionally cause the release to fail, meaning we'd have to manually fix our release process' commit history and publish to Docker.

This change moves the step to publish to Docker to its own action, which is run only if the actual release is fully successful. Because this is now a separate job in the release process, we can rerun that individual job if it fails.

Fixes #3546.


MicaiahReid:
Notes for reviewers:

CI stuff is always weird to test. Here were the steps that I took.
### Testing the new action
I created a temporary `test-publish-action.yml` (removed in e7929ef2659e0e1bfe0ec9bd016b3d0ede94cf4f) that would run the new action every time I pushed to the branch. With this, I was able to test calling the action and making sure I was passing/using the inputs to the action correctly. With some finagling, I was able to call the Docker publish action succesffully.

Since we don't want to _actually_ publish to Docker, I didn't let the step complete. However, I let it run deep into the process before cancelling. This run can be found [here](https://github.com/trufflesuite/ganache/runs/7867654078?check_suite_focus=true).

### Testing dependent jobs
We also want to be sure that the new release action still works, and that the action will successfully be called when the release completes. If you look at the diff of the `release.yml`, I've mainly removed stuff and added the dependent `publish-docker` job.

To test the dependency, I commented out most of the meaningful code in the `release` job, and the actual calling of the new action in `publish-docker` job (calling this job was copied verbatim from my above test using `test-publish-action.yml`. All that should happen with these commented out jobs is: 1. `release` runs `npm ci` 2. once the `release` is finished, `publish-docker` would checkout the branch. This can be seen working [here](https://github.com/trufflesuite/ganache/actions/runs/2882887531). Note that the `publish-docker` job can be rerun independently if it fails.

### Summary
My goal with the above tests was to prove that the new action can successfully publish to docker, that we are correctly calling this new action from the `release` job, and that the `release` job no longer relies on the `publish-docker` job to finish the release.

Let me know if there are more tests you'd like to see to feel comfortable with this.

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
+name: "Docker Publish"
+description: "Builds and publishes a Docker image for the Ganache repo."
+inputs:
+  VERSION:
+    description: "The release version to be tagged to the Docker image, e.g. vX.x.x"
```
davidmurdoch: `vX.x.x` would be incorrect and result in a docker tag with `vvX.x.x` getting published. I think it should be just `X.x.x`.
```suggestion
    description: "The release version to be tagged to the Docker image, e.g., 4.0.1"
```

p.s., e.g., means "for example"; I think the example value should be a valid example.

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    required: true
+  TAG:
+    description: "The tag to be tagged to the Docker image, e.g. latest."
```
davidmurdoch:
```suggestion
    description: "The tag to be tagged to the Docker image, e.g., latest."
```

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    required: true
+  DOCKER_USERNAME:
+    description: "Username to log into Docker hub."
```
davidmurdoch:
```suggestion
    description: "Username to login to Docker hub."
```

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    required: true
+  DOCKER_ACCESS_TOKEN:
+    description: "Access token to log into Docker hub."
```
davidmurdoch:
```suggestion
    description: "Access token to login to Docker hub."
```

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    required: true
+  GITHUB_TOKEN:
+    description: "Github token used to log into the container registry."
```
davidmurdoch:
```suggestion
    description: "Github token used to login to the container registry."
```

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    required: true
+  INFURA_KEY:
+    description: "Infura key used to build the Docker image."
```
davidmurdoch:
```suggestion
    description: "Infura key used to build Ganache within Docker image."
```

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    required: true
+runs:
+  # this means we are using yml (rather than js or docker) for our action
```
davidmurdoch:
```suggestion
  # "composite" means we are using yml (rather than js or docker) for our action
```

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+  using: "composite"
+  steps:
+    - name: Use npm 14
+      uses: actions/setup-node@v1
+      with:
+        node-version: 14
+
+    - name: Run installation
+      run: npm ci
+      shell: bash
```
davidmurdoch, line 34: Do we need to `npm ci` here , outside of Docker, since we also `npm ci` within Docker?

davidmurdoch (changes requested):
<3

.github/actions/docker-publish/action.yml
MicaiahReid: Yes. Why? I don't know.

.github/actions/docker-publish/action.yml
jeffsmale90:
```suggestion
    description: "Infura key used to build Ganache within the Docker image."
```

jeffsmale90 (approved):

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+description: "Builds and publishes a Docker image for the Ganache repo."
+inputs:
+  VERSION:
+    description: "The release version to be tagged to the Docker image, e.g., 4.0.1"
+    required: true
+  TAG:
+    description: "The tag to be tagged to the Docker image, e.g., latest."
+    required: true
+  DOCKER_USERNAME:
+    description: "Username to login to Docker hub."
```
davidmurdoch: sorry.

```suggestion
    description: "Username to log in to Docker hub."
```

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    required: true
+  DOCKER_ACCESS_TOKEN:
+    description: "Access token to login to Docker hub."
```
davidmurdoch: sorry.
```suggestion
    description: "Access token to log in to Docker hub."
```

.github/actions/docker-publish/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    required: true
+  GITHUB_TOKEN:
+    description: "Github token used to login to the container registry."
```
davidmurdoch: sorry.

```suggestion
    description: "Github token used to log in to the container registry."
```

davidmurdoch (changes requested):
lol

davidmurdoch (approved):
yay

PR #3600, opened 2022-09-01
fix: implement backoff retry policy for websocket handler

### LINKED ISSUE #3476: websocket fails to reconnect after disconnect because `this.connection` is closed remotely
The new'd WebSocket is passed to `this.connect` to get its handlers wired up. `this.connection`'s `onClose` attempts to reconnect by passing `this.connection` back to `this.connect`.

### LINKED ISSUE #3477: implement backoff in ws-handler `onClose`
Implement an exponential backoff for reconnection attemps.

satyajeetkolhapure:
Web socket handler receives closed connection event more frequently than expected from infura server. So to make ganache more robust a backoff retry mechanism is implemented.

Retry configuration - 3 retries with 2 seconds initial retry and with exponential attempts followed.

When connection is closed, in flight requests are stored temporarily and processed once connection retry attempt succeeds.

Note: Once connection is successful, the retry counter is reset to initial counter.


src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
```diff
@@ -46,10 +52,21 @@ export class WsHandler extends BaseHandler implements Handler {
     this.open = this.connect(this.connection, logging);
     this.connection.onclose = () => {
       // try to connect again...
-      // Issue: https://github.com/trufflesuite/ganache/issues/3476
-      // TODO: backoff and eventually fail
-      // Issue: https://github.com/trufflesuite/ganache/issues/3477
-      this.open = this.connect(this.connection, logging);
+      // backoff and eventually fail
+      if( this.retryCounter > 0 || this.retryCounter == -1 ) {
```
tenthirtyone: Just a suggestion, you might be able to reduce the variables/closures if something like this works for you:

``` javascript
const expBackoff = (fn, exponent = 0) => {
  setTimeout(() => {
    if (exponent > 8) return;
    if (!fn()) expBackoff(fn, exponent + 1);
  }, Math.max(1, 2 ** exponent) * 1000);
};
```

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
```diff
@@ -46,10 +52,21 @@ export class WsHandler extends BaseHandler implements Handler {
     this.open = this.connect(this.connection, logging);
     this.connection.onclose = () => {
```
tenthirtyone: This will run on SIGINT. When a user `ctrl+c`'s out the retry will run once. To test/verify:

``` javascript
this.retryTimeoutId = setTimeout(
          () => console.log("I ran"),
          Math.pow(
            this.retryIntervalBase,
            this.initialRetryCounter - this.retryCounter
          ) * 1000
        );
```

ER:
``` bash
RPC Listening on 127.0.0.1:8645
^C
Received shutdown signal: SIGINT
Shutting down…

Received shutdown signal: SIGINT
Server has been shut down

$ I ran      
```

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
```diff
@@ -46,10 +52,21 @@ export class WsHandler extends BaseHandler implements Handler {
...
-      // TODO: backoff and eventually fail
-      // Issue: https://github.com/trufflesuite/ganache/issues/3477
-      this.open = this.connect(this.connection, logging);
+      // backoff and eventually fail
+      if( this.retryCounter > 0 || this.retryCounter == -1 ) {
+        if( this.retryCounter !== -1 ) this.retryCounter--;
+        clearTimeout( this.retryTimeoutId );
+        this.retryTimeoutId = setTimeout( () => {
+          const onCloseEvent = this.connection.onclose;
+          this.connection = new WebSocket(url.toString(), {
```
tenthirtyone: Just for conversation - Would it make sense to move the constructor's `new Websocket...` into `this.connect`? 

`options` does not immediately appear to be available after the constructor finishes. E.g. no `this.options` so maybe the `url` and `origin` could be plucked out of the existing, closed, `this.connection`?  

Then it might just be `this.connect(url, origin, logging)`

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
```diff
@@ -46,10 +52,21 @@ export class WsHandler extends BaseHandler implements Handler {
...
+            origin,
+            headers: this.headers
+          });
+          this.connection.binaryType = "nodebuffer";
+          this.connection.onclose = onCloseEvent;
+          this.open = this.connect(this.connection, logging);
```
tenthirtyone: There is also a `inFlightRequests` that may need care/attention here. I am not sure if the desired behavior would be to kill/cancel them or attempt to rebroadcast if/once the socket reconnects and if that's possible. I don't know this part of code. I would assume the latter but @davidmurdoch will need to clarify

tenthirtyone (changes requested):

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
satyajeetkolhapure: I agree. it will be simpler and also as name suggests connect will actually initiate the connection.

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
satyajeetkolhapure: Initially I had similar function 

```
const exponentialBackoff = async (fn, depth = 0) => {
    try {
      return await fn();
    }catch(e) {
      if (depth > 7) {
        throw e;
      }
      await wait(2 ** depth * 10);
    
      return exponentialBackoff(fn, depth + 1);
    }
  }
```

But with this it became more complicated. I will try again when I move the web socket instantiation inside connect method.

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
satyajeetkolhapure: On every request the connect is called to verify if connection is still active. So cant really move `new websocket...` into `this.connect`.

satyajeetkolhapure: I have moved `new websocket...` into `this.connect`. Also I added reconnect method.

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
satyajeetkolhapure: Hi @davidmurdoch : Will this change have any impact on inFlightRequests?

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
satyajeetkolhapure: Yes tested it. It works as expected

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
satyajeetkolhapure: I have made few changes. I have moved some code in methods like reconnect, connect etc. Can you have a look again?

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
```diff
@@ -20,6 +20,12 @@ export class WsHandler extends BaseHandler implements Handler {
     }>
   >();
 
+  // retry configuration
+  private retryIntervalBase: number = 2;
+  private retryCounter: number = 3;
+  private initialRetryCounter = this.retryCounter;
```
cds-amal: What's the advantage of using `this` on the RHS, vs the same value already in scope?

cds-amal (commented):
Not a review, I just had a question.

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
satyajeetkolhapure: do you mean instead of using retry counter why not directly use 3?

if I have understood the question correctly - Actually at the start I am setting the retry counter which is used as variable and on every attempt it is reduced by 1. To compare it with initial retry counter I added one more property initial retry counter. It can be implemented with different logic as well. Like using only one retry counter and starting with 1 check against retry counter but there will little bit more code to add one more variable counter. 

So initial condition is to have both retry counter and initial retry counter same. If we decide to change the value, it will be only at one place.

cds-amal: Thanks! It took a minute to reason about `this` outside the constructor. Those declarations will be relocated into the constructor during transpilation.

MicaiahReid:
Would it be possible to get tests added for this change?

satyajeetkolhapure:
> Would it be possible to get tests added for this change?
Yes. I am going to add few tests to check InFlightRequests too.

src/chains/ethereum/ethereum/src/forking/handlers/ws-handler.ts
satyajeetkolhapure: @tenthirtyone , @davidmurdoch : In flight requests are now stored when connection is closed and processed once connection retry attempt succeeds

PR #3619, opened 2022-09-06
ci: fix docker publish

MicaiahReid:
This code was tested as an alpha release. The run can be seen here: https://github.com/trufflesuite/ganache/runs/8215680896?check_suite_focus=true

The problem with the original PR (#3547) that this fixes was that the environment variables are not persisted from job to job. This fix saves those environment variables that are needed in the subsequent docker publish job.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,15 +42,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Run installation
-        run: npm ci
-
-      - name: Test
-        run: npm test
-        env:
-          FORCE_COLOR: 1
-          INFURA_KEY: ${{ secrets.TEST_INFURA_KEY }}
-
       - name: Set TAG for master to latest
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -94,6 +92,15 @@ jobs:
...
 
+      - name: Run installation
+        run: npm ci
+
+      - name: Test
+        run: npm test
+        env:
+          FORCE_COLOR: 1
+          INFURA_KEY: ${{ secrets.TEST_INFURA_KEY }}
+
```
davidmurdoch, line 103: What was the reason for moving these?

MicaiahReid: Peace of mind! The test step takes a few minutes, and everything after it moves pretty quick. It's nice to be able to confirm the output of the previous steps (such as the `TAG` and that the release is correctly registered as pre-release). We've now confirmed that the pre-release stuff works for alpha, so it's a bit less of a concern, but we've still never done a beta or rc release so it will be nice for when we run those for the first time.

davidmurdoch (approved):

jeffsmale90 (approved):

PR #3534, opened 2022-08-12
chore: set default hardfork to `grayGlacier`; update `@ethereumjs/*` packages

davidmurdoch:
This changes the default hardfork to `grayGlacier` (was previously `london`). This change also updates `@ethereumjs/*` packages to their latest versions, which is what allows us to add the `grayGlacier` hardfork (`ganache --hardfork grayGlacier`). This hardfork does not change ganache behavior, other than make the hardfork name available.

PR #3629, opened 2022-09-07
fix: set default hardfork to london

MicaiahReid:
In #3534, we updated Ganache to use Gray Glacier as the default hardfork. This broke forking testnets using Ganache because the test nets do not support Gray Glacier. This PR keeps support for Gray Glacier, but reverts the default hardfork to London so that forking testnets will work out of the box.


davidmurdoch (approved):
Approved so in not a blocker.

tenthirtyone (approved):
Pulled down, tested `truffle test` for metacoin against sepolia fork

PR #3656, opened 2022-09-14
chore: upgrade @ethereumjs/vm to v6.0.0

### LINKED ISSUE #3473: gas could theoretically go over `NUMBER.MAX_SAFE_INTEGER` in `blockchain.traceTransaction`
Linking [TODO](https://github.com/trufflesuite/ganache/blob/develop/src/chains/ethereum/ethereum/src/blockchain.ts#L1210) to issue.

If blocks grow large enough, `gasUsed` could exceed `NUMBER.MAX_SAFE_INTEGER`

### LINKED ISSUE #3578: Rename `Account.stateRoot` to `storageRoot`
See: https://github.com/trufflesuite/ganache/blob/591d0256975dc80bf2cec90ad8be67704ab4556b/src/chains/ethereum/utils/src/things/account.ts#L18

`Account`'s `storageRoot` property is incorrectly named `stateRoot`. 

This probably has flow on impacts, as references to `Account` (or references to `RLP.decode()`) will likely use the same nomenclature.

MicaiahReid:
Fixes #3473 
Fixes #3578


src/chains/ethereum/ethereum/src/leveldown-to-level.ts
```diff
@@ -0,0 +1,69 @@
...
+const levelup = require("levelup");
+
+type GanacheLevelDown = AbstractLevelDOWN<Buffer, Buffer>;
+export class UpgradedLevelDown {
+  public db: LevelUp<GanacheLevelDown> = null;
+  #rootStore: GanacheLevelDown;
+
+  constructor(store: GanacheLevelDown) {
+    this.#rootStore = store;
+    this.db = levelup(store, LEVEL_OPTIONS);
```
MicaiahReid, line 15: We need to return an actual location

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1127,30 +1148,28 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
-      fromAccount.nonce.iaddn(1);
-      const txCost = new BN(
-        (gasLimit * transaction.gasPrice.toBigInt()).toString()
-      );
-      fromAccount.balance.isub(txCost);
-      await vm.stateManager.putAccount({ buf: caller } as any, fromAccount);
+      fromAccount.nonce += 1n;
+      const txCost = gasLimit * transaction.gasPrice.toBigInt();
+      const startBalance = fromAccount.balance;
+      // TODO: should we throw if insufficient funds?
```
MicaiahReid, line 1176: In previous ejs versions, if we subtracted from a balance such that the result would be negative, they just parsed it as a positive number. So we never had to worry about a negative balance. Now, however, if we pass on a negative number ejs will throw. 

If we run the transaction with a zero balance, the vm will throw with insufficient funds anyways, so I think we can possibly remove this todo.

src/chains/ethereum/ethereum/src/leveldown-to-level.ts
davidmurdoch: is this comment in the wrong place?

src/chains/ethereum/ethereum/src/leveldown-to-level.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  get location() {
+    console.log("LOCATION RETRIEVED");
```
MicaiahReid: We need to return an actual location

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -394,19 +398,21 @@ export default class EthereumApi implements Api {
...
+    const eei = blockchain.vm.eei;
+    const account = await eei.getAccount(ejsAddress);
 
-    account.balance = {
-      toArrayLike: () => Quantity.toBuffer(balance)
-    } as any;
+    account.balance = Quantity.toBigInt(balance);
 
-    await stateManager.putAccount({ buf: buffer } as any, account);
+    await eei.checkpoint();
```
jeffsmale90, line 412: Does this mean that we are now required to checkpoint and commit any changes made to the vmstate?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -136,33 +134,45 @@ export type BlockchainOptions = {
  * @param stateRoot -
  */
 function setStateRootSync(stateManager: StateManager, stateRoot: Buffer) {
-  (stateManager as any)._trie.root = stateRoot;
-  (stateManager as any)._cache.clear();
-  (stateManager as any)._storageTries = {};
+  (stateManager as DefaultStateManager)._trie.root(stateRoot);
```
jeffsmale90: nit: Can we do the type assertion just once here?

src/chains/ethereum/ethereum/src/helpers/precompiles.ts
```diff
@@ -31,26 +28,26 @@ const makeAccount = (i: number): Address => {
 
 /**
  * Puts the precompile accounts into the state tree
- * @param stateManager -
+ * @param eei -
  */
-export const activatePrecompiles = async (stateManager: StateManager) => {
-  await stateManager.checkpoint();
-  const cache = (stateManager as any)._cache;
+export const activatePrecompiles = async (eei: EEI) => {
```
jeffsmale90: Should eei be of type `EEIInterface` rather than `EEI`? This would allow us to remove a bunch of type assertions when calling this function - `vm.eei as EEI`. (but not the `eei as any` assertion _in_ this function sadly :( )

src/chains/ethereum/ethereum/src/helpers/precompiles.ts
```diff
@@ -31,26 +28,26 @@ const makeAccount = (i: number): Address => {
...
+  await eei.commit();
 };
 
 /**
  * Puts the precompile accounts into the warmed addresses
- * @param stateManager -
+ * @param eei -
  */
-export const warmPrecompiles = async (stateManager: DefaultStateManager) => {
+export const warmPrecompiles = async (eei: EEI) => {
```
jeffsmale90: Same as above - should the parameter be of type `EEIInterface`?

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: I guess the upside for throwing here is it's less work for the case where funds are insufficient, but it's more complexity, and more work for the common case where funds _are_ sufficient, so I agree with your suggestion.

src/chains/ethereum/ethereum/src/helpers/run-call.ts
```diff
@@ -68,28 +66,23 @@ export function runCall(
...
 ) {
-  const caller = { buf: transaction.from.toBuffer() };
+  const caller = { buf: transaction.from.toBuffer() } as EthereumJsAddress;
   const to =
-    transaction.to == null ? undefined : { buf: transaction.to.toBuffer() };
-  const value = new BN(
-    transaction.value == null ? 0 : transaction.value.toBuffer()
-  );
+    transaction.to == null
+      ? undefined
```
jeffsmale90, line 72: Should this be `null` rather than `undefined`? As I understand `undefined` should be implicit, rather than explicitly assigned.

Within `EVM.runCall` it's coercing `transaction.to` to boolean to check whether it's set, so behaviour should be the same.

src/chains/ethereum/ethereum/src/helpers/run-call.ts
```diff
@@ -128,29 +121,23 @@ export async function applySimulationOverrides(
...
-        } as any;
+        account.nonce = Quantity.toBigInt(nonce === "" ? "0x0" : nonce);
       }
       if (balance != null) {
-        account.balance = {
-          toArrayLike: () =>
-            // geth treats empty strings as "0x0" balance for overrides
-            balance === "" ? BUFFER_EMPTY : Quantity.toBuffer(balance)
-        } as any;
+        account.balance = Quantity.toBigInt(balance === "" ? "0x0" : balance);
```
jeffsmale90:
```suggestion
        account.balance = balance === "" ? 0n : Quantity.toBigInt(balance);
```
skips the string=>buffer=>bigint conversion of "0x0"

src/chains/ethereum/ethereum/src/leveldb.ts
```diff
@@ -0,0 +1,49 @@
...
+import type { AbstractLevel } from "abstract-level";
+
+const ENCODING_OPTS = { keyEncoding: "binary", valueEncoding: "binary" };
+
+export type EJSLevel = AbstractLevel<
+  string | Buffer | Uint8Array,
+  string | Buffer,
+  string | Buffer
+>;
+export class LevelDB implements DB {
```
jeffsmale90, line 13: I think it would be worthwhile to explain why we have to wrap DB here.

src/chains/ethereum/ethereum/src/leveldown-to-level.ts
```diff
@@ -0,0 +1,61 @@
+import type { AbstractLevelDOWN } from "abstract-leveldown";
+import type { BatchDBOp } from "@ethereumjs/trie";
+import sub from "subleveldown";
+import type { LevelUp } from "levelup";
+import { LEVEL_OPTIONS } from "./database";
+const levelup = require("levelup");
+
+export type GanacheLevelDown = AbstractLevelDOWN<Buffer, Buffer>;
+export class UpgradedLevelDown {
```
jeffsmale90, line 9: A comment explaining why `UpgradedLevelDown` would be worthwhile.

Level/Up/Down is confusing _anyways_ and we're having to add another layer of complexity :/

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -19,14 +19,14 @@ import {
   EthereumRawAccount,
   TraceTransactionResult
 } from "@ganache/ethereum-utils";
-import type { Address as EthereumJsAddress } from "ethereumjs-util";
-import type { InterpreterStep } from "@ethereumjs/vm/dist/evm/interpreter";
+import type { Address as EthereumJsAddress } from "@ethereumjs/util";
+import type { InterpreterStep } from "@ethereumjs/evm/";
```
jeffsmale90: There's a couple of different imports here that are importing from the same place (a couple of which are `import` vs `import type`). Can we collapse those down?

Also there's `"@ethereumjs/evm/"` and `"@ethereumjs/evm"` - are those the same?

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -40,13 +40,13 @@ export default class BlockManager extends Manager<Block> {
 
   #blockchain: Blockchain;
   #common: Common;
-  #blockIndexes: LevelUp;
+  #blockIndexes: GanacheSublevel;
```
jeffsmale90, line 43: I'm not super sure that we should define a dependency on a _sub_level - I feel that the consumer should just depend on a level (Ganache- or otherwise), and the fact that it's _sub_ should be irrelevant?

src/chains/ethereum/ethereum/src/data-managers/blocklog-manager.ts
```diff
@@ -8,7 +8,7 @@ import { Ethereum } from "../api-types";
 
 export default class BlockLogManager extends Manager<BlockLogs> {
   #blockchain: Blockchain;
-  constructor(base: LevelUp, blockchain: Blockchain) {
+  constructor(base: GanacheSublevel, blockchain: Blockchain) {
```
jeffsmale90, line 11: Similar comment above about depending explicitly on _Sub_level

src/chains/ethereum/ethereum/src/data-managers/manager.ts
```diff
@@ -7,9 +7,9 @@ export type Instantiable<T> = { new (...args: any[]): T };
 export default class Manager<T> {
   #Type: Instantiable<T>;
   #options: {};
-  protected base: LevelUp;
+  protected base: GanacheSublevel;
```
jeffsmale90, line 10: Similar comment above about depending explicitly on _Sub_level

src/chains/ethereum/ethereum/src/data-managers/transaction-manager.ts
```diff
@@ -34,7 +34,7 @@ export default class TransactionManager extends Manager<NoOp> {
     options: EthereumInternalOptions["miner"],
     common: Common,
     blockchain: Blockchain,
-    base: LevelUp
+    base: GanacheSublevel
```
jeffsmale90, line 37: Similar comment above about depending explicitly on _Sub_level

src/chains/ethereum/ethereum/src/data-managers/transaction-receipt-manager.ts
```diff
@@ -7,7 +7,7 @@ import { Address } from "@ganache/ethereum-address";
 
 export default class TransactionReceiptManager extends Manager<InternalTransactionReceipt> {
   #blockchain: Blockchain;
-  constructor(base: LevelUp, blockchain: Blockchain) {
+  constructor(base: GanacheSublevel, blockchain: Blockchain) {
```
jeffsmale90, line 10: Similar comment above about depending explicitly on _Sub_level

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
tenthirtyone: iirc sublevel is just a namespaced level

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -1,34 +1,51 @@
...
-  public db: LevelUp = null;
-  public blocks: LevelUp;
-  public blockIndexes: LevelUp;
-  public blockLogs: LevelUp;
-  public transactions: LevelUp;
-  public transactionReceipts: LevelUp;
-  public storageKeys: LevelUp;
-  public trie: LevelUp;
+  public db: GanacheLevel;
+  public blocks: GanacheSublevel;
```
jeffsmale90, line 40: Similar comment about _Sub_Level types here as above.

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -1,34 +1,51 @@
...
+  public blockIndexes: GanacheSublevel;
+  public blockLogs: GanacheSublevel;
+  public transactions: GanacheSublevel;
+  public transactionReceipts: GanacheSublevel;
+  public storageKeys: GanacheSublevel;
+  public trie: GanacheSublevel;
   public readonly initialized: boolean;
-  #rootStore: AbstractLevelDOWN;
+  public type: DBType = DBType.Level;
```
jeffsmale90, line 48: bit of a nit: Should we leave this unassigned until we call `initialise()`? Then we've got consistence assignment for both database types.

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -49,15 +66,15 @@ export default class Database extends Emittery {
...
-      valueEncoding: "binary"
-    };
     const store = this.#options.db;
-    let db: LevelUp;
+    let db: GanacheLevel;
     if (store) {
-      this.#rootStore = encode(store as AbstractLevelDOWN, levelupOptions);
-      db = levelup(this.#rootStore, {});
+      if (typeof store === "string") {
+        db = <GanacheLevel>new Level(store, LEVEL_OPTIONS);
```
jeffsmale90:
```suggestion
        db = new Level(store, LEVEL_OPTIONS) as GanacheLevel;
```
For consistencies sake

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -70,31 +87,28 @@ export default class Database extends Emittery {
       }
       this.directory = directory;
 
-      // specify an empty `prefix` for browser-based leveldown (level-js)
-      const leveldownOpts = { prefix: "" };
-      const store = encode(leveldown(directory, leveldownOpts), levelupOptions);
-      this.#rootStore = store;
-      db = levelup(store);
+      const store = <GanacheLevel>new Level(directory, LEVEL_OPTIONS);
```
jeffsmale90:
```suggestion
      const store = new Level(directory, LEVEL_OPTIONS) as GanacheLevel;
```
For consistencies sake

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -70,31 +87,28 @@ export default class Database extends Emittery {
...
+      db = store;
     }
 
     // don't continue if we closed while we were waiting for the db
     if (this.#closed) return this.#cleanup();
 
     const open = db.open();
-    this.trie = sub(db, "T", levelupOptions);
+    this.trie = <GanacheSublevel>db.sublevel("T", LEVEL_OPTIONS);
```
jeffsmale90:
```suggestion
    this.trie = db.sublevel("T", LEVEL_OPTIONS) as GanacheSublevel;
```

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -70,31 +87,28 @@ export default class Database extends Emittery {
...
-    this.blockLogs = sub(db, "l", levelupOptions);
-    this.transactions = sub(db, "t", levelupOptions);
-    this.transactionReceipts = sub(db, "r", levelupOptions);
-    this.storageKeys = sub(db, "s", levelupOptions);
+    this.blocks = <GanacheSublevel>db.sublevel("b", LEVEL_OPTIONS);
+    this.blockIndexes = <GanacheSublevel>db.sublevel("i", LEVEL_OPTIONS);
+    this.blockLogs = <GanacheSublevel>db.sublevel("l", LEVEL_OPTIONS);
+    this.transactions = <GanacheSublevel>db.sublevel("t", LEVEL_OPTIONS);
+    this.transactionReceipts = <GanacheSublevel>db.sublevel("r", LEVEL_OPTIONS);
+    this.storageKeys = <GanacheSublevel>db.sublevel("s", LEVEL_OPTIONS);
```
jeffsmale90:
```suggestion
    this.blocks = db.sublevel("b", LEVEL_OPTIONS) as GanacheSublevel;
    this.blockIndexes = db.sublevel("i", LEVEL_OPTIONS) as GanacheSublevel;
    this.blockLogs = db.sublevel("l", LEVEL_OPTIONS) as GanacheSublevel;
    this.transactions = db.sublevel("t", LEVEL_OPTIONS) as GanacheSublevel;
    this.transactionReceipts = db.sublevel("r", LEVEL_OPTIONS) as GanacheSublevel;
    this.storageKeys = db.sublevel("s", LEVEL_OPTIONS) as GanacheSublevel;
```
Multiline suggestion FTW! https://github.blog/changelog/2019-10-01-multi-line-references-and-comments-beta/

src/chains/ethereum/ethereum/src/data-managers/transaction-receipt-manager.ts
jeffsmale90: I assume this touches everything, so I'm going to stop commenting on it

src/chains/ethereum/ethereum/src/forking/persistent-cache/helpers.ts
```diff
@@ -56,11 +56,12 @@ export function getBlockNumberFromParams(method: string, params: any[]) {
   }
 }
 
-export async function setDbVersion(db: LevelUp, version: Buffer) {
+export async function setDbVersion(db: GanacheLevel, version: Buffer) {
   // set the version if the DB was just created, or error if we already have
   // a version, but it isn't what we expected
+  const VERSION = Buffer.from("version", "utf-8");
```
jeffsmale90: `VERSION_KEY`?

helps clarify with the different `version`, `VERSION`, and `recordedVersion` vars

src/chains/ethereum/ethereum/src/forking/state-manager.ts
```diff
@@ -1,23 +1,19 @@
...
   /**
-   * Parameters of the chain ([`Common`](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common))
-   */
-  common: Common;
-  /**
-   * An [`merkle-patricia-tree`](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/trie) instance
+   * An [`@ethereumjs/trie`](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/trie) instance
    */
   trie: ForkTrie;
+  prefixCodeHashes?: boolean;
```
jeffsmale90, line 21: A comment explaining what and why `prefixCodeHashes` might be worthwhile. Something like https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/statemanager/src/stateManager.ts#L40

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -38,34 +43,38 @@ export class ForkTrie extends GanacheTrie {
...
 
-  constructor(db: LevelUp | null, root: Buffer, blockchain: Blockchain) {
+  constructor(
+    db: LevelDB | UpgradedLevelDown | null,
+    root: Buffer,
+    blockchain: Blockchain
+  ) {
     super(db, root, blockchain);
-
+    this.db = db;
```
jeffsmale90, line 54: Instead of storing an additional reference to db, can we access the same instance via the checkpoint db at `this._db.db` ? Could wrap it in a getter property, or getter function.

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -38,34 +43,38 @@ export class ForkTrie extends GanacheTrie {
...
-
-  get root() {
-    return (this as any)._root;
+    // a LevelDB has a _leveldb property, otherwise this is an upgraded
+    // leveldown instance
+    const leveldb =
+      "_leveldb" in db
+        ? (db._leveldb as GanacheSublevel)
+        : // @ts-ignore
+          (db as GanacheSublevel);
```
jeffsmale90:
```suggestion
        : (db as unknown as GanacheSublevel);
```

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -94,15 +103,18 @@ export class ForkTrie extends GanacheTrie {
     startBlockNumber: Quantity,
     endBlockNumber: Quantity
   ) {
-    const db = this.metadata._leveldb;
-    const stream = db.createKeyStream({
+    //@ts-ignore
+    const db = this.metadata.db._leveldb;
```
jeffsmale90:
```suggestion
    const db = (this.metadata.db as any)._leveldb;
```
I prefer this, as @ts-ignore is very broad

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -139,12 +151,13 @@ export class ForkTrie extends GanacheTrie {
     // because it just checks every single key we've ever deleted (before this
     // one).
     // Issue: https://github.com/trufflesuite/ganache/issues/3484
-    const stream = this.metadata._leveldb.createReadStream({
+    // @ts-ignore
+    const db = this.metadata.db._leveldb;
```
jeffsmale90:
```suggestion
    const db = (this.metadata.db as any)._leveldb;
```

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -166,16 +179,21 @@ export class ForkTrie extends GanacheTrie {
...
+      }
 
       await metaDataPutPromise;
     } else {
       const hash = keccak(key);
       const { node, stack } = await this.findPath(hash);
-      if (node) await this._deleteNode(hash, stack);
+      if (node) {
+        await this._deleteNode(hash, stack);
+        await this.persistRoot();
```
jeffsmale90, line 193: Is it worth commenting on why we need to `persistRoot` explicitly?

src/chains/ethereum/ethereum/src/helpers/gas-estimator.ts
```diff
@@ -1,12 +1,36 @@
...
+import BN from "bn.js";
 import { RuntimeError, RETURN_TYPES } from "@ganache/ethereum-utils";
 import { Quantity } from "@ganache/utils";
+import { RunTxOpts, RunTxResult, VM } from "@ethereumjs/vm";
+import { EVM } from "@ethereumjs/evm";
+import type { InterpreterStep } from "@ethereumjs/evm/";
+import { RuntimeBlock } from "@ganache/ethereum-block";
 
-const bn = (val = 0) => new (BN as any)(val);
+const bn = (val = 0): BN => new BN(val);
```
jeffsmale90, line 8:
```suggestion
const bn = (val = 0) => new BN(val);
```
The return type is implicit and reasonably obvious with the BN ctor right there

src/chains/ethereum/ethereum/src/helpers/gas-estimator.ts
```diff
@@ -1,12 +1,36 @@
-import { BN } from "ethereumjs-util";
+import BN from "bn.js";
 import { RuntimeError, RETURN_TYPES } from "@ganache/ethereum-utils";
 import { Quantity } from "@ganache/utils";
+import { RunTxOpts, RunTxResult, VM } from "@ethereumjs/vm";
+import { EVM } from "@ethereumjs/evm";
```
jeffsmale90:
```suggestion
```

src/chains/ethereum/ethereum/src/helpers/gas-estimator.ts
```diff
@@ -1,12 +1,36 @@
...
+import type { InterpreterStep } from "@ethereumjs/evm/";
+import { RuntimeBlock } from "@ganache/ethereum-block";
 
-const bn = (val = 0) => new (BN as any)(val);
+const bn = (val = 0): BN => new BN(val);
 const STIPEND = bn(2300);
+export type EstimateGasRunArgs = {
+  tx: any;
```
jeffsmale90:
```suggestion
  tx: { gasLimit: bigint };
```

src/chains/ethereum/ethereum/src/helpers/gas-estimator.ts
```diff
@@ -1,12 +1,36 @@
...
+  block: RuntimeBlock;
+  skipBalance: boolean;
+  skipNonce: boolean;
+};
+
+export type EstimateGasResult =
+  | RunTxResult & {
+      gasEstimate?: bigint;
+    };
+// proof that I don't know what I'm doing:
```
jeffsmale90, line 21: ❤️

src/chains/ethereum/ethereum/src/provider-events.ts
```diff
@@ -37,27 +37,25 @@ function normalizeEvent(event: InterpreterStep) {
...
+    memoryWordCount: event.memoryWordCount,
     opcode: {
       name: opcode.name,
       fee: opcode.fee
     },
     pc: BigInt(event.pc),
-    returnStack: event.returnStack.map(rs => rs.toArrayLike(Buffer)),
-    stack: event.stack.map(s => s.toArrayLike(Buffer))
+    returnStack: event.returnStack.map(r => r),
+    stack: event.stack.map(s => s)
```
jeffsmale90, line 58:
```suggestion
    returnStack: event.returnStack,
    stack: event.stack
```

src/chains/ethereum/ethereum/tests/api/eth/call.test.ts
```diff
@@ -909,10 +910,13 @@ describe("api", () => {
...
 
         const getDbData = async (trie: GanacheTrie) => {
-          let dbData: (string | Buffer)[] = [];
-          for await (const data of trie.db._leveldb.createReadStream()) {
-            dbData.push(data);
-          }
+          const dbData: [Buffer, Buffer][] = [];
+          const stream = trie.createReadStream();
+          // @ts-ignore TODO: why is this necessary? seems like a bug on ejs' end
+          stream.on("data", ({ key, value }) => {
```
jeffsmale90, line 916: I suspect it's because of the conditional export in readable-stream is muddling the types. How about this in the meantime?
```suggestion
          (stream as any).on("data", ({ key, value }) => {
```

src/chains/ethereum/ethereum/tests/api/eth/sign.test.ts
```diff
@@ -40,14 +41,11 @@ describe("api", () => {
...
-        sgn = sgn.slice(2);
+        const { v, r, s } = fromRpcSig(sgn);
 
-        const r = Buffer.from(sgn.slice(0, 64), "hex");
-        const s = Buffer.from(sgn.slice(64, 128), "hex");
-        const v = parseInt(sgn.slice(128), 16);
         const pub = ecrecover(msgHash, v, r, s);
         const addr = fromSigned(pubToAddress(pub));
-        const strAddr = "0x" + addr.toString("hex");
+        const strAddr = Data.toString(Quantity.toBuffer(addr), 20);
```
jeffsmale90, line 48: That's a bit of a pain that Data doesn't accept the bigint arg, especially now that the type is being used more widely.

Maybe we should support bigint in Data ctor?

In the meantime, I was initially confused by why we were doing this - given it's a test, would it be clearer to do this?
```suggestion
        const strAddr = Data.toString(`0x${addr.toString(16)}`, 20);
```

src/chains/ethereum/ethereum/tests/api/eth/sign.test.ts
```diff
@@ -60,15 +58,12 @@ describe("api", () => {
...
 
-        const r = Buffer.from(sgn.slice(0, 64), "hex");
-        const s = Buffer.from(sgn.slice(64, 128), "hex");
-        const v = parseInt(sgn.slice(128), 16);
+        const { v, r, s } = fromRpcSig(sgn);
         const pub = ecrecover(msgHash, v, r, s);
         const addr = fromSigned(pubToAddress(pub));
-        const strAddr = "0x" + addr.toString("hex");
-        assert.deepStrictEqual(strAddr, accounts[0].toLowerCase());
+        const strAddr = Data.toString(Quantity.toBuffer(addr), 20);
```
jeffsmale90, line 65: Same as previous comment:
```suggestion
        const strAddr = Data.toString(`0x${addr.toString(16)}`, 20);
```

src/chains/ethereum/ethereum/tests/forking/forking.test.ts
```diff
@@ -1237,6 +1237,14 @@ describe("forking", function () {
...
     KNOWN_NETWORKS.forEach(network => {
       describe(network, () => {
+        // TODO: possibly reinstate if this
+        // https://github.com/ethereumjs/ethereumjs-monorepo/issues/2286
+        // gets resolved
+        before("skip kovan", function () {
+          if (network === "kovan") {
+            this.skip();
+          }
+        });
```
jeffsmale90: We decided to wait so that we don't need to support kovan right? So we can just drop this?
```suggestion
```

src/chains/ethereum/transaction/src/signing.ts
```diff
@@ -14,8 +14,8 @@ import { digest, encodeRange } from "@ganache/rlp";
 import { Address } from "@ganache/ethereum-address";
 import secp256k1 from "@ganache/secp256k1";
 
-const intToBuffer = (value: number) =>
-  value === 0 ? BUFFER_EMPTY : uintToBuffer(value);
+const _bigintToBuffer = (value: bigint) =>
+  value === 0n ? BUFFER_EMPTY : bigIntToBuffer(value);
```
jeffsmale90, line 18: Should we build this into `bigIntToBuffer`? Seems like a logical thing to do in all cases, not just this one.

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -381,7 +381,7 @@ describe("@ganache/ethereum-transaction", async () => {
         assert.notDeepStrictEqual(vmTx.hash().toString(), "");
       });
       it("has nonce property", () => {
-        assert.strictEqual(vmTx.nonce.toString(), "0");
+        assert("nonce" in vmTx);
```
jeffsmale90, line 384: We are no longer asserting the value of nonce here

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -456,7 +456,7 @@ describe("@ganache/ethereum-transaction", async () => {
...
```
jeffsmale90, line 459: Also, here - no longer asserting the value

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -532,7 +532,7 @@ describe("@ganache/ethereum-transaction", async () => {
...
```
jeffsmale90, line 535: ...and here...

src/chains/ethereum/ethereum/src/helpers/gas-estimator.ts
```diff
@@ -34,7 +62,10 @@ const stepTracker = () => {
         preCompileCheck = false;
       }
       if (isCall(info.opcode.name)) {
-        info.stack = info.stack.map(val => val.clone());
+        // this might not be necessary. previously this was a BN that we were
+        // cloning, I'm guessing to get the values and not have refs. I didn't
+        // know of a good way to do this with bigints, or if it's needed at all
+        info.stack = info.stack.map(val => Quantity.toBigInt(val));
```
jeffsmale90: You shouldn't have to clone the bigints themselves as they're immutable primitives. You might need to clone the array though, something like:

```suggestion
        info.stack = [...info.stack];
```

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: Yep :/
`await eei.putAccount(ejsAddress, account);` will be undone when we mine a block unless we checkpoint/commit.

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Good call on collapsing the duplicates, and on removing the trailing `/`. It doesn't look like we can collapse `import` and `import type` (at least as far as I could tell)

src/chains/ethereum/ethereum/src/blockchain.ts
MicaiahReid: Instead I updated the function to receive a `DefaultStateManager` and did the assertion when calling the function (which only happens once).

src/chains/ethereum/ethereum/src/database.ts
MicaiahReid: I had that originally and am open to putting it back. I liked initializing to `DBType.Level` because 2 out of the 3 code paths in the `initialize` function end up being `Level`, only one is `LevelDown`. Eventually we'll support someone providing their _own_ level instance, so then it will be 3 out of 4 code paths that are all `Level`. So I think starting with `DBType.Level` gets across that this is our "default type". Open to disagreement if you still think it should change.

src/chains/ethereum/ethereum/src/forking/persistent-cache/helpers.ts
MicaiahReid: I like it, done! Also moved that confusingly placed comment to be a js-doc comment.

src/chains/ethereum/ethereum/src/forking/state-manager.ts
MicaiahReid: Done!

src/chains/ethereum/ethereum/src/forking/trie.ts
MicaiahReid: Done ish. I added a note for the inspiration for this function. Some digging would be needed for why `persistRoot` needs to be called, the honest answer is because that's what they do!

src/chains/ethereum/ethereum/src/helpers/gas-estimator.ts
MicaiahReid: `new BN(val)` returns type `any`, so that's why I'm setting explicitly to return a BN

src/chains/ethereum/ethereum/src/helpers/gas-estimator.ts
MicaiahReid: Good eye!

src/chains/ethereum/ethereum/src/helpers/precompiles.ts
MicaiahReid: Fixed, good find.

src/chains/ethereum/ethereum/src/helpers/run-call.ts
MicaiahReid: good find!

src/chains/ethereum/ethereum/src/provider-events.ts
MicaiahReid: This one makes me nervous because gas estimation is really complicated and finicky and we don't have any tests for it yet :/ I don't know if we were doing this to prevent passing those arrays by reference

PR #3658, opened 2022-09-15
feat: add support for the `merge` hardfork

davidmurdoch:
closes #3658

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -38,7 +38,8 @@ import {
   BUFFER_32_ZERO,
   BUFFER_256_ZERO,
   findInsertPosition,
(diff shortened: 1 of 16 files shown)


src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1447,6 +1461,8 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     };
   };
 
+  isPostMerge: boolean;
```
MicaiahReid, line 1464: For later when we clean up tech debt: does this make sense to define here? This flag is used all throughout the file, so it seems like it would make sense at the top.

davidmurdoch: probably a good idea to move it up in the file, yeah.

PR #3660, opened 2022-09-15
fix: use zero address when unable to calculate 'from'

gBaGu:
I'm trying to fork a evm-compatible chain with ganache-ui, but my chain contains custom signatures and when block that is used as a fork point has transaction with custom signature ganache application crashes on Transactions screen.
The problem comes from Transaction.from getter function which throws if signature is invalid and causes ui to crash.
This PR simply uses zero address as 'from' field instead of throwing an exception.

### DIFF
--- a/lib/utils/transaction.js
+++ b/lib/utils/transaction.js
@@ -235,10 +235,17 @@ module.exports = class Transaction extends EthereumJsTransaction {
    * into an identical Transaction via `Transaction.decode(encodedTx)`.
    */
   encode() {
+    var from;
+    try {
+      from = to.rpcDataHexString(this.from);
+    }
+    catch {
+      from = "0x0000000000000000000000000000000000000000";
+    }
     const resultJSON = {
       hash: to.nullableRpcDataHexString(this.hash()),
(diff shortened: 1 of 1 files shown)


davidmurdoch:
Ganache-UI hasn't been updated to use the latest version of Ganache. Your fix here would certainly make UI stop crashing... but would ultimately just make the implementation completely incorrect.

gBaGu:
Is there another way to handle invalid signatures without crashing?

davidmurdoch:
Does Ganache v7 (`npm install ganache --global` and then run `ganache --fork.url <YOUR URL HERE>`) return errors when you try to use it?

What chain are you using?

gBaGu:
@davidmurdoch cli Ganache works fine. So my only option is to use command line Ganache instead of UI?
I'm using [Velas chain](https://github.com/velas/velas-chain)

davidmurdoch:
Ganache UI hasn't been updated yet.

gBaGu:
@davidmurdoch will it be updated any time soon?

PR #3712, opened 2022-09-23
feat: package ganache as a snap

tenthirtyone:
Adds a `snapcraft.yaml` to the `src/packages/ganache` directory that contains the configuration for building snaps.


tenthirtyone:
You will need to install

``` sh
$ sudo apt install snapd snapcraft
```

The `cd` into `src/packages/ganache` and run

``` sh
$ snapcraft
```

It will ask you to install `multipass`, say yes. After the build completes run:

``` sh
$  sudo snap install ganache_1.0.0_amd64.snap --devmode --dangerous
```

`--devmode` runs unconfined, `--dangerous` acknowledges this is not a signed snap. Then run:

``` sh
$ ganache
```

PR #3719, opened 2022-09-24
feat: Add remaining txpool methods

### LINKED ISSUE #763: Add `txpool` methods to RPC
https://geth.ethereum.org/docs/rpc/ns-txpool

- [x] txpool_content
- [ ] txpool_contentFrom
- [ ] txpool_inspect
- [ ] txpool_status

amandesai01:
Resolves #763 

Breaking down problem to these sub-tasks:

- [x] Refactor `processMap`
    - [ ] Rename to something better
    - [x] move to TransactionPool class
- [x] create txpool_contentFrom
- [ ] create txpool_inspect
- [x] create txpool_status


amandesai01:
Not sure exactly which params should be displayed for [txpool_inspect](https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_inspect). Rest can be merged. Also, not sure with naming of processMap().

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -453,6 +455,33 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
...
+        }
+        
+        const from = transaction.from.toString();
+        if (res[from] === undefined) {
+          res[from] = {};
+        }
+        // The nonce keys are actual decimal numbers (as strings) and not
+        // hex literals (based on what geth returns).
+        const nonce = transaction.nonce.toBigInt().toString();
+        res[from][nonce] = transaction.toJSON() as Ethereum.Pool.Transaction<"private">;
```
MicaiahReid, line 479: I believe we will still need to pass in a `Common` to the transaction's `toJSON` field. This is used in the legacy transaction type to determine if the user is running Ganache with EIP-2718 activated. If not, transaction types did not yet exist and should not be included on the return object.

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -453,6 +455,33 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
     return null;
   }
 
+  public processMap(map: Map<string, Heap<TypedTransaction>>, address?: Address) {
```
MicaiahReid, line 458: We probably should have come up with a better name for this function when we first reviewed it. But especially now that it's being used elsewhere, I think we should rename this. The trouble is deciding on the name! Perhaps `groupTransactionsByOriginAndNonce`? @davidmurdoch opinions?

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -453,6 +455,33 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
     return null;
   }
```
MicaiahReid, line 457: I think we'll need a js-doc comment here. Could you add something like this? (feel free to refine)
```suggestion
  /**
   * Iterates over all transactions in each Heap for each key in the map.
   * Returns the transactions grouped by origin and nonce in the format: `{
   * origin1: { nonceA: ...txData, ... }, ...}`
   *
   * If `address` is specified, filters the result to only show transactions
   * from `address`
   * @param map
   * @param address
   * @returns Transactions grouped by origin and nonce.
   */
```

src/chains/ethereum/ethereum/src/transaction-pool.ts
```diff
@@ -453,6 +455,33 @@ export default class TransactionPool extends Emittery<{ drain: undefined }> {
...
+    let res: Record<
+      string,
+      Record<string, Ethereum.Pool.Transaction<"private">>
+    > = {};
+    
+    for (let [_, { array, length }] of map) {
+      for (let i = 0; i < length; ++i) {
+        const transaction = array[i];
+
+        if(address && transaction.from.toString() != address.toString()) {
```
MicaiahReid, line 468: We're getting `transaction.from.toString()` twice, could we store this as a variable?

MicaiahReid: Could we also store `address.toString()` outside of the loop so it's only computed once?

src/chains/ethereum/ethereum/tests/api/txpool/txpool.test.ts
```diff
@@ -120,4 +120,110 @@ describe("txpool", () => {
       assert.deepStrictEqual(queued, {});
     });
   });
+
```
MicaiahReid, line 123: For some reason github isn't formatting this file as typescript. I wonder what's going on with that :thinking:

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -3426,39 +3426,83 @@ export default class EthereumApi implements Api {
...
-
     return {
       pending: processMap(executables.pending),
       queued: processMap(origins)
     };
   }
 
+
+  /**
+   * Returns the number of transactions created by specified address currently pending for inclusion in the next block(s), as well as the ones that are being scheduled for future execution only.
```
MicaiahReid:
```suggestion
   * Returns transactions created by the specified address currently pending inclusion in the next blocks, as well as the ones that are scheduled for future execution.
```

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -3426,39 +3426,83 @@ export default class EthereumApi implements Api {
...
+    const fromAddress = Address.from(address);
+
+    return {
+      pending: processMap(executables.pending, fromAddress),
+      queued: processMap(origins, fromAddress)
+    };
+   }
+
+  /**
+   * Returns the number of transactions currently pending for inclusion in the next block(s), as well as the ones that are being scheduled for future execution only.
```
MicaiahReid, line 3471:
```suggestion
   *  Returns the number of transactions currently pending inclusion in the next blocks, as well as the ones that are scheduled for future execution.
```

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -3426,39 +3426,83 @@ export default class EthereumApi implements Api {
...
       queued: processMap(origins)
     };
   }
 
+
+  /**
+   * Returns the number of transactions created by specified address currently pending for inclusion in the next block(s), as well as the ones that are being scheduled for future execution only.
+   * 
+   * @param address - The account address
+   * @returns The transactions currently pending or queued in the transaction pool by address.
```
MicaiahReid, line 3445:
```suggestion
   * @returns The transactions from the specified address that are currently pending or queued in the transaction pool.
```

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -3426,39 +3426,83 @@ export default class EthereumApi implements Api {
...
+   * @example
+   * ```javascript
```
MicaiahReid, line 3447: For some reason these examples aren't working in the generated docs - but the `txpool_content` example isn't working either. I can look into this!

MicaiahReid: I found it!
We'll need to add this to the top of  all three txpool examples to keep the transactions from being mined so they actually stay in the pool:
```TypeScript
await provider.send("miner_stop")
```

Also, for the `txpool_contentFrom` example can we send _another_ transaction from a _different_ address to showcase the fact that it filters by address?

MicaiahReid (changes requested):
First I want to say thank you so much for making this PR! This is great work and is really appreciated.

I have a few small comments, a lot of which is about legacy code that I'd like to get fixed up before merging :smile: 

Aside from those comments, I think I'd like to see slightly more robust tests. I think the previous test for `txpool_content` are a good starting point to work off of.

If you come to a point where this is too much, let us know, we'd be happy to take over. But we really appreciate the work you've done and would love to have you take it over the finish line.

amandesai01:
Hello team. Apologies for leaving this PR un-resolved. Is it still needed? If yes, I can complete it. Thanks in advance.

PR #3723, opened 2022-09-26
docs: update README.md to clarify usage

emilyJLin95:
Just clarifying how to run Ganache for usage on the CLI as well as in programmatic usage.

### DIFF
--- a/src/packages/ganache/README.md
+++ b/src/packages/ganache/README.md
@@ -60,7 +60,12 @@ In case you're upgrading from a previous version of Ganache, we've also written 
 Once installed globally, you can start ganache right from your command line:
 
 ```console
-$ ganache
+ganache
+```
+
+Your output should look something like this:
+
+```
 Ganache CLI v6.12.1 (ganache-core: 2.13.1)
 
 Available Accounts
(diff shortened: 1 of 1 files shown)


davidmurdoch:
The section you're updating is about adding `ganache` to an _npm_ project (it gets added to their `package.json` as a dependency).

There is already a section in the README.md about command line use that says:


> Once installed globally, you can start ganache right from your command line:

```bash
$ ganache
Ganache CLI v6.12.1 (ganache-core: 2.13.1)
```

I welcome improvements to the readability here so it is more clear on the different ways ganache can be used.

emilyJLin95:
ah okay updated section titles

also, just wondering, is line 50 still up to date? do we support node v12

@davidmurdoch

davidmurdoch:
> ah okay updated section titles
> 
> also, just wondering, is line 50 still up to date? do we support node v12
> 
> @davidmurdoch

Yes, Ganache still works with v12.0.0+. This will be changing soon though.

davidmurdoch (approved):

eggplantzzz (approved):
I like

tenthirtyone (approved):

PR #3735, opened 2022-09-28
test: adds test coverage report to @ganache/cli, improve coverage

### LINKED ISSUE #3978: Update weird promises import in detach once node v12 is dropped
See: https://github.com/trufflesuite/ganache/pull/3568/files#diff-7caf617963066f26d673e6beceec6bbae9814dc730358a6e2f4bcd7e49377102R6

jeffsmale90:
Makes some of the components in @ganache/cli more testable, adds coverage report, and improves test coverage to `process-name.ts` and `args.ts`. The tests for `args.ts` cover all of the "general" configuration cases, but does not cover the flavor specific arguments.

image

Fixes: #3978


src/packages/cli/src/args.ts
```diff
@@ -160,8 +165,9 @@ export default function (
 
   // disable dot-notation because yargs just can't coerce args properly...
   // ...on purpose! https://github.com/yargs/yargs/issues/1021#issuecomment-352324693
-  yargs
+  const yargsParser = yargs()
     .parserConfiguration({ "dot-notation": false })
+    .exitProcess(false)
```
jeffsmale90, line 170: By default Yargs terminates the process if `--help` provided, or if validation fails.

With this change, we will pass through if `--help` is provided (and return `{ action: "help }`), and throw if validation fails.

This allows us to test args validation more thoroughly.

src/packages/cli/src/args.ts
```diff
@@ -188,12 +194,9 @@ export default function (
         command = ["$0", flavor];
         defaultPort = 8545;
         break;
-      default:
```
jeffsmale90, line 191: As far as I can tell, this code path is unreachable, as the only valid flavors are "ethereum" and "filecoin", which are handled above.

src/packages/cli/src/args.ts
jeffsmale90, line 170: By default Yargs terminates the process if `--help` provided, or if validation fails.

With this change, we will pass through if `--help` is provided (and return `{ action: "help }`), and throw if validation fails.

This allows us to test args validation more thoroughly, and we handle the thrown error in [cli.ts](https://github.com/trufflesuite/ganache/pull/3735/files#diff-d5cc68b0e026fee131b90a36b9f5bb25b719c702a557cb904ecde02dda59f78cR64)

src/packages/cli/src/args.ts
jeffsmale90, line 191: As far as I can tell, this code path is unreachable, as the only valid flavors are "ethereum" and "filecoin", which are handled above.

davidmurdoch:
just FYI, another massive change to this file is also underway: https://github.com/trufflesuite/ganache/pull/3449/files#diff-d5cc68b0e026fee131b90a36b9f5bb25b719c702a557cb904ecde02dda59f78c

src/packages/cli/src/types.ts
```diff
@@ -21,6 +21,7 @@ export type StartArgs<TFlavorName extends FlavorName> =
 export type GanacheArgs =
   | (AbstractArgs<"stop"> & { name: string })
   | AbstractArgs<"list">
+  | AbstractArgs<"none">
```
jeffsmale90, line 24: This is required because Yargs parser no longer terminates the process when argument validation fails (or `--help`) is passed.

src/packages/cli/src/args.ts
```diff
@@ -214,14 +217,19 @@ export default function (
...
+            coerce: port => (isFinite(port) ? +port : port)
           })
           .check(argv => {
             const { "server.port": port, "server.host": host } = argv;
-            if (port < 1 || port > 65535) {
-              throw new Error(`Invalid port number '${port}'`);
+            if (!isFinite(port) || port < 1 || port > 65535) {
+              throw new Error(
+                `Port should be >= 0 and < 65536. Received ${port}.`
+              );
```
jeffsmale90, line 231: This will conflict with #4070

src/packages/cli/package.json
```diff
@@ -71,7 +73,7 @@
     "@trufflesuite/ps-list": "0.0.3",
     "@types/node": "17.0.0",
     "chalk": "4.1.0",
-    "cli-table": "0.3.11",
-    "marked-terminal": "4.1.0"
+    "marked-terminal": "4.1.0",
```
MicaiahReid: Was this change on purpose? It seems like the versions are the same and they are now out of alphabetical order.

MicaiahReid (changes requested):

src/packages/cli/src/args.ts
```diff
@@ -160,8 +165,9 @@ export default function (
 
   // disable dot-notation because yargs just can't coerce args properly...
   // ...on purpose! https://github.com/yargs/yargs/issues/1021#issuecomment-352324693
-  yargs
+  const yargsParser = yargs()
```
davidmurdoch, line 168: Why shouldn't we use the singleton?

src/packages/cli/src/args.ts
davidmurdoch: This is all being removed in another PR. This change will just add to the painful merge conflicts to come.

src/packages/cli/package.json
jeffsmale90: Nope!

src/packages/cli/src/args.ts
jeffsmale90: Cos singletons are the literal devil 👿 

But actually, because the singleton makes it much harder to test. Because the yargs singleton underlying the args parser would be shared between each instance under test, we would need to reset it's state between tests (probably reaching into the yargs module directly from the test, which makes it more fragile, because the test is making assumptions about the implementation that aren't enforced). It's much nicer to just have each instance of the args parser to be self contained.

PR #3692, opened 2022-09-20
feat: get logs from a fork network directly

adjisb:

PR #3739, opened 2022-09-29
fix: eth get block transaction count by hash to work with forks

adjisb:


src/chains/ethereum/ethereum/tests/forking/block.test.ts
```diff
@@ -68,5 +68,12 @@ describe("forking", function () {
       const block = await provider.send("eth_getBlockByNumber", ["0x0", true]);
       assert.deepStrictEqual(block, block0);
     });
+
+    it("should get TransactionCountByHash from the original chain", async () => {
```
MicaiahReid:
```suggestion
    it("should get transaction count by hash from the original chain", async () => {
```
Just to be more in line with how other tests are formatted!

MicaiahReid (changes requested):
This is so awesome, thanks for the PR!

MicaiahReid (approved):

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1266,14 +1266,12 @@ export default class EthereumApi implements Api {
...
     const { blocks } = this.#blockchain;
-    const blockNum = await blocks.getNumberFromHash(hash);
-    if (!blockNum) return null;
-
-    const rawBlock = await blocks.getRawByBlockNumber(Quantity.from(blockNum));
-    if (!rawBlock) return null;
-
-    const [, rawTransactions] = decode<GanacheRawBlock>(rawBlock);
-    return Quantity.from(rawTransactions.length);
+    const block = await blocks
```
davidmurdoch, line 1277: Looks like the bug is really in `getNumberFromHash`, if we fix that I suspect the fix for https://github.com/trufflesuite/ganache/pull/3692 will become slightly simpler.

Do you want to try to tackle fixing `getNumberFromHash` to fetch from the fallback?

adjisb: getByHash does the right thing (calling eth_getBlockByHash on the fork) and getNumberFromHash end up being just a helper function we can even declare it private in the future.

davidmurdoch (approved):

PR #3755, opened 2022-10-03
fix: set the `earliest` block tag on startup when forking

adjisb:
This change updates Ganache's startup procedure when forking to retrieve the `earliest` block from the remote and cache this block as the `earliest` block in Ganache's block manager. This fixes a bug where calling `eth_getBlockByNumber` with the `"earliest"` block tag parameter yielded no result.

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -332,7 +332,7 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
             ({ block }) => block
           );
           // when we are forking, blocks.earliest is already set to what was
-          // retrieve from the fork
+          // retrieved from the fork
           if (!blocks.earliest) {
             blocks.earliest = blocks.latest;
           }
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -328,8 +328,10 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
             options.miner.blockGasLimit,
             initialAccounts
           );
-          blocks.earliest = blocks.latest =
-            await this.#blockBeingSavedPromise.then(({ block }) => block);
+          blocks.latest = await this.#blockBeingSavedPromise.then(({ block }) => block);
+          if (!blocks.earliest) {
+            blocks.earliest = blocks.latest;
+          }
         }
       }
 
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -331,6 +331,8 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
           blocks.latest = await this.#blockBeingSavedPromise.then(
             ({ block }) => block
           );
+          // when we are forking, blocks.earliest is already set to what was
+          // retrieve from the fork
           if (!blocks.earliest) {
             blocks.earliest = blocks.latest;
           }
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -328,7 +328,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
             options.miner.blockGasLimit,
             initialAccounts
           );
-          blocks.latest = await this.#blockBeingSavedPromise.then(({ block }) => block);
+          blocks.latest = await this.#blockBeingSavedPromise.then(
+            ({ block }) => block
+          );
           if (!blocks.earliest) {
             blocks.earliest = blocks.latest;
           }
--- a/src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
+++ b/src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
@@ -15,6 +15,7 @@ import {
   TypedDatabaseTransaction
 } from "@ganache/ethereum-transaction";
 import { GanacheLevelUp } from "../database";
+import { Ethereum } from "../api-types";
 
 const LATEST_INDEX_KEY = BUFFER_ZERO;
 
@@ -277,7 +278,7 @@ export default class BlockManager extends Manager<Block> {
   async getEarliest() {
     const fallback = this.#blockchain.fallback;
     if (fallback) {
-      const json = await fallback.request<any>(
+      const json = await fallback.request<Ethereum.Block<true, "public">>(
         "eth_getBlockByNumber",
         [Tag.earliest, true],
         // TODO: re-enable cache once this is fixed
--- a/src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
+++ b/src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
@@ -274,17 +274,34 @@ export default class BlockManager extends Manager<Block> {
     await this.#blockIndexes.put(LATEST_INDEX_KEY, number);
   }
 
+  async getEarliest() {
+    const fallback = this.#blockchain.fallback;
+    if (fallback) {
+      const json = await fallback.request<any>(
+        "eth_getBlockByNumber",
+        [Tag.earliest, true],
+        { disableCache: true }
+      );
+      if (json) {
+        const common = fallback.getCommonForBlockNumber(
+          this.#common,
+          BigInt(json.number)
+        );
+        return new Block(BlockManager.rawFromJSON(json, common), common);
+      }
+    }
+    for await (const data of this.base.createValueStream({ limit: 1 })) {
+      return new Block(data as Buffer, this.#common);
+    }
+  }
+
   /**
    * Updates the this.latest and this.earliest properties with data
    * from the database.
    */
   async updateTaggedBlocks() {
     const [earliest, latestBlockNumber] = await Promise.all([
-      (async () => {
-        for await (const data of this.base.createValueStream({ limit: 1 })) {
-          return new Block(data as Buffer, this.#common);
-        }
-      })(),
+      this.getEarliest(),
       this.#blockIndexes.get(LATEST_INDEX_KEY).catch(e => null)
     ]);
 
--- a/src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
+++ b/src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
@@ -292,9 +292,12 @@ export default class BlockManager extends Manager<Block> {
         );
         return new Block(BlockManager.rawFromJSON(json, common), common);
       }
-    }
-    for await (const data of this.base.createValueStream({ limit: 1 })) {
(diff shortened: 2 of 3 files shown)


src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -272,26 +272,44 @@ export default class BlockManager extends Manager<Block> {
     await this.#blockIndexes.put(LATEST_INDEX_KEY, number);
   }
 
+  async getEarliest() {
+    const fallback = this.#blockchain.fallback;
+    if (fallback) {
+      const json = await fallback.request<any>("eth_getBlockByNumber", [
+        Tag.earliest,
+        true
+      ], { disableCache: true });
```
MicaiahReid, line 281: I don't think we should be disabling the cache here. Maybe I could be convinced, though! What was your reasoning here?

src/chains/ethereum/ethereum/tests/forking/block.test.ts
```diff
@@ -3,15 +3,15 @@ import getProvider from "../helpers/getProvider";
...
+describe("forking", function() {
   this.timeout(10000);
 
   describe("blocks", () => {
     const blockNumber = 0xb77935;
     const blockNumHex = `0x${blockNumber.toString(16)}`;
     const URL = "https://mainnet.infura.io/v3/" + process.env.INFURA_KEY;
     let provider: EthereumProvider;
-    before(async function () {
+    before(async function() {
```
MicaiahReid:
```suggestion
    before(async function () {
```

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
MicaiahReid: I don't think we should be disabling the cache here. Maybe I could be convinced, though! What was your reasoning here?

MicaiahReid (changes requested):
Great stuff, thanks for the PR! Just a tiny change and a question from me. Would love to see what @davidmurdoch thinks.

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
MicaiahReid:
```suggestion
      const json = await fallback.request<any>(
        "eth_getBlockByNumber",
        [Tag.earliest, true],
        { disableCache: true }
      );
```
Prettier change

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -322,8 +322,10 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
             options.miner.blockGasLimit,
             initialAccounts
           );
-          blocks.earliest = blocks.latest =
-            await this.#blockBeingSavedPromise.then(({ block }) => block);
+          blocks.latest = await this.#blockBeingSavedPromise.then(({ block }) => block);
```
```suggestion          
          blocks.latest = await this.#blockBeingSavedPromise.then(
            ({ block }) => block
          );
```
Prettier change

MicaiahReid (changes requested):
Also, we use prettier for our linting. I just ran prettier on the files you changed and added some suggested changes accordingly.

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
adjisb: The cache don't support "earliest" it assumes that the first argument for eth_getBlockByNumber is a number (maybe another thing to fix in the cache code?)
On the other hand the block is cached in block-manager (my understanding is that the initialization routine is called once).

adjisb:
> Great stuff, thanks for the PR! Just a tiny change and a question from me. Would love to see what @davidmurdoch thinks.

To be honest the important stuff is in: [PR3692](https://github.com/trufflesuite/ganache/pull/3692), this PRs are just to simplify 3692

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -272,26 +272,45 @@ export default class BlockManager extends Manager<Block> {
     await this.#blockIndexes.put(LATEST_INDEX_KEY, number);
   }
 
+  async getEarliest() {
+    const fallback = this.#blockchain.fallback;
+    if (fallback) {
+      const json = await fallback.request<any>(
+        "eth_getBlockByNumber",
+        [Tag.earliest, true],
+        { disableCache: true }
```
MicaiahReid, line 286:
```suggestion
        // TODO: re-enable cache once this is fixed
        // https://github.com/trufflesuite/ganache/issues/3773
        { disableCache: true }
```
I see the reasoning for the disabled cache now, and think it's worth moving forward on this fix without also fixing the cache.

MicaiahReid (changes requested):

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -272,26 +272,47 @@ export default class BlockManager extends Manager<Block> {
     await this.#blockIndexes.put(LATEST_INDEX_KEY, number);
   }
 
+  async getEarliest() {
+    const fallback = this.#blockchain.fallback;
+    if (fallback) {
+      const json = await fallback.request<any>(
```
davidmurdoch: We actually have types for these now. You'll need to add

```
import { Ethereum } from "../api-types";
```
at the top, and then you can do:

```
      const json = await fallback.request<Ethereum.Block<true, "public">>(
```

Do you mind making this change?

src/chains/ethereum/ethereum/tests/forking/block.test.ts
```diff
@@ -3,7 +3,7 @@ import getProvider from "../helpers/getProvider";
 import { EthereumProvider } from "../../src/provider";
 import request from "superagent";
 
-describe("forking", function () {
+describe("forking", function() {
```
davidmurdoch:
```suggestion
describe("forking", function () {
```

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -322,8 +322,12 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
             options.miner.blockGasLimit,
             initialAccounts
           );
-          blocks.earliest = blocks.latest =
-            await this.#blockBeingSavedPromise.then(({ block }) => block);
+          blocks.latest = await this.#blockBeingSavedPromise.then(
+            ({ block }) => block
+          );
+          if (!blocks.earliest) {
```
davidmurdoch, line 336: I believe if we have a `latest` we will always have an `earliest` too; I don't think this extra check here is needed (and I think it would be impossible to test for).

If we do think the `blocks.earliest` check is needed, it should go outside this `if (!latest) {` block and be checked independently of `latest`. @MicaiahReid any thoughts?

davidmurdoch (changes requested):
Crazy that this bug has been in here for so long! Thanks so much for this PR!

I've requested a couple of changes, as well as pinged others for their opinions. If you don't want to make the changes yourself let me know and we can take over. Thanks again!

src/chains/ethereum/ethereum/src/blockchain.ts
adjisb: Please can you take over this part. 
I thought that the code wasn't necessary. In theory the values are initialized in BlockManager.initialize. My first attempt was to remove the lines, but then some test failed.

MicaiahReid: I had the same thought that this check will never happen, but I checked out code coverage and it is hit many times. This might take a bit of investigation.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -328,8 +328,12 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+            blocks.earliest = blocks.latest;
+          }
```
MicaiahReid, line 338:
```suggestion
          blocks.earliest = blocks.latest;
```

MicaiahReid: Previously I took @davidmurdoch's comment above to mean that we shouldn't be _setting_ earliest here. I think what he was actually saying is that we should _always_ be setting earliest here. If we've had to create a genesis block, we don't have an earliest or a latest, so they should be equal upon initialization. This seems to be the right approach to me.

MicaiahReid (changes requested):

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: Yes, that is what I was saying. Sorry I wasn't clear. When you checked the code coverage for this do remember seeing if the (implicit) `else` of the `if (!blocks.earliest) {` was hit (which is what I thought you were saying in your previous comment)? IIRC, it'll show a black `E` on the same line as the `if` statement.

MicaiahReid: image
Nope, no implicit else.

But we check that if 423x and only enter is 342x :thinking:

src/chains/ethereum/ethereum/src/blockchain.ts
adjisb: Always assigning here overwrites the value set in BlockManager.initialize (via updateTaggedBlocks). So if there is a value we must keep it. 
Maybe there is a better solution, but I don't have the understanding of the whole code to implement it.

src/chains/ethereum/ethereum/src/blockchain.ts
adjisb: Always assigning here overwrites the value set in BlockManager.initialize (via updateTaggedBlocks). So if there is a value we must keep it.
Maybe there is a better solution, but I don't have the understanding of the whole code to implement it.

MicaiahReid: Okay, I think I understand now. This check is relevant for the following case:

When we are forking and initialize the fork, we get the remote's "latest" block, or whatever number the user forked at, but we don't actually set `this.latest` to that result (this is on purpose, read on). After the fork is initialized, we initialize the block manager. In this forking case, the block manager's `updateTaggedBlocks` function will now return a result for `earliest` because, thanks to this change, we now check the fallback for an earliest block. However, that function will _not_ set a value for `this.latest` yet because no block has been saved to the database. After the block manager is initialized, we'll get here and find that `latest` has not yet been set, but `earliest` has. Because `latest` has not been set, `initializeGenesisBlock` will be called, which uses the fallback information to mine one block _above_ the remote's latest (or the fork block number). So this case would yield the expected result - `earliest` is whatever the remote returns for fetching earliest, and `latest` is one block mined on top of the fork block number.

@davidmurdoch  does that all seem right to you?

If so, this check should remain as-is. Though maybe to your original point this could be moved outside of the `if(!latest)` check.

davidmurdoch (approved):

MicaiahReid (approved):

jeffsmale90 (approved):

PR #3777, opened 2022-10-07
refactor: rework `@ganache/ethereum-address` to extend `ethereumjs/util`'s `Address`

MicaiahReid:


src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -93,7 +93,7 @@ export default class BlockManager extends Manager<Block> {
     const extraTxs: GanacheRawBlockTransactionMetaData[] = [];
     json.transactions.forEach((tx, index) => {
       const blockExtra = [
-        Quantity.toBuffer(tx.from),
+        Address.toBuffer(tx.from),
```
MicaiahReid, line 96: Using `Quantity` here previously definitely was a bug :sweat_smile:

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -948,7 +939,7 @@ export default class EthereumApi implements Api {
       const block = new RuntimeBlock(
         Quantity.from((parentHeader.number.toBigInt() || 0n) + 1n),
         parentHeader.parentHash,
-        parentHeader.miner,
+        Address.from(parentHeader.miner.toString()),
```
davidmurdoch: Is `parentHeader.miner` a Data here? If so, I think we can call `toBuffer()` here.

src/chains/ethereum/transaction/src/eip1559-fee-market-transaction.ts
```diff
@@ -115,7 +116,7 @@ export class EIP1559FeeMarketTransaction extends RuntimeTransaction {
       blockNumber: this.blockNumber ? this.blockNumber : null,
       transactionIndex: this.index ? this.index : null,
       from: this.from,
-      to: this.to.isNull() ? null : this.to,
+      to: this.to ? this.to : null,
```
davidmurdoch: Can this be simplified to:

```suggestion
      to: this.to,
```

?

davidmurdoch: (same in the other transaction types that do this same check)

davidmurdoch (changes requested):
Looks so much better now! Loving it!

davidmurdoch (approved):

src/chains/ethereum/address/index.ts
```diff
@@ -1,14 +1,15 @@
...
+import { Address as EJSAddress } from "@ethereumjs/util";
+import { JsonRpcDataInputArg, Data } from "@ganache/utils";
 
-export class Address extends Data {
+export class Address extends EJSAddress {
   static ByteLength = 20;
 
-  constructor(value: string | Buffer) {
-    super(value, Address.ByteLength);
+  constructor(value: Buffer) {
```
jeffsmale90, line 7: Why change the signature of the function? Wouldn't it be nice to be able to `new Address("0x123123123123");`?

Additionally, this throws if the length of value is not 20 - do we want to allow different length buffers to be passed?

Not saying that we definitely _should_ do these things, but it would maintain consistency with the Data / Quantity types 🤷

jeffsmale90 (changes requested):
Other than my one comment - this is 🔥 . I love not having to hack around the ejs-Address type.

src/chains/ethereum/address/index.ts
MicaiahReid: > Why change the signature of the function? Wouldn't it be nice to be able to new Address("0x123123123123");?

I don't have particularly strong opinions on this, but to me it seemed to make sense for the constructor to match the `EJSAddress` that we are extending, and the `.from` function do the conversions. If the `.from` only calls `new Address`, why does it need to exist?

> Additionally, this throws if the length of value is not 20 - do we want to allow different length buffers to be passed?

I would rather not allow different length buffers to be passed because: 
1. we didn't have any existing cases where this happened
2. I don't want to increase the performance overhead of this change if it isn't necessary

jeffsmale90 (approved):

PR #3796, opened 2022-10-12
refactor: add a type to an internal request

davidmurdoch:


src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -130,10 +132,10 @@ export default class BlockManager extends Manager<Block> {
       blockNumber = tagOrBlockNumber.toString();
     }
 
-    const json = await fallback.request<any>("eth_getBlockByNumber", [
-      blockNumber,
-      true
-    ]);
+    const json = await fallback.request<Ethereum.Block<true, "public">>(
```
jeffsmale90, line 135: Could we do something similar to the Ganache provider where it infers the return type from the `method` parameter?

jeffsmale90 (changes requested):

PR #3797, opened 2022-10-12
Where's the flavor?

davidmurdoch:
I don't taste anything

(draft)


src/chains/ethereum/ethereum/index.ts
```diff
@@ -5,5 +5,159 @@
...
+interface EthereumFlavor
+  extends Flavor<EthereumProvider, ServerOptionsConfig, EthereumDefaults> {
+  flavor: "ethereum";
+  Connector: typeof Connector;
+  serverOptions: ServerOptionsConfig;
+  initialize: typeof initialize;
+}
+
+const EthereumFlavor: EthereumFlavor = {
+  flavor: "ethereum",
```
tenthirtyone, line 35: non-technical thought

I think flavor makes sense. For the uninitiated, they would understand flavors means types of something. Would it be worthwhile to alias the branding to the specific thing for clarity? E.g. support both `pluginName` & `flavor` and they do the same thing?

Does this translate well to non-English users?

davidmurdoch: `npmPackageNameOrInstallationPath` is what `flavor` means :-p

davidmurdoch:
closing in favor of https://github.com/trufflesuite/ganache/pull/4362

PR #3820, opened 2022-10-19
perf: use keccak from `@ganache/utils` where possible

MicaiahReid:
The latest releases of `@ethereumjs/util` and `@ethereumjs/trie` swap out their previous keccak function to use the less performant [ethereum-cryptography](https://github.com/ethereum/js-ethereum-cryptography) package instead. This change uses the keccak exported from `@ganache/utils` instead, where possible.

### DIFF
--- a/src/packages/utils/src/utils/signature.ts
+++ b/src/packages/utils/src/utils/signature.ts
@@ -5,7 +5,12 @@ type ECDSASignOutput = {
   recid: number;
 };
 
-export function ecsign(msgHash: Uint8Array, privateKey: Uint8Array) {
+export type ECSignResult = { v: bigint; r: Buffer; s: Buffer };
+
+export function ecsign(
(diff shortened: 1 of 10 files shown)


davidmurdoch (approved):

jeffsmale90 (approved):

src/packages/utils/src/utils/signature.ts
```diff
@@ -0,0 +1,37 @@
+import secp256k1 from "@ganache/secp256k1";
+
+type ECDSASignOutput = {
+  signature: Uint8Array;
+  recid: number;
+};
+
+export function ecsign(msgHash: Uint8Array, privateKey: Uint8Array) {
```
jeffsmale90: Could potentially define (and export) a `Signature` return type for this function, which you could use to avoid defining the type inline in `legacy-transaction.ts` 🤷

PR #3821, opened 2022-10-19
chore: upgrade ethereumjs packages

MicaiahReid:


src/chains/ethereum/transaction/package.json
```diff
@@ -48,18 +48,19 @@
     "truffle"
   ],
   "dependencies": {
-    "@ethereumjs/common": "3.0.0",
+    "@ethereumjs/common": "3.0.1",
+    "@ethereumjs/util": "8.0.2",
```
MicaiahReid: I have added `@ethereumjs/util` to this package because I was getting compilation errors. The `@ethereumjs/common` dependency used stuff from the `@ethereumjs/util` library, but I believe it was using an older version, which was causing type errors. It was hard to pin down the exact cause, so that reason could be a little off. But this does fix it.

tenthirtyone (approved):
:+1: been rolling on this branch for genesis.json work

jeffsmale90 (approved):
Like!

PR #3825, opened 2022-10-20
test: fix eth_call test race conditions

MicaiahReid:


MicaiahReid:
Closing cause I'm bad at git

PR #3853, opened 2022-10-28
fix: set default hardfork for each network

### LINKED ISSUE #3631: Set default network hardfork for each network instead of for the entire application
The latest hardfork to land on mainnet, `gray glacier`, [does not exist on any testnet](https://blog.ethereum.org/2022/06/16/gray-glacier-announcement). Setting the default hardfork in ganache to `gray glacier` breaks testnet forking unless a valid hardfork, e.g. `london` is supplied at startup.

satyajeetkolhapure:
fixes https://github.com/trufflesuite/ganache/issues/3631 depending on the network hardfork default value should be set.

The latest hardfork to land on mainnet, gray glacier, [does not exist on any testnet](https://blog.ethereum.org/2022/06/16/gray-glacier-announcement). Setting the default hardfork in ganache to gray glacier breaks testnet forking unless a valid hardfork, e.g. london is supplied at startup.


src/chains/ethereum/options/src/helpers.ts
```diff
@@ -1 +1,14 @@
+import { Hardfork } from "./chain-options"
+import { KnownNetworks } from "./fork-options"
+
+export const getDefaultForkByNetwork = (network : KnownNetworks) : Hardfork => {
+    switch(network){
+        case "mainnet" : return "grayGlacier"
+        case "sepolia" : return "london"
+        case "goerli" : return "london"
+        case "görli" : return "london"
```
jeffsmale90: Could probably remove all of these cases, as they'll get picked up by the default.

Alternatively, you could argue that you want to ensure that all cases are covered explicitly, but then you'd probably want to remove the default case, and could use the `assertExhaustive()` pattern. https://www.fullstory.com/blog/discriminated-unions-and-exhaustiveness-checking-in-typescript/

src/chains/ethereum/options/src/chain-options.ts
```diff
@@ -196,7 +198,12 @@ export const ChainOptions: Definitions<ChainConfig> = {
   hardfork: {
     normalize,
     cliDescription: "Set the hardfork rules for the EVM.",
-    default: () => "london",
+    default: () => {
+      const forkArg = yargs.parse(process.argv)['fork']
```
jeffsmale90: As a general comment: There's a couple of code lines missing trailing ; - if you install `Prettier - Code formatter` in VS Code, and enable "formatOnSave" (https://www.robinwieruch.de/how-to-use-prettier-vscode/), it'll standardise styling for you every time you save.

jeffsmale90 (changes requested):
I know that this area isn't very well covered in tests, but would it be possible to add some for this behaviour?

It's also very _simple_ behaviour, but at least if we document it in tests, we reduce the risk of accidentally changing the behaviour in the future, or becoming confused about the desired outcome as/if it becomes more complicated.

src/chains/ethereum/options/src/helpers.ts
satyajeetkolhapure: Yes. Actually I wanted to cover all cases but I think its better to add them later when there is requirement to do so.

src/chains/ethereum/options/src/chain-options.ts
satyajeetkolhapure: I accidentally disabled prettier. Thank you for noticing it. Will fix it

src/chains/ethereum/options/tests/index.test.ts
```diff
@@ -78,5 +78,18 @@ describe("EthereumOptionsConfig", () => {
...
+        process.argv.push("--fork", "mainnet");
+        const options = EthereumOptionsConfig.normalize({} as Object);
+        assert.strictEqual(options.chain.hardfork, "grayGlacier");
+      });
+      it("when network is not mainnet then hardfork defualt should be london", () => {
+        process.argv.push("--fork", "sepolia");
+        const options = EthereumOptionsConfig.normalize({} as Object);
+        assert.strictEqual(options.chain.hardfork, "london");
+      });
+    });
```
jeffsmale90: Great to get some tests! It'd be nice to name them consistently with the rest of the tests here - `describe()` defines the "unit" that's being tested (property, function, class etc), and `it()` describes the desired behavior, in a way that reads logically with the `it...` prefix.

ie.

```
describe("hardfork", () => {
    it("defaults to `grayGlacier` when network is `mainnet`", () => { ... });
});
```

src/chains/ethereum/options/tests/index.test.ts
```diff
@@ -78,5 +78,18 @@ describe("EthereumOptionsConfig", () => {
...
       });
     });
+
+    describe("set hardfork default value depending on network", () => {
+      it("when network is mainnet then hardfork defualt should be grayGlacier", () => {
+        process.argv.push("--fork", "mainnet");
+        const options = EthereumOptionsConfig.normalize({} as Object);
+        assert.strictEqual(options.chain.hardfork, "grayGlacier");
+      });
+      it("when network is not mainnet then hardfork defualt should be london", () => {
```
jeffsmale90: defualt > default

src/chains/ethereum/options/tests/index.test.ts
jeffsmale90: Would it be worthwhile adding a test to ensure that the default is over-ridden by an explicit hardfork value? Maybe cover some of the other argument values such as "--fork" etc?

src/chains/ethereum/options/src/chain-options.ts
```diff
@@ -196,7 +198,13 @@ export const ChainOptions: Definitions<ChainConfig> = {
   hardfork: {
     normalize,
     cliDescription: "Set the hardfork rules for the EVM.",
-    default: () => "london",
+    default: () => {
+      const forkArg = yargs.parse(process.argv)["fork"];
+      // If fork argument is blank, default value is mainnet
+      const fork =
+        !forkArg || forkArg.toString() == "true" ? "mainnet" : forkArg;
```
jeffsmale90: What's up with checking `toString()` here? I assume the value would be either Boolean, or empty string (or an explicit value)?

It might be worthwhile making this check more explicit, for readability if nothing else.

jeffsmale90 (changes requested):

src/chains/ethereum/options/tests/index.test.ts
satyajeetkolhapure: Yes, I will add more scenarios.

src/chains/ethereum/options/src/chain-options.ts
satyajeetkolhapure: Actually there are 3 scenarios 
1. where --fork option is not provided 
2. --fork option without any value
3. --fork option with value
To cover second scenario where the value is not string I have added that change. Also yargs parse converts it to string. 
If I use process.argv directly then Boolean can be used, but it needs additional parsing logic to check all possibilites.

PR #3880, opened 2022-11-07
fix: simplify database implementation to allow removal of type assertions

MicaiahReid:


src/chains/ethereum/ethereum/src/forking/persistent-cache/persistent-cache.ts
```diff
@@ -1,6 +1,9 @@
 import { Tree } from "./tree";
```
MicaiahReid, line 1: Most of these changes are just reverting back to what we did in `develop`. The only real change is using the `GanacheLevelUp` type instead of `LevelUp`

src/chains/ethereum/ethereum/src/forking/trie.ts
```diff
@@ -42,44 +37,35 @@ export class ForkTrie extends GanacheTrie {
   private isPreForkBlock = false;
   private forkBlockNumber: bigint;
   public blockNumber: Quantity;
-  private metadata: CheckpointDB;
-  private db: LevelDB | UpgradedLevelDown;
+  private checkpointedMetadata: CheckpointDB;
+  /** The underlying database for `checkpointedMetadata */
```
MicaiahReid, line 41: The `.db` property of a `CheckpointDB` strips the types so that we can't iterate over the whole database, so now we store the metadata's database separately from the checkpointed version. This is similar to the added `.db` property of the `GanacheTrie`

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -87,31 +76,32 @@ export default class Database extends Emittery {
...
+      db = levelup(store);
     }
 
     // don't continue if we closed while we were waiting for the db
     if (this.#closed) return this.#cleanup();
 
     const open = db.open();
-    this.trie = db.sublevel("T", LEVEL_OPTIONS) as GanacheSublevel;
+    const sublevelTrie = sub(db, "T", levelupOptions);
+    this.trie = new TrieDB(sublevelTrie);
```
davidmurdoch, line 91: Note for other reviewers: these are the only two runtime lines that have changed in this file when compared against develop.

davidmurdoch (approved):
Useful when reviewing: https://github.com/trufflesuite/ganache/compare/develop...fix-db

Also: great job walking those initial changes back to something so so manageable!

tenthirtyone (approved):

PR #3913, opened 2022-11-14
build: remove hacks from process to build types

MicaiahReid:
This change essentially removes some packages from our list of bundled packages in the `api-extractor.json`, because API extractor wasn't handling the bundling correctly. Instead we add those dependencies to our `src/packages/ganache/package.json` so we include the entire set of packages in our exported ganache bundle.

To test the actual types I ran `npm run reinstall && INFURA_KEY=... npm run build && npm run prepublishOnly && cd src/packages/ganache/ && npm pack && cd ../../..` and imported the generated tarball into a project. I tested the types and that Ganache works in node and in the browser.


src/packages/ganache/package.json
```diff
@@ -28,8 +28,7 @@
   },
   "scripts": {
     "build": "npm run tsc && npx shx rm -rf ./dist && webpack && npm run types",
-    "types": "api-extractor run && npm run hack-types",
-    "hack-types": "npx cpy-cli@3.1.1 './extra-types/**/*.d.ts' ./dist",
```
jeffsmale90, line 32: ❤️

src/packages/ganache/api-extractor.json
```diff
@@ -69,15 +69,12 @@
...
-    "@ethereumjs/vm",
     "@ethereumjs/statemanager",
-    "@ethereumjs/evm",
     "@ethereumjs/util",
     "@ethereumjs/tx",
     "@ethereumjs/common",
     "@ethereumjs/block",
     "@ethereumjs/blockchain",
-    "@ethereumjs/trie",
     "@ethereumjs/ethash",
```
jeffsmale90: I don't understand why _some_ @ethereumjs packages get treated as "bundledPackages"?

src/chains/ethereum/utils/package-lock.json
```diff
@@ -1,2156 +1,2156 @@
 {
-	"name": "@ganache/ethereum-utils",
```
jeffsmale90: Is this file _just_ whitespace changes (I guess it's not, given the package.json changes)? It's real hard to tell.

edit: same for the other package-locks

jeffsmale90 (changes requested):

src/packages/ganache/api-extractor.json
MicaiahReid: Because those packages don't break our types ¯\_(ツ)_/¯

API extractor doesn't work correctly for the vm, evm, and trie packages, but it works fine for the rest. We'd rather have them all here in the extractor config than in the package.json as a dependency, because adding to the package.json increases our total bundle size. But alas, since API extractor doesn't handle these packages correctly, we're removing them.

src/chains/ethereum/utils/package-lock.json
MicaiahReid: If you want to confirm that the package-locks are inline with the package.json, you can always `npm run reinstall` locally to confirm that there are no diffs. Other than that, I don't know what we could do to confirm the validity of these files.

davidmurdoch: I think you need to restore the original whitespace. I think what happened here is that your version of npm has different rules for which whitespace to use in a package-lock. I think some versions of npm preserve what was there before, but others like to change it _sometimes_.

This has happened to me a few times and what I've done in the past is disabled prettier (just in case it is doing weird things), then did a find/replace for `<new whitespace>`/`<old whitespace>` individually within each package-lock npm screwed up. If using VM code you have to do this many many times, since it only will find/replace a few hundred matches at a time.

MicaiahReid: ahhhh got ya. Looks like that worked, thanks @davidmurdoch!

src/packages/ganache/webpack/webpack.node.config.ts
```diff
@@ -51,6 +53,11 @@ const config: webpack.Configuration = merge({}, base, {
   externals: [
     //#region dependencies that have the potential to compile something at install time
     "@trufflesuite/bigint-buffer",
+    "@trufflesuite/uws-js-unofficial",
+    "abstract-leveldown",
+    "async-eventemitter",
+    "leveldown",
+    "levelup",
```
davidmurdoch: Should these fall under the `things api-extractor can't handle, so we don't bundle them` section?

src/packages/ganache/package.json
```diff
@@ -110,9 +114,16 @@
...
+    "@trufflesuite/uws-js-unofficial",
+    "@ethereumjs/vm",
+    "@ethereumjs/trie",
+    "@ethereumjs/evm",
     "abstract-leveldown",
+    "async-eventemitter",
     "emittery",
     "keccak",
+    "leveldown",
+    "levelup",
```
davidmurdoch: Do you remember why `levelup` was added? It isn't in develop and the version, 5.1.1, is the same as before.

src/packages/ganache/webpack/webpack.node.config.ts
```diff
@@ -60,7 +67,20 @@ const config: webpack.Configuration = merge({}, base, {
     //#endregion
     "@ganache/filecoin",
     // things api-extractor can't handle, so we don't bundle them:
-    "emittery"
+    "emittery",
+    "@ethereumjs/vm",
+    // because we're calling @ethereumjs/vm is already added as an external,
```
davidmurdoch:
```suggestion
    // because @ethereumjs/vm is already added as an external,
```

src/packages/ganache/webpack/webpack.node.config.ts
MicaiahReid: I've moved `abstract-leveldown` and `async-eventemitter`. `levelup` is being removed and `leveldown` should remain there (it's already there in develop)

src/packages/ganache/package.json
MicaiahReid: I don't and was just wondering the same thing. I tried removing it and adding it back to `src/chains/ethereum/ethereum` as it is in develop, but I'm getting compilation errors on a project using Ganache:
```
node_modules/ganache/dist/ganache.d.ts:13:25 - error TS2307: Cannot find module 'levelup' or its corresponding type declarations.

13 import { LevelUp } from 'levelup';
```

davidmurdoch: Strange! `levelup` is in api-extractor, but slaps it into the generated `.d.ts` anyway? :thinking:

src/chains/ethereum/ethereum/package.json
```diff
@@ -75,6 +75,7 @@
     "eth-sig-util": "2.5.3",
     "ethereumjs-abi": "0.6.8",
     "leveldown": "6.1.0",
+    "levelup": "5.1.1",
```
MicaiahReid, line 78: This change is reverting back to what's in develop

src/packages/ganache/package.json
```diff
@@ -57,11 +56,10 @@
     "truffle"
   ],
   "devDependencies": {
-    "@ethereumjs/vm": "6.2.0",
     "@ganache/cli": "0.6.0",
     "@ganache/core": "0.6.0",
     "@ganache/flavors": "0.6.0",
-    "@microsoft/api-extractor": "7.31.1",
+    "@microsoft/api-extractor": "7.20.1",
```
MicaiahReid, line 62: This is reverting back to the version that's in develop. Upgrading broke bundling of levelup

src/packages/ganache/package.json
MicaiahReid: It seems that reverting the api-extractor version has fixed this issue

src/chains/ethereum/ethereum/package.json
```diff
@@ -90,6 +91,7 @@
     "@types/encoding-down": "5.0.0",
     "@types/fs-extra": "9.0.2",
     "@types/keccak": "3.0.1",
+    "@types/levelup": "5.1.0",
```
MicaiahReid, line 94: This change is reverting back to what's in develop

src/chains/ethereum/ethereum/package.json
```diff
@@ -50,7 +51,6 @@
     "tooling"
   ],
   "dependencies": {
-    "@ethereumjs/block": "4.0.1",
```
MicaiahReid, line 53: This dependency wasn't used anywhere

src/packages/ganache/api-extractor.json
```diff
@@ -347,9 +348,6 @@
         // api-extractor _really_ wants to complain about this one, so much so
         // that setting the `"default"` to "none" (above) has no affect!
         "logLevel": "none"
-      },
-      "ae-wrong-input-file-type": {
-        "logLevel": "none"
```
MicaiahReid, line 352: We previously added this flag because we were updating the api-extractor version. Now we aren't, so we can remove

src/packages/ganache/api-extractor.json
```diff
@@ -79,6 +79,7 @@
     "@ethereumjs/blockchain",
     "@ethereumjs/trie",
     "@ethereumjs/ethash",
+    "@ethereumjs/rlp",
```
MicaiahReid, line 82: I don't know if we need this (or some of the other) `@ethereumjs` packages, but I thought I'd be thorough. Do we think I should generate types one by one with each of these to see what's actually needed?

src/chains/ethereum/ethereum/patches/@ethereumjs+block+4.0.1.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/node_modules/@ethereumjs/block/dist/block.d.ts b/node_modules/@ethereumjs/block/dist/block.d.ts
+index 4614cd8..1e33678 100644
+--- a/node_modules/@ethereumjs/block/dist/block.d.ts
++++ b/node_modules/@ethereumjs/block/dist/block.d.ts
+@@ -1,6 +1,13 @@
+ /// <reference types="node" />
+ import { Trie } from '@ethereumjs/trie';
+-import { ethers } from 'ethers';
++// we aren't actually using ethers, so we don't want to have to add
```
MicaiahReid:
```suggestion
+// we aren't actually using ethers, so we don't want to have to add it
```

PR #3971, opened 2022-11-30
feat: add support for the `merge` hardfork

### LINKED ISSUE #2006: Remove Node 12 support and add Node 18
Blocked by Node v18 actually being released and Node v12 reaching EOL. :-) This should be possible in May of 2022, as long as Truffle can follow suit.

### LINKED ISSUE #3578: Rename `Account.stateRoot` to `storageRoot`

### LINKED ISSUE #3693: Upgrade of Ethereum JS to v6 has introduced technical debt
Various changes to types and member accessibility in Ethereum JS has required changes to Ganache. In #3656, an amount of technical debt was introduced in order to support the merge.

Including, but not limited to:
- `any` type assertions
- private / protected members accessed from outside of the instance
- weird and convoluted shimming / facade wrapping of level database stuff
- `@ts-ignore`

MicaiahReid:
This release (re)introduces support for the Merge! In preparation for the merge, our friends at [ethereumjs](https://github.com/ethereumjs/ethereumjs-monorepo) were working hard on a huge new release with quite a few breaking changes. Because we rely so heavily on the work that ethereumjs does, these breaking changes impacted just about every nook and cranny of the Ganache code base. We wanted to get the merge features to our users as soon as possible, while also doing our due diligence to make sure the sweeping changes made by the ethereumjs upgrade don't affect Ganache's stability. With all that in mind, we released [Ganache v7.4.4-alpha.0](https://github.com/trufflesuite/ganache/releases/tag/v7.4.4-alpha.0)  back in September with support for the Merge. This PR brings that set of features to our stable release.

This introduces a few important changes and features that are worth explaining.

## Important Changes
 - Upgrading ethereumjs required dropping support for Node.js v12. In Ganache v7.0.0 we gave a deprecation notice for Node.js v12, and this release finalizes the removal of support for this version. To use Ganache, you'll need to install Node.js v14 or greater.

## New Feature
The only new feature that could potentially impact our users is the introduction of the `finalized` and `safe` block tags. These can be used with `eth_getBlockByNumber`, or any other RPC method that receives a block tag:
```TypeScript
const finalizeBlock = await provider.request({ method: "eth_getBlockByNumber", params: ["finalized", true] } );
const safeBlock = await provider.request({ method: "eth_getBlockByNumber", params: ["safe", true] } );
```
Note: These new block tags are aliases for the existing `latest` block tag. 

Fixes #2006, fixes #3693, fixes #3578


src/chains/ethereum/console.log/README.md
```diff
@@ -49,7 +49,7 @@ import { ConsoleLogs, maybeGetLogs } from "@ganache/console.log";
 // ... initialize vm ...
 
 // vm should be an instance of @ethereumjs/vm
-vm.on("step", event => {
+vm.evm.on("step", event => {
```
MicaiahReid:
```suggestion
vm.evm.events.on("step", event => {
```
By the way, @davidmurdoch, is this a breaking change?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -915,8 +914,15 @@ export default class EthereumApi implements Api {
     const parentHeader = parentBlock.header;
     const options = this.#options;
 
-    const generateVM = () => {
-      return blockchain.vm.copy();
+    const generateVM = async () => {
+      // note(hack): blockchain.vm.copy() doesn't work so we just do it this way
```
MicaiahReid, line 918: I just noticed this hack that has slipped through, though I don't know if I'd call it a hack... More a just a hit against performance for gas estimation. Thoughts on removing comment and/or making an issue to look into it at a later date?
@davidmurdoch @tenthirtyone @jeffsmale90

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -597,6 +609,11 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
   ) => {
     const nextBlock = this.#readyNextBlock(this.blocks.latest, timestamp);
 
+    // if block time is incremental, adjustments should only apply once,
+    // otherwise they accumulate with each block.
+    if (this.#options.miner.timestampIncrement !== "clock") {
+      this.#timeAdjustment = 0;
+    }
```
```suggestion
```
This is duplicating code below and can be removed. Seems like a rebase artifact.

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -93,16 +93,16 @@ function hookEventSystem(
 
 type Primitives = string | number | null | undefined | symbol | bigint;
 
-// Externalize changes any `Quantity`, `Data`, `ITraceData` types into `string`
-// as that's how they are after being serialized to JSON. It's be nice if
-// `JSON.stringify` did that for us, as our types implement `toJSON()`, but it
-// doesn't
+// Externalize changes any `Quantity`, `Data`, `ITraceData`, `Address` types
+// into `string` as that's how they are after being serialized to JSON. It's be
```
MicaiahReid:
```suggestion
// into `string` as that's how they are after being serialized to JSON. It'd be
```

src/chains/ethereum/address/tests/index.test.ts
```diff
@@ -42,23 +37,18 @@ describe("@ganache/ethereum-address", () => {
...
 
       assert.deepStrictEqual(bufferAddress, expected);
     });
 
-    it("should truncate an address to the specified length", () => {
-      const address = new Address("0x2104859394604359378433865360947116707876");
-      const stringifiedAddress = address.toString(1);
-
-      assert.strictEqual(stringifiedAddress, "0x21");
-    });
```
jeffsmale90: This test _could_ still make sense in a modified state:

If you pass a string that's longer than 20 bytes long, you would expect the call to toString() to return a 42 char long result.

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -113,25 +98,25 @@ export class RuntimeBlock {
     timestamp: Quantity,
     difficulty: Quantity,
     previousBlockTotalDifficulty: Quantity,
-    baseFeePerGas?: bigint
+    mixHash: Buffer,
+    baseFeePerGas: bigint | undefined
```
jeffsmale90, line 102: Not _strictly_ your change, but:

When calling this constructor, we are converting Quantity to Buffer, and _in_ the constructor we are using `Quantity.toBigInt()` to convert the buffer to BigInt. If you change the parameter types for gasLimit and gasUsed to `Quantity`, we can just call `toBigInt()` directly on that value.

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -113,25 +98,25 @@ export class RuntimeBlock {
...
   ) {
     const ts = timestamp.toBuffer();
```
jeffsmale90: we convert timestamp (`Quantity`) to `Buffer` here, and to `BigInt` on :116 - we could just go directly from `Quantity` to `BigInt`.

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -113,25 +98,25 @@ export class RuntimeBlock {
...
-      timestamp: new BnExtra(ts),
-      baseFeePerGas:
-        baseFeePerGas === undefined
-          ? new BnExtra(BUFFER_ZERO)
-          : new BnExtra(Quantity.toBuffer(baseFeePerGas))
+      ),
+      gasLimit: Quantity.toBigInt(gasLimit),
+      gasUsed: Quantity.toBigInt(gasUsed),
+      timestamp: Quantity.toBigInt(ts),
+      baseFeePerGas: baseFeePerGas === undefined ? 0n : baseFeePerGas,
```
jeffsmale90: I think this could be simplified to `baseFeePerGas || 0n`

src/chains/ethereum/console.log/src/maybe-get-logs.ts
```diff
@@ -35,7 +30,7 @@ type LogsStack = [
 export const maybeGetLogs = (event: {
   opcode: { name: string };
   memory: Buffer;
-  stack: BN[];
+  stack: bigint[];
```
jeffsmale90, line 33: Is this event expected to be used externally? If so, is this also a breaking change?

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: I think an issue would be reasonable.

We checkpoint and revert in `binSearch()` (but not in `exactimate()` [which is an excellent name]) - is this not enough? Why do we need a separate vm?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -935,24 +941,29 @@ export default class EthereumApi implements Api {
       const block = new RuntimeBlock(
         Quantity.from((parentHeader.number.toBigInt() || 0n) + 1n),
         parentHeader.parentHash,
-        parentHeader.miner,
+        new Address(parentHeader.miner.toBuffer()),
```
jeffsmale90, line 944: Should `parentHeader.miner` be of type `Address`?

src/chains/ethereum/utils/src/things/account.ts
```diff
@@ -15,7 +15,7 @@ export class Account {
   public balance: Quantity;
   public privateKey: Data;
   public nonce: Quantity;
-  public stateRoot: Buffer = KECCAK256_RLP;
+  public storageRoot: Buffer = KECCAK256_RLP;
   public codeHash: Buffer = KECCAK256_NULL;
 
   constructor(address: Address) {
```
jeffsmale90, line 21: I think this PR should be marked as fixing #3578 ❤️

src/chains/filecoin/filecoin/package.json
```diff
@@ -9,7 +9,7 @@
   "homepage": "https://github.com/trufflesuite/ganache/tree/develop/src/filecoin#readme",
   "license": "MIT",
   "engines": {
-    "node": ">=12.17.0 <17.0.0",
+    "node": ">=12.17.0 <14.0.0 || >=14.0.1 <17.0.0",
```
jeffsmale90, line 12: Should we still be supporting 12.17 here?

src/chains/ethereum/ethereum/tests/merge.test.ts
```diff
@@ -0,0 +1,124 @@
...
+          from: accounts[0],
+          to: contractAddress,
+          gasLimit: "0x2fefd8",
+          data: `0x${contract.contract.evm.methodIdentifiers["getCurrentDifficulty()"]}`
+        }
+      ]);
+      const receipt = await provider.send("eth_getTransactionReceipt", [
+        txHash
+      ]);
+      // make sure it didn't revert, which is will do if `difficulty` is not > `1`
```
MicaiahReid:
```suggestion
      // make sure it didn't revert, which it will do if `difficulty` is not > `1`
```

src/chains/ethereum/console.log/README.md
davidmurdoch: No, this isn't our VM. The example requires the use provider their own instance.

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: Let's add a new issue for it

package.json
```diff
@@ -3,7 +3,7 @@
   "author": "David Murdoch <david@trufflesuite.com> (https://davidmurdoch.com)",
   "private": true,
   "engines": {
-    "node": ">=12.0.0",
+    "node": ">=14.0.0",
     "npm": ">=6.12.0"
```
davidmurdoch:
```suggestion
    "npm": ">=6.14.4"
```

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: Done, #3980

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -306,26 +299,19 @@ export default class BlockManager extends Manager<Block> {
...
-          });
-      });
+        for await (const data of this.base.createValueStream()) {
+          const block = new Block(data as Buffer, this.#common);
+          if (
+            !latest ||
+            block.header.number.toBigInt() > latest.header.number.toBigInt()
+          ) {
+            latest = block;
+          }
```
jeffsmale90, line 311: I made this https://github.com/trufflesuite/ganache/issues/3991 to look into reducing the number of calls to `toBigInt()` while not blowing out this review.

src/chains/ethereum/ethereum/src/helpers/run-call.ts
```diff
@@ -68,28 +65,21 @@ export function runCall(
   transaction: SimulationTransaction,
   gasLeft: bigint
 ) {
-  const caller = { buf: transaction.from.toBuffer() };
+  const caller = new Address(transaction.from.toBuffer());
```
jeffsmale90: I think we can lose this line entirely, as `transation.from` is already an `Address` 😍

src/chains/ethereum/ethereum/src/helpers/run-call.ts
```diff
@@ -68,28 +65,21 @@ export function runCall(
...
   const to =
-    transaction.to == null ? undefined : { buf: transaction.to.toBuffer() };
-  const value = new BN(
-    transaction.value == null ? 0 : transaction.value.toBuffer()
-  );
+    transaction.to == null ? null : new Address(transaction.to.toBuffer());
```
jeffsmale90: I think we can lose this guy too

src/chains/ethereum/ethereum/src/helpers/trie.ts
```diff
@@ -1,33 +1,50 @@
...
+
+export class GanacheTrie extends Trie {
   public readonly blockchain: Blockchain;
+  /**
+   * The database that's returned from this.database() does not have all of
+   * the types of the original input database because ethereumjs doesn't use
+   * generics for their types in the underlying `CheckpointDB`. So, we store the
+   * original db on our trie so we can access those types.
+   */
+  public readonly db: TrieDB;
```
jeffsmale90, line 18: I'm a little uncomfortable with maintaining a second reference here. We _could_ implement an accessor that returns the underlying property, but it appears to be slower. Although I don't know how much performance would matter here.

```typescript
public get db(): TrieDB {
  return super._db as TrieDB;
}
```

:shrug:

src/packages/rlp/tests/dataTypes.test.ts
```diff
@@ -40,33 +40,17 @@ describe("RLP encoding (string):", function () {
...
-      )
-    );
-    assert.strictEqual(72, encodedLongString.length);
-    assert.strictEqual(RLP.getLength(encodedLongString), 2);
-    assert.strictEqual(encodedLongString[0], 184);
-    assert.strictEqual(encodedLongString[1], 70);
-    assert.strictEqual(encodedLongString[2], 122);
-    assert.strictEqual(encodedLongString[3], 111);
-    assert.strictEqual(encodedLongString[12], 53);
-  });
```
jeffsmale90: Why is this test no longer valid?

jeffsmale90 (changes requested):
In progress - 77/133 files viewed :/

src/chains/ethereum/address/tests/index.test.ts
MicaiahReid: Added a test for this!

src/chains/ethereum/block/src/runtime-block.ts
MicaiahReid: Yay node 14!

src/chains/ethereum/console.log/src/maybe-get-logs.ts
davidmurdoch: The `@ganache/console.log` package is pre-1.0, which means every new version is allowed breaking changes.

MicaiahReid: Maybe? Thoughts, @davidmurdoch?

Regardless, the console.log package is at version `0.2.0`, so we can break freely

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: I think it could be! I've made #3994

src/chains/ethereum/utils/src/things/account.ts
MicaiahReid: Yeah boi!!

src/chains/ethereum/ethereum/src/helpers/trie.ts
MicaiahReid: I'm not super opinionated on this. @davidmurdoch, @tenthirtyone, thoughts?

src/packages/rlp/tests/dataTypes.test.ts
MicaiahReid: I just didn't like the look of it.

MicaiahReid: I actually don't know - the test passes just fine. @davidmurdoch you did most of the RLP work, do you know why I would have deleted this?

src/chains/filecoin/filecoin/package.json
MicaiahReid: Hmm, this one was @davidmurdoch's change. Thoughts?

src/chains/ethereum/ethereum/src/helpers/trie.ts
davidmurdoch: I think we should have our own `.db` reference. Relying further on EthereumJS private members (or public ones, for that matter) just leads to pain and agony.

src/chains/filecoin/filecoin/package.json
davidmurdoch: filecoin doesn't work in node 14.0.0. This was just to make CI happy. I'm okay with leaving it as untouched as possible.

src/packages/rlp/tests/dataTypes.test.ts
davidmurdoch: The RLP library changed and it no longer has a `getLength` property. This test was testing `getLength` itself, the other assertions were just checks, and those cases are all tested elsewhere.

MicaiahReid: > This test was testing `getLength` itself

That's not immediately clear to me, isn't this testing that RLP correctly handles longer strings?

davidmurdoch: I did some digging and I take it back. I don't know why I removed this one!

davidmurdoch: I've put it back here: https://github.com/trufflesuite/ganache/pull/3971/commits/295e95196efafddd880de43d374bd02632a315af

I hope it works

src/chains/ethereum/console.log/src/maybe-get-logs.ts
jeffsmale90: I think breaking console package is fine, but if updating the console dep in ganache results in breaking _ganache_ changes, that'd be a breaking change for the `ganache` package too right?

davidmurdoch: It would. But it doesn't break anything, I don't think 🙃

src/chains/ethereum/ethereum/src/helpers/trie.ts
jeffsmale90: You could say the same about the inverse though. The `._db` property is protected, so it's _intended_ to be used by sub classes.

I can imagine a scenario where `super._db` is modified in the constructor to be different to the value passed as an argument, making it really difficult to debug issues, where we expect the super class to be acting on the database instance referred to by our own `.db` property.

But, again 🤷 , I'm not super opinionated on it, so happy if you guys disagree.

MicaiahReid: When there are a few conflicting not-very-opinionated opinions, I tend to go with the "do nothing" route :)

Let's keep it as-is for now.

davidmurdoch: @jeffsmale90 We tried to get ethereumjs to make the db generic so it returns the original type it was given, but they do not won't want to support that. So I think the argument your presented _against_ using `._db` is very valid.

davidmurdoch (approved):
LETS DO THIS (after detach mode is released)

tenthirtyone (approved):

jeffsmale90 (approved):

src/chains/ethereum/console.log/src/maybe-get-logs.ts
cds-amal: Thanks for the heads up @MicaiahReid. I'll test this once the Truffle release ships. Truffle pins [@ganache/console.log](https://github.com/trufflesuite/truffle/blob/a0a005c68e6133d455e39d19623313482a555eed/packages/resolver/package.json#L30)

PR #4013, opened 2022-12-10
docs: expand on troubleshooting node-gyp errors

legobeat:
node running as root is a common gotcha.

While there are workarounds that make that work (`npm config set user 0`), it seems better to guide towards dropping privileges.


CONTRIBUTING.md
```diff
@@ -18,11 +18,13 @@ If installation fails due to a `node-gyp` issue you may need to perform some add
...
 ### on Linux (Ubuntu-based)
 
+- Make sure `npm` commands are not run as `root`.
 - Determine if you have Python 2.7 installed
   - example: `which python2.7`
 - If you do not have Python 2.7 installed, you need to install it
   - example: `sudo apt update && sudo apt install python2.7`
-- Finally, run `npm config set python python2.7`
+- Run `npm config set python python2.7`
+- If you still get errors, [read here](https://blog.logrocket.com/solving-common-issues-node-gyp/).
```
davidmurdoch: I don't think we should link out to a 3rd party blog, as we can't trust that the author of the blog will not use our endorsement to trick users.

davidmurdoch (changes requested):

CONTRIBUTING.md
legobeat:
```suggestion
```

davidmurdoch (approved):

MicaiahReid (approved):

PR #4030, opened 2022-12-14
ci: update docker release action

davidmurdoch:
Some of the GitHub Actions commands used by our Docker Publish Action are deprecated. This PR updates the Actions to versions that have since been updated to stop using these deprecated commands.

### DIFF
--- a/.github/actions/docker-publish/action.yml
+++ b/.github/actions/docker-publish/action.yml
@@ -34,13 +34,13 @@ runs:
       shell: bash
 
     - name: Log in to Docker Hub
-      uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+      uses: docker/login-action@f75d088332b07a08afadf6ac53c74509b9453f12
       with:
         username: ${{ inputs.DOCKER_USERNAME }}
         password: ${{ inputs.DOCKER_ACCESS_TOKEN }}
 
     - name: Log in to the Container registry
-      uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+      uses: docker/login-action@f75d088332b07a08afadf6ac53c74509b9453f12
       with:
         registry: ghcr.io
         username: ${{ github.actor }}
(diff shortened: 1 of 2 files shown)


davidmurdoch:
I "tested" this by disabling npm publishing and Docker push in our release process, then forced the release process to run. Here is the commit to do it: https://github.com/trufflesuite/ganache/pull/4030/commits/9b8b8c8021d022fb8b3b239721ce118db9c1989d, and here is the Action run: https://github.com/trufflesuite/ganache/actions/runs/3696605217/jobs/6260524322

.github/workflows/release.yml
```diff
@@ -177,6 +177,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          ref: v${{ needs.release.outputs.VERSION }}
```
MicaiahReid, line 183: Previously this job was checking out the repo at the _time of running this action_, which is _before_ we've built, committed the new version, and pushed those tags. By specifying the tag of this version as the ref, we're making sure the version we just release to npm is being released to Docker.

MicaiahReid: I've tested this change here:
https://github.com/trufflesuite/ganache/actions/runs/3736450089/jobs/6340773768
(I realized I force-pushed to undo the test changes, so I can't point to the code that was used for the test)

MicaiahReid (approved):

.github/actions/docker-publish/action.yml
```diff
@@ -34,34 +34,34 @@ runs:
...
+      uses: docker/setup-qemu-action@38836d31bd1e0f8f29b8c7ad0412da32020f15b3
 
     - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@91cb32d715c128e5f0ede915cd7e196ab7799b83
+      uses: docker/setup-buildx-action@39a1a82492fd1ad19af19d61b5f748e4cb6cd1af
 
     - name: Build and push Docker image
-      uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+      uses: docker/build-push-action@175d02bffea74695e96b351069ac938b338802f9
       with:
```
jeffsmale90, line 65: I guess risk is fairly low, but if I'm interpreting this correctly, these are `master` commits that are _not_ tagged releases. We might want to consider switching to tagged releases at some stage 🤷

jeffsmale90 (approved):

.github/actions/docker-publish/action.yml
davidmurdoch: Thoughts on this perspective? https://blog.natterstefan.me/why-you-should-not-use-the-version-tag-or-branch-name-when-using-github-actions

jeffsmale90: That's a very interesting take, and definitely makes sense. As an addendum, I imagine it would make sense to use the commit hash of a tagged release - that way the author is stating they intend for this revision to be used (and assumedly it is more stable than intermediate commits :shrug:)

davidmurdoch: I did read through all of the commits (ok... I read their titles) since their previous releases and everything seemed innocuous enough.

PR #4036, opened 2022-12-15
docs: complete macOS node-gyp troubleshooting steps in CONTRIBUTING.md

DiscRisknBisque:
Does what it says on the tin.


CONTRIBUTING.md
```diff
@@ -31,7 +31,14 @@ If installation fails due to a `node-gyp` issue you may need to perform some add
 
 ### on macOS
 
-- I have no idea.
+- Determine if you have Xcode command line tools installed
+  - example the console will tell you if they're already installed): `xcode-select --install`
```
davidmurdoch: Can you make the steps more prescriptive? i.e., tell the user what to do directly.

davidmurdoch (changes requested):

CONTRIBUTING.md
DiscRisknBisque: This style (minus the missing "(") matches the other instructions. Would you like me to modify those too? They're short so it wouldn't take long.

CONTRIBUTING.md
```diff
@@ -31,14 +31,17 @@ If installation fails due to a `node-gyp` issue you may need to perform some add
...
-- Determine if you have Xcode command line tools installed
-  - example the console will tell you if they're already installed): `xcode-select --install`
-- Determine if you have Python 2.7 installed
-  - example: `which python2.7`
-- If you do not have Python 2.7 installed, you need to install it. We recommend `pyenv` for this so you can keep the default MacOS python installation untouched.
+- Attempt to install Xcode command line tools (the console will tell you if they're already installed)
+  - example: `xcode-select --install`
+- Determine if you have Python 2.x installed
+  - example: `which python`
+- If you do not have Python 2.x installed, you need to install it. We recommend `pyenv` for this so you can keep the default MacOS python installation untouched.
```
davidmurdoch: I don't know Python things... but does this need to be Python `2.7.x`?

davidmurdoch:
```suggestion
- If you do not have Python 2.x installed, you need to install it. We recommend `pyenv` for this so you can keep the default macOS Python installation untouched.
```

jeffsmale90: `which python` won't necessarily tell you which version is installed (for example, if you use pyenv, it'll just show you the shim).

`python --version` is probably more useful.

jeffsmale90 (changes requested):

MicaiahReid (approved):

CONTRIBUTING.md
```diff
@@ -16,22 +16,39 @@
 
 If installation fails due to a `node-gyp` issue you may need to perform some additional system configuration.
 
-### on Linux (Ubuntu-based)
+Ganache uses [node-gyp v7.1.2](https://github.com/nodejs/node-gyp/tree/v7.1.2) as part of it's build system, which requires Python v2.7, v3.5, v3.6, v3.7, or v3.8 to be installed on the system.
```
davidmurdoch:
```suggestion
Ganache uses [node-gyp v7.1.2](https://github.com/nodejs/node-gyp/tree/v7.1.2) as part of its build system, which requires Python v2.7, v3.5, v3.6, v3.7, or v3.8 to be installed on the system.
```

jeffsmale90 (approved):

davidmurdoch (approved):

davidmurdoch:
image

@MicaiahReid <3

PR #4045, opened 2022-12-15
fix: add `"EADDRINUSE"` as `code` to server `Error`

### LINKED ISSUE #4020: EADDRINUSE error is uncoded, most of the time
Occupying the port defined by `server.port` results in an error:

```javascript
{
  code: undefined,
  message: 'listen EADDRINUSE: address already in use 127.0.0.1:5001'
}
```

Whereas occupying the port defined by `chain.ipfsport` (with `flavor:filecoin`) results in an error:
```javascript
{
  code: 'EADDRINUSE',
  message: 'listen EADDRINUSE: address already in use 127.0.0.1:5001'
}
```

Reproduced with Ganache 7.5.0 and @ganache/filecoin 0.1.5
```javascript
const g = require("ganache");

const handler = (e) => {
  console.error({
    code: e.code,
    message: e.message,
  });
  process.exit(1);
};

const options = {
  flavor: "filecoin",
  server: { port: 8545, host: "127.0.0.1" },
  chain: { ipfsPort: 5001 },
};

g.server(options)
  .listen(options.server.port, options.server.host)
  .catch(handler);
```

davidmurdoch:
This PR sets the `code` property related to the error `"listen EADDRINUSE: address already in use { address }"` to "EADDRINUSE", making detecting this kind of "address in use" error programmatically much easier and straightforward. This property existed in Ganache v6 and earlier and was removed in v7.0.0 in error.

Previously:

```json
{
  "message": "listen EADDRINUSE: address already in use 127.0.0.1:5001"
}
```

Now:

```json
{
  "code": "EADDRINUSE",
  "message": "listen EADDRINUSE: address already in use 127.0.0.1:5001"
}
```

Fixes #4020

### DIFF
--- a/src/packages/core/src/server.ts
+++ b/src/packages/core/src/server.ts
@@ -245,7 +245,7 @@ export class Server<
             `listen EADDRINUSE: address already in use ${
               host || DEFAULT_HOST
             }:${portNumber}.`
-          ) as any;
+          ) as NodeJS.ErrnoException;
           // emulate part of node's EADDRINUSE error:
           err.code = "EADDRINUSE";
           throw err;
--- a/src/packages/core/src/server.ts
+++ b/src/packages/core/src/server.ts
@@ -245,7 +245,9 @@ export class Server<
             `listen EADDRINUSE: address already in use ${
               host || DEFAULT_HOST
             }:${portNumber}.`
-          );
+          ) as any;
+          // emulate part of node's EADDRINUSE error:
+          err.code = "EADDRINUSE";
           throw err;
         }
       })
--- a/src/packages/core/tests/server.test.ts
+++ b/src/packages/core/tests/server.test.ts
@@ -493,10 +493,10 @@ describe("server", () => {
       server.listen(port);
 
       try {
-        await assert.rejects(
-          setup(),
-          `Error: listen EADDRINUSE: address already in use 127.0.0.1:${port}.`
-        );
+        await assert.rejects(setup(), {
+          message: `Error: listen EADDRINUSE: address already in use 127.0.0.1:${port}.`,
+          code: "EADDRINUSE"
+        });
       } finally {
         await Promise.all([
           teardown(),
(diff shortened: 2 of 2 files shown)


src/packages/core/tests/server.test.ts
```diff
@@ -493,10 +493,10 @@ describe("server", () => {
       server.listen(port);
 
       try {
-        await assert.rejects(
-          setup(),
-          `Error: listen EADDRINUSE: address already in use 127.0.0.1:${port}.`
-        );
+        await assert.rejects(setup(), {
+          message: `Error: listen EADDRINUSE: address already in use 127.0.0.1:${port}.`,
```
davidmurdoch:
```suggestion
          message: `listen EADDRINUSE: address already in use 127.0.0.1:${port}.`,
```

src/packages/core/src/server.ts
```diff
@@ -255,7 +255,9 @@ export class Server<
             `listen EADDRINUSE: address already in use ${
               host || DEFAULT_HOST
             }:${portNumber}.`
-          );
+          ) as any;
+          // emulate part of node's EADDRINUSE error:
+          err.code = "EADDRINUSE";
```
davidmurdoch, line 250: So the actual error has a bunch of other fields in it, like a platform specific `errno`, a system call (`"listen"`), `address`, `port`, and more.

I don't mind adding the ones that make sense (some we can't add, because they use nodejs internals), but they aren't all necessary to fix the issue with Ganache UI.

jeffsmale90: From my POV, they don't bring necessarily bring any value, as that information would probably be accessible to the caller anyways.

src/packages/core/src/server.ts
```diff
@@ -255,7 +255,9 @@ export class Server<
             `listen EADDRINUSE: address already in use ${
               host || DEFAULT_HOST
             }:${portNumber}.`
-          );
+          ) as any;
```
jeffsmale90: If you like, this type assertion could be to `NodeJS.ErrnoException`, but I don't think I care strongly either way.

jeffsmale90 (approved):

src/packages/core/src/server.ts
davidmurdoch: fixed!

tenthirtyone (approved):

PR #4056, opened 2022-12-17
fix: handle `eth_gasEstimate` reverts correctly to prevent potential process crash

davidmurdoch:
When updating to support the Merge hardfork code that handled `eth_gasEstimate` was refactored. A code block that previously "short circuited" the function no longer did, causing a JavaScript Promise to be fullfilled with a _handled_ rejection, but then also an _unhandled_ rejection.

We've fixed the short circuit code and added a test to prevent this from regressing in the future.

### DIFF
--- a/src/chains/ethereum/ethereum/src/api.ts
+++ b/src/chains/ethereum/ethereum/src/api.ts
@@ -960,7 +960,7 @@ export default class EthereumApi implements Api {
         generateVM,
         runArgs,
         (err: Error, result: EstimateGasResult) => {
-          if (err) reject(err);
+          if (err) return void reject(err);
           resolve(Quantity.from(result.gasEstimate));
         }
       );
--- a/src/chains/ethereum/ethereum/tests/api/eth/estimateGas.test.ts
+++ b/src/chains/ethereum/ethereum/tests/api/eth/estimateGas.test.ts
@@ -1,0 +1,45 @@
+import assert from "assert";
+import getProvider from "../../helpers/getProvider";
+
+describe("api", () => {
+  describe("eth", () => {
+    describe("estimateGas", () => {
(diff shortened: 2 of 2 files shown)


jeffsmale90:
@davidmurdoch originally I was deploying a contract that throws in the constructor, but figured this was a simpler way to cause the estimator to fail. LMK what you think.

src/chains/ethereum/ethereum/tests/api/eth/estimateGas.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        try {
+          const estimatingGas = provider.request({
+            method: "eth_estimateGas",
+            params: [transaction]
+          });
+
+          await assert.rejects(estimatingGas);
+          await new Promise(resolve => setImmediate(resolve));
+
+          assert(
```
davidmurdoch, line 35: Oh wow, this rejection test was actually kind of tricky!

davidmurdoch: Wonder if there is a mocha flag to always fail if an Unhandled rejection occurs?

davidmurdoch:
Code coverage in the problematic function with @jeffsmale90's test is now:

image

MicaiahReid (approved):

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -960,7 +960,7 @@ export default class EthereumApi implements Api {
         generateVM,
         runArgs,
         (err: Error, result: EstimateGasResult) => {
-          if (err) reject(err);
+          if (err) return void reject(err);
```
tenthirtyone, line 963: I think you are safe to `return reject(err)` if you want to.

tenthirtyone (approved):

PR #4069, opened 2022-12-20
fix: make empty string mnemonic use default path

davidmurdoch:
For reviewers:

I don't know if this is the best approach. Maybe we should throw instead?

### DIFF
--- a/src/chains/ethereum/options/src/wallet-options.ts
+++ b/src/chains/ethereum/options/src/wallet-options.ts
@@ -263,7 +263,17 @@ export const WalletOptions: Definitions<WalletConfig> = {
     conflicts: ["mnemonic", "deterministic"]
   },
   mnemonic: {
-    normalize,
+    // if the mnemonic is empty, we want to throw an error because it is likely
+    // unexpected behavior
+    normalize: (mnemonic, config) => {
+      // if the mnemonic is `""` use `default` mnemonic generation.
+      if (mnemonic === "") {
+        // NOTE: the `flavor` param here isn't used by `default` but is
+        // required by the `default` function signature, so we pass in ""
+        return WalletOptions.mnemonic.default(config, "");
+      }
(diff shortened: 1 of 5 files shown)


MicaiahReid (changes requested):
@davidmurdoch just a heads up, it looks like this isn't passing CI

src/chains/ethereum/options/src/wallet-options.ts
```diff
@@ -263,7 +263,17 @@ export const WalletOptions: Definitions<WalletConfig> = {
     conflicts: ["mnemonic", "deterministic"]
   },
   mnemonic: {
-    normalize,
+    // if the mnemonic is empty, we want to throw an error because it is likely
+    // unexpected behavior
+    normalize: (mnemonic, config) => {
+      // if the mnemonic is `""` use `default` mnemonic generation.
+      if (mnemonic === "") {
```
davidmurdoch, line 270: I'm not so sure this is even a good idea.

We probably should _validate_ mnemonics... and `""` is actually an invalid mnemonic.

Maybe we should close this PR.

jeffsmale90: the library we use exposes the function `validateMnemonic()`, so I vote that we do that (either in this PR, or a separate one), and fail if an empty string is provided (unless there's some interface where it would be difficult to _not_ provide an empty string 🤷)

davidmurdoch: Yeah, `validateMnemonic` is quite the busy function. Reading it at a glance it looks like it'd be costly, but even if it is most users will only need to call it once per start up, so it's probably fine?

I think it should be in another PR, so I'll close this one. Issue created: https://github.com/trufflesuite/ganache/issues/4075

davidmurdoch:
Closing as this was a bad idea :badger:

Issue #3805, opened 2022-10-15
0 should be a supported port


sewinter:
1. 0 should be a supported port.
2. Yikes to this error message:

```
Error: Port should be >= 0 and < 65536. Received 0.
    at Server.listen (webpack://Ganache/../core/lib/src/server.js:157:25)
    at ForkManager.createNewServer (/Users/samuelwinter/src/backrun/src/fork/ForkManager.ts:60:12)
    at ForkManager.simulateTx (/Users/samuelwinter/src/backrun/src/fork/ForkManager.ts:26:35)
    at handleMempoolTx (/Users/samuelwinter/src/backrun/src/backrun/mempool.ts:54:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

davidmurdoch:
Hey, can you provide reproduction steps? Also, what version are you using?

sewinter:
Hi David, I'm on `7.4.4`. Repro:

```
 const server = ganache.server({
      fork: {
        url: FORK_URL,
      },
      server: {
        ws: true,
      },
    });
    server.listen(0);
```

davidmurdoch:
Looks like a duplicate of https://github.com/trufflesuite/ganache/issues/2643. I'll try to get the priority bumped up on this for you. I suspect it will be an easy fix. Thanks for reporting!

sewinter:
Thank you! I ended up writing a utility to find a free port, so it's not super urgent for me.

PR #4070, opened 2022-12-20
fix: make port `0` bind to any available port

### LINKED ISSUE #2643: Support binding to port 0
If you ask linux to bind to port 0, it will automatically pick an open port for you. This can be really helpful when running multiple instances of ganache on the same machine. Currently, ganache says `Invalid port number '0'`

Please make ganache support binding to port 0.

It would also be helpful (though not necessary) if it wrote this port to a file somewhere.

davidmurdoch:
Specifying a port of `0` is a way to tell a server that you want the Operating System to let you bind to any open port. Ganache claimed to support binding to port 0, but would rudely yell at you if you tried: `Port should be >= 0 and < 65536. Received 0` :facepalm:! This PR rights our wrongs.

Using port `0` is useful when you don't care which port you bind to, you just want an open port.

You can use port `0` in a few ways:

### On the cli:

```console
$ ganache --port 0
…
RPC Listening on 127.0.0.1:35006
```

### In detach mode:

$$\color{#8b949e}\textsf{note: port 0 binding in detach mode doesn't yet provide a scripting friendly way of returning the port}$$
```console
$ ganache --detach --port 0
frozen_caramel_doughnut

$ ganache instances list
┌───────┬─────────────────────────┬──────────┬─────────┬─────────────────┬────────┐
│   PID │ Name                    │ Flavor   │ Version │ Host            │ Uptime │
├───────┼─────────────────────────┼──────────┼─────────┼─────────────────┼────────┤
│ 58078 │ frozen_caramel_doughnut │ ethereum │ 7.7.3   │ 127.0.0.1:35006 │    14s │
└───────┴─────────────────────────┴──────────┴─────────┴─────────────────┴────────┘
```

### and programmatically:

```javascript
import ganache from "ganache";

const server = ganache.server();
await server.listen(0);
console.log("Listening on port: ", server.address().port);
```

Fixes #2643 and #3805

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Check bundle size
         # this should match the os and version used in the release.yml
-        if: startsWith(matrix.os, 'ubuntu-22.04') && startsWith(matrix.node, '14.')
+        if: startsWith(matrix.os, 'ubuntu-22.04') && startsWith(matrix.node, '14.') && matrix.node != '14.0.0'
         # 1. build ganache
         # 2. pack it into a tarball
         # 3. measure the _unpacked_ tarball's size
@@ -65,7 +65,7 @@ jobs:
           INFURA_KEY: "badc0de0deadc0debadc0de0deadc0de"
       - name: Upload artifact
         # this should match the os and version used in the release.yml
-        if: startsWith(matrix.os, 'ubuntu-22.04') && startsWith(matrix.node, '14.')
+        if: startsWith(matrix.os, 'ubuntu-22.04') && startsWith(matrix.node, '14.') && matrix.node != '14.0.0'
         uses: actions/upload-artifact@v3
         with:
           name: Candidate
--- a/src/packages/cli/src/cli.ts
+++ b/src/packages/cli/src/cli.ts
@@ -158,7 +158,7 @@ if (argv.action === "start") {
     // instance), so we need to notify that we are ready.
     const isDetachedInstance = process.send !== undefined;
     if (isDetachedInstance) {
-      notifyDetachedInstanceReady();
+      notifyDetachedInstanceReady(port);
     }
   });
 } else if (argv.action === "stop") {
--- a/src/packages/cli/src/cli.ts
+++ b/src/packages/cli/src/cli.ts
@@ -138,17 +138,18 @@ if (argv.action === "start") {
       return;
     }
     started = true;
+    const { address: host, port } = server.address();
     switch (flavor) {
       case FilecoinFlavorName: {
-        await initializeFilecoin(
-          server.provider as FilecoinProvider,
-          cliSettings
-        );
+        await initializeFilecoin(server.provider as FilecoinProvider, {
+          host,
+          port
+        });
         break;
       }
       case EthereumFlavorName:
       default: {
-        initializeEthereum(server.provider as EthereumProvider, cliSettings);
+        initializeEthereum(server.provider as EthereumProvider, { host, port });
         break;
       }
     }
--- a/src/packages/cli/src/detach.ts
+++ b/src/packages/cli/src/detach.ts
@@ -20,7 +20,6 @@ export type DetachedInstance = {
 };
 
 const FILE_ENCODING = "utf8";
-const READY_MESSAGE = "ready";
 const START_ERROR =
   "An error occurred spawning a detached instance of Ganache:";
 const dataPath = envPaths(`Ganache/instances`, { suffix: "" }).data;
@@ -32,10 +31,10 @@ function getInstanceFilePath(instanceName: string): string {
 /**
  * Notify that the detached instance has started and is ready to receive requests.
  */
-export function notifyDetachedInstanceReady() {
+export function notifyDetachedInstanceReady(port: number) {
   // in "detach" mode, the parent will wait until the "ready" message is
   // received before disconnecting from the child process.
-  process.send(READY_MESSAGE);
+  process.send(port);
 }
 
 /**
@@ -113,11 +112,14 @@ export async function startDetachedInstance(
   // event is emitted) will be streamed to stderr on the parent.
   child.stderr.pipe(process.stderr);
 
-  await new Promise<void>((resolve, reject) => {
-    child.on("message", message => {
-      if (message === READY_MESSAGE) {
-        resolve();
-      }
+  // Wait for the child process to send the "ready" message, which indicates
+  // that the Ganache server has started and is ready to receive RPC requests.
+  // It signals by sending the port number to which it was bound to back to us;
+  // this is needed because Ganache may bind to a random port if the user
+  // specified port 0.
+  const actualPort = await new Promise<number>((resolve, reject) => {
+    child.on("message", port => {
+      resolve(port as number);
     });
 
     child.on("error", err => {
@@ -159,7 +161,7 @@ export async function startDetachedInstance(
     pid,
     name: createInstanceName(),
     host,
-    port,
+    port: actualPort,
     flavor,
     cmd,
     version
--- a/src/packages/cli/src/detach.ts
+++ b/src/packages/cli/src/detach.ts
@@ -114,7 +114,7 @@ export async function startDetachedInstance(
 
   // Wait for the child process to send the "ready" message, which indicates
   // that the Ganache server has started and is ready to receive RPC requests.
-  // It signals by sending the port number to which it was bound to back to us;
+  // It signals by sending the port number to which it was bound back to us;
   // this is needed because Ganache may bind to a random port if the user
   // specified port 0.
   const actualPort = await new Promise<number>((resolve, reject) => {
--- a/src/packages/cli/src/detach.ts
+++ b/src/packages/cli/src/detach.ts
@@ -117,7 +117,7 @@ export async function startDetachedInstance(
   // It signals by sending the port number to which it was bound back to us;
   // this is needed because Ganache may bind to a random port if the user
   // specified port 0.
-  const actualPort = await new Promise<number>((resolve, reject) => {
+  const port = await new Promise<number>((resolve, reject) => {
     child.on("message", port => {
       resolve(port as number);
     });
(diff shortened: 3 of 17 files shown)


.github/workflows/pr.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
       - name: Check bundle size
         # this should match the os and version used in the release.yml
-        if: startsWith(matrix.os, 'ubuntu-20.04') && startsWith(matrix.node, '14.')
```
davidmurdoch: This stops the `check bundle size` run from happening twice; we don't want it to run on `14.0.0`

src/packages/cli/src/cli.ts
```diff
@@ -157,7 +158,7 @@ if (argv.action === "start") {
     // instance), so we need to notify that we are ready.
     const isDetachedInstance = process.send !== undefined;
     if (isDetachedInstance) {
-      notifyDetachedInstanceReady();
+      notifyDetachedInstanceReady(port);
```
davidmurdoch, line 161: < controversy >

src/packages/core/src/server.ts
```diff
@@ -33,6 +34,13 @@ import WebsocketServer, { WebSocketCapableFlavor } from "./servers/ws-server";
 import HttpServer from "./servers/http-server";
 import Emittery from "emittery";
 
+// not using the "net" node package in order to avoid having to polyfill this
+// for the browser build.
+// isIPv4 taken from https://github.com/nodejs/node/blob/01323d50c4b24cf730a651d06ba20633905ecbed/lib/internal/net.js#L31
+const v4Seg = "(?:[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])";
+const IPv4Reg = new RegExp(`^(${v4Seg}[.]){3}${v4Seg}$`);
+const isIPv4 = (s: string) => IPv4Reg.test(s);
```
davidmurdoch, line 42: While `server` isn't useful in the browser, our webpack step still processes it. I could have made it ignore the `net` package, or used a polyfill, but this code was so simple I didn't think it was necessary.

I think copy+paste is ultimately better than ignoring or partially polyfilling `net` because if we ever _do_ utilize the `net` package somewhere else the webpack process to succeed, but the actual probably won't work (because the polyfill is incomplete or just missing). Note: polyfilling the entire `net` package isn't possible, since we can't create actual sockets/pipes/etc in the browser.

You insights on this will be much appreciated!

src/packages/core/src/server.ts
```diff
@@ -43,13 +51,13 @@ export type Callback = (err: Error | null) => void;
...
- *  * closing: `status === Status.closing` or `status & Status.closing !== 0`
- *  * closed: `status === Status.closed` or `status & Status.closed !== 0`
- *  * closing || closed: `status & Status.closingOrClosed !== 0` or `status & (Status.closing | Status.closed) !== 0`
+ *  * ready: `status === ServerStatus.ready` or `status & ServerStatus.ready !== 0`
+ *  * opening: `status === ServerStatus.opening` or `status & ServerStatus.opening !== 0`
+ *  * open: `status === ServerStatus.open` or `status & ServerStatus.open !== 0`
+ *  * opening || open: `status & ServerStatus.openingOrOpen !== 0` or `status & (ServerStatus.opening | ServerStatus.open) !== 0`
+ *  * closing: `status === ServerStatus.closing` or `status & ServerStatus.closing !== 0`
+ *  * closed: `status === ServerStatus.closed` or `status & ServerStatus.closed !== 0`
+ *  * closing || closed: `status & ServerStatus.closingOrClosed !== 0` or `status & (ServerStatus.closing | ServerStatus.closed) !== 0`
```
davidmurdoch, line 60: This is just a drive-by fix up. A version of this PR used `ServerStatus`, and I had tried to copy paste from the examples here... but it didn't work because the names were wrong.

Let me know if you want this in another PR.

src/packages/core/src/server.ts
```diff
@@ -283,6 +292,19 @@ export class Server<
     }
   }
 
+  public address() {
```
davidmurdoch, line 297: This PR copies Node's own http `createServer`'s `address()` API: https://nodejs.org/api/net.html#net_server_address

With one exception: we do _not_ support named pipes and Unix domain sockets, so we do need to support returning a string.

src/packages/core/src/server.ts
```diff
@@ -283,6 +292,19 @@ export class Server<
...
+    if (this.#listenSocket) {
```
davidmurdoch, line 298: My first iteration of this checked the server's `#status` to see if we were "Open". This `this.#listenSocket` check is effectively the same (the two properties are always set together). I chose this check instead only because passing `null` to `us_socket_local_port` triggers what seems to be undefined behavior in the uWS library, so this is slightly more explicit and less prone to bugs due to future refactoring.

src/packages/core/tests/server.test.ts
```diff
@@ -369,8 +425,7 @@ describe("server", () => {
         "-0o1",
         "-0b1",
         "0o",
-        "0b",
-        0
+        "0b"
```
davidmurdoch, line 428: Yes, these are both an encoding of `0` in octal and binary, respectively, but we don't support that, and neither does node's `http.createServer.listen(<port>)`

src/packages/cli/src/detach.ts
```diff
@@ -113,11 +112,14 @@ export async function startDetachedInstance(
...
   child.stderr.pipe(process.stderr);
 
-  await new Promise<void>((resolve, reject) => {
-    child.on("message", message => {
-      if (message === READY_MESSAGE) {
-        resolve();
-      }
+  // Wait for the child process to send the "ready" message, which indicates
+  // that the Ganache server has started and is ready to receive RPC requests.
+  // It signals by sending the port number to which it was bound to back to us;
```
MicaiahReid:
```suggestion
  // It signals by sending the port number to which it was bound back to us;
```
I think?

src/packages/core/src/server.ts
jeffsmale90: I think this is fine

src/packages/cli/src/detach.ts
```diff
@@ -159,7 +161,7 @@ export async function startDetachedInstance(
     pid,
     name: createInstanceName(),
     host,
-    port,
```
MicaiahReid: Because we are no longer using the `port` variable, it can be deleted a few lines above

src/packages/core/src/server.ts
MicaiahReid: I'm cool with the approach. I'm particularly cool with _not_ making our webpack process any more complicated

src/packages/cli/src/cli.ts
jeffsmale90: I don't like this. Sending an object has _minimal_ additional overhead in terms of complexity and performance, and makes this statement much more obvious:

```
child.on("message", port => {
  resolve(port as number);
});
```

```
child.on("message", message => {
  if (message.message === READ_MESSAGE) {
    resolve(message.port as number);
  }
});
```

Granted, this area is not _super_ complicated, and the surrounding code is quite explanatory, but :shrug:./

Keen to hear other people's opinions...

jeffsmale90 (changes requested):

src/packages/core/tests/server.test.ts
```diff
@@ -336,15 +336,71 @@ describe("server", () => {
...
+          const address = s.address();
+          assert.notEqual(
+            address,
+            null,
+            "server.address() should not be null after listening"
+          );
+          assert.strictEqual(
+            address!.port,
+            +specificPort,
+            "server.address().port should be non-zero after server.listen(0)"
```
```suggestion
            "server.address().port should be equal to the specified port after server.listen(nonZeroPort)"
```
Or something like that, since this test isn't testing `server.listen(0)`

MicaiahReid (changes requested):

src/packages/cli/src/cli.ts
davidmurdoch: If we can't trust that `message` is what we expect it to be (since we have to check against `READY_MESSAGE`), then do we also need to check the entire payload? Basically:

```
try {
  if (message && typeof message === "object" && message.message === READ_MESSAGE && typeof message.port === "number" && isValidPort(message.port)) {
    resolve(message.port);
  }
} catch {
  // ignore unexpected messages
}
function isValidPort(port: number): boolean {
   // ...
}
```

.github/workflows/pr.yml
```diff
@@ -43,7 +43,7 @@ jobs:
 
       - name: Check bundle size
         # this should match the os and version used in the release.yml
-        if: startsWith(matrix.os, 'ubuntu-22.04') && startsWith(matrix.node, '14.')
+        if: startsWith(matrix.os, 'ubuntu-22.04') && startsWith(matrix.node, '14.') && matrix.node != '14.0.0'
```
davidmurdoch, line 46: This stops the check bundle size run from happening twice; we don't want it to run on 14.0.0

MicaiahReid (approved):
I think this is a fix because we claimed to support binding to port 0.

src/packages/cli/src/detach.ts
```diff
@@ -113,11 +112,13 @@ export async function startDetachedInstance(
...
 
-  await new Promise<void>((resolve, reject) => {
-    child.on("message", message => {
-      if (message === READY_MESSAGE) {
-        resolve();
-      }
+  // Wait for the child process to send its port, which indicates that the
+  // Ganache server has started and is ready to receive RPC requests. It signals
+  // by sending the port number to which it was bound back to us; this is needed
+  // because Ganache may bind to a random port if the user specified port 0.
```
jeffsmale90, line 118: 👍

src/packages/cli/src/cli.ts
jeffsmale90: After much discussions, I'm happy with the solution as is, with the updated comment in detach.ts

src/packages/core/tests/server.test.ts
```diff
@@ -336,15 +336,71 @@ describe("server", () => {
...
+      );
+      await s.listen(zeroPort);
+
+      try {
+        const address = s.address();
+        assert.notEqual(
+          address,
+          null,
+          "server.address() should not be null after listening"
+        );
```
jeffsmale90, line 387: Would it be worthwhile asserting that the value is a non-negative integer? Maybe not because we do ensure that sending a request there is successful 🤷

jeffsmale90 (approved):

PR #4112, opened 2023-01-04
fix: store transactions with the correct `effectiveGasPrice` 

### LINKED ISSUE #4094: transaction gasPrice does not equal baseFeePerGas + maxPriorityFeePerGas
Running on the latest version of ganache, `ganache v7.7.2 (@ganache/cli: 0.8.1, @ganache/core: 0.8.1)`

This could be my misunderstanding, but it appears that `gasPrice != min(maxPriorityFeePerGas + baseFeePerGas, maxFeePerGas)` when multiple transactions are mined in the same block.

command: `ganache-cli --chain.vmErrorsOnRPCResponse true --miner.blockTime 3600 --chain.chainId 1 -v --miner.timestampIncrement 12 --wallet.deterministic`

repro:
```txnHashes = []                                                                                                                                                                                                                                                     
accounts = w3.eth.accounts                                                                                                                                                                                                                                         
for i in range(1,4):                                                                                                                                                                                                                                               
    txnHashes.append(w3.eth.send_transaction({'from'                 : accounts[0],                                                                                                                                                                                
                                              'to'                   : accounts[1],                                                                                                                                                                                
                                              'value'                : w3.toWei('1', 'ether'),                                                                                                                                                                     
                                              'gas'                  : 21000,                                                                                                                                                                                      
                                              'maxPriorityFeePerGas' : w3.toWei(i , 'gwei'),                                                                                                                                                                       
                                              'maxFeePerGas'         : w3.toWei(30 , 'gwei')}))                                                                                                                                                                    
w3.provider.make_request('evm_mine', [])                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                   
for txnHash in txnHashes:                                                                                                                                                                                                                                          
    txn = w3.eth.get_transaction(txnHash)                                                                                                                                                                                                                          
    block = w3.eth.get_block(txn.blockNumber)                                                                                                                                                                                                                      
    print(f'{block.baseFeePerGas=} {txn.maxPriorityFeePerGas=} {txn.maxFeePerGas=} {txn.gasPrice=}')
```

Output:
```
block.baseFeePerGas=875000000 txn.maxPriorityFeePerGas=1000000000 txn.maxFeePerGas=30000000000 txn.gasPrice=1875000000
block.baseFeePerGas=875000000 txn.maxPriorityFeePerGas=2000000000 txn.maxFeePerGas=30000000000 txn.gasPrice=3000000000
block.baseFeePerGas=875000000 txn.maxPriorityFeePerGas=3000000000 txn.maxFeePerGas=30000000000 txn.gasPrice=4000000000
```
For the 2nd and 3rd transactions I expected the gasPrice to be 2875000000 and 3875000000 respectively.
When the transaction are mined individually, the invariant holds.

The example uses the web3.py library but the raw json is:
```
{'from': '0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1', 'to': '0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0', 'value': '0xde0b6b3a7640000', 'gas': '0x5208', 'maxPriorityFeePerGas': '0x3b9aca00', 'maxFeePerGas': '0x6fc23ac00'}
{'from': '0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1', 'to': '0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0', 'value': '0xde0b6b3a7640000', 'gas': '0x5208', 'maxPriorityFeePerGas': '0x77359400', 'maxFeePerGas': '0x6fc23ac00'}
{'from': '0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1', 'to': '0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0', 'value': '0xde0b6b3a7640000', 'gas': '0x5208', 'maxPriorityFeePerGas': '0xb2d05e00', 'maxFeePerGas': '0x6fc23ac00'}
```

And the raw responses are:
```
{'id': 12, 'jsonrpc': '2.0', 'result': {'type': '0x2', 'hash': '0x79f205173e9ff6878320b5a6167a4a07d3c65669fdf995293c88cb7aa79facf7', 'chainId': '0x1', 'nonce': '0x0', 'bloc
kHash': '0xcd96e8f2140ddce2e1aa86129fcd30d9f0ee33e6870af3a9190e589326568c19', 'blockNumber': '0x1', 'transactionIndex': '0x0', 'from': '0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1', 'to': '0xffcf8fdee72ac11b5c542428b35eef5769c409f0', 'value': '0xde0b6b3a764000
0', 'maxPriorityFeePerGas': '0x3b9aca00', 'maxFeePerGas': '0x6fc23ac00', 'gasPrice': '0x6fc23ac0', 'gas': '0x5208', 'input': '0x', 'accessList': [], 'v': '0x1', 'r': '0xf53bbc8b944d320b2d5917ac8b2bf1bd3e0e6d14c39e740c532198c7a4cb5f88', 's': '0x3fd99bceb278e1b
97ac98a9c7a67f6dcf9f10ae85fd4ffa255a8a1fce72bde2f'}}

{'id': 14, 'jsonrpc': '2.0', 'result': {'type': '0x2', 'hash': '0xf8c20f31c957d9a7418e71161b407b08bca2dd44e921439af3c8bd7c1157c8c2', 'chainId': '0x1', 'nonce': '0x1', 'bloc
kHash': '0xcd96e8f2140ddce2e1aa86129fcd30d9f0ee33e6870af3a9190e589326568c19', 'blockNumber': '0x1', 'transactionIndex': '0x1', 'from': '0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1', 'to': '0xffcf8fdee72ac11b5c542428b35eef5769c409f0', 'value': '0xde0b6b3a764000
0', 'maxPriorityFeePerGas': '0x77359400', 'maxFeePerGas': '0x6fc23ac00', 'gasPrice': '0xb2d05e00', 'gas': '0x5208', 'input': '0x', 'accessList': [], 'v': '0x0', 'r': '0x6d2e48588b35d7e1ce71db5eb9a7f2c94e8f6e061eca15242a2409c39440cee', 's': '0x158d1ec71e465519
d076bf8e592b3a455e5d0499a8ee229aa3da7a416b9991f7'}}

{'id': 16, 'jsonrpc': '2.0', 'result': {'type': '0x2', 'hash': '0xf1fbd2ace4bc613672fdec399afc301081c404451b84324ee880f79c88a84b31', 'chainId': '0x1', 'nonce': '0x2', 'bloc
kHash': '0xcd96e8f2140ddce2e1aa86129fcd30d9f0ee33e6870af3a9190e589326568c19', 'blockNumber': '0x1', 'transactionIndex': '0x2', 'from': '0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1', 'to': '0xffcf8fdee72ac11b5c542428b35eef5769c409f0', 'value': '0xde0b6b3a764000
0', 'maxPriorityFeePerGas': '0xb2d05e00', 'maxFeePerGas': '0x6fc23ac00', 'gasPrice': '0xee6b2800', 'gas': '0x5208', 'input': '0x', 'accessList': [], 'v': '0x1', 'r': '0x4e3fc279f2dd436837461fd641d7de0cb0f5f562a3db0243101dd8372b43048c', 's': '0x3984d95009ac9bd
1ab9ceb0dc2a6614169d31f15111048c0f9f759fb8329fe1c'}}
```

MicaiahReid:
Previously we were returning the wrong `gasPrice` when a transaction was fetched via `eth_getTransactionByHash`. This was happening because when we saved the transaction to the chain, we didn't update the transaction's `effectiveGasPrice` field based off of the `baseFeePerGas` of the block that the transaction was mined onto in some cases. 

However, the `effectiveGasPrice` returned on a receipt was correct because when the receipt was fetched, we would calculate the `effectiveGasPrice` on the fly just to correct what we return to the user, rather than directly using what the database returned.

This change removes the on-the-fly calculation of the `effectiveGasPrice` on the receipt, and presets the transaction's `effectiveGasPrice` based off of the _next_ block's `baseFeePerGas` when the transaction is added to the pool. As block's are mined, the miner updates all pending transaction's `effectiveGasPrice` based off of the next block's `baseFeePerGas`.

Fixes #4094.

### DIFF
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -93,7 +93,7 @@ export class Block {
       // get the maxFeePerGas and maxPriorityFeePerGas, use those to calculate
       // the effectiveGasPrice and add it to `extra` above, or we can just
       // leave it out of extra and update the effectiveGasPrice after like this
-      tx.updateEffectiveGasPrice(header.baseFeePerGas.toBigInt());
+      tx.updateEffectiveGasPrice(header.baseFeePerGas?.toBigInt());
       return txFn(tx);
     }) as IncludeTransactions extends true ? TypedTransactionJSON[] : Data[];
 
--- a/src/chains/ethereum/block/src/block.ts
+++ b/src/chains/ethereum/block/src/block.ts
@@ -93,7 +93,7 @@ export class Block {
       // get the maxFeePerGas and maxPriorityFeePerGas, use those to calculate
       // the effectiveGasPrice and add it to `extra` above, or we can just
       // leave it out of extra and update the effectiveGasPrice after like this
-      tx.updateEffectiveGasPrice(header.baseFeePerGas);
+      tx.updateEffectiveGasPrice(header.baseFeePerGas.toBigInt());
       return txFn(tx);
     }) as IncludeTransactions extends true ? TypedTransactionJSON[] : Data[];
 
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -429,6 +429,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         const hash = tx.hash.toBuffer();
         const index = Quantity.from(i);
 
+        // update the effective gas price of the transaction based off of the
+        // baseFeePerGas of the block that the transaction is actually on
+        tx.updateEffectiveGasPrice(block.header.baseFeePerGas);
         // save transaction to the database
         const serialized = tx.serializeForDb(blockHash, blockNumberQ, index);
         this.transactions.set(hash, serialized);
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -429,9 +429,6 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         const hash = tx.hash.toBuffer();
         const index = Quantity.from(i);
 
-        // update the effective gas price of the transaction based off of the
-        // baseFeePerGas of the block that the transaction is actually on
-        tx.updateEffectiveGasPrice(block.header.baseFeePerGas);
         // save transaction to the database
         const serialized = tx.serializeForDb(blockHash, blockNumberQ, index);
         this.transactions.set(hash, serialized);
--- a/src/chains/ethereum/ethereum/src/miner/miner.ts
+++ b/src/chains/ethereum/ethereum/src/miner/miner.ts
@@ -93,7 +93,7 @@ export default class Miner extends Emittery<{
   #isBusy: boolean = false;
   #paused: boolean = false;
   #resumer: Promise<void>;
-  #currentBlockBaseFeePerGas: Quantity;
+  #currentBlockBaseFeePerGas: bigint;
   #resolver: (value: void) => void;
 
   /**
@@ -424,7 +424,7 @@ export default class Miner extends Emittery<{
           // if baseFeePerGas is undefined, we are pre london hard fork.
           // no need to refresh the order of the heap because all Txs only have gasPrice.
           if (this.#currentBlockBaseFeePerGas !== undefined) {
-            priced.refresh(this.#currentBlockBaseFeePerGas.toBigInt());
+            priced.refresh(this.#currentBlockBaseFeePerGas);
           }
         } else {
           // reset the miner
@@ -533,9 +533,7 @@ export default class Miner extends Emittery<{
       if (next && !next.locked) {
         const origin = next.from.toString();
         origins.add(origin);
-        next.updateEffectiveGasPrice(
-          this.#currentBlockBaseFeePerGas.toBigInt()
-        );
+        next.updateEffectiveGasPrice(this.#currentBlockBaseFeePerGas);
         priced.push(next);
         next.locked = true;
       }
@@ -573,9 +571,7 @@ export default class Miner extends Emittery<{
           continue;
         }
         origins.add(origin);
-        next.updateEffectiveGasPrice(
-          this.#currentBlockBaseFeePerGas.toBigInt()
-        );
+        next.updateEffectiveGasPrice(this.#currentBlockBaseFeePerGas);
         priced.push(next);
         next.locked = true;
       }
@@ -594,6 +590,6 @@ export default class Miner extends Emittery<{
     const baseFeePerGas = block.header.baseFeePerGas;
     // before london hard fork, there will be no baseFeePerGas on the block
     this.#currentBlockBaseFeePerGas =
-      baseFeePerGas === undefined ? undefined : Quantity.from(baseFeePerGas);
+      baseFeePerGas === undefined ? undefined : baseFeePerGas;
   };
 }
--- a/src/chains/ethereum/ethereum/src/miner/miner.ts
+++ b/src/chains/ethereum/ethereum/src/miner/miner.ts
@@ -146,7 +146,7 @@ export default class Miner extends Emittery<{
     this.#executables = executables;
     this.#createBlock = (previousBlock: Block) => {
       const newBlock = createBlock(previousBlock);
-      this.#setCurrentBlockBaseFeePerGas(newBlock);
+      this.#currentBlockBaseFeePerGas = newBlock.header.baseFeePerGas;
       return newBlock;
     };
 
@@ -179,7 +179,7 @@ export default class Miner extends Emittery<{
       this.#updatePricedHeap();
       return;
     } else {
-      this.#setCurrentBlockBaseFeePerGas(block);
+      this.#currentBlockBaseFeePerGas = block.header.baseFeePerGas;
       this.#setPricedHeap();
       const result = await this.#mine(block, maxTransactions, onlyOneBlock);
       this.emit("idle");
@@ -581,15 +581,4 @@ export default class Miner extends Emittery<{
   public toggleStepEvent(enable: boolean) {
     this.#emitStepEvent = enable;
   }
-
-  /**
-   * Sets the #currentBlockBaseFeePerGas property if the current block
-   * has a baseFeePerGas property
-   */
-  #setCurrentBlockBaseFeePerGas = (block: RuntimeBlock) => {
-    const baseFeePerGas = block.header.baseFeePerGas;
-    // before london hard fork, there will be no baseFeePerGas on the block
-    this.#currentBlockBaseFeePerGas =
-      baseFeePerGas === undefined ? undefined : baseFeePerGas;
-  };
 }
--- a/src/chains/ethereum/ethereum/src/miner/miner.ts
+++ b/src/chains/ethereum/ethereum/src/miner/miner.ts
@@ -71,7 +71,7 @@ const updateBloom = (blockBloom: Buffer, bloom: Buffer) => {
 const sortByPrice = (values: TypedTransaction[], a: number, b: number) =>
   values[a].effectiveGasPrice > values[b].effectiveGasPrice;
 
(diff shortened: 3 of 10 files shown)


jeffsmale90 (approved):

src/chains/ethereum/transaction/src/transaction-receipt.ts
```diff
@@ -136,7 +136,7 @@ export class InternalTransactionReceipt {
   public toJSON(
     block: {
       hash(): Data;
-      header: { number: Quantity; baseFeePerGas?: Quantity };
+      header: { number: Quantity };
```
davidmurdoch: Since we no longer need `baseFeePerGas` we should change the receipt's `toJSON` method to not require the `block`, as the `transaction` property already has `blockHash` and `blockNumber` properties we can use.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -429,6 +429,9 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         const hash = tx.hash.toBuffer();
         const index = Quantity.from(i);
 
+        // update the effective gas price of the transaction based off of the
+        // baseFeePerGas of the block that the transaction is actually on
+        tx.updateEffectiveGasPrice(block.header.baseFeePerGas);
```
davidmurdoch: In other places we only do set this if the block header has a `baseFeePerGas`. 

Actually, since `updateEffectiveGasPrice` is already called in the miner itself, why wouldn't the `effectiveGasPrice` be updated already :thinking:

davidmurdoch (changes requested):

src/chains/ethereum/transaction/src/transaction-receipt.ts
MicaiahReid: Removed!

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -427,7 +424,7 @@ export default class Miner extends Emittery<{
           // if baseFeePerGas is undefined, we are pre london hard fork.
           // no need to refresh the order of the heap because all Txs only have gasPrice.
           if (this.#currentBlockBaseFeePerGas !== undefined) {
-            priced.refresh(this.#currentBlockBaseFeePerGas);
+            priced.refresh(this.#currentBlockBaseFeePerGas.toBigInt());
```
davidmurdoch: Maybe `#currentBlockBaseFeePerGas` should be updated to be a `bigint` instead of a `Quantity`?

src/chains/ethereum/transaction/src/eip1559-fee-market-transaction.ts
```diff
@@ -253,8 +253,8 @@ export class EIP1559FeeMarketTransaction extends RuntimeTransaction {
     return computeIntrinsicsFeeMarketTx(v, <EIP1559FeeMarketDatabaseTx>raw);
   }
 
-  public updateEffectiveGasPrice(baseFeePerGas: Quantity) {
-    const baseFeePerGasBigInt = baseFeePerGas.toBigInt();
+  public updateEffectiveGasPrice(baseFeePerGas: bigint) {
+    const baseFeePerGasBigInt = baseFeePerGas;
```
davidmurdoch: Instead of reassigning this variable can we just use the `baseFeePerGas` variable as is?

davidmurdoch (changes requested):

src/chains/ethereum/transaction/src/eip1559-fee-market-transaction.ts
MicaiahReid: Yes, obviously. :sweat_smile:

src/chains/ethereum/ethereum/src/miner/miner.ts
MicaiahReid: Yes, my bad for not self-reviewing this yet.

src/chains/ethereum/ethereum/src/miner/miner.ts
```diff
@@ -593,6 +590,6 @@ export default class Miner extends Emittery<{
     const baseFeePerGas = block.header.baseFeePerGas;
     // before london hard fork, there will be no baseFeePerGas on the block
     this.#currentBlockBaseFeePerGas =
-      baseFeePerGas === undefined ? undefined : Quantity.from(baseFeePerGas);
+      baseFeePerGas === undefined ? undefined : baseFeePerGas;
```
davidmurdoch: I think this can just be `baseFeePerGas`

src/chains/ethereum/block/src/block.ts
```diff
@@ -93,7 +93,7 @@ export class Block {
       // get the maxFeePerGas and maxPriorityFeePerGas, use those to calculate
       // the effectiveGasPrice and add it to `extra` above, or we can just
       // leave it out of extra and update the effectiveGasPrice after like this
-      tx.updateEffectiveGasPrice(header.baseFeePerGas);
+      tx.updateEffectiveGasPrice(header.baseFeePerGas?.toBigInt());
```
davidmurdoch, line 96: What this an existing bug?

src/chains/ethereum/transaction/src/runtime-transaction.ts
```diff
@@ -256,5 +256,5 @@ export abstract class RuntimeTransaction extends BaseTransaction {
   );
 
   protected abstract toVmTransaction();
-  protected abstract updateEffectiveGasPrice(baseFeePerGas?: Quantity);
+  protected abstract updateEffectiveGasPrice(baseFeePerGas?: bigint);
```
davidmurdoch: This should probably be:

```suggestion
  protected abstract updateEffectiveGasPrice(baseFeePerGas: bigint);
```

src/chains/ethereum/ethereum/tests/api/eth/eth.test.ts
```diff
@@ -543,34 +543,67 @@ describe("api", () => {
...
+          type: "0x2",
+          from: accounts[0],
+          to: accounts[1],
+          maxPriorityFeePerGas: "0x77359400",
+          maxFeePerGas: "0x6FC23AC00"
+        } as any;
+        // we previously had a bug where the `gasPrice` field returned from
+        // `eth_getTransactionByHash` was incorrect when multiple transactions
+        // were sent, because we weren't saving the updated `effectiveGasPrice`.
+        // this tests against that. see:
```
davidmurdoch: I think something like is more accurate according to the actual fix:

```suggestion
        // were sent, because we weren't assigning the correct `effectiveGasPrice`
        // when adding the transaction to the pool. see:
```

davidmurdoch (changes requested):

src/chains/ethereum/block/src/block.ts
MicaiahReid: No, this was introduced when I changed the `updateEffectiveGasPrice` function to receive a bigint rather than a quantity. For transactions types that are pre-1559, the `updateEffectiveGasPrice` function is a no-op (the `effectiveGasPrice` is just the static `gasPrice` param passed by the user), so the `.toBigInt()` was never called within the function for those. If we called the function passing an undefined `Quantity`, it didn't matter. Now, we're converting on the outside so we have to prevent calling `.toBigInt` on an undefined base fee.

src/chains/ethereum/transaction/src/runtime-transaction.ts
MicaiahReid: Fixed

jeffsmale90 (approved):

davidmurdoch (approved):
I wonder if there's a measurable performance improvement when fetching receipts since we don't query the db for the block anymore.

PR #4132, opened 2023-01-06
docs: add console.log to all solidity examples in interactive docs

MicaiahReid:
All of our RPC methods are documented with examples in our [interactive documentation](https://trufflesuite.github.io/ganache/). Some of these examples even have a simple Solidity contract that is precompiled and is deployed as you run the example. This change updates those precompiled Solidity contracts to include an import of Ganache's `console.sol` package and calls to `console.log`. When you run an example, go to the Ganache Logs tab and you'll see these Solidity `console.log` calls being logged! This change impacts the following RPC methods:
- `eth_call`
- `eth_getBlockByHash`
- `eth_getBlockTransactionCountByHash`
- `eth_getCode`
- `eth_getFilterChanges`
- `eth_getLogs`
- `eth_getFilterLogs`
- `eth_getProof`
- `eth_getStorageAt`
- `debug_traceTransaction`
- `debug_storageRangeAt`


tenthirtyone (approved):
Pulled down and tested against 4 at random. Looks good. 

One nit: The new logging increase the logs beyond the height of the window for some examples. On xl screens the page is very very wide and very very thin:

image

 I think it's benign, no recommendation just bringing it up

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -1100,15 +1100,18 @@ export default class EthereumApi implements Api {
...
    * // // SPDX-License-Identifier: MIT
-   * //  pragma solidity ^0.7.4;
+   * //  pragma solidity >= 0.4.22 <0.9.0;
+   * //
+   * // import "console.sol";
    * //
    * //  contract Simple {
    * //      uint256 public value;
    * //      constructor() payable {
+   * //          console.log("Called Simple contract constructor. Setting value to 5.");
```
jeffsmale90:
```suggestion
   * //          console.log("Called Simple contract constructor. Setting `value` to 5.");
```
I think this is a little clearer (that "value" refers to the variable name)

Same as per other instances. It's not a significant issue though, as clearly the contracts are fairly simple.

jeffsmale90 (changes requested):

src/chains/ethereum/ethereum/src/api.ts
MicaiahReid: ¯\_(ツ)_/¯ works for me

MicaiahReid: (I don't have the option to use the suggested change)

davidmurdoch: Seeing you here made my heart hurt.

PR #4145, opened 2023-01-10
fix: remove setting of private property

MicaiahReid:
Before the latest upgrade of etheremjs-vm, we would manually set the `_isInitialized` property of the `VM` to avoid some extra computation that came along with initialization that we didn't need. After the ethereumjs-vm upgrde, creating the `VM` also initializes it, so this manual setting of the private property is unnecessary, but we didn't remove it. This change removes the manual setting of `vm._isInitialized`.


jeffsmale90 (approved):
Looks good to me

tenthirtyone (approved):
2 under par

davidmurdoch (approved):

PR #4166, opened 2023-01-17
fix: log warning regarding transactions with future-nonces when mining in eager mode

### LINKED ISSUE #4165: Sending future nonce transaction in eager mode causes Ganache to hang
If a transaction is sent with a future nonce while in eager mining mode, Ganache hangs indefinitely and never returns a response for the transaction.

This happens the blockchain's [`queueTransaction` function](https://github.com/trufflesuite/ganache/blob/b816de218202e3f10182c61de9bae505b3c42710/src/chains/ethereum/ethereum/src/blockchain.ts#L1039-L1072). A future nonce transaction will return `false` for `isExecutable`, yet the code path still leads to awaiting `transaction.once("finalized")`, which won't happen until the nonce gap is filled.

MicaiahReid:
@benams ran into a confusing situation in which future-nonce transactions that are sent to Ganache running in eager mining mode appear to hang indefinitely, rather than being mined or rejected. This is expected behavior in eager mode, as Ganache does not return the hash until after it is mined. To help alleviate the confusion we now log a message about the status of these kinds of transactions.

Closes #4165


src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1058,7 +1058,12 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
     if (this.#isPaused() || !this.#instamine) {
       return hash;
     } else {
-      if (this.#instamine && this.#options.miner.instamine === "eager") {
+      // if the transaction is not executable, we just have to return the hash
+      if (
+        this.#instamine &&
+        this.#options.miner.instamine === "eager" &&
+        isExecutable
```
MicaiahReid: My main concern with this fix is that I don't think we currently have a way to return VM errors to transactions where we return the hash first. With eager mode we advertise that you can use the `vmErrorsOnRPCResponse` flag to get those VM errors, but that flag doesn't work with strict mode. This is pretty much reverting to strict mode functionality for future-nonce transactions; we aren't able to return VM errors.

src/chains/ethereum/ethereum/tests/api/eth/instamine.test.ts
```diff
@@ -0,0 +1,138 @@
+import getProvider from "../../helpers/getProvider";
```
MicaiahReid, line 1: I renamed this to `instamine.test.ts` because the previous name was based off of now-obsolete iterations on the name for the instamine modes.

I've also rearranged the describe blocks a bit to make more sense. I'll mark below where the actual new test is.

src/chains/ethereum/ethereum/tests/api/eth/instamine.test.ts
```diff
@@ -0,0 +1,138 @@
...
+              value: "0x1"
+            }
+          ]);
+          const receipt = await provider.send("eth_getTransactionReceipt", [
+            hash
+          ]);
+          assert.notStrictEqual(receipt, null);
+        });
+
+        it("returns the tx hash before mining for future-nonce transactions", async () => {
```
MicaiahReid: This is the new test

davidmurdoch (approved):

jeffsmale90 (approved):

PR #4195, opened 2023-01-30
feat: allow logs to be written to a file by providing the `--logging.file` argument

jeffsmale90:
Introducing the ability to specify a log file for Ganache to write its logs to. No more sifting through endless terminal output to find that one critical message you need. Just provide a path to your desired log file with `ganache --logging.file <path-to-log-file>`, and you're all set! Now you can sit back, relax, and let Ganache do the heavy lifting while you focus on building awesome applications.


src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -92,23 +109,55 @@ export const LoggingOptions: Definitions<LoggingConfig> = {
...
     cliDescription: "Set to `true` to log detailed RPC requests.",
     default: () => false,
     legacyName: "verbose",
     cliAliases: ["v", "verbose"],
     cliType: "boolean"
+  },
+  filePath: {
+    normalize,
+    cliDescription: "The path to a file to log to.",
+    cliAliases: ["log-file"],
```
davidmurdoch: Why the rename here?

jeffsmale90: the `log-file` / `filePath` - that's purely because I changed it only in one place :). It's now fixed.

jeffsmale90:
Sorry for the force push, but it had to be in order to remove the detached instances logs logic :/

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -164,6 +164,7 @@ export class EthereumProvider
       providerOptions.fork.url ||
       providerOptions.fork.provider ||
       providerOptions.fork.network;
+
```
jeffsmale90: I'll remove this but waiting for feedback so that I can couple it with any additional changes.

src/chains/ethereum/options/tests/index.test.ts
```diff
@@ -1,32 +1,7 @@
...
-
-    it("disables the logger when the quiet flag is used", () => {
-      const message = "message";
-      const options = EthereumOptionsConfig.normalize({
-        logging: { quiet: true }
-      });
-      options.logging.logger.log(message);
-      assert.strictEqual(spy.withArgs(message).callCount, 0);
-    });
-  });
```
jeffsmale90, line 29: these tests are moved to `logging-options.test.ts`

src/packages/cli/src/cli.ts
```diff
@@ -54,6 +55,17 @@ if (argv.action === "start") {
...
+  let closeLogHandle = null;
+  if (loggingOptions?.file) {
+    const { log, close } = createLogger({
+      baseLog: console.log,
+      file: loggingOptions.file
+    });
+    loggingOptions.logger = { log };
+    closeLogHandle = close;
+  }
+
```
jeffsmale90: This is clearly terrible, as the logging configuration is, untyped. The logging config is currently flavor specific - I think we should pull it up to be a "top level" Ganache config similar to Server config. I think we should do that in a separate PR though.

Thoughts?

src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -110,5 +123,23 @@ export const LoggingOptions: Definitions<LoggingConfig> = {
     legacyName: "verbose",
     cliAliases: ["v", "verbose"],
     cliType: "boolean"
+  },
+  file: {
+    normalize: rawInput => {
+      // this will throw if the file is not writable, and creates the log file for later appending
+      try {
+        const descriptor = openSync(rawInput, "w");
```
cds-amal: This will truncate the file if it exists. Is that the desired behavior? flag `wx` would fail if the file exists, or, `a` for appending, but won't truncate if the file exists.

src/packages/cli/src/cli.ts
cds-amal: I have no opinion, but curious about your reasoning to defer to follow on work.

cds-amal (commented):
Incomplete review, will finish it tomorrow.

src/chains/ethereum/options/src/logging-options.ts
jeffsmale90: 'a' is the desired behaviour - thanks for pointing that out!

jeffsmale90: I'll add a test for that too

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -432,6 +432,12 @@ export class EthereumProvider
     this.#executor.stop();
     await this.#blockchain.stop();
 
+    // only need to do this if it's an `AsyncronousLogger`
```
cds-amal: Does the test `'getCompletionHandle' in this.#options.logging.logger` indicate it's an `AsynchronousLogger`?

src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -74,6 +72,15 @@ export type LoggingConfig = {
...
       hasDefault: true;
     };
+
+    /**
+     * If you set this option, Ganache will write logs to a file located at the
+     * specified path.
+     */
+    readonly file: {
+      type: number;
+      rawType: PathLike;
```
cds-amal, line 88: Neat! URLs too?

src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -92,23 +99,48 @@ export const LoggingOptions: Definitions<LoggingConfig> = {
...
     legacyName: "verbose",
     cliAliases: ["v", "verbose"],
     cliType: "boolean"
+  },
+  file: {
+    normalize: (raw: PathLike): number => {
+      let descriptor: number;
+
+      try {
+        descriptor = openSync(raw, "a");
```
cds-amal, line 121: Oh, URL is constrained to the `file:` protocol.

src/chains/ethereum/options/src/miner-options.ts
```diff
@@ -220,7 +220,7 @@ const toNumberOrString = (str: string) => {
     return parseInt(str);
   }
 };
-
+const normalizeQuantity = value => Quantity.from(value);
```
cds-amal, line 229: Can you explain the reasoning behind adding this additional conversion abstraction? I spent some time distinguishing the usage of this conversion for the `normalize` property from the general transformation that `Quantity.from` provides for other properties. I'm unsure if there's a compelling reason for this added complexity, unless I'm overlooking something.

src/chains/ethereum/options/tests/logging-options.test.ts
```diff
@@ -0,0 +1,203 @@
...
+              "File descriptor not valid"
+            );
+          } finally {
+            await unlink(validFilePath);
+          }
+        });
+
+        it("resolves a file URL as Buffer to descriptor", async () => {
+          const options = EthereumOptionsConfig.normalize({
+            logging: { file: new URL(`file://${validFilePath}`) }
```
cds-amal: aha, I should have scanned the code before commenting!

src/chains/ethereum/options/tests/logging-options.test.ts
```diff
@@ -0,0 +1,203 @@
...
+            await unlink(validFilePath);
+          }
+        });
+
+        it("fails if the process doesn't have write access to the file path provided", async () => {
+          const file = resolve("./eperm-file.log");
+          try {
+            const handle = await open(file, "w");
+            // set no permissions on the file
+            await handle.chmod(0);
```
cds-amal, line 167: How does `handle.chmod()` behave on Windows? is it necessary or does the test work because the handle is closed in the next line?

src/chains/ethereum/options/tests/logging-options.test.ts
```diff
@@ -0,0 +1,203 @@
...
+
+            assert.deepStrictEqual(calls, [["message", "param1", "param2"]]);
+
+            const fromFile = await readFile(validFilePath, "utf8");
+            assert(fromFile !== "", "Nothing written to the log file");
+
+            const timestampPart = fromFile.substring(0, 24);
+
+            const timestampRegex =
+              /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/;
```
cds-amal:
```suggestion
              /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/;
```

A period can match any character, escaping it scopes it to match a literal dot.

src/chains/filecoin/options/tests/logging-options.test.ts
```diff
@@ -0,0 +1,177 @@
...
+            await unlink(validFilePath);
+          }
+        });
+
+        it("fails if the process doesn't have write access to the file path provided", async () => {
+          const file = resolve("./eperm-file.log");
+          try {
+            const handle = await open(file, "w");
+            // set no permissions on the file
+            await handle.chmod(0);
```
cds-amal, line 93: Similar question about `chmod()` on Windows.

src/chains/filecoin/options/tests/logging-options.test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+            assert.deepStrictEqual(calls, [["message", "param1", "param2"]]);
+
+            const fromFile = await readFile(validFilePath, "utf8");
+            assert(fromFile !== "", "Nothing written to the log file");
+
+            const timestampPart = fromFile.substring(0, 24);
+
+            const timestampRegex =
+              /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/;
```
cds-amal:
```suggestion
              /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/;
```

Restrict wildcard match

src/chains/filecoin/options/tsconfig.json
```diff
@@ -15,6 +15,10 @@
     {
       "name": "@ganache/options",
       "path": "../../../packages/options"
+    },
+    {
+      "name": "@ganache/utils",
+      "path": "../../../packages/utils"
     }
   ]
 }
```
cds-amal: Does Ganache care about eof newlines?

src/packages/utils/tests/logger.test.ts
```diff
@@ -0,0 +1,182 @@
...
+const getFileDescriptor = (slug: string) => {
+  const path = `./tests/test-${slug}.log`;
+  return {
+    path,
+    descriptor: openSync(path, "a")
+  };
+};
+
+describe("createLogger()", () => {
+  const timestampRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/;
```
cds-amal:
```suggestion
  const timestampRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/;
```

src/packages/utils/tests/logger.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      const logLines = fileContents.split("\n");
+
+      // 4, because there's a \n at the end of each line
+      assert.strictEqual(logLines.length, 4);
+      assert.strictEqual(logLines[3], "");
+
+      logLines.slice(0, 3).forEach((logLine, lineNumber) => {
+        const timestampPart = logLine.slice(0, 24);
+        const messagePart = logLine.slice(25);
+        const delimiter = logLine[24];
```
cds-amal: Should probably add a comment for these magic numbers since you documented `4`?

src/packages/utils/tests/logger.test.ts
```diff
@@ -0,0 +1,182 @@
...
+
+      // length == 4, because there's a \n at the end (string.split() results
+      // in each log line, follwed by an empty string)
+      assert.strictEqual(loggedLines.length, 4);
+      assert.strictEqual(loggedLines[3], "");
+
+      loggedLines.slice(0, 3).forEach((logLine, lineNumber) => {
+        const timestampPart = logLine.slice(0, 24);
+        const messagePart = logLine.slice(25);
+        const delimiter = logLine[24];
```
cds-amal: What is this [kind of magic](https://www.youtube.com/watch?v=0p_1QSUsbsM)?

src/packages/utils/src/things/logger.ts
```diff
@@ -0,0 +1,61 @@
...
+      throw new Error(
+        `'config.file' was not correctly noramlized to a file descriptor. This should not happen. ${
+          config.file
+        }: ${typeof config.file}`
+      );
+    }
+    const descriptor = config.file;
+
+    const diskLogFormatter = (message: any) => {
+      const linePrefix = `${new Date().toISOString()} `;
```
cds-amal, line 31: Probably worth documenting if that trailing space is necessary so a future maintainer won't trim it.

cds-amal (commented):
Nice @jeffsmale90. I have some questions and feedback.

src/chains/ethereum/ethereum/src/provider.ts
jeffsmale90: Yep - that's correct. I've updated the comment to be explicit about that.

It's not the nicest assertion, happy to change it if anyone suggests something better.

src/chains/ethereum/options/src/miner-options.ts
jeffsmale90: Yeah, so previously, this was:

`normalize: Quantity.from`

but the signature of the `normalize` function has changed - an additional `config` parameter has been added. This gets passed as the second argument to the `Quantity.from` constructor function (`nullable`).

src/chains/ethereum/options/src/logging-options.ts
jeffsmale90: I've added a comment in each of the `logging-options.ts` files.

Originally this was `number | PathLike` (where `number` is a filedescriptor (because sync functions work with number descriptors, and yargs works with async normalisers).

I removed this option, because it makes it hard to decide whether we should close the filedescriptor ourselves or not, and I don't see a huge amount of value to providing this behaviour. If anyone disagrees, I'm happy to look at it again.

src/chains/filecoin/options/tsconfig.json
jeffsmale90: Generally, yes :)

src/packages/utils/src/things/logger.ts
jeffsmale90: Done

src/packages/utils/tests/logger.test.ts
jeffsmale90: 👑 

Done!

src/packages/utils/tests/logger.test.ts
jeffsmale90: Done!

src/chains/ethereum/options/tests/logging-options.test.ts
jeffsmale90: `chmod` with `0` will add the `r` (readonly) flag, causing writes to the file to fail:
```
> fs.chmodSync("./testfile.log", 0);
> fs.appendFileSync("./testfile.log", "hi");
Uncaught Error: EPERM: operation not permitted, open './testfile.log'
    at Object.openSync (fs.js:498:3)
    at Object.writeFileSync (fs.js:1529:35) {
  errno: -4048,
  syscall: 'open',
  code: 'EPERM',
  path: './testfile.log'
}

...

-ar---         3/13/2023  12:33 PM             10 testfile.log
```

cds-amal: TIL node has a `chown` for windows! Thanks @jeffsmale90

src/chains/ethereum/options/src/logging-options.ts
cds-amal: Thanks for the explanation!

src/chains/ethereum/options/src/miner-options.ts
cds-amal: Could you add a comment to explain this is for 1) optimization, 2) to work make the arities of `Quantity.from` and `normalize` play nice?

src/chains/ethereum/ethereum/src/provider.ts
cds-amal: Did the commit make it into this branch? I don't see it.

src/packages/utils/tests/logger.test.ts
cds-amal: There are a few commits that didn't make it, including this one it seems.

src/packages/utils/src/things/logger.ts
cds-amal: image
:smiley:

cds-amal (commented):
Another pass @jeffsmale90 , Thanks for the explanations! I requested a clarifying comment, and noticed some previous resolution commits are missing?

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -644,6 +646,34 @@ describe("provider", () => {
       });
     });
 
+    it("closes the logging fileDescriptor", async () => {
+      const filePath = "./closes-logging-descriptor.log";
+      const provider = await getProvider({ logging: { file: filePath } });
+
+      try {
+        const descriptor = (await provider).getOptions().logging.file;
```
davidmurdoch:
```suggestion
        const descriptor = provider.getOptions().logging.file;
```

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -644,6 +646,34 @@ describe("provider", () => {
       });
     });
 
+    it("closes the logging fileDescriptor", async () => {
+      const filePath = "./closes-logging-descriptor.log";
```
davidmurdoch:
```suggestion
      const filePath = "./closes-logging-descriptor.log";
```

Should this be written to a temp location? I could imagine:

* this test failing midway through
* the file remaining
* the developer not noticing
* committing this test file to the repo
* file making its way through review

it's happened before to me. 'twas embarrassing

src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -92,23 +100,48 @@ export const LoggingOptions: Definitions<LoggingConfig> = {
...
+      } catch (err) {
+        throw new Error(
```
davidmurdoch, line 124: We shouldn't throw away the original `err`. It likely has very useful information in it.

src/chains/ethereum/options/tests/logging-options.test.ts
```diff
@@ -0,0 +1,203 @@
...
+        it("disables the logger when the quiet flag is used", () => {
+          const message = "message";
+          const options = EthereumOptionsConfig.normalize({
+            logging: { quiet: true }
+          });
+          options.logging.logger.log(message);
+          assert.strictEqual(spy.withArgs(message).callCount, 0);
+        });
+
+        it("calls the provided logger when quiet flag is used", () => {
```
davidmurdoch, line 111: Should it do the opposite? if the user specified `quiet` then we shouldn't log, right?

src/chains/ethereum/options/tests/logging-options.test.ts
```diff
@@ -0,0 +1,203 @@
...
+            const options = EthereumOptionsConfig.normalize({
+              logging: { file: validFilePath }
+            });
+            options.logging.logger.log(message);
+            closeSync(options.logging.file);
+
+            const readHandle = await open(validFilePath, "r");
+            const content = await readHandle.readFile({ encoding: "utf8" });
+            await readHandle.close();
+            assert(
```
davidmurdoch: I think this should test that the `message` `"message"` was also appended to the file.

src/chains/filecoin/filecoin/tsconfig.json
```diff
@@ -7,14 +7,8 @@
...
-  "include": [
-    "index.ts",
-    "src/**/*"
-  ],
-  "typeRoots": [
-    "./node_modules/@types",
-    "./src/@types"
-  ],
+  "include": ["index.ts", "src/**/*"],
+  "typeRoots": ["./node_modules/@types", "./src/@types"],
```
davidmurdoch: I don't think this file should be included in this PR

src/chains/filecoin/options/package.json
```diff
@@ -66,6 +68,7 @@
     "webpack-cli": "4.9.1"
   },
   "dependencies": {
-    "keccak": "3.0.2"
+    "keccak": "3.0.2",
+    "@ganache/utils": "0.7.1"
```
davidmurdoch: `@ganache/utils` should appear first, otherwise it will likely get moved there later and it'll annoy us (npm or lerna alphabetized the list sometimes).

src/chains/filecoin/options/tsconfig.json
davidmurdoch: this file is generated. so the `scripts/link-ts-references.ts` script should be fixed before adding newlines to these.

src/packages/cli/src/cli.ts
```diff
@@ -89,6 +89,7 @@ if (argv.action === "start") {
         case ServerStatus.open:
           console.log("Shutting down…");
           await server.close();
+
```
davidmurdoch: meh

src/packages/core/tests/connector.test.ts
```diff
@@ -9,29 +9,33 @@ describe("connector", () => {
...
-      logger.log = msg => {
-        const expected =
-          "   >  web3_sha3: [\n" + '   >   "Tim is a swell guy."\n' + "   > ]";
-        assert.strictEqual(msg, expected, "doesn't work with params");
-        resolve();
-      };
-      await p.send("web3_sha3", ["Tim is a swell guy."]);
-    });
+    // clear the logLines
+    logLines.splice(0, logLines.length);
```
davidmurdoch: `logLines.length = 0;` would work, too. I don't care either way.

src/packages/core/tests/connector.test.ts
```diff
@@ -9,29 +9,33 @@ describe("connector", () => {
...
+
+    const message = `0x${Buffer.from("Tim is a swell guy.").toString("hex")}`;
+    await p.send("web3_sha3", [message]);
+    assert.deepStrictEqual(
+      logLines,
+      [`   >  web3_sha3: [\n   >   \"${message}\"\n   > ]`],
+      "doesn't work with params"
+    );
```
davidmurdoch, line 38: I like your new test improvements!

src/packages/utils/src/things/logger.ts
```diff
@@ -0,0 +1,62 @@
...
+}): Logger {
+  if (config.file === undefined) {
+    return config.baseLogger;
+  } else {
+    if (typeof config.file !== "number") {
+      throw new Error(
+        `'config.file' was not correctly noramlized to a file descriptor. This should not happen. ${
+          config.file
+        }: ${typeof config.file}`
+      );
```
davidmurdoch: I'm not convinced this check is needed. `config.file` is always normalized before we can get here, right?

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -35,7 +35,7 @@ import {
   MessageEvent,
   VmConsoleLogEvent
 } from "./provider-events";
-
+import { closeSync } from "fs";
```
jeffsmale90:
```suggestion
```

src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -74,6 +73,16 @@ export type LoggingConfig = {
       type: boolean;
       hasDefault: true;
     };
+
+    /**
+     * If you set this option, Ganache will write logs to a file located at the
+     * specified path.
+     * Note: If you provide a `URL` it must use the `path://` protocol.
```
davidmurdoch: The tone/style of this doc comment is different than the rest of the options docs comments. Maybe something like:

```
The file append logs to.

Can be a filename, url, or descriptor.

If specifying a url, the path must begin with `path://`.

By default no log file is created.
```

src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -92,23 +101,48 @@ export const LoggingOptions: Definitions<LoggingConfig> = {
...
+        descriptor = openSync(raw, "a");
+      } catch (err) {
+        throw new Error(
+          `Failed to open log file ${raw}. Please check if the file path is valid and if the process has write permissions to the directory.`
+        );
+      }
+      return descriptor;
+    },
+    cliDescription:
+      "If set, Ganache will write logs to a file located at the specified path.",
```
davidmurdoch: Maybe something like `The filename to append logs to.`?

src/chains/ethereum/options/src/logging-options.ts
davidmurdoch: Does this accept a string that starts with `path://`, or does it have to be a `URL`?

src/chains/ethereum/options/tests/logging-options.test.ts
davidmurdoch: On second thought, I also see value in this. `--quiet` quiets the console, but the user can still get the logs. Hmmmmm

src/packages/utils/src/things/logger.ts
```diff
@@ -0,0 +1,70 @@
...
+    };
+
+    const writeStream = createWriteStream(null, { fd });
+
+    const onError = err =>
+      console.error(`Error writing to log file: ${err.message}`);
+    writeStream.on("error", onError);
+
+    const log = (message: any, ...optionalParams: any[]) => {
+      // we are logging to a file, but we still need to writing to console
```
davidmurdoch:
```suggestion
      // we are logging to a file, but we still need to write to console
```

src/packages/utils/src/things/logger.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (typeof config.file !== "number") {
+      throw new Error(
+        `'config.file' was not correctly normalized to a file descriptor. This should not happen. Value: ${
+          config.file
+        } of type ${typeof config.file}`
+      );
+    }
+    const fd = config.file;
+
+    const diskLogFormatter = (message: any) => {
```
davidmurdoch: Is `message` always a string? And if so, we don't need the `toString()` call below.

src/packages/utils/src/things/logger.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const onError = err =>
+      console.error(`Error writing to log file: ${err.message}`);
+    writeStream.on("error", onError);
+
+    const log = (message: any, ...optionalParams: any[]) => {
+      // we are logging to a file, but we still need to writing to console
+      baseLog(message, ...optionalParams);
+
+      const formattedMessage: string = format(message, ...optionalParams);
```
davidmurdoch: A thought crossed my mind, i'm not sure if it is a good thought. But here is an alternative to ponder:

```suggestion
    const log = (...params: any[]) => {
      // we are logging to a file, but we still need to write to console
      baseLog(...params);

      const formattedMessage: string = format(...params);
```

src/packages/utils/src/things/logger.ts
```diff
@@ -0,0 +1,70 @@
...
+      writeStream.write(diskLogFormatter(formattedMessage) + "\n");
```
davidmurdoch: Should we use `os.EOL` here? https://nodejs.org/api/os.html#oseol

@cds-amal thoughts?

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: Since writing the above comment we've spoken about `tmp-promise`. I agree that should be used here (and in the other tests) as well.

davidmurdoch (commented):
I still need to pull this down and review, and take a look at code coverage, but here is a high level review. Looking pretty darn good!

src/chains/ethereum/options/src/logging-options.ts
jeffsmale90: That's certainly less convoluted than my description :P

src/chains/ethereum/options/tests/logging-options.test.ts
jeffsmale90: I like it as is for the reason you point out - also it becomes more complicated if the `--quiet` flag stops output to the file, because we'd maybe want to make those conflict or something, because `--quiet --logging.file` together wouldn't make sense if nothing was logged 🤷 

I've changed the description of `--quiet` to be more explicitly _only_ the console.

src/packages/utils/src/things/logger.ts
jeffsmale90: Yeah it _should_ be. I added this while developing, to ensure that it is always normalized. It's defensive, but maybe not necessary, given our confidence that it is. 🤷

src/chains/ethereum/options/src/logging-options.ts
jeffsmale90: It does not - I've clarified that in the description. We _could_ implement this, but I feel like it's not going to be very widely used.

src/packages/utils/src/things/logger.ts
jeffsmale90: I actually prefer having the full signature, because this is more explicit, and easier to understand how the different functions with the same signature are used together. I'm happy to hear alternative viewpoints though if you or anyone else disagrees.

I updated another signature within this file that had the `(...params)` signature to include the `optionalParams` parameter as well.

src/chains/filecoin/options/tsconfig.json
jeffsmale90: Done - it only writes changes if they are materially different, so they will get newlines as, and when we change their contents.

scripts/link-ts-references.ts
```diff
@@ -137,7 +138,8 @@ function saveConfigs(configs: PackageInfo[]) {
   configs.forEach(({ modified, path, tsConfig }) => {
     if (modified) {
       const tsConfigFile = join(path, "tsconfig.json");
-      writeFileSync(tsConfigFile, JSON5.stringify(tsConfig, null, 2));
+      const body = JSON5.stringify(tsConfig, null, 2) + EOL;
+      writeFileSync(tsConfigFile, body);
```
jeffsmale90: This is the only meaningful change in this file (plus the import on line 8).

The rest is just auto-formatting prettier.

src/chains/ethereum/options/src/index.ts
```diff
@@ -7,14 +7,12 @@ import { ForkConfig, ForkOptions } from "./fork-options";
...
   Base,
   Defaults,
-  Definitions,
   ExternalConfig,
   InternalConfig,
   Legacy,
   LegacyOptions,
   OptionName,
   OptionRawType,
-  Options,
```
jeffsmale90, line 17: Not really relevant to this change, but cleaning up unnecessary imports.

scripts/link-ts-references.ts
davidmurdoch: The EOL shouldn't be OS independent, otherwise we'll be going back and forth on which EOL character gets used depending on which OS a PR is for.

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -767,6 +771,37 @@ describe("provider", () => {
       });
     });
 
+    it("closes the logging fileDescriptor", async () => {
+      await tmp.withDir(
```
davidmurdoch, line 775: cool!

src/chains/ethereum/options/package.json
```diff
@@ -53,7 +53,8 @@
     "@ganache/options": "0.8.0",
     "@ganache/utils": "0.8.0",
     "bip39": "3.0.4",
-    "seedrandom": "3.0.5"
+    "seedrandom": "3.0.5",
+    "tmp-promise": "3.0.2"
```
davidmurdoch: is this supposed to be in devDependencies?

src/chains/filecoin/options/package.json
```diff
@@ -66,6 +68,8 @@
     "webpack-cli": "4.9.1"
   },
   "dependencies": {
-    "keccak": "3.0.2"
+    "@ganache/utils": "0.8.0",
+    "keccak": "3.0.2",
+    "tmp-promise": "3.0.2"
```
davidmurdoch: Should this be in dev deps?

src/chains/ethereum/options/src/logging-options.ts
```diff
@@ -87,28 +100,53 @@ export const LoggingOptions: Definitions<LoggingConfig> = {
   },
   quiet: {
     normalize,
-    cliDescription: "Set to `true` to disable logging.",
+    cliDescription: "Set to `true` to disable writing logs to the terminal.",
```
davidmurdoch: Maybe something like:

```suggestion
    cliDescription: "Set to `true` to disable writing logs to `logger.log` (`stdout` by default).",
```

src/chains/ethereum/options/src/logging-options.ts
davidmurdoch: I was going to suggest https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError, but since the stack trace of the original error is probably not useful I think what you have here is better.

scripts/link-ts-references.ts
jeffsmale90: Yes it is!

src/chains/filecoin/options/package.json
davidmurdoch: this still needs to be moved

src/packages/utils/package.json
```diff
@@ -50,7 +50,8 @@
     "@ganache/secp256k1": "0.5.0",
     "emittery": "0.10.0",
     "keccak": "3.0.2",
-    "seedrandom": "3.0.5"
+    "seedrandom": "3.0.5",
+    "tmp-promise": "3.0.2"
```
davidmurdoch: this one too? myabe?

src/chains/ethereum/ethereum/package.json
```diff
@@ -83,10 +83,10 @@
     "seedrandom": "3.0.5",
     "semaphore": "1.1.0",
     "subleveldown": "6.0.1",
-    "tmp-promise": "3.0.2",
     "ws": "8.2.3"
   },
   "devDependencies": {
+    "tmp-promise": "3.0.2",
```
davidmurdoch: You need to run `npm run reinstall` to update the package locks, and since dependencies will be alphabetically ordered when someone installs something new into these packages via `npm install ___` (or `npx lerna add ___...`) these really should be put into deps alphabetically as from the start, otherwise they'll be needlessly moved in some future unrelated PR.

src/chains/ethereum/options/package-lock.json
```diff
@@ -1,326 +1,402 @@
 {
 	"name": "@ganache/ethereum-options",
 	"version": "0.8.0",
-	"lockfileVersion": 1,
+	"lockfileVersion": 3,
```
davidmurdoch: I'd love to update to v3, but it should be all at once and in a different PR (the one that adds support for node 20, probably?)

davidmurdoch (commented):
you accidentally updated to package lock v3, we're still on v1 :-(

src/chains/ethereum/ethereum/package.json
```diff
@@ -110,6 +109,7 @@
     "shx": "0.3.3",
     "solc": "0.8.11",
     "superagent": "6.1.0",
+    "tmp-promise": "3.0.2",
```
davidmurdoch: Want to hear a funny story?

This one is an actual dependency.

davidmurdoch (changes requested):
😅

davidmurdoch (approved):

PR #4199, opened 2023-02-01
feat: improved user experience for detached instances with prefix matching and suggestions

jeffsmale90:
Detached instances are pretty awesome, but they have long names that you have to type all the time!

With fuzzy matching (and some helpful suggestions), you can get by without typing so much:

```bash
$ ganache instances stop hot_lim_cheesecacke

hot_lim_cheesecacke found.

Did you mean:
 - hot_lime_cheesecake
 - hot_lemon_loaf

$ ganache instances stop hot_li
hot_lime_cheesecake stopped
```

note: now you only need to enter the unique prefix of the instance name. Ganache will use this prefix to identify the instance.


jeffsmale90:
This will have more thorough testing added after #3735 is merged

leeoniya:
maybe a bit late for a pitch, but https://github.com/leeoniya/uFuzzy#a-biased-appraisal-of-similar-work

jeffsmale90:
> https://github.com/leeoniya/uFuzzy#a-biased-appraisal-of-similar-work

Thanks for the suggestion! I intentionally _didn't_ introduce a library for this feature, in an effort to limit the impact on the bundle size.

In addition, I'm not sure this is a good fit for your library, based on your comment "match a relatively short search phrase (needle) against a large list of short-to-medium phrases (haystack)" whereas this use case is a very short search phrase against a small list of very short phrases.

I'll keep it in mind for future work though.

src/packages/cli/src/detach.ts
```diff
@@ -322,3 +428,34 @@ export function formatUptime(ms: number) {
 
   return isFuture ? `In ${duration}` : duration;
 }
+
+export function levenshteinDistance(a: string, b: string): number {
```
jeffsmale90:
```suggestion
/**
 * This function calculates the Levenshtein distance between two strings.
 * Levenshtein distance is a measure of the difference between two strings, 
 * defined as the minimum number of edits (insertions, deletions or substitutions) 
 * required to transform one string into another.
 * 
 * @param {string} a - The first string to compare.
 * @param {string} b - The second string to compare.
 * @return {number} The Levenshtein distance between the two strings.
 */
export function levenshteinDistance(a: string, b: string): number {
```

src/packages/cli/src/detach.ts
```diff
@@ -62,25 +70,123 @@ export async function removeDetachedInstanceFile(
...
+
+/*
+Find instances with names similar to `instanceName`.
+
+If there is a single instance with an exact prefix match, it is returned as the
+`match` property in the result. Otherwise, up to `MAX_SUGGESTIONS` names that
+are similar to `instanceName` are returned as `suggestions`. Names with an exact
+prefix match are prioritized, followed by increasing Levenshtein distance, up to
+a maximum distance of `MAX_LEVENSHTEIN_DISTANCE`.
+*/
```
jeffsmale90:
```suggestion
@param {string} instanceName the name for which similarly named instance will be searched
@returns {{ match: string } | { suggestions: string[] }} either a single exact `match` or a number of `suggestions`
*/
```

jeffsmale90:
I've converted this to draft - I'll refactor it once the `instances logs` changes land

src/packages/cli/src/detach.ts
```diff
@@ -19,6 +19,8 @@ export type DetachedInstance = {
   version: string;
 };
 
+const MAX_SUGGESTIONS = 4;
+const MAX_LEVENSHTEIN_DISTANCE = 10;
```
jeffsmale90: This is a _huge_ distance. I'm unsure about how to tune this. 

I'd rather we give false positives, than false negatives - if I _forget_ the name of the instance, and get something that's truly wrong, I'd still like a recommendation.

I'm open to suggestions on this. At the same time, it's something that we can tune later, and hopefully we'll get some user feedback on the experience.

davidmurdoch: agreed. we can tune it later.

src/packages/cli/src/detach.ts
```diff
@@ -61,25 +69,127 @@ export async function removeDetachedInstanceFile(
...
+ * the `match` property in the result. Otherwise, up to `MAX_SUGGESTIONS` names
+ * that are similar to `instanceName` are returned as `suggestions`. Names with
+ * an exact prefix match are prioritized, followed by increasing Levenshtein
+ * distance, up to a maximum distance of `MAX_LEVENSHTEIN_DISTANCE`.
+ * @param {string} instanceName the name for which similarly named instance will
+ * be searched
+ * @returns {{ match: string } | { suggestions: string[] }} an object
+ * containiner either a single exact `match` or a number of `suggestions`
+ */
+async function getSimilarInstanceNames(
```
davidmurdoch: Oh man I really want to optimize this function! haha. But since the actual number of files is typically going to be on the order of `1` I'll hold back! haha

p.s., using our `findInsertPosition` util function is always a good time.

src/packages/cli/src/cli.ts
```diff
@@ -181,7 +192,8 @@ if (argv.action === "start") {
     })
     .catch(err => {
       // the child process would have output its error to stdout, so no need to
-      // output anything more
+      // do anything more other than set the exitCode
+      process.exitCode = 1;
```
davidmurdoch: smart!

src/packages/cli/src/cli.ts
```diff
@@ -164,11 +164,22 @@ if (argv.action === "start") {
...
+    if ("instance" in instanceOrSuggestions) {
+      const highlightedName = porscheColor(instanceOrSuggestions.instance.name);
+      console.log(`${highlightedName} stopped.`);
     } else {
-      console.error("Instance not found");
+      process.exitCode = 1;
+      console.log(`${porscheColor(instanceName)} not found.`);
+      if (instanceOrSuggestions.suggestions?.length > 0) {
+        console.log();
+        console.log("Did you mean:");
```
davidmurdoch: Since this isn't interactive (oh that'd be cool!), and the user can't actually answer the question. Should we use a different phrase here? git uses `The most similar command is`.

image

src/packages/cli/src/detach.ts
```diff
@@ -61,25 +69,127 @@ export async function removeDetachedInstanceFile(
...
-    await removeDetachedInstanceFile(instanceName);
+    if ((err as NodeJS.ErrnoException).code === "ENOENT") {
+      // instances directory does not exist, so there can be no suggestions
+      return { suggestions: [] };
+    }
+  }
+
+  const prefixMatches = [];
+  for (const name of filenames) {
+    if (name.startsWith(instanceName)) {
```
davidmurdoch: Probably not the right PR for this question since all names are generated by ganache at this time, but should we be concerned with case sensitivity?

src/packages/cli/src/detach.ts
```diff
@@ -323,3 +433,44 @@ export function formatUptime(ms: number) {
...
+
+/**
+ * This function calculates the Levenshtein distance between two strings.
+ * Levenshtein distance is a measure of the difference between two strings,
+ * defined as the minimum number of edits (insertions, deletions or substitutions)
+ * required to transform one string into another.
+ *
+ * @param {string} a - The first string to compare.
+ * @param {string} b - The second string to compare.
+ * @return {number} The Levenshtein distance between the two strings.
```
davidmurdoch: I think I've mentioned this before, but including the types in these definitions are kinda redundant. I don't think you need to remove them, but I don't think you need to add them in the first place either.

src/packages/cli/src/detach.ts
jeffsmale90: I sort of assumed that they would just be case sensitive. That could be problematic on Windows (and sometimes Mac?) operating systems however. I wonder if we need to lowercase them? Let's pick this up on the `--name <name>` PR

PR #4272, opened 2023-03-02
feat: support `shanghai` hardfork

davidmurdoch:
* [EIP-3651: Warm COINBASE](https://eips.ethereum.org/EIPS/eip-3651)
  * Needed to handle warming ourselves in simulateTransaction (`eth_call`). A new test was added for this.
* [EIP-3855: PUSH0 instruction](https://eips.ethereum.org/EIPS/eip-3855)
  * No changes required
* [EIP-3860: Limit and meter initcode](https://eips.ethereum.org/EIPS/eip-3860)
  * We've added a new option `allowUnlimitedInitCodeSize` to disable this EIP to continue to allow for uploading very large contracts. In most cases it will need to be combined with the `allowUnlimitedContractSize` option.  Two new tests were added for this.
* [EIP-4895: Beacon chain push withdrawals as operations](https://eips.ethereum.org/EIPS/eip-4895)
  * Needed to add new fields to Block: `withdrawals` and `withdrawalsRoot`.
* [EIP-6049: Deprecate SELFDESTRUCT](https://eips.ethereum.org/EIPS/eip-6049)
  * No changes required. We could technically hook into the EVM steps and `console.log` a deprecation notice if the SELFDESTRUCT opcode is hit, but I feel like this sort of warning should happen at compilation, and not by a dev net client like Ganache. Also, maintaining the deprecation notice forever would suck. :thinking: 
  
This also fixes two bugs:

* block `size` was not computed correctly (persisted databases will be updated and migrated automatically)
  * the database migration may add a one-time delay to startup when using a persistent database (--database.dbPath). This should be insignificant for the majority of cases. The migration was benchmarked with 180000 blocks, and completed in 6.4 seconds.
* internal "intrinsic gas" checks now correctly include `accessList` fees, if there are any.


davidmurdoch:
Note: CI failed due to time outs several times. I'm concerned that this PR introduces a performance regression and need to test Test timings in a more controlled manner (locally) before merging.

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -34,7 +34,10 @@ describe("@ganache/console.log", () => {
       wallet: { deterministic: true, totalAccounts: 1 },
       miner: { blockGasLimit: "0xfffffffff" },
       logging: { logger },
-      chain: { allowUnlimitedContractSize: true },
+      chain: {
+        allowUnlimitedContractSize: true,
+        allowUnlimitedInitCodeSize: true
```
davidmurdoch, line 39: The console.log test contract is so big we need to lift the initcode limit as well as the contract size limit.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1334,15 +1348,19 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
         // We get the memory as one large array.
         // Let's cut it up into 32 byte chunks as required by the spec.
+        const limit = Number(event.memoryWordCount);
+        memory = Array(limit);
         let index = 0;
-        while (index < event.memory.length) {
-          const slice = event.memory.slice(index, index + 32);
-          memory.push(TraceData.from(Buffer.from(slice)));
-          index += 32;
+        while (index < limit) {
```
davidmurdoch, line 1373: Ethereum JS changed the way they slice and dice memory, and the `event.memory` Buffer here might be waaaaayyy too big (like 80x+ times too big), causing us to copy the whole Buffer, instead of just the actual current memory range (which can be computed by using `memoryWordCount`).

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1284,7 +1298,14 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
       this.emit("ganache:vm:tx:before", {
         context: transactionEventContext
       });
-      await vm.runTx({ tx, block: block as any });
+      await vm.runTx({
+        skipHardForkValidation: true,
+        skipNonce: true,
+        skipBalance: true,
+        skipBlockGasLimitValidation: true,
```
davidmurdoch, line 1311: We don't need to check these since we've already run all these transactions before, these validations shouldn't be needed.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1431,6 +1456,10 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
       context: transactionEventContext
     });
     await vm.runTx({
+      skipHardForkValidation: true,
```
davidmurdoch, line 1465: We don't need to check these since we've already run all these transactions before, these validations shouldn't be needed.

src/chains/ethereum/ethereum/src/forking/fork.ts
```diff
@@ -297,18 +302,22 @@ export class Fork {
...
-        {
-          chainId: this.chainId,
-          defaultHardfork: common.hardfork()
-        },
-        { baseChain: 1 }
-      );
+      if (this.#options.chain.allowUnlimitedInitCodeSize) {
+        removeEIP3860InitCodeSizeLimitCheck(forkCommon);
+      }
+      return forkCommon;
```
davidmurdoch, line 320: just reorging some code so we can call `removeEIP3860InitCodeSizeLimitCheck`

src/chains/ethereum/ethereum/src/helpers/disable-common-eip.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Effectively removes the 3860 InitCode Size Limit check by increasing the init
+ * code size to Number.MAX_SAFE_INTEGER.
+ *
+ * This number works because an initcode can practically never be that large
+ * (approx 9 PetaBytes!).
+ *
+ * @param common
+ * @returns
+ */
+export function removeEIP3860InitCodeSizeLimitCheck(common: Common) {
```
davidmurdoch: this is so we can support our new `allowUnlimitedInitCodeSize` option.

src/chains/ethereum/ethereum/src/helpers/disable-common-eip.ts
```diff
@@ -0,0 +1,67 @@
...
+  const original = common.paramByEIP.bind(common);
+  common.paramByEIP = function (
+    paramTopic: string,
+    paramName: string,
+    paramEip: number
+  ) {
+    if (paramTopic === topic && paramName === name && paramEip === eip) {
+      return value;
+    }
+    return original.call(common, paramTopic, paramName, paramEip);
```
davidmurdoch: duck punching FTW!

src/chains/ethereum/ethereum/tests/api/debug/debug.test.ts
```diff
@@ -112,14 +112,18 @@ describe("api", () => {
...
         if (memory.length > 0) {
           // check formatting of memory
-          assert.strictEqual(memory[0].length, 64);
-          assert.notStrictEqual(memory[0].substr(0, 2), "0x");
-          break;
+          assert.strictEqual(memory[0].length, 64); // one word
+          assert.notStrictEqual(memory[0].substring(0, 2), "0x"); // 0x prefixed
         }
       }
+      assert.strictEqual(actualMemoryWordsLength, expectedMemoryWordsLength);
```
davidmurdoch, line 126: this is a test to ensure the `debug_traceTransaction` code change I made actually worked.

src/chains/ethereum/ethereum/tests/api/eth/contracts/EthCall.sol
```diff
@@ -2,17 +2,23 @@
...
-}
\ No newline at end of file
+    function doARevert() public pure {
+        revert("you are a failure");
+    }
+
+    function getCoinbase() public view virtual returns (uint256) {
+        uint256 balance = address(block.coinbase).balance;
+        require(balance != 0, "coinbase balance is not zero");
+        return gasleft();
```
davidmurdoch, line 22: A weird test that gets the `coinbase` address (since this had a gas sideeffect) but returns the `gasLeft`. New fn names welcome!

src/packages/ganache/package.json
```diff
@@ -102,7 +102,8 @@
     "emittery": "0.10.0",
     "keccak": "3.0.2",
     "leveldown": "6.1.0",
-    "secp256k1": "4.0.3"
+    "secp256k1": "4.0.3",
+    "@trufflesuite/uws-js-unofficial": "20.10.0-unofficial.2"
```
davidmurdoch, line 106: EJS got a little bigger, putting us over our size limit. So, for now, we've booted `uws-js-unofficial` into npm download step. This brings us down to 56 MB (used to be around 96MB, I think)

src/chains/ethereum/ethereum/tests/contracts/Create.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.11;
+
+contract Create {
+    function create() public returns (address addr) {
```
davidmurdoch, line 5: for testing EIP-3860

src/chains/ethereum/ethereum/src/helpers/disable-common-eip.ts
jeffsmale90: 🦆 👊

src/chains/ethereum/transaction/tests/index.test.ts
```diff
@@ -419,6 +425,22 @@ describe("@ganache/ethereum-transaction", async () => {
       it("has a function to check capability support", () => {
         assert.strictEqual(vmTx.supports(2930), false);
       });
+      describe("EIP-3860", () => {
+        it("computes intrinstic gas correctly pre-shanghai (EIP-3860)", () => {
+          const common= new Common({chain: "mainnet", hardfork: "london"});
```
jeffsmale90:
```suggestion
          const common = new Common({chain: "mainnet", hardfork: "london"});
```

jeffsmale90: I think prettier isn't working on this for some reason, as there's a couple of prettier issues.

src/chains/ethereum/transaction/src/params.ts
```diff
@@ -56,7 +58,14 @@ export const Params = {
   /**
    * Fee for creation a transaction
    */
-  TRANSACTION_CREATION: 32000n,
+  TRANSACTION_CREATION_GAS: 53000n,
```
jeffsmale90, line 61: I was a little confused by this, cos all the documentation says "Creation of a contract costs 32,000 gas", but turns out it's + base tx fee.

I feel that is worth a comment.

src/chains/ethereum/transaction/src/base-transaction.ts
```diff
@@ -22,11 +22,15 @@ export const calculateIntrinsicGas = (
...
-  // if it doesn't have a "to" address this is a contract creation and it costs
-  // `TRANSACTION_CREATION` more gas.
-  if (!hasToAddress) gas += Params.TRANSACTION_CREATION;
+  let gas: bigint;
+  // set the starting gas for the transaction
+  if (hasToAddress) {
+    gas = Params.TRANSACTION_GAS;
+  } else {
+    // if it doesn't have a "to" address this is a contract creation and it costs
+    // `TRANSACTION_CREATION` gas.
```
jeffsmale90:
```suggestion
    // `TRANSACTION_CREATION_GAS`.
```

jeffsmale90: Additionally, I wonder if this is easier to read as a conditional:
```
let gas = hasToAddress
  ? Params.TRANSACTION_GAS
  : Params.TRANSACTION_CREATION_GAS;
```

🚴🏚

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -337,6 +338,128 @@ describe("provider", () => {
...
+            createTx
+          ]);
+        const { status: unlimitedStatus } =
+          await unlimitedInitCodeProvider.send("eth_getTransactionReceipt", [
+            unlimitedInitCodeCreateHash
+          ]);
+        assert.strictEqual(
+          unlimitedStatus,
+          "0x1",
+          "It failed when it should work"
```
jeffsmale90, line 459: I love this assertion message 😍

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -337,6 +338,128 @@ describe("provider", () => {
...
+        const { status: limitedStatus } = await limitInitCodeProvider.send(
+          "eth_getTransactionReceipt",
+          [limitInitCodeCreateHash]
+        );
+
+        // it should fail
+        assert.strictEqual(
+          limitedStatus,
+          "0x0",
+          "It worked when it should fail"
```
jeffsmale90, line 425: I love this one even more 😍 😍

src/chains/ethereum/ethereum/tests/api/eth/contracts/EthCall.sol
```diff
@@ -2,17 +2,23 @@
...
-  }
+    constructor() payable {
+        value = 5;
+    }
 
-  function getBaseFee() public view virtual returns (uint256) {
-    return block.basefee;
-  }
+    function getBaseFee() public view virtual returns (uint256) {
+        return block.basefee;
```
jeffsmale90, line 12: Is the indentation here squiffy or is it just me?

jeffsmale90 (changes requested):

src/chains/ethereum/block/src/block.ts
```diff
@@ -116,23 +166,6 @@ export class Block {
     }
   }
 
-  static fromParts(
```
davidmurdoch, line 119: this isn't used anywhere

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -115,13 +135,9 @@ export class RuntimeBlock {
...
       mixHash,
-      prevRandao: mixHash
+      prevRandao: mixHash,
+      withdrawalsRoot
     };
-    // When forking we might get a block that doesn't have a baseFeePerGas value,
-    // but EIP-1559 might be active on our chain. We need to keep track on if
-    // we should serialize the baseFeePerGas value or not based on that info.
-    // this will be removed as part of https://github.com/trufflesuite/ganache/pull/1537
-    if (baseFeePerGas === undefined) this.serializeBaseFeePerGas = false;
```
davidmurdoch, line 124: We just set `baseFeePerGas` to `undefined` in the header, instead of setting it to `0` and adding this flag.

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -113,15 +131,11 @@ export class RuntimeBlock {
       gasLimit: gasLimit.toBigInt(),
       gasUsed: gasUsed.toBigInt(),
       timestamp: timestamp.toBigInt(),
-      baseFeePerGas: baseFeePerGas ?? 0n,
+      baseFeePerGas: baseFeePerGas ?? undefined,
```
davidmurdoch: We just set `baseFeePerGas` to undefined here in the `header`, instead of setting it to `0` and adding a `serializeBaseFeePerGas` flag (below).

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -156,21 +171,27 @@ export class RuntimeBlock {
...
-      rawHeader,
-      txs,
-      [],
+    const txs: BlockRawTransaction[] = Array(transactions.length);
+    const extraTxs: GanacheRawBlockTransactionMetaData[] = Array(
+      transactions.length
+    );
+    for (let i = 0; i < transactions.length; i++) {
+      const tx = transactions[i];
+      txs[i] = tx.raw.length === 9 ? tx.raw : tx.serialized;
```
davidmurdoch: `tx.raw.length === 9` means it's a legacy transaction and we should use the raw array of Buffers.

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -156,21 +171,27 @@ export class RuntimeBlock {
...
+      extraTxs[i] = [tx.from.toBuffer(), tx.hash.toBuffer()];
+    }
+    let rawBlock: EthereumRawBlock;
+    rawBlock = isEip4895 ? [rawHeader, txs, [], []] : [rawHeader, txs, []];
```
davidmurdoch: So this on two lines only because if I put it on one it puts it on lots of lines, and I didn't like it. haha. Not sure I like this either though. So it is up to you, dear reviewer, to decide which way you hate more.

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -179,20 +200,16 @@ export class RuntimeBlock {
...
-    (block as any)._rawTransactions = txs;
-    (block as any).header = makeHeader(rawHeader, totalDifficulty);
-    (block as any).serializeBaseFeePerGas = rawHeader[15] === undefined;
-    (block as any)._rawTransactionMetaData = extraTxs;
-    (block as any)._size = size;
+    const block: any = new Block(null, this._common);
+    block._raw = rawHeader;
+    block._rawTransactions = txs;
+    block.header = makeHeader(rawHeader, totalDifficulty);
+    block._rawWithdrawals = [];
```
davidmurdoch, line 204: This is the only new line in this mess. The `any`'s were annoying me so I switched it up.

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -179,20 +200,16 @@ export class RuntimeBlock {
...
-      // TODO(hack)!
-      transactions.length > 0 ? transactions[0].common : null
-    );
-    (block as any)._raw = rawHeader;
-    (block as any)._rawTransactions = txs;
-    (block as any).header = makeHeader(rawHeader, totalDifficulty);
-    (block as any).serializeBaseFeePerGas = rawHeader[15] === undefined;
-    (block as any)._rawTransactionMetaData = extraTxs;
-    (block as any)._size = size;
+    const block: any = new Block(null, this._common);
```
davidmurdoch, line 200: We use to not pass a `common` to the block, but since we do now we don't need to yank it from the transactions.

src/packages/utils/src/utils/constants.ts
```diff
@@ -8,3 +8,8 @@ export const BUFFER_8_ZERO = BUFFER_256_ZERO.slice(0, 8);
 export const WEI = 1000000000000000000n as const;
 
 export const KNOWN_CHAINIDS = new Set([1, 3, 4, 5, 42, 11155111]);
+
+/**
+ * Buffer representation of the string "version":
+ */
+export const VERSION_KEY = Buffer.from([118, 101, 114, 115, 105, 111, 110]);
```
davidmurdoch, line 15: We use `version` this in two database tables now, so it earned itself a `constant` home.

src/chains/ethereum/block/src/serialize.ts
```diff
@@ -26,24 +43,48 @@ export type EthereumRawBlockHeader = [
...
+ * @param start
+ * @param end
+ * @returns
+ */
 export function serialize(
-  raw: Head<GanacheRawBlock>
-): { serialized: Buffer; size: number } {
-  const serializedStart = encodeRange(raw, 0, 3);
+  start: Head<EthereumRawBlock> | EthereumRawBlock,
+  end: Head<GanacheRawBlockExtras>
```
davidmurdoch, line 86: Previously, we passing in an array of Buffers and then encoding it into two parts. Now we just pass in the two parts separately.

src/chains/ethereum/block/src/serialize.ts
```diff
@@ -53,3 +94,50 @@ export function serialize(
     size
   };
 }
+
+function isLegacyRawTransaction(
+  raw: Buffer | LegacyRawTransaction
+): raw is LegacyRawTransaction {
+  return raw.length === 9;
```
davidmurdoch, line 109: This function is really just for the `: raw is LegacyRawTransaction` part that is used to just help TypeScript figure out the types correctly.

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -949,7 +951,8 @@ export default class EthereumApi implements Api {
         options.miner.difficulty,
         parentHeader.totalDifficulty,
         blockchain.getMixHash(parentHeader.parentHash.toBuffer()),
-        0n // no baseFeePerGas for estimates
+        0n, // no baseFeePerGas for estimates
+        KECCAK256_RLP
```
davidmurdoch, line 955: `KECCAK256_RLP` is the empty trie root: https://github.com/ethereumjs/merkle-patricia-tree/blob/master/src/baseTrie.ts#L56

src/chains/ethereum/transaction/src/transaction-factory.ts
```diff
@@ -193,19 +200,57 @@ export class TransactionFactory {
     switch (txType) {
       case TransactionType.EIP1559AccessList:
         return EIP1559FeeMarketTransaction.fromTxData(
-          txData.slice(1) as EIP1559FeeMarketDatabasePayload,
+          txData.slice(1) as EIP1559FeeMarketRawTransaction,
```
davidmurdoch, line 215: This diff looks weird, but all that really happened here is that I added a new function, `fromTypeAndTxData` (below), and some type names have changed.

src/chains/ethereum/transaction/src/signing.ts
```diff
@@ -190,35 +191,74 @@ export const computeIntrinsicsLegacyTx = (
...
-  const encodedData = encodeRange(raw, 1, 8);
-  const encodedSignature = encodeRange(raw, 9, 3);
-  const serialized = Buffer.concat([
-    typeBuf,
-    digest(
-      [encodedData.output, encodedSignature.output],
-      encodedData.length + encodedSignature.length
-    )
-  ]);
+  const encodedData = encodeRange(raw, 0, 8);
```
davidmurdoch, line 240: raw data doesn't include the type anymore.

src/chains/ethereum/transaction/src/signing.ts
```diff
@@ -166,15 +167,15 @@ export const computeFromAddress = (
...
 ) => {
   const encodedData = encodeRange(raw, 0, 6);
   const encodedSignature = encodeRange(raw, 6, 3);
-  const serialized = digest(
-    [encodedData.output, encodedSignature.output],
-    encodedData.length + encodedSignature.length
-  );
+  const ranges = [encodedData.output, encodedSignature.output];
+  const length = encodedData.length + encodedSignature.length;
+  const serialized = digest(ranges, length);
```
davidmurdoch, line 177: Just making this one match the others' format

src/chains/ethereum/transaction/src/legacy-transaction.ts
```diff
@@ -187,10 +187,10 @@ export class LegacyTransaction extends RuntimeTransaction {
...
     this.raw = raw;
     const encodedSignature = encodeRange(raw, 6, 3);
-    this.serialized = digest(
-      [data.output, encodedSignature.output],
-      dataLength + encodedSignature.length
-    );
+    const ranges = [data.output, encodedSignature.output];
+    const length = dataLength + encodedSignature.length;
+    // serialized is the rlp encoded raw data
+    this.serialized = digest(ranges, length);
```
davidmurdoch, line 194: no real changes other that changing the format to match the others

src/chains/ethereum/transaction/src/eip2930-access-list-transaction.ts
```diff
@@ -162,6 +158,9 @@ export class EIP2930AccessListTransaction extends RuntimeTransaction {
       }
     };
   }
+  public calculateIntrinsicGas(): bigint {
+    return super.calculateIntrinsicGas() + this.accessListDataFee;
```
davidmurdoch, line 162: This is a bug fix. We weren't include these costs in intrinsic costs when we should.

src/chains/ethereum/transaction/src/eip1559-fee-market-transaction.ts
```diff
@@ -178,6 +172,9 @@ export class EIP1559FeeMarketTransaction extends RuntimeTransaction {
       }
     };
   }
+  public calculateIntrinsicGas(): bigint {
+    return super.calculateIntrinsicGas() + this.accessListDataFee;
```
davidmurdoch, line 176: This is a bug fix. We weren't include these costs in intrinsic costs when we should.

src/chains/ethereum/transaction/src/base-transaction.ts
davidmurdoch: but where would the comments go, hmmmmmm? :-D

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: its my finest work

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: okay THIS is my finest work

src/chains/ethereum/ethereum/tests/api/eth/contracts/EthCall.sol
davidmurdoch: solidity is supposed to be 4 spaces. this file missed that memo, i guess.

src/chains/ethereum/ethereum/src/helpers/common-helpers.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Effectively removes the 3860 InitCode Size Limit check by increasing the init
+ * code size to Number.MAX_SAFE_INTEGER.
+ *
+ * This number works because an initcode can practically never be that large
+ * (approx 9 PetaBytes!).
+ *
+ * @param common
+ * @returns
+ */
+export function removeEIP3860InitCodeSizeLimitCheck(common: Common) {
```
davidmurdoch, line 13: this is a hack until EJS ships a `allowUnlimitedInitCodeSize` option.

src/chains/ethereum/ethereum/src/forking/persistent-cache/helpers.ts
```diff
@@ -56,7 +56,6 @@ export function getBlockNumberFromParams(method: string, params: any[]) {
   }
 }
 
-const VERSION_KEY = Buffer.from("version", "utf-8");
```
davidmurdoch, line 59: moved to `constants.ts`

src/chains/ethereum/block/src/block.ts
```diff
@@ -41,13 +51,42 @@ export class Block {
...
+   *
+   * This migration updates the `size` value of the block to the correct value
+   * by re-serializing the block for storage in the db.
+   * @param serialized
+   * @returns
+   */
+  static migrate(serialized: Buffer) {
+    const deserialized = decode<GanacheRawBlock>(serialized);
+    return serialize(
+      deserialized.slice(0, 3) as Head<EthereumRawBlock>,
```
jeffsmale90, line 85: I don't think this is super critical, but as far as I can tell the `Head<T>` here is to remove the `withdrawals` from `EthereumRawBlock`, but unless I'm mistaken, it's already optional, based on the definition of `EthereumRawBlock`?

https://github.com/trufflesuite/ganache/pull/4272/files#diff-97cedef4b674bd7c1fceaad9d0b9dde6d59bd8ce65756c0bac54dedab35900bdR66
```
export type EthereumRawBlock = _EthereumRawBlock | Head<_EthereumRawBlock>;
```

src/chains/ethereum/ethereum/tests/database.test.ts
```diff
@@ -1,126 +1,252 @@
...
-    tmp.setGracefulCleanup();
-    dbPath = (await tmp.dir()).path;
-    provider = await startProvider();
-    accounts = Object.keys(provider.getInitialAccounts());
-  });
-
-  afterEach(async () => {
-    if (provider) {
-      await provider.disconnect();
+  describe("resumption", () => {
```
davidmurdoch, line 9: nothing in this `describe` block is a new test, I just needed two `describe`s.

src/chains/ethereum/ethereum/tests/database.test.ts
```diff
@@ -1,126 +1,252 @@
...
-    });
-    const latestBlockAfterRevert = await provider.request({
-      method: "eth_blockNumber",
-      params: []
-    });
-
-    // make sure the revert worked (only checks in memory state, not the db):
-    assert.strictEqual(endingBalanceAfterRevert, endingBalance);
-    assert.strictEqual(latestBlockAfterRevert, latestBlock);
+    it("resumes blockchain from a prior state", async () => {
```
davidmurdoch, line 34: nothing here has changed other than indentation.

src/chains/ethereum/ethereum/tests/database.test.ts
```diff
@@ -1,126 +1,252 @@
...
+      });
+      assert.strictEqual(latestBlockAfterRestart, latestBlock);
     });
-    assert.strictEqual(balanceAfterRestart, endingBalance);
+  });
 
-    const latestBlockAfterRestart = await provider.request({
-      method: "eth_blockNumber",
-      params: []
+  describe("migration", () => {
```
davidmurdoch, line 130: this is the new test

davidmurdoch (commented):
added migration tests

src/chains/ethereum/block/src/block.ts
```diff
@@ -41,13 +52,57 @@ export class Block {
...
+   * @returns
+   */
+  static migrate(serialized: Buffer) {
+    const deserialized = decode<GanacheRawBlock>(serialized);
+    const start = deserialized.slice(0, 3) as EthereumRawBlock;
+    start[1] = start[1].map((oldRawTx: any) => {
+      if (oldRawTx.length === 9) {
+        return oldRawTx; // legacy transactions are fine
+      } else {
+        // `type` is always `< 0x27`, so we can yank the first byte from the
```
jeffsmale90: Should this comment be:

```suggestion
        // `type` is always `< 0x7F`, so we can yank the first byte from the
```

??

src/chains/ethereum/block/src/block.ts
```diff
@@ -57,16 +112,18 @@ export class Block {
 
   getTransactions() {
     const common = this._common;
+    const hashBuffer = this.hash().toBuffer();
```
jeffsmale90: How about calling this `blockHash`? Might reduce confusion around `hash` and `hashBuffer`.

src/chains/ethereum/block/src/block.ts
```diff
@@ -101,8 +158,11 @@ export class Block {
       hash,
       ...header,
       size: Quantity.from(this._size),
-      transactions: jsonTxs,
-      uncles: [] as Data[] // this.value.uncleHeaders.map(function(uncleHash) {return to.hex(uncleHash)})
+      transactions,
+      uncles: [] as Data[], // this.value.uncleHeaders.map(function(uncleHash) {return to.hex(uncleHash)})
```
jeffsmale90, line 162: #not-your-change _but_ is there a reason we ignore uncles?

src/chains/ethereum/block/src/block.ts
```diff
@@ -101,8 +158,11 @@ export class Block {
...
+      // if `this._rawWithdrawals` is not set we should not include it in the
+      // JSON response ()`undefined` gets stripped when JSON.stringify is called).
```
jeffsmale90:
```suggestion
      // JSON response (`undefined` gets stripped when JSON.stringify is called).
```

src/chains/ethereum/block/src/block.ts
```diff
@@ -101,8 +158,11 @@ export class Block {
...
+      withdrawals: this._rawWithdrawals?.map(convertRawWithdrawals) || undefined
```
jeffsmale90: won't `this._rawWithdrawals?.map(convertRawWithdrawals)` resolve to `undefined` anyways if `_rawWithdrawals` is not set, making `|| undefined` unnecessary?

src/chains/ethereum/block/src/runtime-block.ts
jeffsmale90: Isn't `baseFeePerGas` going to be `undefined` anyways?
```suggestion
      baseFeePerGas: baseFeePerGas
```

src/chains/ethereum/block/src/runtime-block.ts
```diff
@@ -156,21 +164,30 @@ export class RuntimeBlock {
...
-      [],
+    const txs: BlockRawTransaction[] = Array(transactions.length);
+    const extraTxs: GanacheRawBlockTransactionMetaData[] = Array(
+      transactions.length
+    );
+    for (let i = 0; i < transactions.length; i++) {
+      const tx = transactions[i];
+      txs[i] =
+        tx.raw.length === 9
+          ? tx.raw
```
jeffsmale90: I like a comment, although I'm not sure this is the correct one?
```suggestion
          ? tx.raw // legacy transactions are correctly encoded
```

src/chains/ethereum/block/src/runtime-block.ts
jeffsmale90: I'd vote for this, because (although not _super_ tidy) functionally different (better).
```suggestion
    const rawBlock: EthereumRawBlock = isEip4895
      ? [rawHeader, txs, [], []]
      : [rawHeader, txs, []];
```

src/chains/ethereum/block/src/serialize.ts
```diff
@@ -1,9 +1,33 @@
...
+  index: Buffer,
+  validatorIndex: Buffer,
+  address: Buffer,
+  amount: Buffer
+];
+
+export type Withdrawal = {
+  index: Quantity;
+  validatorIndex: Quantity;
+  address: Data;
```
jeffsmale90: Why not `: Address`?

src/chains/ethereum/block/src/serialize.ts
jeffsmale90: We `.length === 9` in a bunch of places throughout the code, it'd be nice if we could actually use this function in all of those places, but 🤷 it may be more complicated than it's worth.

src/chains/ethereum/block/src/serialize.ts
```diff
@@ -53,3 +101,50 @@ export function serialize(
     size
   };
 }
+
+function isLegacyRawTransaction(
+  raw: Buffer | LegacyRawTransaction
```
jeffsmale90: could this be `raw: BlockRawTransaction`?

src/chains/ethereum/block/src/serialize.ts
```diff
@@ -53,3 +101,50 @@ export function serialize(
...
+
+/**
+ * Converts a raw transaction encoded for use in a raw block into a `Transaction`
+ *
+ * @param raw the raw transaction data after the block has been rlp decoded.
+ * @param common
+ * @param extra
+ * @returns
+ */
+export function convertRawBlockTransaction(
```
jeffsmale90: I'm not 100% on the naming of this function, as `convert` doesn't specify what it's converting _to_.

`blockTransactionFromRaw` might work?

src/chains/ethereum/ethereum/src/api.ts
```diff
@@ -2883,7 +2885,8 @@ export default class EthereumApi implements Api {
       options.miner.difficulty,
       parentHeader.totalDifficulty,
       blockchain.getMixHash(parentHeader.parentHash.toBuffer()),
-      baseFeePerGasBigInt
+      baseFeePerGasBigInt,
+      KECCAK256_RLP
```
jeffsmale90, line 2892: Should this check whethere EIP4895 is active, in which case `undefined` should be passed for `withdrawalsRoot`? Does it matter in any case?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1159,6 +1174,11 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         warmPrecompiles(eei);
         eei.addWarmedAddress(caller);
         if (to) eei.addWarmedAddress(to.buf);
+
+        // shanghai hardfork warm coinbase address
+        if (common.isActivatedEIP(3651) === true) {
```
jeffsmale90: Why the ` === true` ?

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -1159,6 +1174,11 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
         warmPrecompiles(eei);
```
jeffsmale90, line 1174: For some reason the ethereumjs types don't seem to be importing, and the return type of `createVmFromStateTrie` is `any`, which means that `vm`, and `eei` are both any :(

Dunno if it's just me.

src/chains/ethereum/ethereum/src/blockchain.ts
jeffsmale90: With this reverted, we get a test failure:
```
api
 debug
  should trace a successful transaction without changing state:
  TypeError: Cannot read property 'length' of null
```

But, what do you think about adding an explicit test for this? Seems something that we would want to safeguard against reverting.

src/chains/ethereum/ethereum/src/data-managers/block-manager.ts
```diff
@@ -77,43 +85,68 @@ export default class BlockManager extends Manager<Block> {
...
+      ];
+      const tx = TransactionFactory.fromRpc(txJson, common, txExtra);
+      txs[index] =
+        tx.raw.length === 9
+          ? tx.raw
+          : tx.serialized ?? encodeWithPrefix(tx.type.toNumber(), tx.raw);
+      extraTxs[index] = blockExtra;
+    }
+
+    let start: EthereumRawBlock | Head<EthereumRawBlock>;
```
jeffsmale90: Can this just be:
```suggestion
    let start: EthereumRawBlock;
```

Because `EthereumRawBlock` already is `_EthereumRawBlock | Head<_EthereumRawBlock>`

src/chains/ethereum/utils/src/errors/runtime-error.ts
```diff
@@ -39,7 +39,10 @@ export class RuntimeError extends CodedError {
 
     this.data = {
       hash: hash,
-      programCounter: execResult.runState.programCounter,
+      // in some failure scenarios, like when the initcode is too large,
+      // `runState` is undefined. In that case, we'll just use 0 for the
+      // programCounter.
+      programCounter: execResult.runState?.programCounter || undefined,
```
jeffsmale90: why the `|| undefined`?

src/chains/ethereum/transaction/src/transaction-factory.ts
```diff
@@ -193,19 +200,57 @@ export class TransactionFactory {
...
+    }
+  }
+  /**
+   * Create a transaction from a `txData` object without the type field in the first position (for type 1 and 2 txs)
+   *
+   * @param txData - The raw transaction data. The `type` field will determine which transaction type is returned (if undefined, creates a legacy transaction)
+   * @param common - Options to pass on to the constructor of the transaction
+   */
+  public static fromTypeAndTxData(
+    txType: number,
```
jeffsmale90: Can this / should this be `:TransactionType`?

jeffsmale90: Also, this looks real similar to 
```
private static _fromData(
    txData: Transaction | TypedRawTransaction,
    txType: TransactionType,
    common: Common,
    extra?: GanacheRawExtraTx
  )
```

Can they be combined / reused in some way?

src/chains/ethereum/ethereum/src/api.ts
jeffsmale90: I don't think it does - so feel free to ignore this comment if that's true in all cases.

src/chains/ethereum/ethereum/src/database.ts
```diff
@@ -44,35 +50,96 @@ export default class Database extends Emittery {
...
+      ops.push({
+        type: "put",
+        key: Buffer.concat([prefix, key]),
+        value: Block.migrate(value)
+      });
+    }
+    //#endregion MIGRATION 0b
+
+    // save all in one atomic operation:
+    await this.db.batch(ops);
```
jeffsmale90, line 115: can we `this.blocks.batch(ops);` to avoid the `prefix` magic above on :100 and the key concatenation on :108?

Ah - I see that `this.db.batch` is intended to be a single operation for _all_ migrations. Maybe not worth it for the (potential) minor performance improvement vs the maintenance cost next time we need to add a migration 🤷

src/chains/ethereum/ethereum/src/helpers/common-helpers.ts
jeffsmale90: Worth a comment?
```suggestion
export function removeEIP3860InitCodeSizeLimitCheck(common: Common) {
  // this is a hack until EJS ships `allowUnlimitedInitCodeSize` option https://github.com/ethereumjs/ethereumjs-monorepo/issues/2588
```

jeffsmale90: Also, looks like they've settled on `allowUnlimitedInitcodeSize` lowercase `code` - do we want to match?

src/chains/ethereum/options/src/chain-options.ts
```diff
@@ -151,6 +167,13 @@ export const ChainOptions: Definitions<ChainConfig> = {
     legacyName: "allowUnlimitedContractSize",
     cliType: "boolean"
   },
+  allowUnlimitedInitCodeSize: {
+    normalize,
+    cliDescription:
+      "Allows unlimited initcode (`transaction.data`) sizes while debugging. Setting this to `true` will cause ganache to behave differently than production environments.",
```
jeffsmale90: I think the description suggests that this only works in debugging context - I understand the intention, and feel that we've given sufficient warning with "...behave differently...".

```suggestion
      "Allows unlimited initcode (`transaction.data`) sizes. Setting this to `true` will cause ganache to behave differently than production environments.",
```

jeffsmale90 (changes requested):
Partial review - 47/75

src/chains/ethereum/block/src/block.ts
davidmurdoch: never got around to implementing them when I added forking in. we don't generate uncles ourselves. I think it's trivial to do though.

(resolve this comment once you read it)

src/chains/ethereum/block/src/runtime-block.ts
davidmurdoch:
```suggestion
          ? tx.raw // legacy transactions don't have their own encoding
```

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: image

Works on my machine :-) Maybe the version of TypeScript your editor is using is different from the one our code uses?

src/chains/ethereum/ethereum/src/blockchain.ts
```suggestion
        if (common.isActivatedEIP(3651)) {
```

src/chains/ethereum/ethereum/src/blockchain.ts
davidmurdoch: Normally, yes, however, EJS has already fixed this bug upstream - they just haven't released the fix yet.

Anyway, I _do_ test this change now in the "should trace a successful transaction without changing state" test. It's just not a test specifically for this workaround, but more of a "does tracing work as expected" test.

src/chains/ethereum/ethereum/src/database.ts
davidmurdoch: We actually do have a mechanism for batching across all db "sublevels" that Just Works :tm: . But it's supposed to beused synchronously, but we're streaming the values here. It's possible to get it to work in this specific case since migrations happen before the db can be written to from anywhere else ... but it seemed really fragile so I didn't use it (that and it's actually a bit cumbersome to use here).

See blockchain.ts's use of `this.#database.batch` for an example.

In implementing this I did dive into the implementation details of `sublevel`, and it does seem to be a bit inefficient for what it does (pretty much just prepends a prefix when writing, and then removes the prefix when reading) -- and the above is likely slightly faster than going through the block sublevel.

src/chains/ethereum/ethereum/src/helpers/common-helpers.ts
davidmurdoch: They switched to `allowUnlimitedInitCodeSize`: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2597

src/chains/ethereum/transaction/src/transaction-factory.ts
davidmurdoch: Good catch. I've updated this code. So one of these validates user data, and the other trusts it and uses the type as is. The new function names and comments hopefully reflects how to use these two methods appropriately.

I think the types for `_fromData` are actually too wide. Made an issue for this: https://github.com/trufflesuite/ganache/issues/4347

src/chains/ethereum/block/src/serialize.ts
davidmurdoch: Made an issue for it: https://github.com/trufflesuite/ganache/issues/4348

src/chains/ethereum/options/src/chain-options.ts
davidmurdoch: The description mirrors the description for `allowUnlimitedContractSize`. Should we update both?

src/chains/ethereum/utils/src/errors/runtime-error.ts
davidmurdoch: My original choice was to use `0`, and I switched it to `undefined` at one point. I've removed the `undefined` and update the code comment.

src/chains/ethereum/ethereum/src/blockchain.ts
```suggestion
        if (common.isActivatedEIP(3651)) {
```

davidmurdoch: GitHub is acting up, and it appears like this is still `=== true` in the PR view, but if you go to the actual code at this branch's `HEAD` it is fixed.

src/chains/ethereum/ethereum/src/api.ts
davidmurdoch: It doesn't, but I wonder if `undefined // eth_call doesn't need withdrawals, since the block isn't saved` is better? Thoughts?

src/chains/ethereum/ethereum/tests/api/eth/contracts/EthCall.sol
jeffsmale90: `accessCoinBase` might be better, but it doesn't hint at what the return value is :shrug:

src/chains/ethereum/ethereum/tests/database.test.ts
```diff
@@ -1,126 +1,257 @@
...
+      assert.strictEqual(latestBlockAfterRestart, latestBlock);
     });
-    const latestBlockAfterRevert = await provider.request({
-      method: "eth_blockNumber",
-      params: []
+  });
+
+  describe("migration", () => {
+    before(function () {
+      if (process.platform === "win32") {
```
jeffsmale90, line 132: https://github.com/trufflesuite/ganache/issues/4350

src/chains/ethereum/ethereum/tests/forking/forking.test.ts
```diff
@@ -1277,4 +1277,154 @@ describe("forking", function () {
...
+        amount: "0x1b7227",
+        index: "0x1a9ec5",
+        validatorIndex: "0x3a995"
+      }
+    ];
+    before("skip if we don't have the INFURA_KEY", function () {
+      // this test uses the `network: "goerli"` option, which requires an
+      // infura key; when run our tests it must be provided as an environment
+      // variable.
+      if (!process.env.INFURA_KEY) this.skip();
```
jeffsmale90, line 1390: Are we able to check if we are in CI and fail if the INFURA_KEY is not set?

src/chains/ethereum/ethereum/tests/provider.test.ts
```diff
@@ -337,6 +338,128 @@ describe("provider", () => {
       );
       await provider.disconnect();
     }).timeout(10000);
+
+    it("allows unlimited init code in transaction when the allowUnlimitedInitCodeSize option is set", async () => {
+      const largeInitCode = "0x" + "00".repeat(49153); // larger than init code allowance
```
jeffsmale90: `Data.toString(0, 49153)` would do the same 🤷

src/chains/ethereum/ethereum/tests/forking/forking.test.ts
davidmurdoch: Yes, github sets an env var we could check. This is a pattern we copied from elsewhere to skip tests that a run without providing an Infura key.

src/chains/ethereum/ethereum/tests/provider.test.ts
davidmurdoch: Seems like a reasonable change to me! I think this test was a copy+paste of a similar test that predates Data.toString , so I just didn't bother to change the format.

src/chains/ethereum/options/src/chain-options.ts
jeffsmale90: Sure!

src/chains/ethereum/ethereum/tests/forking/forking.test.ts
jeffsmale90: Sounds like a good future improvement :)

jeffsmale90 (approved):

PR #4285, opened 2023-03-02
fix: `node-gyp-build: Permission denied` error

0x416e746f6e:
- Use non-root user to build/run ganache.
- Use newer base image so that python3 is 3.6+.
- Pin base images to tags (not SHA's) since we build cross-platform.


0x416e746f6e:
🙃

PR #4287, opened 2023-03-04
fix: use adjusted time on estimate gas when latest block is being used

### LINKED ISSUE #3528: `eth_estimateGas` does not use blockchain's `adjustedTime`
In the blockchain, when a block is being prepared to be mined, the timestamp is adjusted by the chain's internal offset:
```TypeScript
if (timestamp == null) {
    timestamp = this.#adjustedTime(previousHeader.timestamp);
  }

  return new RuntimeBlock(
    Quantity.from(previousNumber + 1n),
    previousBlock.hash(),
    this.coinbase,
    minerOptions.blockGasLimit.toBuffer(),
    BUFFER_ZERO,
    Quantity.from(timestamp),
    minerOptions.difficulty,
    previousHeader.totalDifficulty,
    Block.calcNextBaseFee(previousBlock)
  );
```
However, `eth_estimateGas` just uses the parent block's timestamp:
```TypeScript
  parentHeader.parentHash,
  parentHeader.miner,
  tx.gas.toBuffer(),
  parentHeader.gasUsed.toBuffer(),
  parentHeader.timestamp,
  options.miner.difficulty,
  parentHeader.totalDifficulty,
  0n // no baseFeePerGas for estimates
);
```

This is particularly annoying if you are using Truffle to test a contract that is time dependent because truffle calls an `eth_estimateGas` before sending your actual transaction. If you use `evm_setTime` to the time needed for your contract to do what it needs to do, the actual transaction would pass, but Truffle doesn't let you get that far because the `eth_estimateGas` fails beforehand.

luzzif:
When estimating gas, use the adjusted timestamp in case the `latest` block tag is being used. This helps avoid some errors in time-dependent contracts when using Ganache in fork mode. 

Should fix https://github.com/trufflesuite/ganache/issues/3528.

This is my first PR here, would love some guidance.

### DIFF
--- a/src/chains/ethereum/ethereum/src/blockchain.ts
+++ b/src/chains/ethereum/ethereum/src/blockchain.ts
@@ -16,7 +16,9 @@ import {
   StructLog,
   TraceTransactionOptions,
   EthereumRawAccount,
-  TraceTransactionResult
+  TraceTransactionResult,
+  QUANTITY,
+  Tag
 } from "@ganache/ethereum-utils";
 import type { InterpreterStep } from "@ethereumjs/evm";
 import { decode } from "@ganache/rlp";
@@ -49,7 +51,10 @@ import {
   calculateIntrinsicGas,
   InternalTransactionReceipt,
   VmTransaction,
-  TypedTransaction
+  TypedTransaction,
+  LegacyTransaction,
+  EIP2930AccessListTransaction,
+  EIP1559FeeMarketTransaction
 } from "@ganache/ethereum-transaction";
 import { Block, RuntimeBlock, Snapshots } from "@ganache/ethereum-block";
 import {
@@ -77,6 +82,7 @@ import { dumpTrieStorageDetails } from "./helpers/storage-range-at";
 import { GanacheStateManager } from "./state-manager";
 import { TrieDB } from "./trie-db";
 import { Trie } from "@ethereumjs/trie";
+import { Ethereum } from "./api-types";
 
 const mclInitPromise = mcl.init(mcl.BLS12_381).then(() => {
   mcl.setMapToMode(mcl.IRTF); // set the right map mode; otherwise mapToG2 will return wrong values.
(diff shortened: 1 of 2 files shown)


src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -577,6 +583,40 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
 
   coinbase: Address;
 
+  gasEstimateBlock = async (
+    tx:
+      | LegacyTransaction
+      | EIP2930AccessListTransaction
+      | EIP1559FeeMarketTransaction,
+    blockNumber: QUANTITY | Ethereum.Tag = Tag.latest
```
jeffsmale90, line 591: `QUANTITY` type shouldn't be used inside of the application logic - it's an alias type that's used to describe parameters in the API layer. Instead, accept type `Quantity` here, and do the conversion in the caller.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -577,6 +583,40 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
...
+  ) => {
+    const previousBlock = await this.blocks.get(blockNumber);
+    const previousHeader = previousBlock.header;
+    const options = this.#options;
+
+    let timestamp = previousHeader.timestamp;
+    if (blockNumber === "latest")
+      timestamp = Quantity.from(this.#adjustedTime(previousHeader.timestamp));
```
jeffsmale90, line 599: Seems to me that if the request is for "latest", this should be incrementing the timestamp, otherwise we use the timestamp of the previous block, rather than the "next" block.

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -577,6 +583,40 @@ export default class Blockchain extends Emittery<BlockchainTypedEvents> {
 
   coinbase: Address;
 
+  gasEstimateBlock = async (
```
jeffsmale90, line 586: it'd be nice if this were a verb, ie `createEstimateGasBlock` or something. 

I haven't the time to look into this further today, but I wonder if there's some value in combining the `readyNextBlock` and this function, as they are very similar (it might be that it just becomes too complicated to be worthwhile though).

jeffsmale90 (changes requested):
Thanks for opening this!

I've added a couple of thoughts - can you please also add tests to cover this behaviour?

luzzif:
A bit of delay on this but... I just implemented the same logic on `eth_call` on a local fork, but using a slightly different way. I made the `readyNextBlock` in the `blockchain` class public and when `latest` is passed as the block I use that instead of the standard built `RuntimeBlock`. I wonder if we could simplify the implementation of the `eth_estimateGas` logic by doing the same here. The only thing we probably need to change is that I noticed that in the `eth_estimateGas` case the gas limit defaults to the miner's call gas limit instead of the default block gas limit used in the standard `readyNextBlock` function.

PR #4317, opened 2023-03-21
perf: optimize options normalization

davidmurdoch:


src/packages/options/src/create.ts
```diff
@@ -38,22 +36,22 @@ const checkForConflicts = (
...
   if (hasOwn(options, namespace)) {
     const namespaceOptions = options[namespace];
 
-    for (let i = 0, l = keys.length; i < l; i++) {
-      const key = keys[i];
-      const propDefinition = def[key];
+    for (let i = 0, l = entries.length; i < l; i++) {
+      const entry = entries[i];
+      const key = entry[0];
+      const propDefinition = entry[1];
```
jeffsmale90:
```suggestion
    for (const [key, propDefinition] of entries) {
```

src/packages/options/src/create.ts
```diff
@@ -85,9 +83,10 @@ function fill(defaults: any, options: any, target: any, namespace: any) {
       }
     }
   } else {
-    for (let i = 0, l = keys.length; i < l; i++) {
-      const key = keys[i];
-      const propDefinition = def[key];
+    for (let i = 0, l = entries.length; i < l; i++) {
+      const entry = entries[i];
+      const key = entry[0];
+      const propDefinition = entry[1];
```
jeffsmale90:
```suggestion
    for (const [key, propDefinition] of entries) {
```

src/packages/options/src/create.ts
jeffsmale90: https://github.com/LeanyLabs/js-array-performance tells me that `for of` is the fastest array iteration technique, and it's super tidy, especially with the array destructuring.

jeffsmale90 (changes requested):

jeffsmale90 (approved):

PR #4318, opened 2023-03-21
perf: optimize account init logging

davidmurdoch:


src/packages/cli/src/initialize/ethereum.ts
```diff
@@ -16,21 +16,21 @@ export default function (provider: EthereumProvider, cliSettings: CliSettings) {
...
-  const addresses = Object.keys(accounts);
+  const addresses = Object.entries(accounts);
   const logs = [];
   logs.push("");
   logs.push("Available Accounts");
   logs.push("==================");
   if (addresses.length > 0) {
-    addresses.forEach(function (address, index) {
-      const balance = accounts[address].balance;
+    addresses.forEach(([address, account], index) => {
```
jeffsmale90: I checkout out this `for` benchmark: https://github.com/LeanyLabs/js-array-performance and reran with node 20.2.0 and found that `forEach` is even more slow compared with `for of` than it was in node 14.

```
Array.forEach x 130 ops/sec ±0.22% (84 runs sampled)
for of x 529 ops/sec ±0.57% (94 runs sampled)
```

While we are here, do you think it's worthwhile updating these `Array.forEach`s to `for....of`s?

src/packages/cli/src/initialize/ethereum.ts
```diff
@@ -16,21 +16,21 @@ export default function (provider: EthereumProvider, cliSettings: CliSettings) {
   const liveOptions = provider.getOptions();
   const accounts = provider.getInitialAccounts();
 
-  const addresses = Object.keys(accounts);
+  const addresses = Object.entries(accounts);
   const logs = [];
   logs.push("");
   logs.push("Available Accounts");
```
jeffsmale90: I also found that (on my machine :tm:, node 20.2) with ~100 log lines, just concatenating to a string is ~2x faster than building a dynamic array of strings (up to 100000 log lines): https://github.com/jeffsmale90/js-performance
```
let log = "";
const message = "My message"
log += message + "\n"
```

```
Benchmarking String Builder: 100000:
array.Join with push: 100000 x 279 ops/sec ±0.44% (90 runs sampled)
array.join with indexing: 100000 x 278 ops/sec ±0.46% (90 runs sampled)
string concat: 100000 x 398 ops/sec ±0.39% (92 runs sampled)
string concat with interpolation: 100000 x 380 ops/sec ±0.30% (93 runs sampled)
string concat: 100000 is faster

Benchmarking String Builder: 1000000:
array.Join with push: 1000000 x 22.81 ops/sec ±2.41% (42 runs sampled)
array.join with indexing: 1000000 x 22.81 ops/sec ±2.25% (42 runs sampled)
string concat: 1000000 x 10.83 ops/sec ±6.67% (32 runs sampled)
string concat with interpolation: 1000000 x 10.86 ops/sec ±6.34% (32 runs sampled)
array.join with indexing: 1000000,array.Join with push: 1000000 is faster
```

jeffsmale90 (commented):
Maybe just over optimising here, but these look like some simple changes to bundle together.

jeffsmale90:
I made those additional optimizations in a separate review here: https://github.com/trufflesuite/ganache/pull/4446 incase you don't like them.

jeffsmale90 (approved):

PR #4353, opened 2023-04-13
test: ensure tests that are skipped in development are run in CI

jeffsmale90:
Previously, tests that depend on an INFURA_KEY would be skipped if it were not provided. After this change these tests will fail in CI if no INFURA_KEY is provided.

Additionally, a test which was previously skipped is now run on Windows and Linux agents in Github Actions only. (` > server > listens on given interface only`)

### DIFF
--- a/src/chains/ethereum/ethereum/tests/forking/block.test.ts
+++ b/src/chains/ethereum/ethereum/tests/forking/block.test.ts
@@ -1,5 +1,6 @@
 import assert from "assert";
 import getProvider from "../helpers/getProvider";
+import skipIfNoInfuraKey from "../helpers/skipIfNoInfuraKey";
 import { EthereumProvider } from "../../src/provider";
 import request from "superagent";
 
@@ -11,10 +12,10 @@ describe("forking", function () {
     const blockNumHex = `0x${blockNumber.toString(16)}`;
     const URL = "https://mainnet.infura.io/v3/" + process.env.INFURA_KEY;
     let provider: EthereumProvider;
-    before(async function () {
-      if (!process.env.INFURA_KEY) {
-        this.skip();
-      }
+
+    skipIfNoInfuraKey();
+
+    before(async () => {
       provider = await getProvider({
         fork: {
           url: URL,
@@ -52,7 +53,10 @@ describe("forking", function () {
         "earliest",
         true
       ]);
-      assert.deepStrictEqual(parseInt(block.number), parseInt(remoteBlock.number));
+      assert.deepStrictEqual(
+        parseInt(block.number),
+        parseInt(remoteBlock.number)
+      );
       assert.deepStrictEqual(block.hash, remoteBlock.hash);
     });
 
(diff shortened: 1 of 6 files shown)


src/chains/ethereum/ethereum/tests/forking/block.test.ts
```diff
@@ -86,10 +90,18 @@ describe("forking", function () {
...
+        true
+      ]);
+      const blockTransactionCountByHash = await provider.send(
+        "eth_getBlockTransactionCountByHash",
+        [block.hash]
+      );
+      assert.deepStrictEqual(
+        block.transactions.length,
+        parseInt(blockTransactionCountByHash)
+      );
```
jeffsmale90, line 104: No change here, just reformatting.

src/packages/core/tests/server.test.ts
```diff
@@ -193,7 +193,9 @@ describe("server", () => {
       // skip this test unless in CI, as this test iterates over all available network interfaces
       // and network interfaces on user machines are unpredictible and may behave in ways that
       // we don't care about.
-      if (process.env.CI) this.skip();
+      if (process.env.CI !== "true") {
+        this.skip();
+      }
```
jeffsmale90: Not strictly related to the topic of this PR, but seemed wrong to me.

src/packages/core/tests/server.test.ts
```diff
@@ -116,11 +116,7 @@ describe("server", () => {
     return deferred;
   }
 
-  // skip this test unless in GitHub Actions, as this test iterates over
-  // all available network interfaces and network interfaces on user
-  // machines are unpredictible and may behave in ways that we don't care
-  // about.
-  (process.env.GITHUB_ACTION ? describe : describe.skip)("listen", function () {
+  describe("listen", function () {
```
jeffsmale90, line 119: This was a little confusing because there's a skip at the `describe` level as well as the test level, so I refactored to be just at the test level (even though we end up with a `describe` with no tests that run in a development environment).

src/packages/core/tests/server.test.ts
```diff
@@ -190,10 +192,12 @@ describe("server", () => {
...
-      // skip this test unless in CI, as this test iterates over all available network interfaces
-      // and network interfaces on user machines are unpredictible and may behave in ways that
-      // we don't care about.
-      if (process.env.CI) this.skip();
+      // This test iterates over all available network interfaces. This can be problematic on user
+      // machines (which may be configured in unsupported ways), and also in macOS (which exposes
+      // unsupported interfaces). In these cases, we skip this test.
+      if (process.env.CI !== "true" || process.platform === "darwin") {
+        this.skip();
+      }
```
jeffsmale90, line 200: I'm keen to hear your thoughts on this, but it appears to be problematic only on macOS, so I've allowed it to run on linux / windows in CI only.

src/chains/ethereum/ethereum/tests/helpers/skipIfNoInfuraKey.ts
```diff
@@ -0,0 +1,17 @@
+function skipIfNoInfuraKey() {
+  before("skip if no INFURA_KEY (unless in CI)", function () {
+    // If there is no INFURA_KEY provided, the test should be skipped. Unless running
+    // in CI, where there should _always_ be a key provided.
+    if (!process.env.INFURA_KEY) {
+      if (process.env.CI === "true") {
+        throw new Error(
+          "No INFURA_KEY was provided. When CI is true, an INFURA_KEY must be provided."
```
davidmurdoch:
```suggestion
          `No INFURA_KEY environment variable was provided. When process.env.CI is "true", an INFURA_KEY must be provided.`
```

src/packages/core/tests/server.test.ts
davidmurdoch: 👍

davidmurdoch (changes requested):
just one small nit

davidmurdoch (approved):

PR #4357, opened 2023-04-13
chore: use `localhost` as hostname for `docs.preview` dev command

davidmurdoch:
This is just to make ganache development from WSL easier.

### DIFF
--- a/src/chains/ethereum/ethereum/package.json
+++ b/src/chains/ethereum/ethereum/package.json
@@ -27,7 +27,7 @@
   "scripts": {
     "docs.build": "npx shx rm -rf ./lib/docs ./lib/api.json && npm run docs.typedoc",
     "docs.typedoc": "typedoc --options ./typedoc.json --readme ./README.md --out ../../../../docs/typedoc --json ../../../../docs/typedoc/api.json src/api.ts",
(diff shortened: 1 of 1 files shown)


src/chains/ethereum/ethereum/package.json
```diff
@@ -27,7 +27,7 @@
   "scripts": {
     "docs.build": "npx shx rm -rf ./lib/docs ./lib/api.json && npm run docs.typedoc",
     "docs.typedoc": "typedoc --options ./typedoc.json --readme ./README.md --out ../../../../docs/typedoc --json ../../../../docs/typedoc/api.json src/api.ts",
-    "docs.preview": "ws --open --port 3010 --directory ../../../../docs",
+    "docs.preview": "ws --open --hostname localhost --port 3010 --directory ../../../../docs",
```
jeffsmale90, line 30: Fair enough - I don't use WSL, but trust that this does help :)

jeffsmale90 (approved):

src/chains/ethereum/ethereum/package.json
davidmurdoch: Haha. 127.0.0 doesn't make it to wsl from windows by default, but localhost does.

PR #4362, opened 2023-04-15
feat: add `--flavor` feature (ganache chain plugins)

davidmurdoch:
Refactor ganache to allow for external plugins, aka "flavors".

Ganache Flavors are plugins that can be used to launch test networks for chains
other than Ganache's built-in Ethereum networks. They are loaded at runtime via
Ganache's `--flavor` flag.

We first shipped this as an internal feature in Q2 2021 with the `filecoin` flavor. We are now externalizing this feature by providing a way for the community to build their own ganache flavors.

This first external implementation is experimental. We'd like to collect feedback from users and get an idea for what to add or change to the new `@ganache/flavor` package. If you don't like something about this interface or feel like it's missing things (it is!) please [open a new issue](https://github.com/trufflesuite/ganache/issues) and let us know about it. We'd be happy to chat about this or even hop on a short call to discuss what you are looking for.

What this is _not_: this is not a way of extending our Ethereum implementation. While we'd love to make our default Ethereum flavor even more flexible and configurable, this is not what this change does.

---

To create a new flavor, you must create a new package that exports your "Flavor". A
flavor is a JavaScript object that implements our Flavor interface.

Here is an example of a "Hello World"-style flavor in TypeScript:

```typescript
import { Flavor, Connector, CliSettings } from "@ganache/flavor";

export type Provider = { sayHi: (name: string) => string };
export type RequestPayload = { name: string };
export type ResponsePayload = { result: string };

const provider: Provider = {
  sayHi(name: string) {
    return `Hello, ${name}`;
  }
};

const helloConnector: Connector<Provider, RequestPayload, ResponsePayload> = {
  // the `provider` is yours to implement however you'd like. In programmatic
  // usage it will be returned to the user when they call
  // `Ganache.provider()` or `Ganache.server().provider`
  provider,

  async connect(): Promise<void> {
    // ganache will `await` the return of your `connect` method before
    // forwarding any requests to your connector.
    // if your connector doesn't need to do any async work to initialize you
    // can leave this empty.
  },

  parse(message: Buffer) {
    // the `message` doesn't have to be JSON, you can use any data interchange
    // format you'd like.
    // NOTE: If your connector's `parse` method throws, Ganache does
    // _not_ call your `formatError` and instead returns a `400 Bad Request`
    // response.
    return JSON.parse(message);
  },

  async handle(
    this: typeof helloConnector,
    payload: RequestPayload
  ): Promise<{ value: unknown }> {
    if (!payload || typeof payload.name !== "string") {
      // you can throw an Exception here and ganache will catch the error and
      // pass it to your `formatError` function.
      throw new Error("payload must have a `name` property of type `string`");
    }

    // in this "Hello World" example we only have one method, but you can
    // implement as many as you need in any way you want. Here we just call the
    // our provider's `sayHi` with the user provided `payload`'s `name`
    //  property.
    const value = this.provider.sayHi(payload.name);

    // Your `handle` function MUST always return any object with a `value`
    // property. The value of `value` MAY itself be a `Promise`. The _resolved_
    // `value` will be passed to your connector's `format` function.
    //
    // NOTE: if a client is connected via WebSockets, and the the `value` is a
    // `PromiEvent` (a `Promise` that also has an `on` function) that emits a
    // `"message"` event Ganache will subscribe to the `"message"` event. If
    // the `PromiEvent` then emits a "message" its event data will be sent to
    // the client.
    // ATTENTION: A flavor cannot `format` this message though; it will always
    // be sent as am Ethereum JSON-RPC 2.0 subscription style message. This
    // behavior will change in the future *without notice*. It is not recommened
    // that you use the PromiEvent feature at this time.
    return { value };
  },

  format(result: ResponsePayload, payload: RequestPayload) {
    console.log(
      `formatting result (${result.result}) for payload (${payload.name})`
    );
    // You don't have to return a string here, you can also return a Buffer
    // and the serialization doesn't have to be JSON. However...
    // ATTENTION: ganache flavors don't yet support changing the `content-type`
    // header so it always returns `content-type: application/json`. This will
    // change in the future and flavors will be able to specify their own
    // content-type and other HTTP headers.
    //
    // You can also return a Generator (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator)
    // which will cause ganache to send the response in chunks (one chunk for
    // yield). This is useful if your response data is too large for Node.js to
    // handle (string and Buffer sizes are limited to about 1-2GB, and memory is
    // capped to 4GB on most systems).
    return JSON.stringify({ result });
  },

  formatError(error: Error, payload: RequestPayload) {
    // Ganache calls your `formatError` if any errors occured while processing
    // a request.

    console.log(`formatting error (${error.message}) for payload (${payload})`);
    // you don't have to return a string here, you can also return a Buffer
    // and the serialization doesn't have to be JSON
    return JSON.stringify({ error: error.message });
  },

  async close() {
    // ganache calls your connector's `close` function when shutting down.
    // This is where you'd perform clean up, like closing database connections
    // or cleaning temporary files.
  }
};

type HelloFlavor = Flavor<"hello-chain", typeof helloConnector>;

const HelloFlavor: HelloFlavor = {
  flavor: "hello-chain",
  options: {
    // see the `example/` directory for how Options work
  },
  connect(providerOptions: never) {
    return helloConnector;
  },
  // the `ready` function is required for your flavor to work with ganache on
  // the CLI
  ready: ({
    provider,
    options
  }: {
    provider: Provider;
    options: { server: CliSettings };
  }) => {
    // this function is only called after ganache has fully initialized, and is
    // only called when used via ganache cli (it is not used when your flavor is
    // used programatically)
    console.log(
      `Hello Chain server is running at http://${options.server.host}:${options.server.port}`
    );
  }
};

/**
 * Your flavor needs to be exported as `default` so Ganache can find it.
 */
export default HelloFlavor;
```

Check out the [example implementation](https://github.com/trufflesuite/ganache/tree/flavor2/src/packages/flavor/example) for a more-in-depth example.


src/chains/ethereum/ethereum/src/forking/handlers/http-handler.ts
```diff
@@ -112,6 +112,7 @@ export class HttpHandler extends BaseHandler implements Handler {
       const postData = `${JSONRPC_PREFIX}${this.id++},${key.slice(1)}`;
       this.headers["content-length"] = postData.length;
 
+      // @ts-ignore
       const req = this._request(requestOptions);
```
davidmurdoch, line 116: I think node's types updated and it doesn't like the polyfill we use for the `abortSignal` anymore.

src/chains/filecoin/filecoin/src/connector.ts
```diff
@@ -1,56 +1,48 @@
...
   #provider: FilecoinProvider;
 
-  get provider() {
+  get provider(): FilecoinProvider {
     return this.#provider;
   }
 
   constructor(
-    providerOptions: FilecoinProviderOptions = {},
+    providerOptions: FilecoinProviderOptions | null = {},
```
davidmurdoch:
```suggestion
    providerOptions: FilecoinProviderOptions = {},
```

src/chains/ethereum/ethereum/index.ts
```diff
@@ -5,5 +5,170 @@
...
+    provider: EthereumOptionsConfig,
+    server: ServerOptionsConfig,
+    cli: CliOptionsConfig
+  },
+  ready
+};
+// flavors are exported as a default export
+export default EthereumFlavor;
+
+function ready(provider: EthereumProvider, cliSettings: CliSettings) {
```
davidmurdoch: this was originally in the cli package  (was Ethereum's `initialized` function)

package.json
```diff
@@ -39,7 +40,7 @@
     "husky": "5.2.0",
     "into-stream": "6.0.0",
     "lerna": "4.0.0",
-    "marked": "2.1.3",
+    "marked": "4.3.0",
```
davidmurdoch, line 43: aligning the versions of `marked` we are using elsewhere.

src/chains/ethereum/console.log/package.json
```diff
@@ -66,7 +66,6 @@
     "typescript": "4.7.4"
   },
   "dependencies": {
-    "@ethereumjs/util": "8.0.5",
-    "@ganache/utils": "0.3.0"
+    "@ganache/utils": "0.8.0"
```
davidmurdoch: `@ethereumjs/util` wasn't used here, and `@ganache/utils` was way outdated!

src/chains/ethereum/ethereum/package.json
```diff
@@ -52,28 +52,29 @@
   ],
   "dependencies": {
     "@ethereumjs/common": "3.1.1",
+    "@ethereumjs/statemanager": "1.0.4",
```
davidmurdoch, line 56: I set off looking into making the `ethereum` package itself be able to "ejected" from the monorepo, as that'd enable flavor authors to copy+paste the ethereum package. I don't think it's there just yet, but the changes in this file are aimed at removing unused dependencies and adding the ones we actually are using.

src/chains/ethereum/ethereum/patches/@ethereumjs+block+4.2.1.patch
```diff
@@ -1,20 +1,147 @@
+diff --git a/node_modules/@ethereumjs/block/dist/block.d.ts b/node_modules/@ethereumjs/block/dist/block.d.ts
+index 6fd773b..bbfc908 100644
+--- a/node_modules/@ethereumjs/block/dist/block.d.ts
```
davidmurdoch, line 3: a previous version of this patch only removed the ethers type, this just removes it all together.

src/chains/ethereum/ethereum/src/@types/ethereumjs-block/header.d.ts
```diff
@@ -1,62 +0,0 @@
-declare module "ethereumjs-block/header" {
```
davidmurdoch, line 1: super old types we haven't needed in a long time

src/chains/ethereum/ethereum/src/blockchain.ts
```diff
@@ -58,7 +58,7 @@ import {
   CallOverrides
 } from "./helpers/run-call";
 import { ForkStateManager } from "./forking/state-manager";
-import { DefaultStateManager } from "@ethereumjs/statemanager";
+import type { DefaultStateManager } from "@ethereumjs/statemanager";
```
davidmurdoch, line 62: In the process of understanding what to include as a dependency vs devDependency (type-only imports should be in devDependencies) I marked some type-only imports. Figured it's worth leaving these in (note: @ethereumjs/statemanager didn't end up being a type-only import).

src/chains/ethereum/ethereum/src/provider.ts
```diff
@@ -124,20 +123,17 @@ type RequestParams<Method extends RequestMethods> = {
...
-    data: DataEvent;
-    error: Error;
-    "ganache:vm:tx:step": VmStepEvent;
-    "ganache:vm:tx:before": VmBeforeTransactionEvent;
-    "ganache:vm:tx:after": VmAfterTransactionEvent;
-    "ganache:vm:tx:console.log": VmConsoleLogEvent;
-    connect: undefined;
-    disconnect: undefined;
-  }>
-  implements Provider<EthereumApi>
```
davidmurdoch, line 138: `implements Provider<EthereumApi>` was basically `any`, so was pointless.

src/chains/ethereum/ethereum/tests/api/eth/signTypedData.test.ts
```diff
@@ -1,5 +1,5 @@
 import assert from "assert";
-const createKeccakHash = require("keccak");
+import { Data, keccak } from "@ganache/utils";
```
davidmurdoch, line 2: more dependency cleanup

src/chains/ethereum/options/src/index.ts
```diff
@@ -45,17 +43,16 @@ export type EthereumLegacyProviderOptions = Partial<
     MakeLegacyOptions<ForkConfig>
 >;
 
-export type EthereumProviderOptions = Partial<
-  {
-    [K in keyof EthereumConfig]: ExternalConfig<EthereumConfig[K]>;
-  }
->;
+export type EthereumProviderOptions = Partial<{
```
davidmurdoch: no changes, just prettier doing it's thing

src/chains/filecoin/filecoin/index.ts
```diff
@@ -5,19 +5,35 @@
...
+type FilecoinFlavor = Flavor<
+  "filecoin",
+  Connector,
+  {
+    provider: FilecoinOptionsConfig;
+    server: ServerOptionsConfig;
+    cli: CliOptionsConfig;
+  }
+>;
+const FilecoinFlavor: FilecoinFlavor = {
```
davidmurdoch, line 28: this new interface is technically breaking, but it was already broken (it doesn't work with latest ganache anyway)

davidmurdoch: oh, when loading an old filecoin that doesn't adhere to the new interface ganache yells at the user telling them they need to update, so at least there is that now :-)

src/chains/filecoin/filecoin/package.json
```diff
@@ -50,16 +51,17 @@
     "ipfs": "0.52.3",
     "ipfs-http-server": "0.1.4",
     "ipld-dag-cbor": "0.17.0",
+    "keccak": "3.0.2",
     "leveldown": "6.1.0",
-    "secp256k1": "4.0.3"
+    "secp256k1": "4.0.3",
+    "@trufflesuite/bigint-buffer": "1.1.10"
```
davidmurdoch, line 57: filecoin is webpacked when distributed, so while `keccak` and `bigint-buffer` aren't used directly by filecoin, they _are_ marked as "external" in the webpack (because they have native components that might be compiled at install-time).

src/chains/filecoin/filecoin/package.json
```diff
@@ -50,16 +51,17 @@
...
   },
   "devDependencies": {
-    "@filecoin-shipyard/lotus-client-provider-browser": "0.0.14",
     "@filecoin-shipyard/lotus-client-rpc": "0.2.0",
     "@filecoin-shipyard/lotus-client-schema": "2.0.0",
     "@ganache/filecoin-options": "0.8.0",
+    "@ganache/flavor": "0.9.0",
```
davidmurdoch: the rest of these changes are intended to make it possible to move filecoin out of the ganache monorepo. that work will happen after this PR is merged.

src/chains/filecoin/filecoin/src/connector.ts
```diff
@@ -1,56 +1,48 @@
...
   #provider: FilecoinProvider;
 
-  get provider() {
+  get provider(): FilecoinProvider {
     return this.#provider;
   }
 
   constructor(
-    providerOptions: FilecoinProviderOptions = {},
+    providerOptions: FilecoinProviderOptions | null = null,
```
davidmurdoch, line 38: I can't remember why this changed, however, filecoin types don't work anyway, as it doesn't use api-extractor to roll up the types.

src/chains/filecoin/filecoin/src/connector.ts
```diff
@@ -1,56 +1,48 @@
...
-    super();
-
     this.#provider = new FilecoinProvider(providerOptions, executor);
   }
 
   async connect() {
     await this.#provider.initialize();
-    // no need to wait for #provider.once("connect") as the initialize()
-    // promise has already accounted for that after the promise is resolved
-    await this.emit("ready");
```
davidmurdoch, line 53: nothing listened to this event, so I just removed it.

src/chains/filecoin/filecoin/tests/api/filecoin/deals.test.ts
```diff
@@ -81,7 +81,7 @@ describe("api", () => {
       it("should accept a new deal", async () => {
         const miners = await client.stateListMiners();
         const accounts =
-          await provider.blockchain.accountManager.getControllableAccounts();
+          await provider.blockchain.accountManager?.getControllableAccounts()!;
```
davidmurdoch, line 84: A bunch of tests didn't type check, so I went through and fixed them.

src/chains/filecoin/filecoin/webpack/webpack.node.config.ts
```diff
@@ -13,22 +13,10 @@ const config: webpack.Configuration = merge({}, base, {
...
-        __dirname,
-        "../",
-        "dist",
-        "node",
-        "ganache.min.js"
-      )
-    }
-  ],
-  plugins: [
-    new webpack.BannerPlugin({ banner: "#!/usr/bin/env node", raw: true })
```
davidmurdoch, line 31: `@ganache/filecoin` was never run directly on the cli (it needs `ganache`), so this wasn't needed.

src/chains/filecoin/filecoin/webpack/webpack.node.config.ts
```diff
@@ -13,22 +13,10 @@ const config: webpack.Configuration = merge({}, base, {
     "ipfs-http-client",
     "ipfs-http-server",
     "ipld-dag-cbor",
-    "bigint-buffer",
+    "@trufflesuite/bigint-buffer",
     "leveldown",
     "secp256k1",
-    "keccak",
-    {
-      "@ganache/core": path.resolve(
```
davidmurdoch, line 21: no clue why this was here.

src/chains/filecoin/filecoin/webpack/webpack.node.config.ts
```diff
@@ -13,22 +13,10 @@ const config: webpack.Configuration = merge({}, base, {
     "ipfs-http-client",
     "ipfs-http-server",
     "ipld-dag-cbor",
-    "bigint-buffer",
+    "@trufflesuite/bigint-buffer",
```
davidmurdoch, line 16: filecoin starts up complaining about bigint-buffer not loading, this fixes that!

src/packages/cli/package.json
```diff
@@ -53,25 +53,27 @@
...
-    "@ganache/utils": "0.8.0",
     "@trufflesuite/ps-list": "0.0.3",
     "@types/node": "17.0.0",
     "chalk": "4.1.0",
     "cli-table": "0.3.11",
-    "marked-terminal": "4.1.0"
+    "env-paths": "2.2.1",
+    "marked": "4.3.0",
+    "marked-terminal": "4.2.0",
+    "yargs": "16.2.0"
```
davidmurdoch: While I was removing @gaanche/filecoin from here I fixed a bunch of dependency mismatches.

src/packages/cli/src/args.ts
```diff
@@ -1,161 +1,26 @@
...
-} from "@ganache/options";
-import { FlavorCommand, StartArgs, GanacheArgs } from "./types";
+import yargs from "yargs";
+import { StartArgs, GanacheArgs } from "./types";
 import chalk from "chalk";
 import { EOL } from "os";
-import marked from "marked";
-import TerminalRenderer from "marked-terminal";
-import { _DefaultServerOptions } from "@ganache/core";
+import { configureFlavorOptions, loadFlavorFromArgs } from "./flavors";
```
davidmurdoch: Most of what is removed from this file was moved to these two functions.

src/packages/cli/src/args.ts
```diff
@@ -311,24 +138,18 @@ export default function (
...
-
+    const flavor = (parsedArgs.flavor || "ethereum") as string | "ethereum";
     finalArgs = {
       flavor,
       action,
-      ...(expandArgs(parsedArgs) as Omit<
-        StartArgs<FlavorName>,
-        "flavor" | "action"
-      >)
+      ...(expandArgs(parsedArgs) as Omit<StartArgs<any>, "flavor" | "action">)
```
davidmurdoch, line 114: `StartArgs<any>` because `FlavorName` isn't a thing anymore (ganache cli doesn't know about "all" possible flavors like it did before)

src/packages/cli/src/cli.ts
```diff
@@ -46,17 +43,30 @@ const coreVersion = process.env.CORE_VERSION || "DEV";
...
+  let flavor: AnyFlavor;
+  if (flavorName === "ethereum") {
+    flavor = require("@ganache/ethereum").default;
+  } else {
+    flavor = load(flavorName);
+
+    // if the flavor handler has a cli section, merge it with our cli settings
+    // as it might have a section for overriding the port and the host
+    if (flavor.options.cli) {
+      Object.assign(cliSettings, flavor.options.cli.normalize(argv));
```
davidmurdoch, line 62: For flavors other than `"ethereum"` we parse our cli args using our ganache defaults, and then parse the args using the `flavor`'s cli options parser, if provided, and then merge them together.

src/packages/utils/package.json
```diff
@@ -47,6 +47,7 @@
     "tooling"
   ],
   "dependencies": {
+    "@trufflesuite/bigint-buffer": "1.1.10",
```
davidmurdoch, line 50: its just not optional. /shrug

src/packages/ganache/webpack/webpack.browser.config.ts
```diff
@@ -26,15 +26,13 @@ const config: webpack.Configuration = merge({}, base, {
     },
     alias: {
       "tmp-promise": require.resolve("./polyfills/browser-tmp-promise"),
-      "bigint-buffer": require.resolve("./polyfills/browser-bigint-buffer"),
```
davidmurdoch, line 29: we don't use `bigint-buffer`, so this doesn't do anything

src/packages/ganache/webpack/webpack.browser.config.ts
```diff
@@ -26,15 +26,13 @@ const config: webpack.Configuration = merge({}, base, {
...
       crypto: require.resolve("./polyfills/browser-crypto"),
       // replace leveldown with a browser version
       leveldown: require.resolve("level-js/"),
       // browser version can't start a server, so just remove the websocket server since it can't work anyway
       "@trufflesuite/uws-js-unofficial": false,
       // replace URL with a browser version -- sorta. just look at the polyfill code
       url: require.resolve("./polyfills/url"),
-      "@ganache/filecoin": false,
```
davidmurdoch, line 37: this is the purpose of this PR

src/packages/cli/src/flavors.ts
```diff
@@ -0,0 +1,84 @@
...
+        // load flavor plugin:
+        return load(flavor === "filecoin" ? "@ganache/filecoin" : flavor);
+      }
+    } else {
+      throw new Error("No flavor specified after `--flavor` flag.");
+    }
+  }
+
+  // fallback to ethereum
+  return require("@ganache/ethereum").default;
```
davidmurdoch, line 29: We lazy load the `etheruem` flavor to reduce start up time.

In actuality I'm not sure this helps because our webpack process probably loads everything at once anyway, but doing this now should make it easier to optimize in the future.

src/packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,129 @@
+import { marked } from "marked";
```
davidmurdoch, line 1: this file was is most just moved from args.ts

src/packages/cli/src/types.ts
```diff
@@ -1,28 +1,29 @@
...
 type AbstractArgs<TAction = Action> = {
   action: TAction;
 };
 
-export type StartArgs<TFlavorName extends FlavorName> =
-  ServerOptions<TFlavorName> & {
-    _: [TFlavorName];
-    server: CliServerOptions;
-  } & AbstractArgs<"start" | "start-detached">;
+export type StartArgs<
```
davidmurdoch, line 11: In actual use StartArgs always receives `any` as the `TFlavorName`, so I'm not sure this is useful (but could in the future if we added unit-level tests that utilized these types)

src/packages/cli/src/types.ts
```diff
@@ -1,28 +1,29 @@
...
+  flavor: TFlavorName;
+} & AbstractArgs<"start" | "start-detached">;
 
 export type GanacheArgs =
   | (AbstractArgs<"stop"> & { name: string })
   | AbstractArgs<"list">
-  | StartArgs<FlavorName>;
+  | StartArgs<"ethereum" | string>;
 
-export type CliSettings = CliServerOptions;
```
davidmurdoch, line 26: I moved this to the @ganache/flavoor package.

src/packages/core/index.ts
```diff
@@ -43,9 +43,9 @@ const Ganache = {
    * @returns A provider instance for the flavor
    * `options.flavor` which defaults to `ethereum`.
    */
-  server: <T extends FlavorName = typeof DefaultFlavor>(
-    options?: ServerOptions<T>
-  ) => new Server(options),
+  server: <F extends AnyFlavor = EthereumFlavor>(
```
davidmurdoch, line 46: Notable change: the type here is an entire `Flavor`, not just the name of the flavor.

src/packages/core/src/connector-loader.ts
```diff
@@ -1,25 +1,51 @@
...
+function getConnector<F extends AnyFlavor>(
+  flavorName: F["flavor"],
+  providerOptions: Parameters<F["connect"]>[0],
+  executor: Executor
+): ReturnType<F["connect"]> {
+  if (flavorName === "ethereum") {
+    // lazy load the ethereum flavor to avoid start up overhead for
+    // other flavors.
+    return <ReturnType<F["connect"]>>(
+      require("@ganache/ethereum").default.connect(providerOptions, executor)
```
davidmurdoch: more lazy loading.

src/packages/flavor/src/connector.ts
```diff
@@ -39,19 +43,19 @@ export interface Connector<
     | ((
         payload: RequestFormat,
         connection: HttpRequest
-      ) => Promise<{ value: ReturnType<Api[KnownKeys<Api>]> }>)
+      ) => Promise<{ value: unknown }>)
```
davidmurdoch, line 37: The `ReturnType<Api[KnownKeys<Api>]>` type wasn't useful since the `Api` could be anything.

src/chains/ethereum/console.log/tests/index.test.ts
```diff
@@ -30,6 +30,7 @@ describe("@ganache/console.log", () => {
   let from: string;
 
   before("set up a ganache provider", function () {
+    this.timeout(10000);
```
davidmurdoch, line 33: Because Ethereum ganache is lazy loaded, the first time we run it it might take a while, so I added this timeout. Previously it would take the same about of time, but it was just doing it before the tests started running.

src/packages/core/src/types.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export type ProviderOptions<F extends AnyFlavor> =
+  (F["flavor"] extends "ethereum"
+    ? {
+        flavor?: F["flavor"];
+      }
+    : {
+        flavor: F["flavor"];
+      }) &
+    Parameters<F["options"]["provider"]["normalize"]>[0];
```
davidmurdoch, line 17: The flavor must provide us with an @ganache/options interface, and so we dig into it's normalization method to grab the options the flavor expects from the user before normalization.

davidmurdoch (commented):
i'm so sorry

davidmurdoch (commented):
i'm so sorry

src/chains/ethereum/ethereum/src/forking/handlers/http-handler.ts
davidmurdoch: removed this change in https://github.com/trufflesuite/ganache/pull/4459

src/chains/ethereum/console.log/tests/index.test.ts
jeffsmale90: Is it worth adding this as a comment in code?

src/chains/filecoin/filecoin/README.md
```diff
@@ -18,21 +18,21 @@ To use Filecoin-flavored Ganache via the CLI, follow the below instructions:
    npm uninstall --global ganache-cli
    npm uninstall --global ganache
    ```
-1. Install the `ganache` package globally with the `filecoin` tag (note that we're **not** installing the old `ganache-cli` package)
+1. Install the `ganache` package globally
    ```bash
-   npm install --global ganache@filecoin
+   npm install --global ganache
    ```
 1. Install the `@ganache/filecoin` globally
```
jeffsmale90, line 25:
```suggestion
1. Install the `@ganache/filecoin` package globally
```

src/chains/filecoin/filecoin/README.md
```diff
@@ -18,21 +18,21 @@ To use Filecoin-flavored Ganache via the CLI, follow the below instructions:
...
    ```
 1. Run Filecoin-flavored Ganache
    ```bash
-   ganache filecoin
+   ganache --flavor @ganache/filecoin
    ```
 1. See available [options](#startup-options)
    ```bash
-   ganache filecoin --help
+   ganache --flavor @ganache/filecoin --help
```
jeffsmale90, line 35: Do we still support `ganache filecoin`?? Should we (undocumented)?

src/packages/flavor/example/src/flavor.ts
```diff
@@ -0,0 +1,74 @@
+import type { CliSettings, Flavor } from "../../";
```
jeffsmale90, line 1: can we import from `"@ganache/flavor"` instead so that it's clear what is imported from that module, and what is just local to the example?

src/packages/flavor/example/src/flavor.ts
```diff
@@ -0,0 +1,74 @@
...
+import type { Provider } from "./provider";
+import { NotABlockchainChainConnector } from "./connector";
+import {
+  NotABlockchainChainProviderOptionsConfig,
+  NotABlockchainChainProviderOptions
```
jeffsmale90, line 6: Controversial opinion:

Although I love the name `NotABlockchainChain` - peppering it through the variable names makes it super hard to follow.

I feel that using something more concise like `DemoFlavor`, `DemoConnector` etc (or [less clear] `NabcFlavor`, `NabcConnector`) would make it easier to read and understand.

src/packages/flavor/example/README.md
```diff
@@ -0,0 +1,90 @@
...
+// `provider` is the NotABlockchainChain flavor's `provider`.
+const provider = Ganache.provider<NotABlockchainChain>({
+  // note: flavors can be referenced by path or npm package name, but since
+  // we are running directly from this folder without publishing to npm we must
+  // use the path (`__dirname`) and cast to make TypeScript happy.
+  flavor: __dirname as "not-a-blockchain-chain",
+  // provider options are defined by the flavor
+  wallet: {
+    // our blockchain is silly and let's you create accounts out of any string
+    // you'd like. This accounts will be funded on start up.
```
jeffsmale90, line 40:
```suggestion
    // you'd like. These accounts will be funded on start up.
```

src/chains/ethereum/ethereum/index.ts
```diff
@@ -5,5 +5,177 @@
  * @license MIT
```
jeffsmale90, line 5: Not related to this line, but:

- `--port` shorthand doesn't work
- `--host` shorthand doesn't work
- `instances list` shows `undefined` for the host (no `host` property is persisted to the instance file) when no host is explicitly provided

src/packages/flavor/example/README.md
```diff
@@ -0,0 +1,90 @@
+# Not-a-Blockchain-Chain
```
jeffsmale90, line 1: This project didn't build out of the box for me - it wouldn't install the local reference to `ganache` in `package.json`.

Additionally, the imports to ganache packages are local, which is not immediately obvious to a newcomer - if someone wanted to use this as a starting point for their _actual_ flavor, they'd have to change these to `@ganache/...` imports.

I feel it would be beneficial to move the demo flavor to it's own repo, so that it works out of the box exactly as a real flavor plugin would.

src/packages/flavor/src/connector.ts
```diff
@@ -1,22 +1,17 @@
-import { Provider } from "./provider";
+import { Api, Executor, KnownKeys } from "@ganache/utils";
```
jeffsmale90, line 1:
```suggestion
import { Api, KnownKeys } from "@ganache/utils";
```

src/packages/flavor/src/flavor.ts
```diff
@@ -0,0 +1,57 @@
+import type { OptionsConfig } from "@ganache/options";
+import { Executor } from "..";
+import { Connector } from "./connector";
+
+export type CliSettings = { host: string; port: number };
```
jeffsmale90, line 5: shouldn't this type actually be `ServerSettings`?

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
...
+to understand the needs of other chains and how to extend the Ganache flavor
+model to support transports other than JSON-RPC 2.0 over HTTP/WS.
+
+An example of shortcomings for non JSON-RPC chains:
+
+- ganache only accepts POST requests over a single configurable path (defaults
+  to `/`). If you have a need for something else, like JSON over REST, gRPC, SOAP,
+  etc, please open an issue describing your use case.
+- websocket subscription-based messaging is very limited and only supports
+  responsing like Ethereum JSON-RPC does.
```
jeffsmale90, line 24: `responsing` is this intentional? Regardless, I kinda like it, and will be using it more going forward.

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
+# `@ganache/flavor`
+
+Ganache's flavor TypeScript types, helpers, and utility functions.
+
+Ganache Flavors are plugins that can be used to launch test networks for chains
```
jeffsmale90, line 5: Is "test network" the best term here?

I'd think either "development network" or a rewording to something like "... used to launch Ganache for chains ..."

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
...
+    if (!payload || typeof payload.name !== "string") {
+      // you can throw an Exception here and ganache will catch the error and
+      // pass it to your `formatError` function.
+      throw new Error("payload must have a `name` property of type `string`");
+    }
+
+    // in this "Hello World" example we only have one method, but you can
+    // implement as many as you need in any way you want. Here we just call the
+    // our provider's `sayHi` with the user provided `payload`'s `name`
+    //  property.
```
jeffsmale90, line 81:
```suggestion
    // implement as many as you need in any way you want. Here we just call our
    // provider's `sayHi` with the user provided `payload`'s `name` property.
```

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
...
+    const value = this.provider.sayHi(payload.name);
+
+    // Your `handle` function MUST always return any object with a `value`
+    // property. The value of `value` MAY itself be a `Promise`. The _resolved_
+    // `value` will be passed to your connector's `format` function.
+    //
+    // NOTE: if a client is connected via WebSockets, and the the `value` is a
```
jeffsmale90, line 88:
```suggestion
    // NOTE: if a client is connected via WebSockets, and the `value` is a
```

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
...
+    // `PromiEvent` (a `Promise` that also has an `on` function) that emits a
+    // `"message"` event Ganache will subscribe to the `"message"` event. If
+    // the `PromiEvent` then emits a "message" its event data will be sent to
+    // the client.
+    // ATTENTION: A flavor cannot `format` this message though; it will always
+    // be sent as am Ethereum JSON-RPC 2.0 subscription style message. This
```
jeffsmale90, line 94:
```suggestion
    // be sent as an Ethereum JSON-RPC 2.0 subscription style message. This
```

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
...
+    // behavior will change in the future *without notice*. It is not recommened
```
jeffsmale90, line 95:
```suggestion
    // behavior will change in the future *without notice*. It is not recommended
```

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
...
+    );
+    // You don't have to return a string here, you can also return a Buffer
+    // and the serialization doesn't have to be JSON. However...
+    // ATTENTION: ganache flavors don't yet support changing the `content-type`
+    // header so it always returns `content-type: application/json`. This will
+    // change in the future and flavors will be able to specify their own
+    // content-type and other HTTP headers.
+    //
+    // You can also return a Generator (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator)
+    // which will cause ganache to send the response in chunks (one chunk for
```
jeffsmale90, line 112:
```suggestion
    // which will cause ganache to send the response in chunks (one chunk per
```

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
...
+    // yield). This is useful if your response data is too large for Node.js to
+    // handle (string and Buffer sizes are limited to about 1-2GB, and memory is
+    // capped to 4GB on most systems).
+    return JSON.stringify({ result });
+  },
+
+  formatError(error: Error, payload: RequestPayload) {
+    // Ganache calls your `formatError` if any errors occured while processing
```
jeffsmale90, line 120:
```suggestion
    // Ganache calls your `formatError` if an error is thrown while processing
```

I think this is a little more clear - happy to change my opinion though.

src/packages/flavor/README.md
```diff
@@ -0,0 +1,170 @@
...
+  async close() {
+    // ganache calls your connector's `close` function when shutting down.
+    // This is where you'd perform clean up, like closing database connections
+    // or cleaning temporary files.
+  }
+};
+
+type HelloFlavor = Flavor<"hello-chain", typeof helloConnector>;
+
+const HelloFlavor: HelloFlavor = {
```
jeffsmale90, line 138: I think it would be clearer if we distinguished between the HelloFlavor type and the HelloFlavor implementation... could get confusing.

src/packages/flavor/example/src/options.ts
```diff
@@ -0,0 +1,78 @@
...
+    defaultBalance: {
+      type: bigint;
+      rawType: string;
+      hasDefault: true;
+    };
+  };
+};
+
+function onlyUnique(value: string, index: number, array: string[]) {
+  return array.indexOf(value) === index;
```
jeffsmale90, line 25: pretty

src/packages/flavor/example/src/options.ts
```diff
@@ -0,0 +1,78 @@
...
+export type NotABlockchainChainDefaults =
+  Defaults<NotABlockchainChainDefinition>;
+export const NotABlockchainChainDefaults: NotABlockchainChainDefaults = {
+  wallet: WalletOptions
+};
+
+export type NotABlockchainChainProviderOptionsConfig =
+  OptionsConfig<NotABlockchainChainDefinition>;
+export const NotABlockchainChainProviderOptionsConfig: NotABlockchainChainProviderOptionsConfig =
+  new OptionsConfig(NotABlockchainChainDefaults);
```
jeffsmale90, line 66: I made a similar comment somewhere else about same name type / instance - it freaks me out a little bit.

src/packages/flavor/example/src/provider.ts
```diff
@@ -0,0 +1,108 @@
...
+    const genesisBlock: Map<string, Account> = new Map();
+    this.#options.wallet.accounts.forEach(address =>
+      genesisBlock.set(address, {
+        balance: this.#options.wallet.defaultBalance
+      })
+    );
+    this.#blockchain.set(0n, genesisBlock);
+    this.#latestBlock = 0n;
+  }
+  async send(method: string, params: any[]) {
```
jeffsmale90, line 54:
```suggestion

  async send(method: string, params: any[]) {
```

src/packages/flavor/example/src/provider.ts
```diff
@@ -0,0 +1,108 @@
...
+      }
+      case "sendFunds":
+        const [from, to, strAmount] = params;
+        const amount = BigInt(strAmount);
+        return this.#sendFunds(from, to, amount);
+      default:
+        throw new Error("Unsupported method " + method);
+    }
+  }
+  #getBalance(from: string, blockTag: string) {
```
jeffsmale90, line 70:
```suggestion

  #getBalance(from: string, blockTag: string) {
```

src/packages/flavor/example/src/provider.ts
```diff
@@ -0,0 +1,108 @@
...
+    let blockNumber: bigint =
+      blockTag === "latest" ? this.#latestBlock : BigInt(blockTag);
+    const block = this.#blockchain.get(blockNumber);
+    if (!block) {
+      throw new Error(`Couldn't fetch block "${blockTag}"`);
+    }
+
+    return `0x${(block.get(from)?.balance || 0n).toString(16)}`;
+  }
+  #sendFunds(from: string, to: string, amount: bigint) {
```
jeffsmale90, line 80:
```suggestion

  #sendFunds(from: string, to: string, amount: bigint) {
```

src/packages/flavor/src/load.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (
+      hasOwn(e, "message") &&
+      typeof e.message === "string" &&
+      e.message.includes(`Cannot find module '${flavor}'`)
+    ) {
+      const NEED_HELP = "Need help? Reach out to the Truffle community at";
+      const COMMUNITY_LINK = "https://trfl.io/support";
+
+      // we print and exit rather than throw to prevent webpack output from being
+      // spat out for the line number
```
jeffsmale90, line 41: it'd be nice if we could though

src/packages/flavor/src/load.ts
```diff
@@ -0,0 +1,57 @@
...
+      console.warn(
```
jeffsmale90, line 42: I'm a little uncomfortable with a message from ganache saying `npm install <arbitrary-package>`, as I can imagine a scenario where a new user blindly follows the instructions as it's from Ganache, and installing a package that they do not want.

jeffsmale90: Do we need a caveat to this?

> If you are sure "package-name" is the correct package, you might want to:
>   $ npm install -g package-name

Something like that but with better wording

src/packages/flavor/src/options/cli/cli-config.ts
```diff
@@ -0,0 +1,44 @@
...
+      legacy: {
+        /**
+         * @deprecated Use server.ws instead.
+         */
+        host: boolean;
+      };
+    };
+
+    /**
+     * Host for the server to bind to
```
jeffsmale90:
```suggestion
     * Run Ganache in detached (daemon) mode.
```

src/packages/flavor/src/options/cli/cli-config.ts
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @defaultValue true
```
jeffsmale90: Should this be false by default??

```suggestion
     * @defaultValue false
```

src/packages/flavor/src/options/cli/cli-config.ts
```diff
@@ -0,0 +1,44 @@
...
+    readonly port: {
+      type: number;
+      hasDefault: true;
+      legacy: {
+        /**
+         * @deprecated Use server.ws instead.
+         */
+        port: boolean;
+      };
+    };
```
jeffsmale90, line 17:
```suggestion
    readonly port: {
      type: number;
      hasDefault: true;
      legacy: {
        /**
         * @deprecated Use server.port instead.
         */
        port: number;
      };
    };
```

src/packages/flavor/src/options/cli/cli-config.ts
```diff
@@ -0,0 +1,44 @@
...
+    readonly host: {
+      type: string;
+      hasDefault: true;
+      legacy: {
+        /**
+         * @deprecated Use server.ws instead.
+         */
+        host: boolean;
+      };
+    };
```
jeffsmale90, line 33:
```suggestion
    readonly host: {
      type: string;
      hasDefault: true;
      legacy: {
        /**
         * @deprecated Use server.host instead.
         */
        host: string;
      };
    };
```

src/packages/flavor/src/options/server/server-config.ts
```diff
@@ -0,0 +1,54 @@
...
+      legacy: {
+        /**
+         * @deprecated Use server.ws instead.
+         */
+        ws: boolean;
+      };
+    };
+
+    /**
+     * Whether or not websockets should response with binary data (ArrayBuffers) or
```
jeffsmale90, line 20: respond?

src/packages/flavor/src/options/server/server-options.ts
```diff
@@ -0,0 +1,41 @@
...
+    normalize,
+    cliDescription: "Enable a websocket server.",
+    default: () => true,
+    legacyName: "ws",
+    cliType: "boolean"
+  },
+  wsBinary: {
+    normalize,
+    cliDescription:
+      "Whether or not websockets should response with binary data (ArrayBuffers) or strings.",
```
jeffsmale90, line 19: respond?

src/packages/flavor/src/options/server/server-options.ts
```diff
@@ -0,0 +1,41 @@
...
+    default: () => "auto",
+    cliChoices: ["true", "false", "auto"] as any[]
+  },
+  rpcEndpoint: {
+    normalize,
+    cliDescription:
+      "Defines the endpoint route the HTTP and WebSocket servers will listen on.",
```
jeffsmale90, line 26:
```suggestion
      "Defines the endpoint path the HTTP and WebSocket servers will listen on.",
```

src/chains/ethereum/ethereum/src/connector.ts
```diff
@@ -36,35 +33,28 @@ function isHttp(
...
+> implements IConnector<EthereumProvider, R | R[], JsonRpcResponse>
 {
   #provider: EthereumProvider;
 
   static BUFFERIFY_THRESHOLD: number = 100000;
 
-  get provider() {
+  get provider(): EthereumProvider {
     return this.#provider;
   }
```
jeffsmale90, line 48: why not just change `#provider` to be `readonly public`?

src/chains/filecoin/filecoin/src/ready.ts
```diff
@@ -1,9 +1,15 @@
-import type { FilecoinProvider } from "@ganache/filecoin";
```
jeffsmale90, line 1: Why not put this whole thing into index.ts, in the same way that it is in the ethereum chain?

jeffsmale90 (changes requested):
WIP review

src/packages/cli/tests/args.test.ts
```diff
@@ -1,5 +1,10 @@
 import assert from "assert";
-import args, { expandArgs } from "../src/args";
+import { parseArgs, expandArgs } from "../src/args";
+// pre import "@ganache/ethereum" so that the first test that is run doesn't
+// take a long time (CI might time out). The reason we do this is because cli
+// lazy-loads `require("@ganache/ethereum");`, and "@ganache/ethereum" is a
+// chonker.
+require("@ganache/ethereum");
```
jeffsmale90, line 7: Couldn't we do this same trick here: https://github.com/trufflesuite/ganache/pull/4362/files#r1186532908 so that we don't have silly high timeoutses?

src/chains/filecoin/filecoin/README.md
davidmurdoch: It's too much trouble, and I think it's fine since almost no one uses it.

src/packages/flavor/example/src/flavor.ts
davidmurdoch: As long as it works!

davidmurdoch:
@jeffsmale90 I'm 👍 on all your suggestions, as long as they still work 😁

src/packages/flavor/src/flavor.ts
```diff
@@ -0,0 +1,57 @@
...
+    OptionsConfig<any>,
+    OptionsConfig<any>
+  > = FlavorOptions<never, never, never>
+> = {
+  flavor: F;
+  connect: (
+    providerOptions: Parameters<O["provider"]["normalize"]>[0],
+    executor: Executor
+  ) => C;
+  ready: (config: {
```
jeffsmale90, line 50: I think a comment on here would be worthwhile - the functions a bit funny, and the only place I found an explanation was in README.md

src/packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,129 @@
...
+  const options = { hidden: true, alias: key };
+  return aliases.reduce((args, a) => args.option(a, options), args);
+};
+
+export function applyDefaults<D extends Defaults<any>>(
+  defaults: D,
+  args: yargs.Argv<{}>
+) {
+  for (const category in defaults) {
+    type GroupType = `${Capitalize<typeof category>}:`;
```
jeffsmale90, line 34: #notyourcode - Carrying the `:` here is a bit awkward, but don't feel strongly.

I guess adding it later is more work 🤷

jeffsmale90 (changes requested):

src/packages/cli/tests/args.test.ts
davidmurdoch: Yes

src/packages/cli/src/helpers.ts
davidmurdoch: 🤷

davidmurdoch:
I'm  👍 on all your new suggestions. Feel free to make the changes if you'd like.

PR #4375, opened 2023-04-29
fix: typo in ganache.d.ts

eltociear:
futher -> further


davidmurdoch:
This file is generated. You'll need to update it in the source code.

PR #4384, opened 2023-05-05
docs: preloads docs assets

davidmurdoch:
Reduces page load time by initiating downloads of assets that will get loaded from other CSS and JS files later.


docs/assets/js/inject-editor.js
```diff
@@ -55,7 +58,7 @@ function getTheme() {
     { token: "string", foreground: text4 },
     { token: "keyword", foreground: text6 }
   ];
-  const base = getUserColorTheme() === "light" ? "vs" : "vs-dark";
+  const base = userColorTheme === "light" ? "vs" : "vs-dark";
```
jeffsmale90: I don't know what the implications of `base` in the return object is, but when you change the theme, this property is not updated, i.e.
- Load the page in light mode, `base` is `vs`
- Click the theme, changes to dark, `base` is `vs`

I didn't see any obvious impact, but did confirm that the response from `getTheme()` is incorrect in `setupColorThemeListener()`

scripts/build-docs/index.ts
```diff
@@ -447,6 +447,17 @@ const preamble =
 
 **Pro Tip**: You can define your own provider by adding \`const provider = ganache.provider({})\` to the start of any example and passing in your [startup options](https://trufflesuite.com/docs/ganache/reference/cli-options/).`);
 
+// these are the fonts that are loaded when you navigate to ganache. It's possible others could be
+// loaded if users use the page, but these are the initial ones at time of writing. To verify this
+// is still true, load ganache.dev with devtools open, and check the `Font` tab within the Network
```
jeffsmale90: I have a couple extras on Chome on macOs:
```
https://fonts.gstatic.com/s/opensans/v35/memQYaGs126MiZpBA-UFUIcVXSCEkx2cmqvXlWq8tWZ0Pw86hd0Rk5hkWVAewIMUdjFXmQ.woff2
https://fonts.gstatic.com/s/opensans/v35/memvYaGs126MiZpBA-UvWbX2vVnXBbObj2OVTS-mu0SC55I.woff2
https://fonts.gstatic.com/s/sharetechmono/v15/J7aHnp1uDWRBEqV98dVQztYldFcLowEFA87Heg.woff2
https://fonts.gstatic.com/l/font?kit=7Au7p_IgjDKdCRWuR1azpmQIElsO02s2tozWoxjt&skey=3d5373e9fc70eefe&v=v14
https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.28.0/min/vs/base/browser/ui/codicons/codicon/codicon.ttf
```

docs/assets/js/inject-editor.js
jeffsmale90: Ah, I see now that's a global you are setting at startup time, so maybe it's intentional?

jeffsmale90 (changes requested):

scripts/build-docs/index.ts
davidmurdoch: Turns out google loads different fonts for each browser :-(

From https://developers.google.com/fonts/docs/technical_considerations:

> What is the Google Fonts API serving?
> When a browser sends a request for a Fonts API stylesheet (as specified in a <link> tag in your web page), the Fonts API serves a stylesheet generated for the specific user agent making the request.

So I should probably remove the preload links :-(

docs/assets/js/inject-editor.js
davidmurdoch: It wasn't, but it looks like we don't actually need the fallback theme to be correct, in practice, at least not in an obvious way. Regardless, I've updated it to change when the theme changes.

scripts/build-docs/index.ts
davidmurdoch: I've removed these google font preloads and added one for codicon

scripts/build-docs/index.ts
```diff
@@ -447,6 +447,13 @@ const preamble =
 
 **Pro Tip**: You can define your own provider by adding \`const provider = ganache.provider({})\` to the start of any example and passing in your [startup options](https://trufflesuite.com/docs/ganache/reference/cli-options/).`);
 
+// this font is loaded when you navigate to ganache. Others are loaded as well, but they are dependent on the,
```
jeffsmale90:
```suggestion
// this font is loaded when you navigate to ganache. Others are loaded as well, but they are dependent on the
```

jeffsmale90 (approved):
Approved! Just a suggestion on an errant comma.

PR #4427, opened 2023-06-01
fix: work around a Node v20.0.0-20.2.0 bug

davidmurdoch:
Node v20's `AutoSelectFamily` default is broken (in Node v20.0.0 - 20.2.0). Disabling the setting on start-up fixes it.

See https://github.com/nodejs/node/issues/47822#issuecomment-1564708870

### DIFF
--- a/packages/ethereum/ethereum/src/forking/handlers/http-handler.ts
+++ b/packages/ethereum/ethereum/src/forking/handlers/http-handler.ts
@@ -9,7 +9,7 @@ import { BaseHandler } from "./base-handler";
 import { Handler } from "../types";
 import Deferred from "../deferred";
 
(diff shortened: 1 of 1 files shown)


packages/ethereum/ethereum/src/forking/handlers/http-handler.ts
```diff
@@ -9,6 +9,21 @@ import { BaseHandler } from "./base-handler";
 import { Handler } from "../types";
 import Deferred from "../deferred";
 
+// Work around a node v20.0.0, v20.1.0, and v20.1.2 bug. The issue was fixed
```
davidmurdoch:
```suggestion
// Work around a node v20.0.0, v20.1.0, and v20.2.0 bug. The issue was fixed
```

PR #4430, opened 2023-06-02
fix: ignore stderr EPIPE errors when detached

davidmurdoch:


src/packages/cli/src/detach.ts
```diff
@@ -35,6 +35,14 @@ export function notifyDetachedInstanceReady(port: number) {
...
   // received before disconnecting from the child process.
   process.send(port);
+  process.stderr.on("error", e => {
+    // ignore errors from stderr, as once we tell the parent process we are
+    // ready it stops listening to stderror and results in a broken pipe.
+    // if any code tries to write to it it will fail with EPIPE.
+    if (e.code !== "EPIPE") {
+      throw e;
+    }
+  });
```
jeffsmale90, line 45: It'd be a lot nicer if we could just not raise the error :shrug:

I don't think there's a way to do that where a) we continue to receive stderr on the parent, and b) can disconnect from the child.

Maybe we could pipe the child's error through a logfile, or communicate startup failure over IPC, but in the meantime 🤷

jeffsmale90 (approved):

src/packages/cli/src/detach.ts
davidmurdoch: Yeah, I spent about an hour looking for alternatives to this approach and couldn't find one that didn't also complicate a bunch of other logic.

PR #4533, opened 2023-10-02
Fix: Replace provider.send with provider.request

pclaesen:
fix: replace provider.send with provider.request


davidmurdoch:
Thanks Pieter. The signature for `request` is `request({ method: string, params: any[] })`;

